PR #8, opened 2025-01-30
add zeroFee flag in `wallet.sendOffchain` method

louisinger:
`sendOffchain(params, true)` allows the user to submit a redeem transaction without miner fees. 

@tiero please review

### DIFF
--- a/src/core/wallet.ts
+++ b/src/core/wallet.ts
@@ -250,7 +250,10 @@ export class Wallet implements IWallet {
         return txid;
     }
 
-    async sendOffchain(params: SendBitcoinParams): Promise<string> {
+    async sendOffchain(
+        params: SendBitcoinParams,
+        zeroFee: boolean = false
+    ): Promise<string> {
         if (
             !this.arkProvider ||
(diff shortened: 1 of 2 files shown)


src/core/wallet.ts
```diff
@@ -250,7 +250,10 @@ export class Wallet implements IWallet {
         return txid;
     }
 
-    async sendOffchain(params: SendBitcoinParams): Promise<string> {
+    async sendOffchain(
+        params: SendBitcoinParams,
+        zeroFee: boolean = false
```
tiero, line 255:
```suggestion
        zeroFee: boolean = true
```

test/integration.test.ts
```diff
@@ -157,7 +157,7 @@ describe('Wallet SDK Integration Tests', () => {
     await carolWallet.sendBitcoin({
       address: daveOffchainAddress!,
       amount: sendAmount,
-    })
+    }, false)
```
tiero, line 160:
```suggestion
    }, true)
```

PR #22, opened 2025-03-12
add `getTransactionHistory()`

### LINKED ISSUE #20: Add GetTransactionHistory

louisinger:
This PR adds `getTransactionHistory` computing an ark transaction history from ListVtxos RPC and boarding transactions.

it closes #20 

@altafan @tiero please review


src/providers/ark.ts
```diff
@@ -98,7 +98,10 @@ export interface ArkInfo {
 
 export interface ArkProvider {
     getInfo(): Promise<ArkInfo>;
-    getVirtualCoins(address: string): Promise<VirtualCoin[]>;
+    getVirtualCoins(address: string): Promise<{
+        spendableVtxos: VirtualCoin[];
+        spentVtxos: VirtualCoin[];
+    }>;
```
tiero, line 104: do we want to maintain ListVtxos structure?

louisinger: we don't have a cache layer (let's add it in a dedicated PR ?)
also, `getVirtualCoins` becomes a private function in PR#26 https://github.com/ArkLabsHQ/wallet-sdk/pull/26/files#diff-f5acd2cbbe67faed5694c18e4cd12898151dd7363e77e54d9793a1d8fe8a7e85R136

src/utils/transactionHistory.ts
```diff
@@ -0,0 +1,196 @@
...
+            type: TxType.TxSent,
+            createdAt: vtxo.createdAt.toISOString(),
+            settled: true,
+        });
+    }
+
+    // sort by createdAt in descending order
+    txs.sort(
+        (a, b) =>
+            new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
```
altafan, line 178: with this sorting logic, where do the unconfirmed txs that have no or zero createdAt end up in the history? at the beginning or at the end? we want them to be at the beginning because soon they'll be confirmed and be the newest txs of the history

test/integration.test.ts
```diff
@@ -204,28 +183,161 @@ describe('Wallet SDK Integration Tests', () => {
...
+    expect(vtxo.txid).toBeDefined()
+    expect(vtxo.value).toBe(boardingAmount)
+    expect(vtxo.virtualStatus.state).toBe('settled')
+
+    // Check Alice's balance after funding
+    const aliceBalanceAfterFunding = await alice.wallet.getBalance()
+    expect(aliceBalanceAfterFunding.offchain.total).toBe(boardingAmount)
+
+    // Check history before sending to bob
+    let aliceHistory = await alice.wallet.getTransactionHistory()
```
altafan, line 265: let's make sure the history looks like we want also before sending to boarding address and settle

PR #26, opened 2025-03-16
add `ServiceWorkerWallet` implementation

louisinger:
_This PR defines a new `ServiceWorkerWallet` implementing the `Wallet` interface. It uses a service worker files exported as `@arklabs/wallet-sdk/sw`_.

`src/sw/service.ts` defines a service worker file running a `BareWallet` instance and listening message from browser event listener. `ServiceWorkerWallet` sends requests defined in `src/sw/request.ts` to the service worker and handle the responses defined in `src/sw/response.ts`. 

Service worker is built and bundled via `esbuild`, the lib exports it as `@arklabs/wallet-sdk/sw`.

**Try it**

`test.html` is a simple webpage running a service worker wallet.

```bash
pnpm run build
```

```bash
pnpm run test:sw
```
 
 go to http://localhost:3030/
 
 
image

 @tiero @altafan please review


src/sw/response.ts
```diff
@@ -0,0 +1,213 @@
...
+    SpendableVtxo,
+    VirtualCoin,
+} from "../core/wallet";
+import { SettlementEvent } from "../providers/ark";
+import { AddressInfo } from "../core/wallet";
+
+export namespace Response {
+    export type Type =
+        | "WALLET_INITIALIZED"
+        | "SETTLE_EVENT"
```
tiero, line 14: can't we have multiple types for each MuSig2 session "stages"?

src/core/sw_wallet.ts
```diff
@@ -0,0 +1,287 @@
...
+                        );
+                        reject(new Error((response as Response.Error).message));
+                        return;
+                    }
+
+                    switch (response.type) {
+                        case "SETTLE_EVENT":
+                            if (callback) {
+                                callback(
+                                    (response as Response.SettleEvent).event
```
tiero, line 245: this way we are lifting up the event from server to the consumers of the SDK. I would do an extra step here and try to type the possible "stages" into concrete domain

src/core/wallet.ts
```diff
@@ -134,24 +132,63 @@ export interface VirtualCoin extends Coin {
...
+    // Event handling
+    subscribeToEvents(
+        message: string,
+        signature: string,
+        address: string
+    ): Promise<void>;
+}
+
+// BareWallet does not store any data and rely on the Ark and onchain providers to fetch utxos and vtxos
+export class BareWallet implements Wallet {
```
tiero, line 170: Why not keeping Wallet for class and added new WalletInterface?

src/index.ts
```diff
@@ -16,13 +17,15 @@ import {
...
     WalletConfig,
+    Wallet,
     SettlementEvent,
     SettlementEventType,
     OnchainProvider,
     ArkProvider,
 };
 export {
-    Wallet,
+    BareWallet,
```
tiero, line 27: If we don't keep Wallet, I would use BaseWallet or CoreWallet ?

src/core/wallet.ts
louisinger: the interface is the one used everywhere so better to keep it as short as possible? Also "Interface" is redundant. 
anyway I can revert

tiero: IWallet ?

src/core/wallet.ts
```diff
@@ -134,24 +132,63 @@ export interface VirtualCoin extends Coin {
...
 
-    public boardingTapscript?: VtxoTapscript;
-
     static DUST_AMOUNT = BigInt(546); // Bitcoin dust limit in satoshis = 546
     static FEE_RATE = 1; // sats/vbyte
 
-    constructor(config: WalletConfig) {
+    private initialized = false;
+
+    async init(config: WalletConfig): Promise<void> {
```
tiero, line 187: I do prefer using functional options pattern vs the init() approach.
I don't dislike sourcing the public key, if we can avoid the getInfo round-trip?

src/core/wallet.ts
```diff
@@ -373,12 +421,22 @@ export class Wallet {
...
     }
 
-    async sendOnchain(params: SendBitcoinParams): Promise<string> {
+    private async sendOnchain(params: SendBitcoinParams): Promise<string> {
+        if (
+            !this.onchainP2TR ||
+            !this.onchainAddress ||
+            !this.network ||
+            !this.identity ||
+            !this.onchainProvider
```
tiero, line 433: we are not consistent with initialized wallet check.

src/core/sw_wallet.ts
```diff
@@ -0,0 +1,273 @@
...
+            navigator.serviceWorker.addEventListener("message", messageHandler);
+            if (this.serviceWorker) {
+                this.serviceWorker.postMessage(message);
+            } else {
+                reject(new Error("Service worker not initialized"));
+            }
+        });
+    }
+
+    async setupServiceWorker(path = "/sw.js"): Promise<void> {
```
louisinger, line 52: move this into static constructor function

tiero (commented):
Let's be consistent in file name conventions (dashes, underscore)

README.md
```diff
@@ -31,10 +31,42 @@ const wallet = new Wallet({
...
+// initialize the wallet
+await wallet.init({
+  network: 'mutinynet',  // 'bitcoin', 'testnet', 'regtest', 'signet' or 'mutinynet'
+  identity: identity,
+  // Esplora API, can be left empty mempool.space API will be used
+  esploraUrl: 'https://mutinynet.com/api', 
+  // OPTIONAL Ark Server connection information
+  arkServerUrl: 'https://master.mutinynet.arklabs.to',
+  arkServerPublicKey: 'd45fc69d4ff1f45cbba36ab1037261863c3a49c4910bc183ae975247358920b6'
+})
```
tiero, line 69: I would move this to the bottom of the Usage section.

louisinger: done

tiero (approved):

PR #27, opened 2025-03-17
Identity abstraction

louisinger:
This PR moves transaction signing process in the identity interface:

```ts
export interface Identity {
    sign(tx: Transaction, inputIndexes?: number[]): Promise<Transaction>;
    xOnlyPublicKey(): Uint8Array;
    signerSession(): SignerSession;
}
```

TL;DR
* do not expose privateKey.
* add signerSession() getter to handle musig2 tree signing session.
* remove unused sign(message), replace by sign(tx, inputIndexes?).
* `Wallet` uses `Identity.sign` to sign all transactions and `Identity.signerSession` to handle musig2 nonces and sigs in settlement process.

@tiero please review


src/core/identity.ts
```diff
@@ -29,39 +29,30 @@ export class InMemoryKey implements Identity {
...
+                throw new Error("Failed to sign transaction");
+            }
+            return txCpy;
+        }
 
-    privateKey(): Uint8Array {
-        return this.key;
-    }
-}
+        for (const inputIndex of inputIndexes) {
```
tiero, line 42: we should maybe check if inputIndex is between boundaries of the psbt.inputs?

PR #28, opened 2025-03-21
`VtxoScript` abstraction

louisinger:
_This PR proposes different types and classes letting to abstract Vtxo contracts. It allows handling multiparty contract like VHTLC._ 

### ArkTapscript

The `ArkTapscript` interface represents the available tapleaf scripts. The PR adds all scripts defined in https://github.com/ark-network/ark.

### VtxoScript

`VtxoScript` is the base class to implement a Vtxo contract. The PR adds 2 implementations: `DefaultVtxo` and `VHTLC`.

The [VHTLC e2e test](https://github.com/ArkLabsHQ/wallet-sdk/pull/28/files#diff-8fdaf3b73054c861d4b1afc72e76086d85c318c75bde88a452ef344c24b7d8b0R414) shows how to use the new types to create custom vtxo contract.

@tiero @altafan  please review


louisinger:
something weird happening in test, need to fix but it pass locally

src/networks.ts
```diff
@@ -8,6 +8,7 @@ export type NetworkName =
     | "regtest";
 
 export interface Network {
+    ark: string;
```
altafan, line 11: what's this?

src/index.ts
```diff
@@ -38,9 +67,21 @@ export {
...
     Request,
     Response,
+    decodeTapscript,
+    MultisigTapscript,
+    CSVMultisigTapscript,
+    ConditionCSVMultisigTapscript,
+    ConditionMultisigTapscript,
+    CLTVMultisigTapscript,
+    addConditionWitness,
+    CONDITION_WITNESS_KEY_PREFIX,
```
altafan, line 84: isn't it better to just expose a method that takes care of adding the condition witness to the psbt and hide the prefix used in the unknown fields?

src/networks.ts
louisinger: yes

src/index.ts
louisinger: `addConditionWitness` is that method, I've exported the key just in case you want to use this lib to do custom things, maybe too much ?

src/networks.ts
altafan: rename then

src/utils/psbt.ts
```diff
@@ -0,0 +1,79 @@
...
+    tx: Transaction,
+    witness: Uint8Array[]
+): void {
+    const witnessBytes = RawWitness.encode(witness);
+
+    tx.updateInput(inIndex, {
+        unknown: [
+            [
+                {
+                    type: 255,
```
altafan, line 29: can we have a const also for hardcoded value please?

altafan (approved):

PR #37, opened 2025-03-29
Fix exit timelock

louisinger:
This PR updates the exit timelocks of vtxo script exit tapscripts to be compliant with the [latest validation PR](https://github.com/ark-network/ark/pull/505).

@tiero please review


tiero:
I think would be more stable to monitor mempool for inclusion via bitcoin JSONRPC calls, rather than having hardcoded timeouts here

PR #58, opened 2025-05-27
stop managing service worker updates

bordalix:
Fixes the infernal loop of service worker auto updating and reloading the page.

Commented all the code that manages service worker updates as a quick fix.

Let's leave this for the developer to manage, as he would maybe want to add more functionalities to his service worker, and having this management on the SDK could clash with the developer's will.

On the developer's side he could do something like this on his service worker:

```js
import { Worker } from '@arkade-os/ts-sdk'

const worker = new Worker()
worker.start().catch(console.error)

declare const self: ServiceWorkerGlobalScope

// activate service worker immediately
self.addEventListener('install', () => {
  self.skipWaiting()
})

// take control of clients immediately
self.addEventListener('activate', () => {
  self.clients.claim()
})
```

@tiero @louisinger I would like to hear your opinions.

### DIFF
--- a/src/wallet/serviceWorker/wallet.ts
+++ b/src/wallet/serviceWorker/wallet.ts
@@ -119,28 +119,28 @@ export class ServiceWorkerWallet implements IWallet {
             registration = await navigator.serviceWorker.register(path);
 
             // Handle updates
-            registration.addEventListener("updatefound", () => {
-                console.info(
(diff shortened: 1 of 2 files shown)


louisinger:
Can we keep a default update logic (maybe the one you proposed ?) and add a `withServiceWorkerUpdate` boolean in `create()` factory ?

Issue #67, opened 2025-06-05
Transaction V3 and P2A


louisinger:
Virtual transactions are v3 without fees + P2A output everywhere.

louisinger:
closed by #84

PR #78, opened 2025-06-10
New round flow

### LINKED ISSUE #70: Update batch session with new events
v0.7.0 will introduce new events :
- BatchTreeEvent
- BatchTreeSignatureEvent
- BatchConfirmedEvent

louisinger:
This PR implements the incoming v7 events flow in settlement session.

### Remove ping, replace by Confirmation

no more ping is necessary, instead the user reacts to `BatchStartedEvent` via ConfirmRegistration RPC. Also, the initial GetInfo call is removed because BatchStartedEvent contains every data we need to craft forfeits and validate the vtxo tree.

### Receive tx trees via special events

transactions trees (connectors and vtxosTree) are now sent via a list of `BatchTreeEvent` and `BatchTreeSignatureEvent`. 

### Forfeit V3

This PR needs forfeit transaction to make the settlement done, thus it migrates them to P2A/V3. However, it does not close https://github.com/arkade-os/ts-sdk/issues/67 because offchain tx has to be done (coming in another PR).

### Run CI on next-version

CI is now targeting next-version branch instead of master. Should be reverted once `next-version` is merged to `master`.

it closes #70 

@bordalix please review


src/providers/ark.ts
```diff
@@ -114,7 +139,10 @@ export interface ArkProvider {
     }>;
     submitVirtualTx(psbtBase64: string): Promise<string>;
     subscribeToEvents(callback: (event: ArkEvent) => void): Promise<() => void>;
-    registerInputsForNextRound(inputs: Input[]): Promise<{ requestId: string }>;
+    registerInputsForNextRound(
```
Kukks, line 142: why do we still have these?

PR #83, opened 2025-06-13
Drop explorer for indexer

### LINKED ISSUE #76: Move to Indexer endpoints
"Explorer" will be dropped in v7, replace by Indexer

bordalix:
Drops rest endpoints provided by Explorer and uses endpoints provided by Indexer.

Closes #76 

@louisinger please review


src/providers/ark.ts
```diff
@@ -563,14 +567,30 @@ export class RestArkProvider implements ArkProvider {
...
-        address: string,
+    async *subscribeForScripts(
+        scripts: string[],
         abortSignal: AbortSignal
     ): AsyncIterableIterator<{
         newVtxos: VirtualCoin[];
         spentVtxos: VirtualCoin[];
     }> {
-        const url = `${this.serverUrl}/v1/vtxos/${address}/subscribe`;
+        const response = await fetch(`${this.serverUrl}/v1/script/subscribe`, {
```
louisinger, line 577: this should be a specific method in order to allow update an existing  subscription

src/wallet/serviceWorker/worker.ts
```diff
@@ -93,18 +93,15 @@ export class Worker {
...
         try {
             const addressScripts = [...scripts.exit, ...scripts.forfeit];
             const vtxoScript = DefaultVtxo.Script.decode(addressScripts);
             const tapLeafScript = vtxoScript.findLeaf(scripts.forfeit[0]);
 
             const abortController = new AbortController();
-            const subscription = this.arkProvider!.subscribeForAddress(
-                address,
+            const subscription = this.arkProvider!.subscribeForScripts(
+                addressScripts,
```
louisinger, line 104: this address scripts are the taproot scripts, not the address's pkScript. It should use `address.pkScript`.

Important point: now the subscribe RPC is using `taprootKey = address.pkScript[2:]` instead of the whole pkScript. It may change in the future but we should handle this now in order to make it work.

src/providers/ark.ts
```diff
@@ -164,8 +164,8 @@ export interface ArkProvider {
         signedRoundTx?: string
     ): Promise<void>;
     getEventStream(signal: AbortSignal): AsyncIterableIterator<SettlementEvent>;
-    subscribeForAddress(
-        address: string,
+    subscribeForScripts(
```
louisinger, line 167: Let's move all methods related to `indexer.proto` to a specific `InderxerProvider` interface/class. `ArkProvider` should keep only the `service.proto` RPCs. Soon, Indexer and arkd won't be the same host.

louisinger (changes requested):
`IndexerProvider` should be a dedicated interface implementing all the indexer.proto RPCs. 
Let's also add integration tests, at least for the subscription behaviour because it is used widely by the service provider wallet (untested on master)

bordalix:
> this should be a specific method in order to allow updating an existing subscription. it accepts an optional subscriptionId param: if set, it updates the subscription instead of create a new one.

@louisinger did you meant something like [this](https://github.com/arkade-os/ts-sdk/pull/83/commits/b53e23088d72b4c928f01de2ea00a0885d59bdeb)?

src/providers/indexer.ts
```diff
@@ -0,0 +1,227 @@
...
+    spentBy: string;
+    commitmentTxid: string;
+};
+
+type IndexerGetVirtualCoinsResponse = {
+    vtxos: IndexerGetVirtualCoinsVtxoResponse[];
+    page: number;
+};
+
+export interface IndexerProvider {
```
louisinger, line 26: It should implement all indexer RPCs : https://github.com/arkade-os/arkd/blob/master/api-spec/protobuf/ark/v1/indexer.proto#L7

bordalix:
@altafan @Kukks since you guys made the server side of indexer, I think it's a good idea if you also review this.

src/providers/indexer.ts
```diff
@@ -0,0 +1,489 @@
...
+
+    async GetCommitmentTxForfeitTxs(
+        txid: string,
+        opts?: {
+            pageIndex?: number;
+            pageSize?: number;
+        }
+    ): Promise<string[]> {
+        let url = `${this.serverUrl}/v1/commitmentTx/${txid}/forfeitTxs`;
+        if (opts) {
```
louisinger, line 191: it seems to be repeated a lot of time, can we move it to some util function ?

src/providers/indexer.ts
```diff
@@ -0,0 +1,489 @@
...
+    isSpent: boolean;
+    spentBy: string;
+    commitmentTxid: string;
+};
+
+export interface IndexerProvider {
+    GetCommitmentTx(txid: string): Promise<IndexerCommitmentTx>;
+    GetCommitmentTxConnectors(
+        txid: string,
+        opts?: {
```
louisinger, line 79: let's add a `PaginationOptions` type for all the opts args?

src/providers/indexer.ts
```diff
@@ -0,0 +1,489 @@
...
+    ): Promise<Outpoint[]>;
+    GetCommitmentTxSwept(txid: string): Promise<string[]>;
+    GetTransactionHistory(
+        address: string,
+        opts?: {
+            startTime?: number;
+            endTime?: number;
+            pageIndex?: number;
+            pageSize?: number;
+        }
```
louisinger, line 106:
```suggestion
        opts?: {
            startTime?: number;
            endTime?: number;
        } & PaginationOptions
```

louisinger (commented):
can we add e2e test for indexer endpoints ? at least for subscriptions and GetVtxos because the service worker wallet depends on them

examples/spillman.js
```diff
@@ -111,7 +112,10 @@ async function main() {
     // Get the virtual coins for the Spillman Channel address
     console.log("Fetching virtual coins...");
     const arkProvider = new RestArkProvider("http://localhost:7070");
-    const { spendableVtxos } = await arkProvider.getVirtualCoins(address);
+    const indexerProvider = new RestIndexerProvider("http://localhost:7070");
```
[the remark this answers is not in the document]
bordalix: spillman was broken, fixed by 291e8fce66476d4a08cdc5390d6fca5d700d0f4c

src/providers/ark.ts
```diff
@@ -899,27 +798,9 @@ function encodeSignaturesMatrix(signatures: TreePartialSigs): string {
...
-        },
-        spentBy: vtxo.spentBy,
-        createdAt: new Date(vtxo.createdAt * 1000),
-    };
-}
-
 // ProtoTypes namespace defines unexported types representing the raw data received from the server
-namespace ProtoTypes {
-    interface Node {
+export namespace ProtoTypes {
```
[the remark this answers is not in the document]
bordalix: fixed

server.Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM golang:1.23.1 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 
-ARG BRANCH=next-version
+ARG BRANCH=deca5ef4699037dcc38a320e50478fc3eae957ec
```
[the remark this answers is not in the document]
bordalix: It was a test to see if it passes the tests.
I just realized ALL ts-sdk `next-version` is broken against arkd `next-version` :(

altafan (approved):

PR #91, opened 2025-06-19
Fetch forfeit from get infos

### LINKED ISSUE #90: do not use BatchStartedEvent.forfeitAddress
once https://github.com/arkade-os/arkd/pull/622 is merged, we should get forfeit address from getInfos instead of BatchStartedEvent.

louisinger:
it closes #90 

@bordalix please review


louisinger:
Will be included into #83

PR #125, opened 2025-07-17
allow inputIndexes to be present but undefined

bordalix:
@louisinger please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for certain message properties to ensure more accurate type checking.

### DIFF
--- a/src/wallet/serviceWorker/request.ts
+++ b/src/wallet/serviceWorker/request.ts
@@ -166,7 +166,7 @@ export namespace Request {
             message.type === "SIGN" &&
             "tx" in message &&
             typeof message.tx === "string" &&
-            ("inputIndexes" in message && message.inputIndexes !== undefined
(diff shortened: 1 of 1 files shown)


src/wallet/serviceWorker/request.ts
```diff
@@ -166,7 +166,7 @@ export namespace Request {
             message.type === "SIGN" &&
             "tx" in message &&
             typeof message.tx === "string" &&
-            ("inputIndexes" in message
+            ("inputIndexes" in message && message.inputIndexes !== undefined
```
louisinger, line 169:
```suggestion
            ("inputIndexes" in message && message.inputIndexes != undefined
```
non-strict equal should also handle case where inputIndexes is "null", am I right @coderabbitai ?

louisinger: @bordalix let's do that no ?

louisinger (approved):

PR #126, opened 2025-07-17
fix service worker example

louisinger:
Re-export `Transaction` type from btc-signer because it is used as input argument of `Identity.sign` method.

fix and update service worker example file.

@bordalix @tiero please review

## Summary by CodeRabbit

* **New Features**
  * Introduced a "Signing" section in the test UI, allowing users to input a PSBT, specify input indexes, and sign transactions directly from the browser.
* **Improvements**
  * Updated wallet initialization in the test environment for improved integration with service workers.
* **Removals**
  * Removed the "Get Coins" button and its related functionality from the test UI.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,4 @@
+import { Transaction } from "@scure/btc-signer";
 import { SingleKey } from "./identity/singleKey";
 import { Identity } from "./identity";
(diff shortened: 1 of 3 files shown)


tiero:
utACK

PR #128, opened 2025-07-22
Fix subscription

bordalix:
It's hard to identify timeout errors from fetch, but we will assume by context and error description that this is a timeout error (tipically a http code 524 by cloudflare), which will not break the outer while loop, making it re-connect after initial timeout.

TODO: we should find a more robust way to identify a timeout error.

https://github.com/user-attachments/assets/f0c90f71-a27e-44d8-a64f-5bdbd960a25d

@altafan please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of fetch timeout errors, including better handling of specific Cloudflare timeout scenarios.

### DIFF
--- a/src/providers/ark.ts
+++ b/src/providers/ark.ts
@@ -830,13 +830,18 @@ namespace ProtoTypes {
 
 export function isFetchTimeoutError(err: any): boolean {
     const checkError = (error: any) => {
+        if (!(error instanceof Error)) return false;
+
+        // TODO: get something more robust than this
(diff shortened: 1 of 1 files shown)


louisinger:
are we sure this error is thrown by fetch ? in theory error code 5x shouldn't throw :
https://stackoverflow.com/questions/38235715/fetch-reject-promise-and-catch-the-error-if-status-is-not-ok#38236296

altafan (approved):

PR #133, opened 2025-08-05
fix wallet returned status

bordalix:
@louisinger I think I found the bug.

The wallet returns one boolean as state, when in fact it has 3 states:
- no service worker defined
- service worker running but not private key (i.e. locked)
- service worker running with private key (i.e. unlocked)

First bug:
- when returning the status, is only returned true if sw was running, not considering if it was unlocked
- now it also checks if it nows the private key and only return if BOTH conditions are true

Second bug:
- when calling for init, if the sw was already running we were not passing it the private key
- now we check if it's running and unlocked, or else we pass it the private key again

Wallet is working like a charm with this changes

We should publish a new release with this fixes

### DIFF
--- a/src/wallet/serviceWorker/utils.ts
+++ b/src/wallet/serviceWorker/utils.ts
@@ -24,11 +24,10 @@ export async function setupServiceWorker(path: string): Promise<ServiceWorker> {
         throw new Error("Failed to get service worker instance");
     }
     // wait for the service worker to be ready
-    if (serviceWorker.state !== "activated") {
-        await new Promise<void>((resolve) => {
-            if (!serviceWorker) return resolve();
-            serviceWorker.addEventListener("activate", () => resolve());
-        });
-    }
-    return serviceWorker;
+    return new Promise<ServiceWorker>((resolve) => {
+        if (serviceWorker.state === "activated") return resolve(serviceWorker);
+        serviceWorker.addEventListener("activate", () =>
+            resolve(serviceWorker)
+        );
+    });
 }
--- a/src/wallet/serviceWorker/wallet.ts
+++ b/src/wallet/serviceWorker/wallet.ts
@@ -62,7 +62,10 @@ export class ServiceWorkerWallet implements IWallet, Identity {
         };
         const response = await this.sendMessage(message);
         if (Response.isWalletStatus(response)) {
-            return response.status;
+            const { status } = response;
+            status.walletInitialized &&=
+                this.cachedXOnlyPublicKey !== undefined;
+            return status;
         }
         throw new UnexpectedResponseError(response);
     }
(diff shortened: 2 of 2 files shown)


src/wallet/serviceWorker/utils.ts
```diff
@@ -24,11 +24,10 @@ export async function setupServiceWorker(path: string): Promise<ServiceWorker> {
...
-            if (!serviceWorker) return resolve();
-            serviceWorker.addEventListener("activate", () => resolve());
-        });
-    }
-    return serviceWorker;
+    return new Promise<ServiceWorker>((resolve) => {
+        if (serviceWorker.state === "activated") return resolve(serviceWorker);
+        serviceWorker.addEventListener("activate", () =>
+            resolve(serviceWorker)
+        );
```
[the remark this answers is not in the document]
tiero: this sounds plausible. Check? @bordalix

src/wallet/serviceWorker/utils.ts
```diff
@@ -24,11 +24,31 @@ export async function setupServiceWorker(path: string): Promise<ServiceWorker> {
...
+            cleanup();
+            resolve(serviceWorker);
+        };
+
+        const onError = () => {
+            cleanup();
+            reject(new Error("Service worker failed to activate"));
+        };
+
+        const timeout = setTimeout(() => {
```
louisinger, line 40: Why timeout ? It's possible that a svc worker registration "hangs" ? Isn't internally handled via an error ?

src/wallet/serviceWorker/wallet.ts
```diff
@@ -62,7 +62,10 @@ export class ServiceWorkerWallet implements IWallet, Identity {
         };
         const response = await this.sendMessage(message);
         if (Response.isWalletStatus(response)) {
-            return response.status;
+            const { status } = response;
+            status.walletInitialized &&=
```
louisinger, line 66: if `walletInitialized = true`, it means the private key is there so it may be possible to get the xonly key. Can't we update the `cachedXOnlyPublicKey` here instead of returning false ? it would save a "load priv key" action from the user just for that xonly cache key.

src/wallet/serviceWorker/wallet.ts
```diff
@@ -80,7 +83,8 @@ export class ServiceWorkerWallet implements IWallet, Identity {
 
         if (
             Response.isWalletStatus(response) &&
-            response.status.walletInitialized
+            response.status.walletInitialized &&
+            this.cachedXOnlyPublicKey !== undefined
```
louisinger, line 87: same here, if walletInitialized = true but no cache xonly, let's update the cache ?

louisinger (changes requested):

src/wallet/serviceWorker/utils.ts
bordalix: Good question. Well, I guess is better to have something that throws after 10 seconds instead of hanging forever.

src/wallet/serviceWorker/wallet.ts
```diff
@@ -85,6 +87,8 @@ export class ServiceWorkerWallet implements IWallet, Identity {
             if (failIfInitialized) {
                 throw new Error("Wallet already initialized");
             }
+
+            this.cachedXOnlyPublicKey = this.xOnlyPublicKey();
```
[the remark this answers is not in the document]
bordalix: Nice catch. Fixed by 8768c35

PR #139, opened 2025-09-07
refactor wallet interfaces

tiero:
## Summary by CodeRabbit

* **New Features**
  * Pluggable storage adapters, cross-environment Wallet.create, ServiceWorkerWallet.setup/create, OnchainWallet.create, wallet & contract repositories, VtxoManager (renewal/recovery), Expo adapters, waitForIncomingFunds, DEFAULT_DB_NAME.

* **Refactor**
  * Repository-backed persistence, messaging-driven service worker lifecycle, async identity/signing APIs, Intent-based signing, SSE/EventSource streaming utilities.

* **Bug Fixes**
  * Streaming resilience, subscription events include sweptVtxos, polling de-duplication, boarding-tx expiry helper.

* **Chores**
  * Version bump (0.3.0-alpha.8), Node >=20, release script and test tooling updates.


README.md
```diff
@@ -213,33 +220,144 @@ import { Worker } from '@arkade-os/sdk'
...
+  storage: new FileSystemStorageAdapter('./wallet-data')
+})
+
+// React Native with AsyncStorage
+const wallet = await Wallet.create({
+  identity,
+  arkServerUrl: 'https://mutinynet.arkade.sh',
+  storage: new AsyncStorageAdapter()
+})
+```
```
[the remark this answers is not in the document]
tiero: This allows to not have to import incompatible adapters and break installs

src/identity/index.ts
```diff
@@ -1,10 +1,13 @@
-import { Transaction } from "@scure/btc-signer";
+import { Transaction } from "@scure/btc-signer/transaction.js";
 import { SignerSession } from "../tree/signingSession";
 
 export interface Identity {
+    signerSession(): SignerSession;
+    xOnlyPublicKey(): Promise<Uint8Array>;
+    compressedPublicKey(): Promise<Uint8Array>;
+    signMessage(message: string): Promise<Uint8Array>;
```
louisinger, line 8:
```suggestion
    signMessage(message: Uint8Array): Promise<Uint8Array>;
```
so we don't have to handle encoding ?

src/identity/index.ts
```diff
@@ -1,10 +1,13 @@
-import { Transaction } from "@scure/btc-signer";
+import { Transaction } from "@scure/btc-signer/transaction.js";
 import { SignerSession } from "../tree/signingSession";
 
 export interface Identity {
+    signerSession(): SignerSession;
+    xOnlyPublicKey(): Promise<Uint8Array>;
+    compressedPublicKey(): Promise<Uint8Array>;
```
louisinger, line 7:
```suggestion
    publicKey(): Promise<Uint8Array>; // returns the compressed 33 bytes
```

xOnly is a subset of the compressedPublicKey, so maybe worth to save 1 method to simplify implementations ?

package.json
```diff
@@ -13,6 +13,36 @@
...
+            "types": "./dist/types/adapters/indexedDB.d.ts",
+            "import": "./dist/esm/adapters/indexedDB.js",
+            "require": "./dist/cjs/adapters/indexedDB.js",
+            "default": "./dist/esm/adapters/indexedDB.js"
+        },
+        "./adapters/asyncStorage": {
+            "types": "./dist/types/adapters/asyncStorage.d.ts",
+            "import": "./dist/esm/adapters/asyncStorage.js",
+            "require": "./dist/cjs/adapters/asyncStorage.js",
+            "default": "./dist/esm/adapters/asyncStorage.js"
```
louisinger, line 45: Because `sideEffect` is set to false, I do think we do not need special export paths to enable "dead code elimination" (https://runebook.dev/en/articles/esbuild/api/index/tree-shaking)

Can we revert to a single import path ? Sounds simpler to me

src/wallet/wallet.ts
```diff
@@ -110,24 +132,57 @@ export class Wallet implements IWallet {
...
         readonly forfeitOutputScript: Bytes,
-        readonly dustAmount: bigint
-    ) {}
+        readonly dustAmount: bigint,
+        walletRepository: WalletRepository,
+        contractRepository: ContractRepository
+    ) {
+        this.walletRepository = walletRepository;
+        this.contractRepository = contractRepository;
+    }
```
louisinger, line 141:
```suggestion
private constructor(
readonly identity: Identity,
readonly network: Network,
@@ -110,24 +132,57 @@
readonly boardingTapscript: DefaultVtxo.Script,
readonly serverUnrollScript: CSVMultisigTapscript.Type,
readonly forfeitOutputScript: Bytes,
        readonly dustAmount: bigint,
        readonly walletRepository: WalletRepository,
        readonly contractRepository: ContractRepository
    ) {}
```

public fields can be delcared in constructor

src/utils/anchor.ts
```diff
@@ -1,5 +1,5 @@
 import { hex } from "@scure/base";
-import { Transaction } from "@scure/btc-signer";
+import { Transaction } from "@scure/btc-signer/transaction.js";
```
louisinger, line 2: why ? please revert if possible 🙏

louisinger (changes requested):
Why all btc-signer imports have been replaced by `.js` ? when it is possible, let's use the default import path

src/wallet/serviceWorker/wallet.ts
```diff
@@ -37,87 +48,144 @@ class UnexpectedResponseError extends Error {
...
+};
+
+export type ServiceWorkerWalletSetupOptions = ServiceWorkerWalletOptions & {
+    serviceWorkerPath: string;
+};
+
+export class ServiceWorkerWallet implements IWallet {
+    public readonly walletRepository: WalletRepository;
+    public readonly contractRepository: ContractRepository;
+    public readonly identity: Identity;
```
louisinger, line 91: having the identity inside the service worker "client" is a regression in terms of security. before that change, the only way to access the secret key was inside the service worker. 

`identity` field should send message to worker instead of being a raw privkey

src/identity/index.ts
bordalix: Nack.
One of the usages of this is to sign an address (string) to implement magic hint with Boltz swaps.
Don't see why we should add an extra burden to the user to encode the address into bytes to sign it.

package.json
bordalix: @tiero what do you think?

PR #152, opened 2025-09-26
update v8 protos

bordalix:
@altafan please review

### DIFF
--- a/src/providers/ark.ts
+++ b/src/providers/ark.ts
@@ -741,20 +741,22 @@ export class RestArkProvider implements ArkProvider {
     }
 }
 
-function encodeMusig2Nonces(nonces: TreeNonces): string {
+function encodeMusig2Nonces(nonces: TreeNonces): Record<string, string> {
     const noncesObject: Record<string, string> = {};
     for (const [txid, nonce] of nonces) {
         noncesObject[txid] = hex.encode(nonce.pubNonce);
     }
-    return JSON.stringify(noncesObject);
+    return noncesObject;
 }
 
-function encodeMusig2Signatures(signatures: TreePartialSigs): string {
+function encodeMusig2Signatures(
+    signatures: TreePartialSigs
+): Record<string, string> {
     const sigObject: Record<string, string> = {};
     for (const [txid, sig] of signatures) {
         sigObject[txid] = hex.encode(sig.encode());
     }
-    return JSON.stringify(sigObject);
+    return sigObject;
 }
 
 function decodeMusig2Nonces(str: string): TreeNonces {


src/providers/ark.ts
```diff
@@ -159,6 +185,9 @@ export interface ArkProvider {
         commitmentTx?: TxNotification;
         arkTx?: TxNotification;
     }>;
+    getPendingTxs(proof: string): Promise<{
+        pendingTxs: PendingTx[];
+    }>;
```
louisinger, line 190:
```suggestion
    getPendingTxs(proof: string): Promise<PendingTx[]>;
```

wdyt ?

src/providers/ark.ts
```diff
@@ -159,6 +185,9 @@ export interface ArkProvider {
         commitmentTx?: TxNotification;
         arkTx?: TxNotification;
     }>;
+    getPendingTxs(proof: string): Promise<{
```
louisinger, line 188:
```suggestion
    getPendingTxs(proof: string, message: string): Promise<{
```

src/providers/ark.ts
```diff
@@ -532,7 +568,44 @@ export class RestArkProvider implements ArkProvider {
...
+        pendingTxs: PendingTx[];
+    }> {
+        const url = `${this.serverUrl}/v1/tx/pending`;
+        const response = await fetch(url, {
+            method: "POST",
+            headers: {
+                "Content-Type": "application/json",
+            },
+            body: JSON.stringify({
+                proof,
```
louisinger, line 581: this will require both proofTx and message to be set, not sure what will be the exact field names after intent refactor / cc @altafan

louisinger (changes requested):

src/providers/ark.ts
bordalix: fixed by eacfd6c836e44fbcfd5226986178a5e24da5190c

src/providers/ark.ts
```diff
@@ -92,6 +92,29 @@ export interface MarketHour {
...
+}
+
+export interface PendingTx {
+    arkTxid: string;
+    finalArkTx: string;
+    signedCheckpointTxs: string[];
+}
+
+export interface DeprecatedSigner {
+    cutoffDate: bigint;
```
Kukks, line 114: this might be optional? @altafan

src/providers/ark.ts
```diff
@@ -532,6 +566,39 @@ export class RestArkProvider implements ArkProvider {
         }
     }
 
+    async getPendingTxs(proof: string): Promise<PendingTx[]> {
```
Kukks, line 569: Maybe we should be passing a better concrete object here?

PR #195, opened 2025-10-18
Computing onchain sending fees from transanction size

### LINKED ISSUE #63: Wallet.sendOnchain should compute the tx size in order to estimate fees
The tx size now is hardcoded, it should use `TxWeightEstimator` class.

https://github.com/arkade-os/ts-sdk/blob/master/src/wallet/onchain.ts#L113

shubertm:
@louisinger, @bordalix, @tiero  could you take a look please

it closes #63 

## Summary by CodeRabbit

* **Bug Fixes**
  * Block-tip validation corrected so each block is consistently evaluated as valid or invalid.

* **New Features**
  * Fee estimation now iterates to converge on accurate fees and surfaces an explicit error when it fails.
  * Change handling is dust-aware via an explicit dust threshold; final outputs consider recipient address type when sizing fees.

* **Tests**
  * Added tests for insufficient funds, amounts below dust, and change/dust scenarios; expanded UTXO coverage.

### DIFF
--- a/src/wallet/onchain.ts
+++ b/src/wallet/onchain.ts
@@ -150,7 +150,10 @@ export class OnchainWallet implements AnchorBumper {
             );
 
             // Check stability of the selected inputs size against loop count
-            if (prevInputsCount === selected.inputs.length) {
+            if (
+                prevInputsCount === selected.inputs.length &&
+                newFee === estimatedFee
+            ) {
                 estimatedFee = Math.ceil(newFee);
 
                 // Compute the value of the selected inputs


src/wallet/onchain.ts
```diff
@@ -109,11 +109,51 @@ export class OnchainWallet implements AnchorBumper {
...
+        // Select coins to the exact send amount
+        const selectedWithoutFee = selectCoins(coins, params.amount);
+
+        // Add weight of each coin
+        for (const _ of selectedWithoutFee.inputs) {
+            txWeightEstimator.addKeySpendInput();
+        }
+
+        // Add weight of send amount output
+        txWeightEstimator.addP2TROutput();
```
louisinger: Output address may be any type of script. Here it admits that the output script is always a P2TR

I suggest you to add a new method to txWeightEstimator `addTxOutput(output)` computing the right size : `8 bytes (output value) + serialized varint size for the length of PkScript + PkScript bytes`.

shubertm:
@jgmcalpine am closing this in favor of #264. Thanks for the improvements :rocket:

PR #223, opened 2025-11-04
Identity decrypt

bordalix:
Adds support for nip04 decrypt in Identity.

It will be useful to decrypt stored messages on Nostr.

@tiero @louisinger do you think this makes sense?

## Summary by CodeRabbit

* **New Features**
  * Introduced public-key-based decryption functionality to the identity management system, allowing secure decryption of encrypted data through identity objects

* **Chores**
  * Version incremented to 0.3.4
  * Package manager updated to pnpm 10.20.0
  * Added new dependencies to support enhanced cryptographic operations

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
         "release:cleanup": "bash scripts/release.sh --cleanup"
     },
     "dependencies": {
+        "@noble/ciphers": "2.0.1",
         "@noble/curves": "2.0.0",
         "@noble/secp256k1": "3.0.0",
         "@scure/base": "2.0.0",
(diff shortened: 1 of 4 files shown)


louisinger (commented):
TBH not sure if I like it. It sounds something really specific to some "nostr implementation". If I'm not using nostr, I still need to implement this decrypt method, which is not used anywhere in the Wallet class 🤔 

EDIT: where and how the method will be used ?

bordalix:
Yeah, I also have mixed feelings about this.

In retrospect I don't like the fact we are able to get the private key from Identity so I was trying not to have another dependency of it by implementing the method inside the class.

We can do this entirely on the wallet, but it will need access to the private key from Identity, which I don't like.

PR #258, opened 2026-01-14
Modular service worker

### LINKED ISSUE #255: Modular service worker
### The problem
The current Service Worker implementation is wallet-specific: it only has access to `IWallet` and it is in charge of managing both the SW lifecycle and the wallet functionalities.

We can achieve modularity by **separating the Service Worker lifecycle logic from the feature-specific one** (such as wallet, contracts, ...) and even inject protocol-specific logic (lightning). This will enable background tasks like settling transactions or polling even when the page is not active.

### Proposed implementation

The Service Worker will implement message routing, lifecycle logic, and scheduling capabilities.

When the app instantiates a new Service Worker it will pass a list of  `IUpdater` where the interface looks like this:

```
export interface IUpdater<REQ,RES> {
    /** unique tag that identifies messages for this recipient **/
    readonly messageTag: string;

    /** Called once when the SW is starting */
    start(): Promise<void>;

    /** Called once when the SW is shutting down */
    stop(): Promise<void>;

    /** Called at arbitrary intervals by the Worker */
    tick(now: number): Promise<RES[]>;

    /** Handle routed messages */
    handleMessage(message: REQ): Promise<RES | null>;
}
```

An example of the usage is:

```
const walletUpdater = new WalletUpdater();
const boltzSwapUpdater = new BoltzSwapUpdater();
const sw = new ArkServiceWorker([walletUpdater, boltzSwapUpdater, /* ... and so on ... */ ]);
```

Additional features include:
- broadcast messages
- solid and extensible typings
- seamless integration in non-browser environments (ie: Node, React Native, ...)

pietro909:
Closes #255 

## Summary by CodeRabbit

* **New Features**
  * Introduced modular service worker architecture with explicit message routing and WalletUpdater-based wallet lifecycle management.
  * Added ArkSW service worker setup and initialization utilities for streamlined wallet integration.

* **Breaking Changes**
  * Service worker integration pattern changed from identity-driven to updater-based messaging flow.
  * Repository migration API updated from array parameter to object with `onchain` and `offchain` fields.

* **Tests**
  * Added comprehensive test coverage for service worker manager, wallet updater, and worker routing.

### DIFF
--- a/src/contracts/contractManager.ts
+++ b/src/contracts/contractManager.ts
@@ -634,22 +634,16 @@ export class ContractManager implements IContractManager {
             return new Map();
         }
 
-        return await this.fetchContractVxosFromIndexer(
-            contracts,
-            false,
-            this.config.extendVtxo
-        );
+        return await this.fetchContractVxosFromIndexer(contracts, false);
     }
 
     private async fetchContractVxosFromIndexer(
         contracts: Contract[],
-        includeSpent: boolean,
-        extendVtxo?: (vtxo: VirtualCoin) => ExtendedVirtualCoin
+        includeSpent: boolean
     ): Promise<Map<string, ContractVtxo[]>> {
         const fetched = await this.fetchContractVtxosBulk(
             contracts,
-            includeSpent,
-            extendVtxo
+            includeSpent
         );
         const result = new Map<string, ContractVtxo[]>();
         for (const [contractScript, vtxos] of fetched) {
@@ -667,8 +661,7 @@ export class ContractManager implements IContractManager {
 
     private async fetchContractVtxosBulk(
         contracts: Contract[],
-        includeSpent: boolean,
-        extendVtxo?: (vtxo: VirtualCoin) => ExtendedVirtualCoin
+        includeSpent: boolean
     ): Promise<Map<string, ContractVtxo[]>> {
         const result = new Map<string, ContractVtxo[]>();
 
@@ -676,8 +669,7 @@ export class ContractManager implements IContractManager {
             contracts.map(async (contract) => {
                 const vtxos = await this.fetchContractVtxosPaginated(
                     contract,
-                    includeSpent,
-                    extendVtxo
+                    includeSpent
                 );
                 result.set(contract.script, vtxos);
             })
@@ -707,9 +699,7 @@ export class ContractManager implements IContractManager {
             });
 
             for (const vtxo of vtxos) {
-                const ext = extendVtxo
-                    ? extendVtxo(vtxo)
-                    : (vtxo as ExtendedVirtualCoin);
+                const ext = this.config.extendVtxo(vtxo);
 
                 allVtxos.push({
                     ...ext,


src/wallet/serviceWorker/service-worker-manager.ts
```diff
@@ -0,0 +1,53 @@
+const registrations = new Map<string, Promise<ServiceWorkerRegistration>>();
+
+const ensureServiceWorkerSupport = () => {
+    if (!("serviceWorker" in navigator)) {
+        throw new Error("Service workers are not supported in this browser");
+    }
+};
```
louisinger:
```suggestion
function ensureServiceWorkerSupport() {
    if (!("serviceWorker" in navigator)) {
        throw new Error("Service workers are not supported in this browser");
    }
};
```
nit: anonymous function can make debugging harder because they do not have "name" so no trace in the stack when an error is thrown. maybe made on purpose ?

src/wallet/serviceWorker/worker.ts
```diff
@@ -1,33 +1,256 @@
...
+> {
+    readonly messageTag: string;
+
+    start(): Promise<void>;
+
+    /** Called once when the SW is shutting down */
+    stop(): Promise<void>;
+
+    /** Called periodically by the Worker */
+    tick(now: number): Promise<RES[]>;
```
louisinger: not clear to me why we need the tick, maybe worth commenting that method explaining why/how ?

louisinger (commented):
I like the abstraction and the modularity. maybe it's missing a README.md in `wallet/serviceWorker` folder describing the architecture ? not in details but the main concepts (tick, registration etc..) ? it's also useful to serve as "LLM doc", wdyt ?

src/wallet/serviceWorker/wallet.ts
```diff
@@ -1,538 +1,747 @@
...
+    | ResponseClear
+    | ResponseReloadWallet
+    | ResponseUtxoUpdate
+    | ResponseVtxoUpdate
+    | ResponseSignTransaction;
+
+export class WalletUpdater
+    implements IUpdater<WalletUpdaterRequest, WalletUpdaterResponse>
+{
+    static messageTag = "WalletUpdater";
```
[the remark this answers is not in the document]
pietro909: That's correct.
Some data collections like `tx`, `vtxo`, `utxo` have `address` in the keypath and will be consistently separated.
Contracts use `script` which is derived from params, possibly generating collisions among different wallets.
Swaps use provider's Ids which are globally unique in the lifecycle of the swap but not sure in the long term.

src/serviceWorker/worker.ts
```diff
@@ -1,33 +1,256 @@
...
+> {
+    readonly messageTag: string;
+
+    start(): Promise<void>;
+
+    /** Called once when the SW is shutting down */
+    stop(): Promise<void>;
+
+    /** Called periodically by the Worker */
+    tick(now: number): Promise<RES[]>;
```
pietro909: @louisinger I've added a dedicated README https://github.com/arkade-os/ts-sdk/pull/258/changes/b2de23dc695e0f766c812710f8105b7e999e8aaa outlining the architectural design

PR #262, opened 2026-01-16
Delegator client

louisinger:
This PR implements the delegator client for the new delegate service implements by fulmine https://github.com/ArkLabsHQ/fulmine/pull/361

The `Wallet` class accepts an optional `DelegatorProvider`. If set, delegator service will be included as extra forfeit tapscript in the offchain ark address. `Wallet.delegate` method can be used to delegate the renewal of a given set of vtxos. e2e test provide a full exemple of a wallet using delegator service.

@tiero @pietro909 @bordalix @Kukks please review

## Summary by CodeRabbit

* **New Features**
  * VTXO delegation: wallet APIs, delegation manager, delegated VTXO scripts, service-worker delegation flow and REST delegator provider; service worker now accepts a delegator URL.

* **Infrastructure**
  * Local Fulmine service added to compose with a multi-stage runtime image; network config updated for external usage.

* **Tests**
  * E2E suite expanded for delegation scenarios; test setup moved to .mjs and a new regtest script added.


src/providers/delegator.ts
```diff
@@ -0,0 +1,71 @@
...
+    pubkey: string;
+    fee: string;
+    delegatorAddress: string;
+}
+
+export interface DelegatorProvider {
+    delegate(
+        intent: SignedIntent<Intent.RegisterMessage>,
+        forfeits: string[]
+    ): Promise<void>;
```
tiero: It's not really helpful for the user if it has to come up with all the parameters at their final state already computed as parameters.

I would expect to initialize a class over-time with a Signer/Identity, a VTXO set, a mapping script template / script paramenters 

It's in the end we need to accomodate all kinds of variants, our class serve as sugar syntax for well known script collections and tx workflows

tiero (changes requested):
My idea is to treat delegates as custom script, rather than a default script with optionality, also because creates an hard dependency to an unknown/spontaneous third party we can't  control.

We can generalize then that a custom script becomes the  "main" script template via an opt-in config of the Wallet instance so that getAdddress (and in the future change address) could be encoded from such a script template (and filled of derived keys when needed).

That's would differ from one-off contracts (HTLC,Escrow...) that aren't spenadable / outgoing 

We assume already a persistent and durable storage for script template / derivation mappings so restore should be already solved. 

nit: I would not run a e2e with real fulmine, rather than mocking these behaviors as in the end are side effects. The SDK test should focus on the restore and encoding, let's mock server-side interaction around registration of scripts and actual delegated intents.

src/wallet/wallet.ts
```diff
@@ -204,20 +213,25 @@ export class ReadonlyWallet implements IReadonlyWallet {
...
         // Generate tapscripts for offchain and boarding address
         const serverPubKey = hex.decode(info.signerPubkey).slice(1);
-        const bareVtxoTapscript = new DefaultVtxo.Script({
-            pubKey: pubkey,
+
+        const delegatePubKey = config.delegatorProvider
+            ? await config.delegatorProvider
+                  .getDelegateInfo()
+                  .then((info) => hex.decode(info.pubkey).slice(1))
+            : undefined;
```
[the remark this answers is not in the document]
louisinger: we can safely accept info.pubkey is 33 bytes

src/providers/utils.ts
```diff
@@ -15,8 +15,9 @@ export async function* eventSourceIterator(
         }
     };
 
-    const errorHandler = () => {
-        const error = new Error("EventSource error");
+    const errorHandler: EventListenerOrEventListenerObject = (event: Event) => {
+        const error = new Error(JSON.stringify(event));
```
pietro909: `event` could be syntetized byt a 3rd party polyfill with unpredictable results. For example, a circular structure will make `JSON.stringify` throw.

It also leads to the creation of an error with a potentially very long string.

If we are interested in preserving the exact origin of the error, a safer approach is `new Error("whatever", { cause: event })` but requires an upgrade of our compile target (> 2025 is think).

Otherwise you may want to extract `event.type` and use it in the error message.

src/intent/index.ts
```diff
@@ -46,15 +50,16 @@ export namespace Intent {
      */
     export function create(
         message: string | Message,
-        inputs: TransactionInput[],
+        ins: (TransactionInput | ExtendedCoin)[],
```
Kukks: why rename?

PR #268, opened 2026-01-28
Update schema and migration for contract repository, remove deprecated methods

### LINKED ISSUE #267: Remove deprecated methods from ContractRepository and fix the migration
Some methods on ContractRepository are specific to `boltz-swap` integration:

```typescript
    getContractData<T>(contractId: string, key: string): Promise<T | null>;
    setContractData<T>(contractId: string, key: string, data: T): Promise<void>;
    clearContractData(): Promise<void>;
    getContractCollection<T>(contractType: string): Promise<ReadonlyArray<T>>;
    saveToContractCollection<T, K extends keyof T>(
        contractType: string,
        item: T,
        idField: K
    ): Promise<void>;
    removeFromContractCollection<T, K extends keyof T>(
        contractType: string,
        id: T[K],
        idField: K
    ): Promise<void>;
```

We're deprecating them but need to remove all before `next` goes live. Only Contracts are to be persisted, the provider-specific information will be handled in a separate repository managed by the package which owns the integration.

While at it, update the DB migration: 

- [x] move `commitmentTxs`  to the WalletRepository (they aren't contracts)
- [x] transform `submarineSwaps` and `reverseSwaps` into Contract type, put the provider-specific data into `metadata`  - addressed in https://github.com/arkade-os/boltz-swap/issues/56
- [x] in `boltz-swap` migrate the two contract types accordingly - addressed in https://github.com/arkade-os/boltz-swap/issues/56

pietro909:
Closes #267 

- Remove all deprecated methods from `src/repositories/contractRepository.ts` minus the two still used for migrating data `saveToContractCollection` and `getContractCollection`
- introduce `commitmentTxs` store in WalletStore and migrate existing data from the legacy ContractRepository
- improve contracts schema in `src/repositories/indexedDB/schema.ts` with indexes for `type`,`state`,`script` and leverage them for filtering
- update tests in `test/storage.test.ts` and extract the ones for ContractRepository to `test/repositories/contractRepository.test.ts` (the file was renamed, leading to a larger diff than I wanted)


pietro909:
@coderabbitai review please with particular attention to the DB schema changes and the migration. It's imperative that no data loss ever occur.

louisinger (approved):
just a nit comment but LGTM

PR #269, opened 2026-01-30
feat: add SeedIdentity for HD wallet derivation

Kukks:
## Summary by CodeRabbit

* **New Features**
  * Add mnemonic- and seed-based identities (BIP39/BIP86), optional passphrase, HD derivation, watch-only descriptor flows, and public API exposure.

* **Documentation**
  * Expanded README with mnemonic/seed workflows, descriptor/watch-only guidance, examples for mnemonic, seed, and readonly flows.

* **Tests**
  * Extensive unit and e2e tests covering identity creation, signing, serialization, readonly flows, and wallet variants.

* **Chores**
  * Added libraries to support descriptors and BIP32/BIP39; updated package metadata.

* **Style**
  * Added .claude to .gitignore.

### DIFF
--- a/src/identity/seedIdentity.ts
+++ b/src/identity/seedIdentity.ts
@@ -230,18 +230,8 @@ export class SeedIdentity implements Identity {
  * ```
  */
 export class MnemonicIdentity extends SeedIdentity {
-    private readonly mnemonic: string;
-    private readonly passphrase?: string;
-
-    private constructor(
-        seed: Uint8Array,
-        descriptor: string,
-        mnemonic: string,
-        passphrase?: string
-    ) {
+    private constructor(seed: Uint8Array, descriptor: string) {
         super(seed, descriptor);
-        this.mnemonic = mnemonic;
-        this.passphrase = passphrase;
     }
 
     /**
@@ -265,7 +255,7 @@ export class MnemonicIdentity extends SeedIdentity {
         const descriptor = hasDescriptor(opts)
             ? opts.descriptor
             : buildDescriptor(seed, (opts as NetworkOptions).isMainnet);
-        return new MnemonicIdentity(seed, descriptor, phrase, passphrase);
+        return new MnemonicIdentity(seed, descriptor);
     }
 }
 


src/identity/seedIdentity.ts
```diff
@@ -0,0 +1,464 @@
...
+    /**
+     * Restores a SeedIdentity from a JSON string.
+     *
+     * The JSON must contain either a `mnemonic` or `seed` field, plus a `descriptor`
+     * field. The network (mainnet/testnet) is inferred from the coin type in the
+     * descriptor path. The xpub in the descriptor is validated against the derived key.
+     *
+     * @param json - JSON string from toJSON()
+     * @returns Restored SeedIdentity instance
+     * @throws Error if JSON is invalid, missing required fields, or xpub doesn't match
```
pietro909: If I understand it correctly, [the regexp](https://github.com/arkade-os/ts-sdk/pull/269/changes#diff-795ecb1c45bb4240841b53ccee4d206e3a8737cc28b7125d028493db09721d5bR49) only supports `/0/*` suffix. 

The function will throw with a `/1/*` suffix which is BIP86-compliant. If that's the case, it should be indicated in the docs.

I may be wrong here though 🙏

src/identity/seedIdentity.ts
```diff
@@ -0,0 +1,449 @@
...
+ *   'abandon abandon abandon ...',
+ *   { isMainnet: true, passphrase: 'secret' }
+ * );
+ *
+ * // toJSON preserves both mnemonic and passphrase
+ * const json = identity.toJSON();
+ * const restored = MnemonicIdentity.fromJSON(json);
+ * ```
+ */
+export class MnemonicIdentity extends SeedIdentity {
```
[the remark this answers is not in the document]
Kukks: I think i have an example of this in the readme. I'd rather leave entropy to other sources :)

src/identity/seedIdentity.ts
```diff
@@ -0,0 +1,449 @@
...
+ * const pubKey = await readonly.xOnlyPublicKey();
+ * ```
+ */
+export class ReadonlyDescriptorIdentity implements ReadonlyIdentity {
+    private readonly xOnlyPubKey: Uint8Array;
+    private readonly compressedPubKey: Uint8Array;
+    readonly descriptor: string;
+
+    private constructor(descriptor: string) {
+        this.descriptor = descriptor;
```
louisinger:
```suggestion
    private constructor(readonly descriptor: string) {
```

louisinger (changes requested):
I'd like to see e2e tests using this mnemonic identity. We could do like [delegator PR ](https://github.com/arkade-os/ts-sdk/pull/262) (let's wait we merge it so simpler) and use a "matrix-like" test (same suite runs with delegator, without delegator, with mnmenonic identity...) ?

PR #276, opened 2026-02-02
Refactor DB management

pietro909:
Introducing [repositories v2 to boltz-swap](https://github.com/arkade-os/boltz-swap/pull/57) requires some boilerplate related to IndexedDB management.
Rather than duplicate the same logic, I've made the `src/db` utils first-class citizens and exposed them.

Any class following our repository pattern will then implement its own DB initialization by passing the name, version, and a migration function:
```
    private async getDB(): Promise<IDBDatabase> {
        if (this.db) return this.db;
        this.db = await openDatabase(this.dbName, 2, initDatabase);
        return this.db;
    }
```

Things like reusing the connection, closing the DB once all consumers are done, and triggering migrations when `onupgradeneeded` event is fired (and queuing the migrations) is handled in one place.

## Summary by CodeRabbit

* **New Features**
  * Enhanced database lifecycle management with reference counting and caching for improved stability.
  * Added wallet state persistence capability (saveWalletState).
  * Expanded public API surface with database utilities and contract manager interface.

* **Chores**
  * Simplified build process by replacing Husky with automatic build step during preparation.
  * Removed unused collection management methods and cleaned up internal dependencies.

### DIFF
--- a/src/repositories/migrations/fromStorageAdapter.ts
+++ b/src/repositories/migrations/fromStorageAdapter.ts
@@ -25,10 +25,18 @@ const requiresMigration = async (
     }
 };
 
+/**
+ * Migrate wallet data from the legacy storage adapter to the new one.
+ * It accepts both onchain and offchain addresses, make sure to pass both.
+ *
+ * @param storageAdapter
+ * @param fresh
+ * @param addresses
+ */
 export async function migrateWalletRepository(
     storageAdapter: StorageAdapter,
     fresh: WalletRepository,
-    addresses: string[]
+    addresses: { onchain: string[]; offchain: string[] }
 ): Promise<void> {
     const migrate = await requiresMigration("wallet", storageAdapter);
     if (!migrate) return;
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -260,12 +266,18 @@ export {
...
+    closeDatabase,
+    openDatabase,
+
     // Repositories
     IndexedDBWalletRepository,
     IndexedDBContractRepository,
     InMemoryWalletRepository,
     InMemoryContractRepository,
     migrateWalletRepository,
+    WalletRepositoryImpl,
```
Kukks: why export these? might be a footgun as people may try to use them?

package.json
```diff
@@ -79,7 +79,7 @@
         "lint": "prettier --check src test examples",
         "audit": "pnpm audit",
         "preinstall": "npx only-allow pnpm",
-        "prepare": "husky",
+        "prepare": "pnpm run build",
```
Kukks: ?

Kukks (approved):

PR #281, opened 2026-02-06
Fix Contract Manager initialization

pietro909:
While testing with the Wallet app I found two issues with the initialization process:
- side effects in ContractRegistry don't run if not imported explicitly - I've marked that file as impure to prevent tree-shaking
- type cast `vtxo as ExtendedVirtualCoin` was breaking at runtime because `vtxo.tapTree` is undefined - fixed by always enriching the `VirtualCoin` with the default function passed in the constructor
- in ContractWatcher, waiting for `listenLoop` causes the initialization to freeze the Service Worker - removed the `await` but maintained error handling

### DIFF
--- a/src/contracts/contractManager.ts
+++ b/src/contracts/contractManager.ts
@@ -633,22 +633,16 @@ export class ContractManager implements IContractManager {
             return new Map();
         }
 
-        return await this.fetchContractVxosFromIndexer(
-            contracts,
-            false,
-            this.config.extendVtxo
-        );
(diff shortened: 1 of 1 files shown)


src/contracts/contractWatcher.ts
```diff
@@ -380,7 +380,19 @@ export class ContractWatcher {
             this.reconnectAttempts = 0;
 
             // Start listening
-            await this.listenLoop();
+            this.listenLoop().catch((e) => {
```
Kukks: Do we have any form of cleanup if we dispose the watcher?

PR #299, opened 2026-02-18
Update db after transaction

louisinger:
This PR makes sure the database is updated after commitment tx or offchain tx.
It aims to fix "double balance" bug on arkade wallet. Tested locally on regtest.

@pietro909 @Kukks please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction matching so previously stored transactions are correctly identified and updated only when all identifying fields match, reducing duplicate entries.

* **Refactor**
  * Centralized and consolidated off-chain transaction and settlement persistence to ensure spent/change outputs and transaction records are reliably saved and state is consistent after sends and settles.

### DIFF
--- a/src/wallet/wallet.ts
+++ b/src/wallet/wallet.ts
@@ -1753,6 +1753,7 @@ export class Wallet extends ReadonlyWallet implements IWallet {
 
         // build change receiver with BTC change and all asset changes
         let changeReceiver: Recipient | undefined;
+        let changeIndex = 0;
         if (changeAmount > 0) {
             const changeAssets: Asset[] = [];
             for (const [assetId, amount] of assetChanges) {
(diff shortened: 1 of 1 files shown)


src/wallet/wallet.ts
```diff
@@ -1908,34 +1848,151 @@ export class Wallet extends ReadonlyWallet implements IWallet {
...
-            inputs.push({
-                txid: hex.decode(input.txid),
-                index: input.vout,
-                witnessUtxo: {
-                    amount: BigInt(input.value),
-                    script: vtxoScript.pkScript,
+            await this.walletRepository.saveVtxos(addr, spentVtxos);
+
+            if (changeVtxo) {
+                await this.walletRepository.saveVtxos(addr, [changeVtxo]);
```
pietro909: If this doesn't have to happen after `this.walletRepository.saveVtxos(addr, spentVtxos);` I'd rather move it into the `if` at line 1922 and reduce complexity (one less variable, one less branch).

PR #305, opened 2026-02-20
Support asset in sendBitcoin

louisinger:
This PR makes `sendBitcoin` using `send` method (supporting asset).

however, because we have an option to spend specific vtxo in sendBitcoin, we need to keep the current implementation if params.selectedVtxos is set. 

@Kukks please review

## Summary by CodeRabbit

* **Refactor**
  * Improved Wallet Bitcoin sending logic to better handle explicit virtual coin selection, resulting in more reliable transaction construction and submission when selecting specific inputs.


Kukks (approved):
We can extend send to also support explicit coin support

PR #319, opened 2026-02-28
Arkade Script support

louisinger:
Adds full client-side support for https://github.com/arkade-os/emulator and ArkadeScript
                                                                                                                                                                                                                            
###   What's included                                                                                                                                                                                                           
                                                                                                                                                                                                                            
  - Arkade opcodes & script codec — encode/decode Arkade extension opcodes (0xc4–0xf2) alongside standard Bitcoin opcodes, with ASM conversion helpers                                                                      
  - Script tweaking — compute introspector-bound public keys via tagged hash (ArkScriptHash) point addition
  - ArkadeVtxoScript — extends VtxoScript to support arkade-enhanced tapscript leaves with automatic key tweaking
  - Introspector provider — REST client for the introspector co-signing service (info, intent submission, finalization)
  - Batch handler — createArkadeBatchHandler handles the full boarding + settlement flow with introspector co-signing
  - PSBT fields — new ArkadeScript and ArkadeScriptWitness unknown fields, plus stricter key matching in existing field decoders
  - Docker setup — docker-compose.yml and introspector.Dockerfile for local development

  Tests

  - Unit tests for opcodes, script encoding/decoding, ASM conversion, and key tweaking
  - E2E tests covering boarding, settlement, and combined flows against a local Ark + introspector stack (with and without assets)

@Kukks @tiero please review

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Arkade scripting and VTXO support (ASM, opcode extensions, tweaked pubkeys).
  * Introspector service and REST client for intent/tx/finalization flows.
  * Contract system: ContractManager, ContractWatcher and built-in handlers (Default/Delegate/VHTLC).
  * Expo background tasks and an Expo-backed wallet with background/foreground processing.
  * MessageBus/service-worker wallet updater for typed wallet messaging.

* **Repositories / Storage**
  * New repository implementations: IndexedDB, SQLite, Realm, and in-memory; migration helpers and storage config.

* **Tests**
  * Extensive unit and end-to-end test coverage for Arkade, contracts, and flows.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


src/arkade/vtxoScript.ts
```diff
@@ -0,0 +1,152 @@
...
+    CLTVMultisigTapscript,
+    type ArkTapscript,
+} from "../script/tapscript";
+import { VtxoScript } from "../script/base";
+import { computeArkadeScriptPublicKey } from "./tweak";
+
+export type ArkadeLeaf = {
+    arkadeScript: Uint8Array;
+    tapscript: ArkTapscript<TapscriptType, any>;
+};
```
louisinger:
```suggestion
export type ArkadeLeaf = {
    arkadeScript: Uint8Array;
    tapscript: ArkTapscript<TapscriptType, any>;
    introspectors: UInt8Array[];
};
```

ghost (approved):
## 🔍 Arkana PR Review — arkade-os/ts-sdk#319 (incremental: `1673ce4d`)

*Review by Arkana 🔍 — [Agent: pr-reviewer]*

**Scope since last review (`8eb98de`):** 1 functional commit + 1 merge from master.

### `1673ce4d` — fix minimal data number encoding (`src/arkade/script.ts:62-73`)

This directly addresses the MINIMALDATA concern I flagged in my previous review. The fix:

- Merges the `number` and `bigint` branches into a single `typeof o === "number" || typeof o === "bigint"` check
- Adds OP_0 handling for `0n` (`o === 0 || o === 0n`) — uses strict equality, correct since `0n === 0` is `false` in JS
- Extends OP_1–OP_16 small-integer encoding to bigint values via `Number(o)` coercion after the range check
- JS comparison operators (`>=`, `<=`) support mixed bigint/number operands, so `1n >= 1` and `16n <= 16` work correctly
- Negative numbers (`-1`, `-1n`) correctly fall through to BigNum encoding

The comment about Arkade VM enforcing MINIMALDATA is a good addition for future maintainers.

**One note:** No unit test was added for bigint MINIMALDATA encoding (e.g., `ArkadeScript.encode([0n, 1n, 16n, 17n])` round-trip). Consider adding one to prevent regression.

### Master merge (`c1b01a86`)

Two PR-touched files were modified by the merge:
- `src/extension/index.ts`: import path `.js` extension fix — trivial, correct
- `src/utils/unknownFields.ts`: `bip68.decode` → centralized `sequenceToTimelock` helper — clean refactor from master, no semantic change

No conflicts. No issues.

**Approving** — the MINIMALDATA fix is correct and addresses the last open concern.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

ghost (approved):
## 🔍 Arkana PR Review — arkade-os/ts-sdk#319 (incremental: `bf176026`)

*Review by Arkana 🔍 — [Agent: pr-reviewer]*

**Scope since last review (`1673ce4d`):** 1 merge from master + 1 fix commit.

### `bf176026` — fix runTask import (`src/worker/expo/index.ts`)

Fixes a real bug I flagged in my first review (concern #8). `runTasks` is an `async function` (value export), not a type — placing it inside `export type { ... }` caused TypeScript to erase it, making it `undefined` at runtime for any consumer importing from the barrel.

The fix:
- Removes `runTasks` from the `export type { ... }` block
- Adds `export { runTasks, createTaskDependencies }` as a value re-export
- `createTaskDependencies` was also missing from the barrel — good catch including it

Verified: both `runTasks` (line 54) and `createTaskDependencies` (line 125) in `taskRunner.ts` are runtime function exports. The type-only re-exports (`TaskProcessor`, `TaskDependencies`, `CreateTaskDependenciesOptions`) remain correctly in the `export type` block.

No protocol-critical code touched. **Approving.**

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Kukks:
Cleaned-up replay in #532 — same arkade/introspector content, replayed onto current master and sequenced by topic (committed by @louisinger via `--author`).

What's different from this PR:
- Drops the committed `dist/` artefacts (~60k LOC) — the un-ignore of `dist` in `packages/ts-sdk/.gitignore` is also gone. Consumers who need a prebuilt for git-based installs should use `prepare`.
- Skips reverts that would undo recent master fixes/perf this branch was unaware of: #511 (delegate-info validation), #512 (`renewVtxos` per-call threshold), #520 (delegate normalisation), #527 (Realm `create` mode), #529/#519 (per-contract tapscript memoisation + clone), #530/#531 (boarding-receive dedup).
- Drops the unrelated boltz-swap refactor and the test renames (`delegate*` → `delegator*`) — those weren't part of the arkade scope.

Leaving this PR open for reference and so you can pick it up if there's anything in #532 that doesn't capture your intent. Happy to close once you've had a look.

tiero (commented):
Why dist folder is committed?

packages/ts-sdk/test/e2e/arkade-onchain.test.ts
```diff
@@ -0,0 +1,297 @@
...
+
+        setArkPsbtField(tx, 0, PrevoutTxField, opts.rawFundingTx);
+
+        if (opts.arkadeScript) {
+            addEmulatorPacket(tx, [
+                {
+                    vin: 0,
+                    script: opts.arkadeScript,
+                    // single empty push for output_index = 0
+                    witness: new Uint8Array([0x01, 0x00]),
```
tiero: can you explain? Seems to be good opportunity to abstract it in a way where we don't make developer aware of the packet and tweaking?

packages/ts-sdk/test/e2e/arkade-onchain.test.ts
```diff
@@ -0,0 +1,297 @@
...
+        const aliceX = await alice.xOnlyPublicKey();
+        const bobP2TR = p2tr(bobX, undefined, networks.regtest).script;
+
+        const arkadeScript = enforcePayTo(bobP2TR, SPEND_AMOUNT);
+        const tweakedIntro = arkade.computeArkadeScriptPublicKey(emulatorPubkey, arkadeScript);
+
+        const vtxoScript = new arkade.ArkadeVtxoScript([
+            {
+                arkadeScript,
+                emulators: [emulatorPubkey],
```
tiero: why plural?

packages/ts-sdk/src/arkade/tweak.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @param pubKey - The emulator's base public key (32 or 33 bytes)
+ * @param script - The raw Arkade script bytes
+ * @returns 32-byte x-only tweaked public key
+ */
+export function computeArkadeScriptPublicKey(pubKey: Uint8Array, script: Uint8Array): Uint8Array {
+    const hash = arkadeScriptHash(script);
+
+    const xOnly = pubKey.length === 33 ? pubKey.subarray(1) : pubKey;
+    const point = secp256k1.Point.fromHex("02" + hex.encode(xOnly));
```
pietro909: This forces even-Y on a real compressed key: do we know if the emulator does exactly the same thing? if yes, we need a unit test and a comment to lock-in the behavior.

packages/ts-sdk/src/arkade/batch.ts
```diff
@@ -0,0 +1,254 @@
...
+
+                if (boardingIdx !== null) {
+                    // Boarding: sign commitment tx directly
+                    commitmentPsbt.updateInput(boardingIdx, {
+                        tapLeafScript: [input.forfeitTapLeafScript],
+                    });
+
+                    boardingIndices.push(boardingIdx);
+                    hasBoardingInputs = true;
+                } else {
```
pietro909: Recoverable and subdust VTXOs would end up here without a filter and likely cause a failure? Or is this expected?

Shall we filter them out?

```
if (isVtxo(input) && (isRecoverable(input) || isSubdust(input, dustAmount))) continue;
```

packages/ts-sdk/src/arkade/tweak.ts
louisinger: emulator is handling every cases while signing. from client point of view, it doesn't matter. here it's only a trick to make `secp256k1.Point.fromHex` succeeds in whatever the pubkey length.

pietro909: fine, a comment would be helpful so that we don't get someone to ask the same question in one month :smile:

pietro909 (approved):

ghost (changes requested):
## 🔍 Arkana PR Review — arkade-os/ts-sdk#319 (incremental: `26804ece`)
*Review by Arkana 🔍 — [Agent: pr-reviewer]*

This is an incremental review covering changes since my last review (`bf176026`). The PR has grown substantially — monorepo restructure, new `boltz-swap` package, contract handlers, signer rotation, and Arkade script high-level APIs.

**⚠️ PROTOCOL-CRITICAL: This PR touches VTXO management, transaction signing, forfeit paths, signer rotation, key derivation, MuSig2, and VHTLC construction. Human review required before merge.**

---

### 🔴 HIGH — Boarding migration settle passes gross amount without fee deduction

`packages/ts-sdk/src/wallet/vtxo-manager.ts:1725-1735`

The boarding migration leg builds a settle intent with `totalAmount = sum(input values)` as the output amount, without deducting intent fees (`evalOnchainInput` / `evalOffchainOutput`). The periodic settle path (lines 2625-2709) correctly uses an `Estimator` to deduct fees. On fee-charging servers, the boarding migration settle will fail with `INTENT_INSUFFICIENT_FEE`, making cooperative signer migration of boarding UTXOs silently broken. Funds are not lost (unilateral exit remains), but the auto-migration feature is dead on any server with non-zero fees.

Same pattern in `recoverVtxos()` (line 1098-1112) and `renewVtxos()` (line 1372-1383).

---

### 🔴 HIGH — `VtxoScript.encode()` hardcodes `depth: 1` for all leaves

`packages/ts-sdk/src/script/base.ts:94-100`

```typescript
encode(): Bytes {
    const tapTree = TapTreeCoder.encode(
        this.scripts.map((script) => ({
            depth: 1,    // <-- always 1 regardless of actual tree depth
            ...
```

The constructor now uses `assembleBtcdTaprootTree` (which produces varying depths for non-power-of-2 leaf counts), but `encode()` still writes every leaf at `depth: 1`. The internal round-trip (`encode → decode → reconstruct`) is safe because decode discards depth, but this encoded TapTree is structurally wrong in the PSBT. External consumers (hardware wallets, other PSBT signers) that interpret the TapTree field literally will derive wrong merkle paths and fail to sign.

---

### 🟡 MEDIUM — URL path injection across providers

`packages/ts-sdk/src/providers/indexer.ts:307, 338, 360, 373, 390, 421, 446, 525, 547`
`packages/ts-sdk/src/providers/onchain.ts:156, 170, 188, 199, 212`

User-supplied values (`txid`, `address`, `subscriptionId`) are interpolated directly into URL paths without sanitization. A malformed txid containing `/` or `..` could redirect requests to arbitrary server endpoints. Note that `assetId` at indexer.ts:649 *does* use `encodeURIComponent`, showing this was considered for one case but not applied consistently. The SDK should not rely on server-side validation for path safety.

---

### 🟡 MEDIUM — `signCooperativeClaimForServer` signs an opaque server-provided hash

`packages/boltz-swap/src/arkade-swaps.ts:2238-2256`

Signs `claimDetails.transactionHash` (a hash from Boltz) using MuSig2 without verifying what transaction the hash corresponds to. The ephemeral key scoping limits blast radius (it's the lockup-side key, not the wallet key), but signing opaque hashes from a counterparty is a security smell. At minimum, add a comment explaining why this is safe (Boltz already controls the refund path after timelock, so worst case is they claim their own funds earlier).

---

### 🟡 MEDIUM — `contract.ts` accepts outputs with `amount: undefined`

`packages/ts-sdk/src/arkade/contract.ts:617`

```typescript
const outputsSum = this.outputs.reduce((s, o) => s + (o.amount ?? 0n), 0n);
```

An output with `amount: undefined` counts as 0 sats in the surplus check but is still included in the transaction. Should validate that all outputs have explicit amounts before building.

---

### 🟡 MEDIUM — Boarding handler doesn't probe deprecated signer keys

`packages/ts-sdk/src/contracts/handlers/boarding.ts:129-145`

Unlike `default` and `delegate` handlers which iterate over `deps.deprecatedSignerPubKeys`, the boarding handler only probes `deps.serverPubKey`. An unspent boarding UTXO anchored to a deprecated signer key becomes undiscoverable during wallet restore, until the user manually provides the old key. This is a gap in the signer rotation story for boarding UTXOs.

---

### 🟡 MEDIUM — Emulator provider doesn't validate `signerPubkey` format

`packages/ts-sdk/src/providers/emulator.ts:67-68`

`getInfo()` validates `signerPubkey` is a non-empty string but doesn't verify it's valid hex encoding a valid public key. A malformed key would produce a tweaked key that nobody can sign for — locking funds.

---

### 🟢 LOW — Duplicate `createVHTLCScript` implementations

`packages/boltz-swap/src/utils/scripts.ts` vs `packages/boltz-swap/src/utils/vhtlc.ts`

Near-duplicate implementations of `createVHTLCScript`, `VhtlcTimeouts`, and `toBip68RelativeTimelock`. A fix applied to one will be missed in the other. Should be deduplicated.

---

### 🟢 LOW — `isMetadata` type guard doesn't validate values

`packages/ts-sdk/src/providers/errors.ts:64-66`

Asserts `Record<string, string>` but only checks it's a non-null, non-array object. Values could be numbers, nested objects, etc. Downstream code treating values as strings would get silent coercion.

---

### 🟢 LOW — Misleading BIP comment in batch.ts

`packages/ts-sdk/src/arkade/batch.ts:82`

Comment references BIP-65 (CLTV / absolute timelocks) but context is CSV (BIP-68 / relative timelocks). The 512 threshold is correct for BIP-68. Fix the comment.

---

### 🟢 LOW — Dead field `arkadeScriptBytes` on `ArkadeExtendedCoin`

`packages/ts-sdk/src/arkade/batch.ts:40`

`arkadeScriptBytes` is required by the type but never read anywhere. Dead code in a protocol-critical type.

---

### 🟢 LOW — No nonce disposal after MuSig session

`packages/boltz-swap/src/utils/musig.ts`

After `signPartial()` produces a `MusigSigned`, the secret nonce remains in memory with no explicit zeroing. The MuSig2 nonce generation relies on `@scure/btc-signer` internally adding `randomBytes(32)` — a correct but undocumented assumption with catastrophic consequences (key leak) if the library removes that randomness. Add a comment documenting this safety dependency.

---

### ℹ️ INFO — Ephemeral key + preimage stored plaintext together

`packages/boltz-swap/src/arkade-swaps.ts:2357-2358`

Both `ephemeralKey` (raw private key) and `preimage` are stored as hex in the same `BoltzChainSwap` row in IndexedDB. An XSS attack reading the swap repository would get both, enabling funds theft. Inherent to browser wallet design but worth noting.

---

### ℹ️ INFO — Master seed serialization across postMessage

`packages/ts-sdk/src/identity/seedIdentity.ts:528-558`

The full 64-byte master seed (or mnemonic+passphrase) is serialized as a plain object for `postMessage` to service workers. The code comments acknowledge this tradeoff. Confirming reviewer awareness: any XSS on the message channel gets the master seed, not just a derived key.

---

### ✅ Verified Correct

- **Key derivation** (`seedIdentity.ts`): Seed length validation, descriptor-xpub cross-check, defensive copies — all correct.
- **Forfeit path construction** (`forfeit.ts`): Amount summation, P2A anchor, v3 transaction — correct.
- **Taproot tree assembly** (`taprootTree.ts`): Correctly implements btcd's `AssembleTaprootScriptTree`.
- **Tree validation** (`validation.ts`): Parent-child linkage, amount conservation, cosigner key checks — correct.
- **PSBT unknown fields** (`unknownFields.ts`): Byte-level key matching and robust decode — correct.
- **Signing sessions** (`signingSession.ts`): 1-input invariant enforced by tree validation, signing loop is safe.
- **VHTLC/delegate/default handlers**: Script construction, role resolution, and forfeit paths are correct.
- **MuSig2 partial sig verification**: Properly validates before aggregation.

---

**Requesting changes** for the boarding migration fee gap (HIGH) and TapTree depth encoding (HIGH). The remaining items are non-blocking but should be addressed.

*— Arkana*

ghost (approved):
## 🔍 Arkana PR Review — arkade-os/ts-sdk#319 (incremental: `98add036`)
*Review by Arkana 🔍 — [Agent: pr-reviewer]*

**Scope since last review (`26804ece`):** boltz-swap optimistic settlement (`waitFor: "funded"`), test sequencer, version bumps, emulator v0.0.3.

### Verdict: ✅ Approve

This is a clean, well-designed feature. The concurrency model is correct and the test coverage is thorough. No protocol-critical paths (VTXO, signing, forfeit, exit) are touched — this is purely swap lifecycle UX.

### What I checked

**1. `waitForSubmarineSwap` state machine** (`arkade-swaps.ts`)
- Two-flag design (`isFinal` / `isSettled`) is correct: `isFinal` gates further processing, `isSettled` gates promise resolution. Optimistic resolution sets only `isSettled`, so the monitor continues persisting updates (preimage on claim, refundable on failure). Terminal handlers set both. No double-resolution risk — JS is single-threaded and flags are set synchronously before any `await`.

**2. `hasSubmarineStatusReached`** (`boltz-swap-provider.ts`)
- Progression order matches the Boltz API. Failure statuses correctly return `false` (not in the array → `indexOf` returns -1). The "at or beyond" semantics handle skipped statuses (0-conf, subscription gaps).

**3. Error handling hardened**
- `saveStatus` now wraps persistence in try/catch — a storage failure no longer leaves the promise dangling. The domain error (swap expired, invoice failed, etc.) always wins.
- `transaction.claimed` handler: preimage fetch failure rejects instead of hanging. Status isn't persisted in that case, but `refreshSwapsStatus` recovers it on next startup. Acceptable trade-off.
- `monitorSwap` callback errors are caught and logged, preventing unhandled rejections after optimistic resolution.

**4. `refreshSwapsStatus` preimage backfill** (`arkade-swaps.ts:~2951`)
- Correctly fetches the preimage for swaps that settled while no monitor was attached (the "funded → app restart" scenario). Gracefully degrades if the preimage fetch fails.

**5. Type overloads**
- `sendLightningPayment` overloads correctly narrow: `waitFor?: "settled"` (or omitted) → `SendLightningPaymentResponse` with required preimage; `waitFor: "funded"` → `OptimisticSendLightningPaymentResponse` with optional preimage. All three consumer surfaces (core, Expo, service worker) and the `IArkadeSwaps` interface are updated consistently.

**6. Test coverage**
- 15+ new test cases covering: happy paths, failure before/after funding, late preimage persistence, monitor errors after resolution, persistence failures, SwapManager-disabled warning, refreshSwapsStatus backfill with and without errors. Edge cases are well covered.

**7. Infra**
- `vitest.config.ts` custom sequencer: clean solution to ensure `deprecatedSignerMigration` runs last and doesn't invalidate the signer key for other tests.
- Emulator v0.0.2 → v0.0.3, regtest submodule bump — no concerns.

No issues found. Ship it. 🚢

*— Arkana*

packages/ts-sdk/src/arkade/batch.ts
```diff
@@ -0,0 +1,271 @@
...
+                return { skip: true };
+            }
+
+            const commitmentTx = Transaction.fromPSBT(base64.decode(event.unsignedCommitmentTx));
+            const sharedOutput = commitmentTx.getOutput(0);
+            if (!sharedOutput?.amount) {
+                throw new Error("Shared output not found");
+            }
+
+            await session.init(vtxoTree, sweepTapTreeRoot, sharedOutput.amount);
```
pietro909, line 119: We initiate a signing session having validated only the amount.
Should we use `validateVtxoTxGraph` like we do in [wallet/wallet.ts](https://github.com/arkade-os/ts-sdk/blob/master/packages/ts-sdk/src/wallet/wallet.ts#L3248C17-L3248C36) ? And maybe validate the recipients too.

packages/ts-sdk/src/arkade/batch.ts
```diff
@@ -0,0 +1,271 @@
...
+            const info = await arkProvider.getInfo();
+            const forfeitOutputScript = OutScript.encode(
+                Address(network).decode(info.forfeitAddress),
+            );
+
+            let commitmentPsbt = Transaction.fromPSBT(base64.decode(event.commitmentTx));
+            const signedForfeits: string[] = [];
+            let hasBoardingInputs = false;
+            let connectorIndex = 0;
+            const connectorLeaves = connectorTree?.leaves() || [];
```
pietro909, line 157: Same as above, we should validate the connectors. This is also what NArk does.

packages/ts-sdk/src/arkade/batch.ts
```diff
@@ -0,0 +1,271 @@
...
+
+            const boardingIndices: number[] = [];
+
+            for (const input of inputs) {
+                // Auto-detect: is this input in the commitment tx? (boarding)
+                let boardingIdx: number | null = null;
+                for (let i = 0; i < commitmentPsbt.inputsLength; i++) {
+                    const psbtInput = commitmentPsbt.getInput(i);
```
pietro909: Shall we use the `!isVtxo(input)` like we do in wallet (see https://github.com/arkade-os/ts-sdk/blob/master/packages/ts-sdk/src/wallet/wallet.ts#L3038)?

packages/ts-sdk/src/arkade/contract.ts
```diff
@@ -0,0 +1,967 @@
...
+        if (coin.sourceTx) {
+            setArkPsbtField(arkTx, 0, PrevArkTxField, coin.sourceTx);
+        }
+
+        // tapscript witness → ConditionWitness on the ark tx and every checkpoint.
+        const condition = (def.tapscript.witness ?? []).map((w) => this.witnessBytes(w));
+        if (condition.length > 0) {
+            setArkPsbtField(arkTx, 0, ConditionWitness, condition);
+            for (const cp of checkpoints) {
+                setArkPsbtField(cp, 0, ConditionWitness, condition);
```
pietro909: Do we want to attach the ConditionWitness to all checkpoints?  Just asking because it feels like belonging to the contract path only.

packages/ts-sdk/src/arkade/contract.ts
```diff
@@ -0,0 +1,967 @@
...
+        const signedArk = await this.signArk(arkTx, userInputs);
+        const res = await client.arkProvider.submitTx(
+            base64.encode(signedArk.toPSBT()),
+            checkpoints.map((c) => base64.encode(c.toPSBT())),
+        );
+        const finalCps = await Promise.all(
+            res.signedCheckpointTxs.map(async (b) =>
+                base64.encode(
+                    (
+                        await client.identity!.sign(Transaction.fromPSBT(base64.decode(b)), [0])
```
pietro909, line 713: This will sign also server-only paths.

Shouldn't we mirror `const userInputs = this.userInputIndexes();`  at line 699?

packages/ts-sdk/src/arkade/contract.ts
```diff
@@ -0,0 +1,967 @@
...
+        return this;
+    }
+
+    /** Assemble the unsigned ark transaction and its checkpoints. */
+    async build(): Promise<{ arkTx: Transaction; checkpoints: Transaction[] }> {
+        if (this.outputs.length === 0) {
+            throw new Error("ArkadeTransactionBuilder: at least one output is required");
+        }
+        const outputsSum = this.outputs.reduce((s, o) => s + (o.amount ?? 0n), 0n);
+        const coin = this.coin ?? (await this.selectCoin(outputsSum));
```
pietro909, line 589: `this.selectCoin` uses `this.contract.getUtxos()` which returns a list of `VirtualCoin` which doesn't have `sourceTx`. 

It duck-types to `Utxo` because it honours the only the mandatory fields: the condition at 657 will always be false for  `this.coin = undefined`

packages/ts-sdk/src/arkade/contract.ts
```diff
@@ -0,0 +1,967 @@
...
+ * `<COSIGNER_KEY>` placeholders from `bind`, pass everything else through to the
+ * Arkade script encoder. The SDK does not interpret opcodes — this is pure
+ * substitution + encoding.
+ */
+export function resolveAsm(asm: AsmToken[], bind: Record<string, ArkadeParamValue>): Uint8Array {
+    const tokens = asm.map((t) => {
+        if (typeof t === "string" && t.startsWith("$")) {
+            return bindValue(bind, t.slice(1));
+        }
+        if (t === "<SELF>") {
```
pietro909: It looks like `<SERVER_KEY>`  and `<COSIGNER_KEY>`  also pass-through, which is opposite to what the docs say above: 

> substitute `$param` / `<SERVER_KEY>` / `<COSIGNER_KEY>` placeholders from `bind`, pass everything else through

pietro909 (changes requested):

pietro909 (approved):

pietro909 (approved):

pietro909 (approved):

packages/ts-sdk/src/arkade/batch.ts
```diff
@@ -0,0 +1,263 @@
...
+import type { TxTree } from "../tree/txTree";
+
+export type ArkadeExtendedCoin = ExtendedCoin & {
+    arkadeScriptBytes: Uint8Array;
+};
+
+const isVtxoCoin = (input: ArkadeExtendedCoin): input is ArkadeExtendedCoin & ExtendedVirtualCoin =>
+    "virtualStatus" in input;
+
+export function createArkadeBatchHandler(
```
Kukks, line 48: these batch handlers are killing me. They feel like such boilerplate

packages/ts-sdk/src/arkade/tweak.ts
```diff
@@ -0,0 +1,81 @@
...
+export function computeArkadeScriptPublicKey(pubKey: Uint8Array, script: Uint8Array): Uint8Array {
+    const hash = arkadeScriptHash(script);
+
+    const xOnly = pubKey.length === 33 ? pubKey.subarray(1) : pubKey;
+    // Force the even-Y ("02") interpretation of the x-only key. This mirrors the
+    // emulator, which lifts its key to even Y (BIP-340 lift_x) before tweaking;
+    // we must derive the tweaked pubkey from the same point or it won't match the
+    // key the emulator signs with. Prepending the prefix also lets `fromHex`
+    // accept both 32-byte (x-only) and 33-byte (compressed) inputs uniformly.
+    const point = secp256k1.Point.fromHex("02" + hex.encode(xOnly));
```
Kukks: this is super confusing... why do we strip it above and then force a 02 right after

packages/ts-sdk/src/arkade/tweak.ts
```diff
@@ -0,0 +1,81 @@
...
+ * result = pubKey + taggedHash("ArkScriptHash", script) * G
+ *
+ * The input pubKey should be a compressed (33-byte) or x-only (32-byte) public key.
+ * Returns a 32-byte x-only public key.
+ *
+ * @param pubKey - The emulator's base public key (32 or 33 bytes)
+ * @param script - The raw Arkade script bytes
+ * @returns 32-byte x-only tweaked public key
+ */
+export function computeArkadeScriptPublicKey(pubKey: Uint8Array, script: Uint8Array): Uint8Array {
```
Kukks, line 53: There's really no quick tweak func in scure?

packages/ts-sdk/src/arkade/tweak.ts
louisinger: that's needed to match the emulator tweak logic (normalize to 02 prefix). however scure has lift_x doing the same without this sloppy piece of code. I pushed the change, no more `"02" + ....`

packages/ts-sdk/src/arkade/tweak.ts
louisinger: yes, but scoped to taproot tweak. so we can't reuse them.

PR #360, opened 2026-03-18
Reduce indexer requests during wallet bootstrap from ~21 to 2 (one per contract)

pietro909:
## Summary 

  Reduce indexer requests during wallet bootstrap from ~21 to 2 (one per contract). The SDK was re-fetching VTXOs from the indexer on every read (`GET_VTXOS`, `GET_BALANCE`,             
  `GET_TRANSACTION_HISTORY`) and during bootstrap, even though the contract manager already populates the repository with VTXO data.
                                                                                                                                                                                          
  After this change, the service worker reads from the `WalletRepository` instead of hitting the indexer. Live updates continue to flow through the existing subscription and contract    
  watcher paths.
                                                                                                                                                                                          
  ## Changes                                                                                                                                                                              
  
  ### `src/contracts/contractManager.ts`                                                                                                                                                  
  - `initialize()` and `createContract()` now fetch with `includeSpent: true` so the repository gets full VTXO history (settled, spent, swept) — not just spendable. This is required for
  transaction history and balance to work from cached data.                                                                                                                               
  - `getContractsWithVtxos()` still uses `spendableOnly: true` (runtime query, not cache hydration).
                                                                                                                                                                                          
  ### `src/wallet/serviceWorker/wallet-message-handler.ts`                                                                                                                                
  - **`onWalletInitialized()`**: removed direct `indexerProvider.getVtxos()` and `readonlyWallet.getTransactionHistory()` calls. Contract manager init (`ensureContractEventBroadcasting`)
   now runs first to populate the repo, then VTXOs are read from the repository.                                                                                                          
  - **`GET_VTXOS`, `GET_BALANCE`**: read from `getVtxosFromRepo()` instead of `readonlyWallet.getVtxos()` (which made per-script indexer calls). Balance previously called `getVtxos()`
  twice (spendable + swept) — now uses a single repo read.                                                                                                                                
  - **`GET_TRANSACTION_HISTORY`**: builds history from cached VTXOs via `buildTransactionHistoryFromCache()` instead of pass-through to `readonlyWallet.getTransactionHistory()`.
  - **New `getVtxosFromRepo()`**: aggregates VTXOs across all contract addresses + wallet address, with outpoint-based deduplication.                                                     
  - **New `buildTransactionHistoryFromCache()`**: resolves `getTxCreatedAt` timestamps from cached VTXOs, falls back to indexer only for uncached txids.                                  
                                                                                                                                                                                          
  ## Open points                                                                                                                                                                          
                                                                                                                                                                                          
  1. **`RELOAD_WALLET`** calls `onWalletInitialized()`, but `ensureContractEventBroadcasting()` short-circuits if a subscription already exists. This means reload reads from whatever is 
  in the repo without forcing a re-fetch. Before this change, reload always did a fresh indexer fetch. Decide whether reload should force a contract manager re-initialization.
                                                                                                                                                                                          
  2. **`handleDelegate`** (line 1240) still calls `wallet.getVtxos()` which hits the indexer per-script. Out of scope for this PR (write operation, not a read handler), but could be     
  changed to read from repo.
                                                                                                                                                                                          
  3. **`getAllBoardingUtxos()`** still calls `readonlyWallet.getBoardingUtxos()` → `onchainProvider.getCoins()` on every `GET_BALANCE`. Boarding UTXOs are already saved to the repo      
  during bootstrap and updated via subscription. Reading from `walletRepository.getUtxos(boardingAddress)` would be consistent. Left as-is because boarding UTXOs use the onchain 
  provider, not the indexer (out of scope per task guardrails).                                                                                                                           
                                                            
  4. **Integration tests** (bootstrap request count, reads-after-bootstrap, live-update reflection) are specified in `TASK-1.agents.md` but require Docker/nigiri infrastructure and are
  not included in this PR.

  ## Risk areas

  - **Stale data window**: reads now depend on the repo being up-to-date. Between subscription disconnects and the next failsafe poll (60s), data can be stale. This window already       
  existed — the previous behavior masked it with redundant fetches, not eliminated it.
                                                                                                                                                                                          
  - **Bootstrap ordering**: `ensureContractEventBroadcasting()` must complete (triggering `ContractManager.initialize()` and its indexer fetches + repo writes) before                    
  `getVtxosFromRepo()` reads. If the contract manager fails to initialize, `getVtxosFromRepo()` falls back to reading from the wallet address only (may return empty on first boot).
                                                                                                                                                                                          
  - **`connection_reset` handler** in `ContractManager` still uses `includeSpent: false` (spendableOnly). After a reconnect, the repo may temporarily lack spent/swept VTXOs until the    
  next full re-fetch or subscription event. Transaction history could be incomplete until then.
                                                                                                                                                                                          
  - **`buildTransactionHistoryFromCache`** calls `readonlyWallet.getBoardingTxs()` which hits `onchainProvider.getTransactions()`. This is not an indexer call but is still a network call
   on every `GET_TRANSACTION_HISTORY`. It was the same before this change.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure complete UTXO/transaction history (including spent/swept entries) for accurate balances and histories.
  * Improved timestamp handling with safer fallbacks when creation times may be missing.

* **Refactor**
  * Balance and transaction history now read from local cached repository for faster, more reliable responses.

* **New Features**
  * Added a wallet refresh action to force a full UTXO/history refresh.

* **Tests**
  * Expanded coverage for bootstrap, cross-contract aggregation, deduplication, and repo-backed reads.

* **Chores**
  * Updated ignore patterns in project configuration.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/src/wallet/vtxo-manager.ts
+++ b/src/wallet/vtxo-manager.ts
@@ -386,8 +386,16 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
     private knownBoardingUtxos = new Set<string>();
     private sweptBoardingUtxos = new Set<string>();
     private pollInProgress = false;
+    private disposed = false;
     private consecutivePollFailures = 0;
+    private startupPollTimeoutId?: ReturnType<typeof setTimeout>;
     private static readonly MAX_BACKOFF_MS = 5 * 60 * 1000; // 5 minutes
+
+    // Guards against renewal feedback loop: when renewVtxos() settles, the
+    // server emits new VTXOs → vtxo_received → renewVtxos() again → infinite loop.
+    private renewalInProgress = false;
+    private lastRenewalTimestamp = 0;
+    private static readonly RENEWAL_COOLDOWN_MS = 30_000; // 30 seconds
 
     constructor(
         readonly wallet: IWallet,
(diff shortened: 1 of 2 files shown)


pietro909:
About https://github.com/arkade-os/ts-sdk/pull/360#issuecomment-4081096445 the comment is worth raising, but I do not think the bootstrap-order concern is a blocker as stated.

  - On the normal SDK path, the UI cannot call these reads before wallet init completes. ServiceWorkerWallet.create() awaits INIT_WALLET before returning, and reinitialize does the same
    on SW restart: src/wallet/serviceWorker/wallet.ts:1128, src/wallet/serviceWorker/wallet.ts:705.
  - INIT_WALLET itself awaits onWalletInitialized(), and that method does not stop after ensureContractEventBroadcasting(). It immediately calls getVtxosFromRepo() afterward, so a real
    getContractManager() / getContracts() failure already makes init fail: src/wallet/serviceWorker/wallet-message-handler.ts:993, src/wallet/serviceWorker/wallet-message-handler.ts:1081, src/wallet/serviceWorker/wallet-message-handler.ts:1365.
  - ReadonlyWallet.getContractManager() is lazy and single-flight, so even if a read somehow lands during bootstrap, it waits on the same initialization promise instead of failing just
    because init is still in progress: src/wallet/wallet.ts:803.
  - The only realistic gap is unsupported usage that bypasses create() and posts messages directly to the worker in the small window after handler start: src/wallet/serviceWorker/wallet-message-handler.ts:524. That is not the normal UI path.

So I would not restore the old catch for this reason alone. The old fallback returned partial wallet-address-only data, which can silently under-report balances/history for contract addresses. If you want stricter confidence, I’d add a test asserting that INIT_WALLET/reinitialize fails when contract-manager init fails. If you want softer UX, I’d prefer an explicit bootstrapping state over a silent partial-data fallback.

  One small wording tweak to the PR comment: these errors are not “unhandled” in the worker; they become rejected responses/promises to the client: src/wallet/serviceWorker/ wallet.ts:500, src/wallet/serviceWorker/wallet.ts:781.

PR #405, opened 2026-04-07
fix(unroll): derive complete-unroll sequence from CSV timelock

brendio:
- Root cause: complete-unroll hardcoded input sequence (0xffffffff - 1) instead of deriving from selected exit timelock.

- Fix: derive sequence with BIP68 from [exit.params.timelock](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) (blocks/seconds).

- Result: sequence matches CSV requirements and complete-unroll broadcasts successfully.

- Scope: targeted change in complete-unroll input construction only; no API changes.

Reproduction
1. Create or select a fully unrolled VTXO that has a CSV exit path.
2. Run complete-unroll against that VTXO.
3. Observe mempool rejection: mempool-script-verify-flag-failed (Locktime requirement not satisfied).

Before
1. complete-unroll input sequence was hardcoded to 0xffffffff - 1.
2. Built transaction could fail CSV validation even when exit maturity checks passed.

After
1. complete-unroll input sequence is derived from the selected exit timelock using BIP68 encoding.
2. Sequence now matches CSV requirements:
   - block timelock -> bip68.encode({ blocks: value })
   - second timelock -> bip68.encode({ seconds: value })
3. Transaction is accepted and broadcast succeeds in the same mature VTXO scenario.

Impact
1. Fixes complete-unroll for mature VTXOs that were incorrectly rejected by CSV validation.
2. Changes only input sequence derivation in the complete-unroll path.
3. No public API or behavior changes outside this spend path.

Risk
1. Low risk and narrowly scoped.
2. Uses the same BIP68 encoding approach already used elsewhere in the SDK.
3. Affects only the sequence value for selected CSV exit-path spends.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed transaction sequence handling so timelocks are encoded correctly for both block-based and time-based exits, improving unroll transaction validity and on-chain reliability.
* **Tests**
  * Added tests that validate timelock (BIP68/CSV) encoding and assert transactions are broadcast with the correct sequence value.


pietro909:
Thank you for your contribution @brendio !
Would you consider adding a dedicated test? There is a `should unroll` in `test/e2e/ark.test.ts` that looks like the best candidate to extend for an e2e test.

Alternatively, you may just work on a faster unit-test: the reusable timelock fixtures are in `test/vtxo-manager.test.ts:1193` for seconds and `test/vtxo-manager.test.ts:1426` for blocks.

Your call :)

brendio:
Thanks for the feedback. I added a dedicated unit test covering the completeUnroll sequence derivation for both block-based and second-based CSV timelocks. The test asserts the built transaction input nSequence matches the expected BIP68 encoding from the selected exit path timelock.


Thanks for the review. I updated completeUnroll to reuse the existing timelockToSequence helper from src/contracts/handlers/helpers.ts instead of duplicating BIP68 encoding logic. The behaviour remains the same; this just centralises the encoding path as suggested.

louisinger (changes requested):
LGTM but unroll.test.ts is unconsistent with other tests and do not pass. let's remove it and update unilateral exit e2e tests (we can do it in another PR)

brendio:
> LGTM but unroll.test.ts is unconsistent with other tests and do not pass. let's remove it and update unilateral exit e2e tests (we can do it in another PR)

Thanks for the review. I removed [test/unroll.test.ts](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) from this PR to keep it aligned with existing test patterns. I’ll handle unilateral-exit e2e test improvements in a separate follow-up PR.

pietro909 (approved):

PR #411, opened 2026-04-12
feat: output descriptors in contract params

Kukks:
## Summary
- Adds `DescriptorProvider` interface for descriptor-based signing with per-request descriptors
- Implements `SeedIdentity` HD wallet (BIP86 taproot derivation) with `signWithDescriptor` and key caching
- Adds `StaticDescriptorProvider` wrapper for legacy single-key `Identity`
- Migrates to `@bitcoinerlab/descriptors-scure` for descriptor parsing
- Updates `DefaultContractHandler` to use descriptor strings instead of raw `Uint8Array`

## Test plan
- [x] 870 tests passing (unit + integration)
- [x] `descriptor.test.ts` — parsing, normalization, HD descriptor support
- [x] `seedIdentity.test.ts` — HD derivation, descriptor signing, `isOurs`
- [x] `staticDescriptorProvider.test.ts` — single-key descriptor wrapping
- [x] `default.test.ts` — contract handler with descriptor strings

## Release Notes

* **New Features**
  * Added support for Bitcoin Taproot descriptors (`tr(...)` format) in wallet identity and contract handling.
  * Introduced hierarchical deterministic (HD) descriptor support with BIP32 derivation path parsing.
  * Added `ReadonlySeedIdentity` class for read-only descriptor-based wallet identities.
  * Introduced `DescriptorProvider` interface enabling descriptor-based signing operations.

* **Bug Fixes**
  * Enhanced public key extraction to support both descriptor and raw hexadecimal formats seamlessly.

* **Tests**
  * Added comprehensive test coverage for descriptor parsing, extraction, and validation utilities.


src/contracts/handlers/default.ts
```diff
@@ -37,13 +49,17 @@ export const DefaultContractHandler: ContractHandler<
 
     createScript(params: Record<string, string>): DefaultVtxo.Script {
         const typed = this.deserializeParams(params);
```
Kukks, line 50: Maybe make deserializeParams use extractPubKeyBytes internally

src/identity/seedIdentity.ts
```diff
@@ -13,51 +14,51 @@ import {
...
-    /** Optional BIP39 passphrase for additional seed entropy. */
     passphrase?: string;
 };
 
-/**
- * Detects the network from a descriptor string by checking for tpub (testnet)
- * vs xpub (mainnet) key prefix.
- * @internal
- */
+/** @deprecated Use {@link DescriptorSigningRequest} instead. */
```
Kukks: Seems out of place

src/identity/seedIdentity.ts
```diff
@@ -13,51 +14,51 @@ import {
...
+export type SigningRequest = DescriptorSigningRequest;
+
+// ── Shared helpers ────────────────────────────────────────────────
+
 function detectNetwork(descriptor: string): Network {
     return descriptor.includes("tpub") ? networks.testnet : networks.bitcoin;
 }
 
+function expandDescriptor(descriptor: string, index?: number) {
```
Kukks: i dont think Identity needs these

src/identity/descriptor.ts
```diff
@@ -0,0 +1,129 @@
...
+    xpub: string;
+    derivationPath: string;
+}
+
+/**
+ * Parse an HD descriptor into its components.
+ * HD descriptors have the format: tr([fingerprint/path']xpub/derivation)
+ * Returns null if the descriptor is not in HD format.
+ */
+export function parseHDDescriptor(
```
Kukks, line 94: this method looks crazy. try find a cleaner way within the descroptors  lib

src/identity/seedIdentity.ts
```diff
@@ -18,32 +18,23 @@ const ALL_SIGHASH = Object.values(SigHash).filter((x) => typeof x === "number");
 
 /** Use default BIP86 derivation with network selection. */
 export interface NetworkOptions {
-    /**
-     * Mainnet (coin type 0) or testnet (coin type 1).
-     * @default true
-     */
+    /** Mainnet (coin type 0) or testnet (coin type 1). @default true */
```
Kukks: revert all the comment removals

src/identity/staticDescriptorProvider.ts
```diff
@@ -0,0 +1,75 @@
...
+        requests: DescriptorSigningRequest[]
+    ): Promise<Transaction[]> {
+        const results: Transaction[] = [];
+        for (const request of requests) {
+            if (!this.isOurs(request.descriptor)) {
+                throw new Error(
+                    `Descriptor ${request.descriptor} does not belong to this provider`
+                );
+            }
+            const signed = await this.identity.sign(
```
Kukks, line 73: should check if we can use signMultiple... and use it if we can

src/identity/seedIdentity.ts
```diff
@@ -101,24 +103,23 @@ export class SeedIdentity implements Identity {
     protected readonly seed: Uint8Array;
     private readonly derivedKey: Uint8Array;
     readonly descriptor: string;
+    readonly isMainnet: boolean;
```
Kukks: why did I add this in seed identity?  Seems too verbose for nothing

Kukks:
CI is all-green on `a05bac05` (build 34s, unit 40s, lint 21s, integration 14m47s, CodeRabbit ✅ — [run #24804426412](https://github.com/arkade-os/ts-sdk/actions/runs/24804426412)).

Recent pushes since the last review round:

- `a05bac05` **fix: address review feedback on descriptor utilities**
  - `isDescriptor` now rejects unclosed `tr(` and empty bodies (no false positives slipping through to `expand()`).
  - `normalizeToDescriptor` throws on empty/non-string instead of silently producing `tr()`.
  - `extractRawPubKey` returns `undefined` for HD descriptors / unparseable input so role resolution stays best-effort.
  - `resolveRole` compares raw keys case-insensitively (mixed-case descriptors still match).
  - Docstrings: `DescriptorProvider` no longer claims `SeedIdentity` implements it directly (that lives in the stacked Phase B PR). `PathContext.role` points at `walletDescriptor` first, `walletPubKey` as fallback.
- `25499db7` **Merge master** (picks up PR #431 VTXO-state mitigations, #432 swept-vtxo fix, #434 e2e robustness, release 0.4.18).

HD work continues stacked on this branch:
- **PR #439** — Phase B (`SeedIdentity` becomes a `DescriptorProvider`): open, targeting `hd-wallet-support-v2`.
- **PR #440** — Phase C1 (`HDDescriptorProvider` standalone utility + 23 unit tests): open, targeting `hd-wallet-v2`.

Ready for another look whenever you have a moment.

src/contracts/handlers/default.ts
pietro909: It was done at line 68

pietro909 (approved):
## Behavior fixes                                                                                                                                                                           
   
  - **`resolveRole` fallback** — `walletDescriptor ?? walletPubKey` was coalescing with `??`, so an unresolvable `walletDescriptor` (notably HD) suppressed an otherwise-usable  `walletPubKey`. Refactored to try descriptor first, fall back to pubkey only when the descriptor cannot be resolved. Extracted a `matchRole` closure so both paths share comparison logic.
  Regression tests added in `test/contracts/handlers.test.ts > resolveRole`: descriptor-match, unresolvable → pubkey fallback, neither resolves, and the "resolved descriptor that doesn't    
  match is authoritative" case.                                       
  - **`StaticDescriptorProvider.isOurs`** — moved `normalizeToDescriptor` inside the try/catch so the method is total. Empty-string regression test added.
  - **`StaticDescriptorProvider.signWithDescriptor`** — mirrored the result-count guard from `src/wallet/wallet.ts:2506-2510`: after `signMultiple(...)`, verify `signed.length ===   requests.length` and throw otherwise. Regression test uses a mock `BatchSignableIdentity` that returns `[]`.                                                                                
                                                                                                                                                                                              
  ## API-surface cleanup (addressing self-review + "i dont think Identity needs these")                                                                                                       
                                                                      
  - **`MnemonicIdentity.mnemonic` field removed.** The BIP39 phrase is no longer retained on the instance — it was enumerable/serializable by default and had no reader in `src/`. 
  - **`SeedIdentity.isMainnet` instance field removed.** Trivially derivable from the descriptor and unread outside a test that existed only to verify storage. The `NetworkOptions.isMainnet`  input option is preserved — only the instance field was dropped.                                                                                                                           
   
  ## Style reverts per review ask                                                                                                                                                             
                                                                      
  Restored braces on all single-line `if` bodies in `seedIdentity.ts`, restored the early-return + for-loop structure in `sign()`, rolled the `// ignore` comment back to its original form,  restored the `const passphrase = opts.passphrase;` destructure, and restored the `private constructor(readonly descriptor: string)` parameter-property in `ReadonlyDescriptorIdentity`.
                                                                                                                                                                                              
@Kukks  please take a look when you have a moment, ready to merge

PR #412, opened 2026-04-12
refactor(timelock): centralise BIP68 encode/decode helpers

brendio:
1. Move BIP68 timelock conversion logic to a shared utility.
2. Reuse the utility from contracts helpers and script base.
3. No functional behaviour change intended; this is dedup/maintainability only.
As suggested in https://github.com/arkade-os/ts-sdk/pull/405#pullrequestreview-4095146296

## Summary by CodeRabbit

* **Refactor**
  * Consolidated timelock ↔ sequence conversion into a dedicated utility, removed the old low-level dependency, and updated usages across the codebase; no user-facing behavior or public APIs changed.
* **Tests**
  * Adjusted tests to import and use the new utility location.

### DIFF
--- a/src/contracts/handlers/helpers.ts
+++ b/src/contracts/handlers/helpers.ts
@@ -1,23 +1,5 @@
-import { RelativeTimelock } from "../../script/tapscript";
-import {
-    timelockToSequence as timelockToSequenceImpl,
-    sequenceToTimelock as sequenceToTimelockImpl,
-} from "../../utils/timelock";
+import { sequenceToTimelock } from "../../utils/timelock";
 import { Contract, PathContext } from "../types";
-
-/**
- * Convert RelativeTimelock to BIP68 sequence number.
- */
-export function timelockToSequence(timelock: RelativeTimelock): number {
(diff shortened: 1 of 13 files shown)


src/contracts/handlers/helpers.ts
```diff
@@ -1,30 +1,22 @@
...
 /**
  * Convert RelativeTimelock to BIP68 sequence number.
  */
 export function timelockToSequence(timelock: RelativeTimelock): number {
-    return bip68.encode(
-        timelock.type === "blocks"
-            ? { blocks: Number(timelock.value) }
-            : { seconds: Number(timelock.value) }
-    );
+    return timelockToSequenceImpl(timelock);
```
louisinger: let's remove this helper and keep only /utils/timelock ones

brendio: Good call. I removed the helper re-exports from src/contracts/handlers/helpers.ts and switched the remaining callers to import directly from src/utils/timelock.ts, so there is now a single source of truth for BIP68 timelock conversion.

PR #415, opened 2026-04-13
fix: use exit() instead of forfeit() for intentTapLeafScript

### LINKED ISSUE #414: intentTapLeafScript is set to forfeit() instead of exit()
Discovered in https://github.com/arkade-os/ts-sdk/pull/413

### `intentTapLeafScript` is set to `forfeit()` instead of `exit()` across the codebase

`intentTapLeafScript` and `forfeitTapLeafScript` are both assigned `vtxoScript.forfeit()` everywhere they're constructed:

- `src/wallet/wallet.ts` (~lines 563–564, ~2188–2189, ~2664–2665)
- `src/wallet/utils.ts` (~lines 25–26, 37–38, 56–57)

`forfeit()` returns the `(Alice + Server)` 2-of-2 multisig leaf (no timelock), while `exit()` returns the `(Alice)` CSV timelock leaf. The intent consumption site in `src/intent/index.ts` calls `getSequence(coin.intentTapLeafScript)` which tries to decode a CSV/CLTV timelock from the leaf script — this always returns `undefined` for the forfeit script since it has no timelock, and the transaction `sequence` is never set.

The intent proof also sets `tapLeafScript: [coin.intentTapLeafScript]`, which determines the tapscript branch revealed in the Bitcoin input. If the verifier expects the exit branch, the control block won't match.

If the two fields were meant to hold the same value there would be no reason for them to exist as separate fields.

**Expected:** `intentTapLeafScript: vtxoScript.exit()`

pietro909:
## Summary

- `intentTapLeafScript` was incorrectly assigned `vtxoScript.forfeit()` (2-of-2 multisig, no timelock) instead of `vtxoScript.exit()` (CSV timelock leaf) in all 7 construction sites across `src/wallet/wallet.ts` and `src/wallet/utils.ts`
- This caused `getSequence()` in intent proof construction to always return `undefined` (no timelock to decode), and produced incorrect tapscript control blocks

Closes #414

## Test plan

- [x] Unit tests pass
- [x] Build succeeds
- [ ] Run integration tests against regtest to verify intent proofs work end-to-end

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated wallet transaction processing to correctly configure virtual transaction outputs during sync operations, pending transaction reconciliation, and transaction finalization.

### DIFF
--- a/src/wallet/utils.ts
+++ b/src/wallet/utils.ts
@@ -23,7 +23,7 @@ export function extendVirtualCoin(
     return {
         ...vtxo,
         forfeitTapLeafScript: wallet.offchainTapscript.forfeit(),
-        intentTapLeafScript: wallet.offchainTapscript.forfeit(),
+        intentTapLeafScript: wallet.offchainTapscript.exit(),
         tapTree: wallet.offchainTapscript.encode(),
     };
 }
@@ -35,7 +35,7 @@ export function extendCoin(
     return {
         ...utxo,
         forfeitTapLeafScript: wallet.boardingTapscript.forfeit(),
-        intentTapLeafScript: wallet.boardingTapscript.forfeit(),
+        intentTapLeafScript: wallet.boardingTapscript.exit(),
         tapTree: wallet.boardingTapscript.encode(),
     };
 }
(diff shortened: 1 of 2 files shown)


louisinger:
nACK @pietro909, the intent should be signed with the collaborative script (forfeit). exit paths are allowed but we planned to drop support for them.

what's the error on master ? this change has been introduced long time ago https://github.com/arkade-os/ts-sdk/pull/238 so I'd expect it to work properly, something changed ?

pietro909:
Closing as false positive: the implementation is intended to work like this.

PR #423, opened 2026-04-16
fix(intent): set proof tx lockTime to 0

pietro909:
## Summary

`craftToSignTx` was deriving `tx.lockTime` from the max of per-input nSequence values, which conflated BIP-68 relative timelocks (CSV leaves) with absolute nLockTime. A VTXO with the typical 605184s CSV timelock produced `lockTime = 4195486` on the signed proof.

Matches BIP-322 and `arkd`'s reference builder by hardcoding `lockTime = 0`. Per-input nSequence continues to carry the CSV value on the PSBT input. `arkd`'s `Verify` does not inspect `LockTime`, so the change is not breaking at the wire level.

Also adds a docstring on `PathSelection.sequence` and a TODO on `getSequence` for a future discriminated-union refactor.

## Test plan
- [x] `pnpm test:unit` (47 files, 790 tests)
- [ ] Regression check against arkd on regtest (register / delete / get-pending-tx intents)

## Release Notes

* **Documentation**
  * Clarified sequence field documentation for relative timelock handling.

* **Tests**
  * Added test suite validating transaction lockTime behavior.

* **Improvements**
  * Refined transaction locking time calculation for timelock inputs.

### DIFF
--- a/src/contracts/types.ts
+++ b/src/contracts/types.ts
@@ -94,7 +94,11 @@ export interface PathSelection {
     /** Additional witness elements, for example a preimage for HTLC-like paths. */
     extraWitness?: Bytes[];
 
-    /** Sequence number override, for example for CSV timelocks. */
+    /**
+     * nSequence for the spending input, BIP-68 encoded when the leaf
+     * uses CSV. Decode with `sequenceToTimelock`; do NOT use as an
+     * absolute `Transaction.lockTime`.
+     */
     sequence?: number;
 }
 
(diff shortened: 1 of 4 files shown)


altafan (approved):
ACK. We do the same thing in go https://github.com/arkade-os/arkd/blob/master/pkg/ark-lib/intent/proof.go#L306-L307

bordalix (approved):

PR #427, opened 2026-04-18
Convert ts-sdk to pnpm monorepo with boltz-swap

pietro909:
## Summary

Converts this repository into a pnpm workspace monorepo that hosts both `@arkade-os/sdk` and `@arkade-os/boltz-swap` as sibling packages, with shared regtest/CI/release tooling at the root.

All ts-sdk sources now live at `packages/ts-sdk/`; boltz-swap joins at `packages/boltz-swap/` via a subtree merge so its full commit history is preserved and reachable through the merge's second parent.

## Structure

The PR opened with four first-parent commits. Since then, the branch has tracked upstream master and progressively consolidated duplicated tooling between the two packages — 24 first-parent commits as of the latest push.

### Initial conversion

1. **Move @arkade-os/sdk under packages/ts-sdk/** — pure renames; `git log --follow` and `git blame` continue to work.
2. **Add monorepo scaffolding** — `pnpm-workspace.yaml`, root `package.json` with workspace scripts, shared `config/` (tsconfig/vitest bases), monorepo CI workflows, root `scripts/release.sh` orchestrator, root README / AGENTS / CLAUDE.
3. **Merge boltz-swap history into packages/boltz-swap/ (subtree)** — two-parent merge; second parent is `arkade-os/boltz-swap@d30cfbde` (v0.3.16 tip at time of PR). Full boltz-swap history available at `packages/boltz-swap/` with rename detection.
4. **Adapt boltz-swap for monorepo** — drops per-package files now hoisted and switches `@arkade-os/sdk` dep to `workspace:*`.

### Subsequent work on this branch

- **Upstream sync** — repeated merges of `master` (v0.4.17 → v0.4.28 + #496 tsup migration) and of `arkade-os/boltz-swap` (v0.3.17 → v0.3.32).
- **Toolchain unification** — both packages now build with `tsup` (was: tsc + post-processors on ts-sdk, tsup on boltz-swap). Per-package tsconfigs extend `config/tsconfig.base.json`; per-package vitest configs `mergeConfig` from `config/vitest.base.ts` (which sets `fileParallelism: false` — boltz-swap was silently running test files in parallel). Single ambient `bip68` declaration at `config/types/bip68.d.ts`.
- **Lockstep release** — `scripts/release.sh` bumps both packages at one shared `v$VERSION` tag (`pnpm release -- X.Y.Z`); per-package release scripts disabled.
- **Hygiene** — root-only prettier config, lockfile, `.husky`, `.npmrc`, `.gitignore`, `.prettierignore`; `engines` consolidated at root (`node >=24.15.0 <25`); `pnpm/action-setup` pinned to `10.25.0` in CI; hoisted dev deps (`tsup`, `prettier`, `typescript`, `vitest`, `fake-indexeddb`, `eventsource`); Node 24 LTS bump; stale upstream-style `FOUNDATION.md` / `GEMINI.md` removed in favour of the monorepo-focused `AGENTS.md`.
- **Code consolidation** — `getRandomId` re-exported from ts-sdk and the three local `Math.random()` copies removed in boltz-swap; `RealmSwapRepository` typed against ts-sdk's `RealmLike` / `RealmResults`; pinned exact versions on shared deps (`bip68`, `@noble/curves`); shared filter helpers (`hasImpossibleSwapsFilter`, `applySwapsFilter`, `applyCreatedAtOrder`) deduped across the four swap-repo backends; unified test polyfills on `fake-indexeddb`; extracted regtest waiters to `scripts/regtest-wait.mjs`.
- **CI** — typecheck step for both packages; `smoke:dist` and `npm pack --dry-run` to validate ts-sdk's publish shape after every build.

## How to review

The **Commits** tab shows ~1100 commits because the boltz-swap subtree merge brings its full history into this PR. That's intentional and desirable for preserving history — but it means the commit tab is not the right place to review the monorepo conversion.

- For the actual changes, review the **Files changed** tab against `master`, or locally:
  - `git log --first-parent master..monorepo-pr` — the 24 conversion + consolidation commits
  - `git diff master..monorepo-pr` — total tree diff

## Merge strategy

**Use *Create a merge commit* — do NOT squash or rebase.** Squashing flattens the subtree merge and loses the boltz-swap history we worked to preserve.

## Test plan

- [x] `pnpm install` (root, workspace mode)
- [x] `pnpm run build` (ts-sdk + boltz-swap, ts-sdk builds first)
- [x] `pnpm run lint`
- [x] `pnpm -C packages/ts-sdk run typecheck` && `pnpm -C packages/boltz-swap run typecheck`
- [x] `pnpm -C packages/ts-sdk run smoke:dist`
- [x] `pnpm run test:unit` (ts-sdk: 60 files / 1198 passing; boltz-swap: 11 files / 303 passing)
- [x] `pnpm run regtest:up && pnpm run regtest:setup && pnpm run test:integration` — needs Docker; CI runs both packages' integration suites against the shared regtest stack
- [x] CI green on this PR

## Deferred (separate work)

- **NPM publish from CI** — `release.yml` creates a GitHub release; `pnpm publish` still runs from developer-local. Wiring it into CI needs an `NPM_TOKEN` secret + a `workflow_dispatch` entry to drive the existing non-interactive `pnpm release -- X.Y.Z` form.


ghost (changes requested):
## 🔍 Arkana Code Review — arkade-os/ts-sdk#427

**Monorepo conversion with boltz-swap subtree merge. 32k+ additions, 14k deletions, ~340 files.**

⚠️ **This PR introduces protocol-critical code (boltz-swap: VTXO handling, MuSig2 signing, swap transactions, refund paths). Requires human sign-off before merge per protocol-critical review policy.**

---

### 🔴 PROTOCOL-CRITICAL FINDINGS

**[M1] MEDIUM — `packages/boltz-swap/src/arkade-swaps.ts:~2053` — Blind signing of opaque Boltz transaction hash**

`signCooperativeClaimForServer` receives `claimDetails.transactionHash` from the Boltz API and uses it directly as the MuSig2 message to co-sign, without ever constructing or verifying the actual Bitcoin transaction this hash represents. The server key check (line ~2039) verifies key consistency but not transaction correctness. In the adversarial model (Boltz is untrusted), a malicious server could substitute a hash for a transaction that redirects the BTC claim to a Boltz-controlled address. The user would unknowingly co-sign it.

**Recommendation:** Reconstruct the expected claim transaction locally and verify its hash matches before signing, or at minimum verify the transaction outputs pay to the expected address.

---

**[M2] MEDIUM — `packages/boltz-swap/src/arkade-swaps.ts:~2262-2265` — `quoteSwap` auto-accepts Boltz-provided amount**

```typescript
async quoteSwap(swapId: string): Promise<number> {
    const { amount } = await this.swapProvider.getChainQuote(swapId);
    await this.swapProvider.postChainQuote(swapId, { amount }); // auto-accepts any amount
    return amount;
}
```

An adversarial Boltz server could return `{ amount: 0 }` or a drastically reduced amount and the client accepts it immediately. No comparison against the original swap amount, no minimum-acceptable-amount check, no user consent.

**Recommendation:** Accept a `minAcceptableAmount` parameter, or at least validate `amount > 0` and `amount >= expectedAmount * (1 - maxSlippage)`.

---

**[M3] MEDIUM — `packages/boltz-swap/src/arkade-swaps.ts:~1673-1779` — `refundArk` only processes `vtxos[0]`, abandons remaining VTXOs**

The chain swap ARK refund path only refunds the first VTXO at the lockup address. If multiple VTXOs exist (Boltz double-fund, or server rounds split), the rest are stranded. Compare with `refundVHTLC` (~line 975-1132) which correctly iterates over all `refundableVtxos`. Additionally, `refundArk` doesn't check whether the CLTV locktime has passed before choosing between the 3-of-3 cooperative refund and the 2-of-2 post-CLTV refund path.

**Recommendation:** Iterate all VTXOs at the lockup address (like `refundVHTLC` does), and check `isSubmarineRefundLocktimeReached` for the CLTV path selection.

---

**[M4] MEDIUM — `packages/boltz-swap/src/arkade-swaps.ts:~547-557` — `claimVHTLC` only claims first VTXO**

Similar to M3. In reverse swap claim, only `vtxos[0]` is claimed. If Boltz funds the lockup address with multiple UTXOs (or double-payment), remaining VTXOs are abandoned.

---

**[M5] MEDIUM — Ephemeral private keys stored in plaintext (`arkade-swaps.ts:~2172`, `types.ts:227`)**

The `ephemeralKey` (secp256k1 secret for MuSig2 signing on BTC chain swaps) is hex-encoded and stored directly in swap objects persisted to IndexedDB/SQLite/Realm. Preimages are similarly stored in plaintext. Any code with storage backend access (browser extensions, XSS on same-origin for IndexedDB) can extract these. This is a known trade-off for persistence across the swap lifecycle but should be documented as a security consideration.

---

### 🟡 CI / RELEASE FINDINGS

**[C1] MEDIUM — No smoke test for `boltz-swap` in CI**

CI runs `smoke:dist` and `npm pack --dry-run` only for `packages/ts-sdk` (ci.yml lines ~43-47). Since `@arkade-os/boltz-swap` is also a published npm package, it should get the same treatment to catch packaging regressions before release.

**[C2] MEDIUM — Release script (`scripts/release.sh`) has no test gate**

The script builds and publishes both packages without running any tests. A broken version could be published. Add at minimum `pnpm run test:unit` before the publish step.

**[C3] LOW — `next` branch dropped from CI triggers**

Old CI triggered on `master` + `next`. New CI only triggers on `master`. If `next` is still used for pre-release development, PRs targeting it lose CI. Intentional?

**[C4] LOW — boltz-swap missing explicit `registry` in `publishConfig`**

ts-sdk has `"publishConfig": { "access": "public", "registry": "https://registry.npmjs.org/" }`. boltz-swap only has `"access": "public"`. Works but inconsistent — add the registry for safety.

**[C5] LOW — `engines` field removed from published `@arkade-os/sdk`**

The `engines` constraint (`node >=22.12.0 <25`) was in the old root `package.json` but not carried to `packages/ts-sdk/package.json`. Consumers on unsupported Node versions will no longer get a warning. The root monorepo sets `>=24.15.0 <25` but that doesn't ship to npm (root is `private: true`).

**[C6] INFO — `.env.regtest` dropped from nigiri cache key**

Old: `key: nigiri-${{ hashFiles('regtest/.env.defaults', '.env.regtest') }}`
New: `key: nigiri-${{ hashFiles('regtest/.env.defaults') }}`
Could serve stale caches if `.env.regtest` changes.

---

### ✅ WHAT LOOKS GOOD

- **No breaking API changes.** `@arkade-os/sdk` exports, main, module, types, files, publishConfig are identical. One additive export (`getRandomId`). All downstream npm consumers (explorer, wdk, BlueWallet, escrow, x402, checkout) are safe.
- **MuSig2 implementation is correct.** Fresh nonces per signing session, `partialSigVerify()` called before accepting remote partials, proper BIP-327 state machine. No nonce reuse vectors found.
- **VHTLC address verification is solid.** Both claim and refund paths reconstruct the VHTLC script locally and verify the resulting address matches Boltz's response, preventing lockup address substitution.
- **Signature verification on all server responses.** `claimVHTLCwithOffchainTx` and `refundVHTLCwithOffchainTx` both verify signatures from all parties before combining.
- **No key material in logs.** Confirmed no private keys, preimages, or secrets passed to logger/console.
- **Build ordering correct.** `pnpm -r` respects topological order; ts-sdk builds before boltz-swap due to `workspace:*` dependency.
- **`workspace:*` is correct.** pnpm replaces with actual version on publish.
- **Per-package release scripts are guard rails** — they error out and point to the root script.
- **ts-sdk source changes are purely formatting** (prettier config: printWidth 80→100, trailingComma es5→all). No logic changes to the SDK.

---

### 📋 MERGE CHECKLIST

- [ ] **Human sign-off required** — protocol-critical code (VTXO handling, MuSig2 signing, refund paths)
- [ ] Address M1 (blind signing) or explicitly document the trust assumption
- [ ] Address M2 (quoteSwap auto-accept) — at minimum validate amount > 0
- [ ] Address M3/M4 (single-VTXO refund/claim) or document why multi-VTXO is impossible
- [x] Add boltz-swap smoke test to CI (C1)
- [x] Add test gate to release script (C2)
- [x] Confirm `next` branch CI drop is intentional (C3)
- [ ] **Merge strategy: Create a merge commit** (do NOT squash — preserves boltz-swap subtree history)
- [ ] Note: All 7 open PRs will need path-based rebasing after merge

---

*🤖 Reviewed by Arkana (model: opus) — aggressive protocol-level review*

ghost (commented):
## 🔍 Arkana Follow-up Review — new commit `fdf5c04e`

**Scope:** Single commit — `ci: run boltz-swap typecheck after build`

### Change

Moves the `Type-check boltz-swap` CI step from before `Build` to after it. This fixes a dependency ordering issue: boltz-swap imports `@arkade-os/sdk` via `workspace:*`, so it needs the built ts-sdk output to resolve types.

### Verdict: ✅ Correct

The CI step ordering is now:
1. `typecheck ts-sdk` (standalone, no build deps)
2. `build` (both packages, topological order)
3. `typecheck boltz-swap` (needs ts-sdk build output ✓)
4. `smoke:dist` / `npm pack` (ts-sdk only)
5. `test:unit`

No issues with this change.

### Outstanding from previous review

My previous findings (M1–M5, C1–C6) remain unaddressed. In particular:
- **C1** is partially helped by this commit (typecheck now works), but boltz-swap still lacks `smoke:dist` and `npm pack --dry-run` validation in CI
- **M1–M4** (protocol-critical: blind signing, auto-accept quote, single-VTXO claim/refund) still need responses

Previous review status (**CHANGES_REQUESTED**) still stands.

*🤖 Reviewed by Arkana (model: opus) — incremental review*

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `3555a9a4`

**Scope:** `ci: smoke boltz-swap dist; restore next, engines, regtest cache key`

### Changes vs previous review findings

| Finding | Status | Detail |
|---------|--------|--------|
| **C1** boltz-swap missing smoke test in CI | ✅ Fixed | `smoke:dist` + `npm pack --dry-run` added for boltz-swap in ci.yml |
| **C3** `next` branch dropped from CI | ✅ Fixed | `next` restored on both `push` and `pull_request` triggers |
| **C4** boltz-swap missing `registry` in publishConfig | ✅ Fixed | `"registry": "https://registry.npmjs.org/"` added |
| **C5** `engines` field removed from ts-sdk | ✅ Fixed | `"node": ">=22.12.0 <25"` added to `packages/ts-sdk/package.json` |
| **C6** `.env.regtest` dropped from nigiri cache key | ✅ Fixed | Restored in ci.yml |

### New code review: `packages/boltz-swap/scripts/smoke-dist.mjs`

Clean adaptation of the ts-sdk smoke script. Validates exports map file existence, declaration relative imports, and CJS/ESM root entry load. No issues found. The `workflow_dispatch` trigger addition is a nice bonus for manual CI runs.

### One minor note

C5: The `engines` constraint on ts-sdk is `>=22.12.0 <25` while the monorepo root uses `>=24.15.0 <25`. The discrepancy is intentional (consumers may still be on Node 22 LTS) — just confirming that's the intent.

### Outstanding findings

**C2** (release script has no test gate) and **M1–M5** (protocol-critical findings) remain unaddressed. Previous **CHANGES_REQUESTED** status still stands.

*🤖 Reviewed by Arkana (model: opus) — incremental review*

ghost (commented):
## 🔍 Arkana Follow-up Review — commits `919d0d1d` + `1e53b733`
*🤖 Reviewed by Arkana (model: opus) — incremental review*

### Commit `919d0d1d` — `chore: move pnpm settings to pnpm-workspace.yaml`

Moves `pnpm.overrides`, `pnpm.onlyBuiltDependencies`, and `pnpm.ignoredBuiltDependencies` from `package.json` to `pnpm-workspace.yaml`. This is the correct location per pnpm v9+ docs — the `pnpm` key in `package.json` is deprecated in favor of workspace-level settings. ✅ No issues.

### Commit `1e53b733` — `test(boltz-swap): isolate swap repo and extend slow timeouts`

Two changes:

1. **Test isolation fix** — Each `ArkadeSwaps` instance in e2e tests now gets a fresh `InMemorySwapRepository` instead of sharing state through `fake-indexeddb`. This prevents cross-test pollution where swap records from one `describe` block leak into another. Good fix — shared mutable state in tests is a classic source of flaky CI. ✅

2. **Timeout bumps** — Four e2e tests (`sendLightningPayment` ×2, `waitForSwapSettlement`, `save submarine swap`) get `{ timeout: 30_000 }`. These involve real Lightning invoice settlement over regtest, so 30s is reasonable. ✅

No protocol-critical changes. No API surface changes. No cross-repo impact.

### Outstanding

**C2** (release script has no test gate) and **M1–M5** (protocol-critical findings) remain unaddressed. Previous **CHANGES_REQUESTED** status still stands — human sign-off required for protocol-critical code.

ghost (commented):
## 🔍 Arkana Follow-up Review — commit `4a680abd`
*🤖 Reviewed by Arkana (model: opus) — incremental review*

### Commit `4a680abd` — `fix(boltz-swap): poll swap status when WebSocket fails`

Converts `monitorSwap` from WS-only (hard fail on error) to WS-preferred with REST polling fallback. Good resilience improvement — WS flakiness in CI/production no longer kills swap monitoring.

**Architecture is sound.** Shared `handleStatus` / `finish` / `cleanup` pattern with `settled` guard prevents double-resolve. Deduplication via `lastStatus` correctly handles polling returning same status across ticks. Fallback triggers on WS error, timeout (5s), and premature close. All good.

### Findings

**[P1] LOW — `boltz-swap-provider.ts:1356-1360` — Silent catch swallows `SwapNotFoundError`**

```typescript
} catch {
    // Transient errors (e.g. 404 during creation race) —
    // keep polling until terminal status or caller times out
}
```

`getSwapStatus` distinguishes `SwapNotFoundError` (Boltz says swap doesn't exist) from transient network errors. Silently swallowing `SwapNotFoundError` means a genuinely invalid `swapId` will poll forever until an external timeout kills it. Consider: after N consecutive `SwapNotFoundError`s (e.g. 10), call `finish(error)` to fail fast rather than polling a swap that Boltz says doesn't exist.

**[P2] LOW — `boltz-swap-provider.ts:1362` — 1-second polling interval is aggressive**

`setInterval(poll, 1000)` means 1 REST request/second to Boltz. Swap state transitions take seconds-to-minutes. Consider 3–5 second intervals, or exponential backoff starting at 1s. This won't matter for a single swap but could be a problem if multiple swaps fall back to polling simultaneously (e.g. if Boltz WS is down globally).

**[P3] LOW — No upper bound on polling lifetime**

If `getSwapStatus` always returns a non-terminal status and the swap never expires, polling runs indefinitely. The comment says "caller times out at a higher layer" — if that's guaranteed, document which layer owns the timeout. If not, add a `maxPollDuration` safety net (e.g. 1 hour) after which `finish(new NetworkError(...))` is called.

**[P4] INFO — Test coverage gaps for new paths**

The updated test covers WS-error→polling, which is the critical path. Missing:
- WS connection timeout (5s) → polling fallback
- WS close mid-stream (after receiving non-terminal status) → polling continues
- Dedupe: polling returns same status twice → callback fires only once

These are secondary, but worth adding for confidence in the state machine.

### Verdict: ✅ No blockers in this commit

This is a solid defensive improvement. P1–P3 are minor hardening suggestions, not blockers. The state machine logic is correct — `settled` flag prevents double-resolve, `cleanup` is thorough, and the WS→polling transition handles all edge cases (error, timeout, premature close, constructor throw).

### Outstanding from previous reviews

**M1–M5** (protocol-critical: blind signing, auto-accept quote, single-VTXO claim/refund, plaintext ephemeral keys) and **C2** (release script test gate) remain unaddressed. Previous **CHANGES_REQUESTED** status still stands — human sign-off required for protocol-critical code.

ghost (approved):
## 🔍 Arkana Follow-up Review — commit `15ee8c63`
*🤖 Reviewed by Arkana (model: opus) — incremental review*

**New since last review:** single commit adding a `pnpm run test:unit` gate to `scripts/release.sh`.

### Verdict: ✅ LGTM

The change adds a unit-test gate before any release mutations (version bump, tag, publish). Placement is correct — it runs after the clean-working-directory check but before version writes, so a test failure aborts cleanly thanks to `set -euo pipefail`. The dry-run path mirrors the real path.

No protocol-critical code, no API changes, no cross-repo impact.

---

*Cumulative status: 6 incremental reviews on this PR. The monorepo conversion is well-structured. Merging with a merge commit (not squash) to preserve boltz-swap subtree history remains the correct strategy. Awaiting human approval.*

ghost (approved):
## 🔍 Arkana Follow-up Review — commit `15ee8c63`
*🤖 Reviewed by Arkana (model: opus) — incremental review*

**New since last review:** single commit adding a `pnpm run test:unit` gate to `scripts/release.sh`.

### Verdict: ✅ LGTM

The change adds a unit-test gate before any release mutations (version bump, tag, publish). Placement is correct — it runs after the clean-working-directory check but before version writes, so a test failure aborts cleanly thanks to `set -euo pipefail`. The dry-run path mirrors the real path.

No protocol-critical code, no API changes, no cross-repo impact.

---

*Cumulative status: 6 incremental reviews on this PR. The monorepo conversion is well-structured. Merging with a merge commit (not squash) to preserve boltz-swap subtree history remains the correct strategy. Awaiting human approval.*

ghost (approved):
## Review: Convert ts-sdk to pnpm monorepo with boltz-swap

**Verdict: APPROVED**

This is a well-executed, large-scale repository restructuring. CI is green across all three check jobs (Lint/Test/Build, Integration ts-sdk, Integration boltz-swap). Prior M1–M5 protocol-critical findings are pre-existing and not introduced here.

### What this PR does

- Converts the repo to a `pnpm` workspace monorepo with `packages/ts-sdk/` and `packages/boltz-swap/` as sibling packages
- Brings boltz-swap history in via a subtree merge (preserves full git history, correct approach)
- Migrates ts-sdk from the old `tsc + post-processors` build pipeline to `tsup` (aligned with boltz-swap)
- Unifies CI, release tooling, prettier config, tsconfig/vitest bases, and shared devDeps at the root
- Adds lockstep release via a rewritten `scripts/release.sh`

### Findings

**No blocking issues found.** Notes below are observations, not requests for changes.

**Monorepo scaffolding — clean**
- `pnpm-workspace.yaml` is correctly structured with sensible `overrides` (esbuild pinned), `onlyBuiltDependencies`, and `ignoredBuiltDependencies`. Root `package.json` is properly `"private": true`.
- `config/tsconfig.base.json` / `config/vitest.base.ts` are minimal and correct. The `fileParallelism: false` global fix for boltz-swap is a genuine correctness improvement.

**Build toolchain (tsup migration) — correct**
- `packages/ts-sdk/tsup.config.ts` cleanly lists all entry points including the `sideEffects` entry (`src/contracts/handlers/index.ts`). The `splitting: true` + `treeshake: true` config is appropriate for a public SDK.
- Expo peer deps remain correctly externalised (`expo`, `expo-sqlite`, etc.).
- `smoke:dist` + `npm pack --dry-run` steps in CI now validate the publish shape on every build — a solid addition.

**Release script — improved**
- The rewritten `scripts/release.sh` is a significant improvement: `set -euo pipefail`, proper flag parsing, the version-bump logic is implemented in an embedded Node script (no `npm version` subprocess side effects), and it guards against unaligned package versions before bumps. The `pnpm run test:unit` gate added in commit `15ee8c63` closes the prior C2 finding.
- `--next` dist-tag is now filtered out of the GitHub Release trigger in `release.yml` — correct.
- The `publish_package` function checks NPM before publishing to avoid double-publishing if the script is re-run. Sensible guard.

**CI — structurally cleaner**
- Old 4-job layout (lint / unit / integration / build) collapsed into a `check` job (lint + typecheck + build + smoke + unit) feeding a matrix `integration` job across both packages. This eliminates the previous diagnostic-dump step and removes fragile per-test env print blocks.
- `pnpm/action-setup` pinned to `10.25.0` — good supply-chain hygiene.
- Node version now driven by `.nvmrc` rather than a hardcoded string — easier to keep in sync.

**boltz-swap additions — reviewed**
- `monitorSwap()` WebSocket-with-REST-polling-fallback implementation is well-structured: proper cleanup, deduplication of repeated statuses via `lastStatus`, 5 s connection timeout, `onerror`/`onclose` both transition to polling. No resource leaks visible.
- `createVHTLCScript` in `swap-helpers.ts` is the authoritative implementation (delegates to ts-sdk's `VHTLC.Script`). The older copy in `scripts.ts` is preserved for export compatibility but delegates the same way — acceptable.
- Shared `swap-repository.ts` filter helpers (`hasImpossibleSwapsFilter`, `applySwapsFilter`, `applyCreatedAtOrder`) cleanly eliminate the per-backend duplication that was flagged in earlier work.
- `musig.ts` MuSig2 wrapper is a clean port of the boltz-core chainable API onto `@scure/btc-signer/musig2.js` primitives, including correct nonce ordering and tweak handling.

**Minor observations (no action needed)**
- `packages/boltz-swap/package.json` `engines` field (`node >=22.12.0`) is not yet aligned with the root (`>=24.15.0 <25`). The PR description notes this as "engines consolidated at root" but the per-package field remains. Low risk since the root constraint applies in the workspace, but worth tidying before the next release.
- `scripts.ts` in boltz-swap (`createVHTLCScript`) and `swap-helpers.ts` both export a `VhtlcTimeouts` type with identical shape. These could be unified in a follow-up if the `scripts.ts` export is retired.

### Summary

The monorepo conversion is structurally sound and the incremental consolidation commits are logically ordered. History preservation via subtree merge is the right call. All CI checks pass. No protocol-critical code paths are changed by this PR.

Approved — merge as a merge commit (not squash) as stated in the PR description to preserve the boltz-swap subtree history.

PR #430, opened 2026-04-20
Fxi issue with default address

pietro909:


src/contracts/contractManager.ts
```diff
@@ -270,9 +269,8 @@ export class ContractManager implements IContractManager {
         // Load persisted contracts
         const contracts = await this.config.contractRepository.getContracts();
 
-        // Delta-sync: fetch only virtual outputs that changed since the last cursor,
-        // falling back to a full bootstrap for scripts seen for the first time.
-        await this.deltaSyncContracts(contracts);
+        // Delta-sync: fetch only virtual outputs that changed since the last cursor.
+        await this.syncContracts({ contracts });
```
Kukks, line 273: remove `contracts` in param and have `syncContracts` fetch it itself else the cursor wont update

src/repositories/realm/schemas.ts
```diff
@@ -77,7 +77,7 @@ export const ArkWalletStateSchema = {
     primaryKey: "key",
     properties: {
         key: "string",
-        lastSyncTime: "int?",
```
Kukks, line 80: meh, if its a schema change we should reconsider?

src/contracts/contractManager.ts
```diff
@@ -363,17 +361,8 @@ export class ContractManager implements IContractManager {
         await this.config.contractRepository.saveContract(contract);
 
         // fetch all virtual outputs (including spent/swept) for this contract
-        const requestStartedAt = Date.now();
         await this.fetchContractVxosFromIndexer([contract], true);
```
Kukks, line 364: Can we just use syncContracts here?

src/contracts/contractManager.ts
```diff
@@ -405,7 +394,8 @@ export class ContractManager implements IContractManager {
         pageSize?: number
     ): Promise<ContractWithVtxos[]> {
         const contracts = await this.getContracts(filter);
-        const vtxos = await this.getVtxosForContracts(contracts, pageSize);
+        await this.syncContracts({ contracts, pageSize });
```
Kukks, line 397: This is bad because everytime we call anything we still hit the indexer?

PR #431, opened 2026-04-20
Mitigate VTXO state / multi-contract / renewal-loop findings

Kukks:
## Summary

Mitigations for the findings in the review doc covering the three symptom clusters on `fxi-issue-with-default-address`:

1. VTXO state out of sync,
2. Multi-contract footguns from the single-contract retrofit,
3. Renewal-loop / register-intent retry cascade.

Per your direction this PR keeps the **single global sync cursor** and takes the lighter *add a `script` column* path rather than re-keying VTXO rows by script (D.2 blocker skipped). The structural C.1/C.2/C.4/C.6 items are left for a follow-up.

### What changed, by commit

| Commit | Theme | Review ref |
|---|---|---|
| `Apply cooldown on failed renewals and surface deleteIntent errors` | Renewal cooldown now updates on failure so transient settle errors can't re-enter renewal on the next `vtxo_received`; `deleteIntent` failures now log instead of silently swallowing | §C.3, §E |
| `Persist vtxo.script in SQLite and Realm wallet repositories` | Adds a `script` column + idempotent migration in SQLite and Realm; InMemory/IndexedDB propagate via the existing `...vtxo` spread. Unblocks contract-aware VTXO extension without the full primary-key swap | §D.2 light path |
| `Attribute VTXO extensions to the contract that owns each script` | New `extendVirtualCoinForContract` + public `Wallet.getContractsByScript()`; the wallet / service-worker callsites that used to stamp every VTXO with the default tapscript now look up the owning contract and fall back to the default only when no match exists | §B.1 |
| `Route subscription VTXOs to the contract that owns each script` | `ContractWatcher.processSubscriptionVtxos` now routes each VTXO via `vtxo.script` (single-script subscription kept as a legacy fallback); no more fan-out writing the same VTXO into every matching contract's `lastKnownVtxos` | §B.2 |
| `Seed watcher lastKnownVtxos and reconcile all contracts on reconnect` | `ContractWatcher.addContract` pre-populates `lastKnownVtxos` from the repo so the first poll doesn't re-emit every cached VTXO as `vtxo_received`; `handleContractEvent('connection_reset')` now refetches + reconciles the pending frontier | §A.3, §A.4 |
| `Keep cursor advancing: monotonic cursor, no spendableOnly` | `advanceSyncCursor` clamped with `Math.max` so concurrent syncs can't rewind it, and removal of the dead `spendableOnly`/`includeSpent` path in `ContractManager` | §A.2, §C.5, §E |
| `Drop spendableOnly from Expo background poll processor` | Upsert-only `saveVtxos` + `spendableOnly: true` meant VTXOs spent between polls stayed marked spendable forever; now fetches the full set so the upsert overwrites stale records | §A.2 |
| `Drop periodic bulk sync interval to consolidate entrypoints` | Removed a 10-min setInterval that was originally added for §A.1 — doubled the periodic-entrypoint count on top of the existing `ContractWatcher` failsafe poll. Cursor now advances via `initialize` + `connection_reset` only | consolidation follow-up |
| `Scope connection_reset reconciliation to active contracts` | `connection_reset` reconciliation narrowed to what the subscription actually watches (was over-broad `getAllContracts`) | scoping follow-up |
| `Scope getContractsByScript to the scripts being resolved` | Helper used to load every contract in the repo on every call; now takes the script set from the VTXOs being processed and uses the repository's existing `script` filter. New `collectVtxoScripts` helper dedupes scripts at each callsite | perf follow-up |
| `Widen connection_reset + failsafe poll to all watched contracts` | Replaced `getActiveContracts` with `getWatchedContracts` (active OR `lastKnownVtxos` non-empty) — matches the subscription's own scope so inactive contracts still holding unspent VTXOs are reconciled after reconnect instead of being orphaned | scope correctness follow-up |
| `Untrack REVIEW.md and add it to .gitignore` | Drop the local review doc from version control | housekeeping |
| `Unify VTXO extension on extendVirtualCoinForContract` | Deprecated `extendVirtualCoin` / `extendVtxoFromContract` in favour of a single entrypoint that accepts either a `Contract` or a `ReadonlyMap<script, Contract>`; migrated every caller (ContractManager, ExpoWallet, expo task runner) and added unit-test coverage for the routing + multi-contract regression case | pietro909 review + arkanaai #3 |
| `Warn when getContractsByScript falls back to default tapscript` | The bare `catch { return new Map() }` swallowed the underlying error and silently rerouted every VTXO through the default tapscript — exactly the regression this refactor prevents. Now logs the error | arkanaai #2 |
| `Log dropped VTXOs in processSubscriptionVtxos` | Subscription VTXOs with no attributable script (legacy indexer, transient issue, unknown script) were silently dropped; add an aggregate debug log so state-drift reports can be correlated with subscription drops | arkanaai #4 |
| `Merge origin/master into mitigations branch` | Pulls in `getDefaultAddress` removal, `fxi-issue-with-default-address` fix (#432), 30m sync overlap note, `vtxos_indexer_updated_at` column rename, and upstream wallet sync simplification | rebase |
| `Funnel VTXO annotation through ContractManager.annotateVtxos` | Adds `IContractManager.annotateVtxos(vtxos)` — single shared API that resolves each VTXO's `script` to its owning contract and stamps the correct tapscripts, throwing when a script has no registered contract. Replaces scattered `extendVirtualCoinForContract` / `getContractsByScript` callsites in `Wallet`, the service-worker handler, and the Expo task runner (whose `extendVtxo` dep is now a required-contract signature). Removes `ReadonlyWallet.getContractsByScript` and per-callsite script collection. Wallet classes always route VTXO annotation through the manager — no more silent default-tapscript fallback | footgun reduction |
| `Revert to single global VTXO sync cursor` | Restores the pre-merge single-cursor sync model (`vtxosIndexerUpdatedAt`) dropped when origin/master's per-script cursors came in. `ContractManager.syncContracts({contracts?, pageSize?, window?})` replaces `deltaSyncContracts`; subset polls intentionally don't advance the cursor so they can't hide data other contracts still need to pick up. Also drops the now-unused `includeSpent` parameter — the manager always fetches full history (spent/swept included), repo stays the source of truth | cursor architecture |
| `Address PR #431 review: annotation reuse, bootstrap cursor, Realm migration` | Fixes pass from reviewer + CodeRabbit: default `annotateVtxos` args to `[]` (avoid `.length` on undefined), let the global cursor leave its bootstrap-zero value by making the `mustUpdateCursor` guard `<= cursor` instead of `<`, and drop the schemaVersion gate in `runArkRealmMigrations` so consumers whose app schema is already past v2 still backfill legacy `ArkVtxo.script` rows. Also consolidated the scripts-from-address migration tests and added the missing `getWalletState` second-call assertion | review feedback |
| `Bundle expiring VTXOs into the boarding-settle poll cycle` | The 60s poll only swept fresh boarding UTXOs into settle intents — VTXOs close to expiry had no periodic renewal path, only the `vtxo_received` event-driven one. Renamed `settleBoardingUtxos` → `runPeriodicSettle` and made it include near-expiry VTXOs alongside unsettled boarding UTXOs in a single settle intent, skipping VTXO collection when the event-driven renewal is mid-flight to avoid double-spending. Unified cooldown state across both paths (`lastPeriodicSettleTimestamp` / `consecutivePeriodicSettleFailures`) since they share the intent; bumps `lastRenewalTimestamp` when VTXOs are part of the intent so the event path respects the recent activity | renewal gap |

### Not in this PR (called out intentionally)

- **§A.1 periodic cursor advance** — intentionally not added as a new interval. The existing `ContractWatcher.startFailsafePolling` (60s) is the natural home if we later want the cursor to progress without a reconnect; keeping entrypoints consolidated for now.
- **§D.2 full primary-key swap** (VTXO rows keyed by `script`) — skipped per your instruction; the `script` column approach covers §B.1/§B.2 without the schema churn.
- **§C.1** (settle error classification) and **§C.2** (scoping `safeRegisterIntent`'s "duplicated input" delete to colliding inputs) — structural changes that deserve their own PR.
- **§C.4** (_txLock observability), **§C.6** (run `finalizePendingTxs` on foreground), **§A.5** (persist `VtxoManager.knownBoardingUtxos`) — lower priority, follow-up.
- **arkanaai #5** (`collectVtxoScripts` reasoning note on change outputs without `script`) — no action needed, the reviewer marked it a nit and the fallback behaviour is correct for locally-created change outputs.

## Test plan

- [x] `pnpm lint` clean
- [x] `pnpm build:types` clean
- [x] `pnpm test:unit` → 819 passed / 3 skipped (49 files)
- [x] CI green on prior commits (build/lint/unit/integration)
- [ ] CI green after latest commits
- [ ] Manual: verify on reconnect every watched contract is reconciled (observe `refreshVtxos`-equivalent hits in indexer logs)
- [ ] Manual: verify `lastKnownVtxos` is warm after restart (no duplicate `vtxo_received` floods at boot)
- [ ] Manual: verify single-cursor sync — kill/relaunch wallet mid-sync and confirm cursor only advances on full-scope refreshes
- [ ] Manual: verify the poll loop renews near-expiry VTXOs in the same intent as a pending boarding UTXO (or alone when no boarding activity)

## Summary by CodeRabbit

* **New Features**
  * Contract-driven VTXO annotation exposed to service workers for accurate output attribution

* **Bug Fixes**
  * Safer migration/backfill for stored VTXO scripts across storage backends
  * More robust global sync cursor, boarding/settle retry/backoff, and renewal cooldowns

* **Refactors**
  * Consolidated contract watch/sync flows and unified VTXO indexing/refresh behavior

* **Docs & Tests**
  * Expanded tests and docs covering annotation, migrations, sync, and cooldowns

* **Breaking Changes**
  * Contract expiration removed; VirtualCoin now requires a script field; wallet sync cursor semantics changed


src/wallet/utils.ts
```diff
@@ -59,6 +59,51 @@ export function extendVtxoFromContract(
...
+ * When the wallet owns multiple contracts (e.g. default + delegate, or
+ * several active vHTLCs) a raw `extendVirtualCoin` call uses only the
+ * default tapscript, which silently overwrites the correct forfeit/intent
+ * data for any VTXO locked to a non-default contract. Resolving by
+ * `vtxo.script` — now always populated by the indexer and persisted —
+ * keeps the extension aligned with the contract that actually owns the
+ * VTXO, which is a correctness requirement before the vtxo is used for
+ * spending or saved back to the repository.
+ */
+export function extendVirtualCoinForContract(
```
pietro909, line 69: I'd deprecate `extendVirtualCoin` at line 19 and replace the usage of `extendVtxoFromContract` (line 43) with this new helper.

src/contracts/contractManager.ts
```diff
@@ -931,11 +838,10 @@ export class ContractManager implements IContractManager {
             for (const vtxo of vtxos) {
                 // Match the virtual output back to its contract via the script field
                 // populated by the indexer.
-                if (!vtxo.script) continue;
                 const contract = scriptToContract.get(vtxo.script);
                 if (!contract) continue;
                 result.get(contract.script)!.push({
-                    ...extendVtxoFromContract(vtxo, contract),
+                    ...extendVirtualCoinForContract(vtxo, contract),
```
pietro909: Can we reuse this.annotateVtxos()

src/wallet/utils.ts
```diff
@@ -40,6 +28,13 @@ export function extendCoin(
...
 }
 
+/**
+ * Internal primitive — prefer {@link extendVirtualCoinForContract}, which
+ * resolves the owning contract via a script→Contract map (or accepts a single
+ * `Contract` directly) and throws when resolution fails. Leaving this
+ * exported for the handful of callsites that already hold a `Contract` and
+ * only need the raw annotation.
+ */
 export function extendVtxoFromContract(
```
pietro909: can we try to not export this?

src/wallet/utils.ts
```diff
@@ -59,6 +54,76 @@ export function extendVtxoFromContract(
...
+    return contractOrMap;
+}
+
+/**
+ * Collect the unique, defined `script` values from one or more batches of
+ * virtual outputs. Callers pass the result to `getContractsByScript` so the
+ * contract lookup is scoped to the scripts actually being processed rather
+ * than every contract the wallet has ever created.
+ */
+export function collectVtxoScripts(
```
pietro909: do we need this? maybe too complex

src/wallet/wallet.ts
```diff
@@ -1828,8 +1630,16 @@ export class Wallet extends ReadonlyWallet implements IWallet {
...
         } catch (error) {
-            // delete the intent to not be stuck in the queue
-            await this.arkProvider.deleteIntent(deleteIntent).catch(() => {});
+            // delete the intent to not be stuck in the queue. If deletion fails
+            // the intent stays on the server and the next settle will hit
+            // "duplicated input" in safeRegisterIntent — surface the failure
+            // rather than silently swallowing it.
+            await this.arkProvider.deleteIntent(deleteIntent).catch((e) => {
+                console.warn(
+                    "Failed to delete intent after settle failure; intent may linger on server and cause 'duplicated input' on next settle",
```
pietro909: can we add id or something to the warning string to help debugging

test/repositories/migrations.test.ts
```diff
@@ -0,0 +1,421 @@
...
+    openDatabase,
+    closeDatabase,
+} from "../../src/repositories/indexedDB/manager";
+import {
+    initDatabase,
+    STORE_VTXOS,
+    backfillVtxoScripts,
+} from "../../src/repositories/indexedDB/schema";
+import { IndexedDBWalletRepository } from "../../src/repositories/indexedDB/walletRepository";
+
```
pietro909, line 20: Don't we have already a file where we test exiting migrations?

test/repositories/migrations.test.ts
```diff
@@ -0,0 +1,421 @@
...
+    // sequence of calls instead, returning pre-scripted responses.
+
+    interface QueryLog {
+        sql: string;
+        params?: unknown[];
+    }
+
+    function makeScriptedExecutor(
+        responses: Map<string, (params?: unknown[]) => unknown>
+    ): { executor: SQLExecutor; log: QueryLog[] } {
```
pietro909: We probably have already these test harness in the existing tests. Please double check it.

test/repositories/migrations.test.ts
```diff
@@ -0,0 +1,421 @@
...
+            ): Promise<T[]> {
+                log.push({ sql: sql.trim(), params });
+                return match(sql)(params) as T[];
+            },
+        };
+        return { executor, log };
+    }
+
+    it("creates a fresh table with NOT NULL script when none exists", async () => {
+        const responses = new Map<string, (params?: unknown[]) => unknown>([
```
pietro909: Existing tests for migrations are found here test/storage.test.ts and they assert on the DB content rather than logs.

Can you check if we can do the same here? Logs are difficult to follow.

src/wallet/utils.ts
```diff
@@ -59,6 +47,58 @@ export function extendVtxoFromContract(
...
+
+function resolveContract(
+    vtxo: VirtualCoin,
+    contractOrMap?: Contract | ReadonlyMap<string, Contract>
+): Contract | undefined {
+    if (!contractOrMap) return undefined;
+    if (isContractMap(contractOrMap)) {
+        return contractOrMap.get(vtxo.script);
+    }
+    return contractOrMap;
```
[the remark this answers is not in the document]
pietro909: It's internal to the module, used as a helper. Can stay like this.

pietro909 (approved):
## Testing

### Regtest

1. onboarding
2. VTXO expiry and renewal, with and without delegate
3. concurrent renewal (no errors)
4. Swap refund
5. Offchain send/receive

### Mainnet

1. Wallet restore
2. DB migration to V3 (with `script` index)
3. Lightning send/receive (Boltz)
4. Offchain send/receive


I feel confident we're ready to merge @Kukks !

PR #439, opened 2026-04-22
feat(identity): HD wallet primitives on SeedIdentity

Kukks:
## Summary

Phase B of the HD-wallet rollout: teach `SeedIdentity` / `MnemonicIdentity` to behave as a `DescriptorProvider`. Strictly additive — existing `Identity` callers are unaffected.

**Stacked on top of #411** — review once the descriptor-first contract params land. Retarget to `master` after #411 merges.

New public surface on `SeedIdentity`:

- `getAccountDescriptor()` — returns the template form with the trailing numeric index replaced by `*`, e.g. `tr([fp/86'/0'/0']xpub/0/*)`.
- `deriveSigningDescriptor(index)` — concrete descriptor at index `N`, validated `0 <= N < 2^31` (unhardened-range guard).
- `getSigningDescriptor()` — current concrete descriptor (the one the identity was constructed with).
- `isOurs(descriptor)` — true iff the descriptor's fingerprint + account xpub match this seed. Works on concrete descriptors, wildcard templates, and simple `tr(pubkey)` fallbacks.
- `signWithDescriptor(requests)` — batched per-request key derivation.
- `signMessageWithDescriptor(descriptor, message, type?)` — per-descriptor message signing.

Internals were refactored so legacy single-key signing and the new descriptor-based signing share the same helpers (`signTxWithKey`, `signMessageWithKey`, `derivePrivateKeyForDescriptor`).

## What's deliberately out of scope

- **`HDDescriptorProvider` + wallet rotation** — Phase C, separate PR.
- **Boarding-as-contract** — Phase D, separate PR.
- **Descriptor format change** — we keep producing `tr([fp/86'/0'/0']xpub/0/0)` by default so existing serialized state keeps working. The template `/*` form is only emitted via `getAccountDescriptor()`.

Full design rationale: `docs/plans/2026-04-22-hd-wallet-v2-design.md`.

## Test plan

- [x] `npm run test:unit` — 898 passing, 1 skipped (57 of those in `test/seedIdentity.test.ts`, 27 new cases covering the HD surface).
- [x] `npm run lint` — clean.
- [x] `npm run build` — builds CJS + ESM + types without errors.
- [ ] Integration tests — will run in CI against a live arkd. Nothing in this PR touches wallet/ark/boarding flows, so integration behavior is unchanged.

## Summary by CodeRabbit

* **New Features**
[...]

PR #440, opened 2026-04-22
feat(wallet): HDDescriptorProvider (Phase C1)

Kukks:
## Summary

- Stacked on top of #439 (Phase B). Introduces `HDDescriptorProvider` in `src/wallet/hdDescriptorProvider.ts`: a `DescriptorProvider` implementation that owns a single global derivation counter (NArk pattern) and rotates the active receive descriptor on demand.
- State persists under `walletRepository.getWalletState().settings.hd` — no storage-schema migration required.
- Concurrency: promise-chain mutex serialises `consumeNextIndex` / `rotateReceive`; template guard rejects HD state written by a different seed.

## Why split from wallet wiring

Phase C1 lands the provider in isolation so it can be reviewed as a focused delta. Wallet integration (rotation on `vtxo_received`, `getAddress()` routing, contract registration) is planned as a follow-up PR (Phase C2).

## Public surface

- `HDDescriptorProvider.create(identity, walletRepository)` — async factory that seeds initial receive index 0 or reuses persisted `currentReceiveIndex`.
- `getSigningDescriptor()` / `getCurrentReceiveIndex()` — sync accessors (the sync `DescriptorProvider.getSigningDescriptor()` contract is why the in-memory `current` cache exists).
- `peekNextIndex()` — advisory read.
- `consumeNextIndex()` — atomic bump + persist, returns `{ index, descriptor }` without touching the active receive slot (intended for change / boarding / self-spend).
- `rotateReceive()` — bump + persist + update active receive slot.
- `isOurs` / `signWithDescriptor` / `signMessageWithDescriptor` — delegate to the underlying `SeedIdentity`.

## Test plan

- [x] `test/hdDescriptorProvider.test.ts` (23 tests, 1.06s): fresh-wallet initialisation, persistence across instances, sequential + concurrent index consumption (10 parallel consumers yield distinct contiguous indexes), rotation semantics (active receive + nextIndex bumps), template-mismatch guard (throws when state was written by another seed), unrelated `settings`/`lastSyncTime` fields preserved across saves, `DescriptorProvider` delegation (schnorr signature verified against the x-only pubkey derived directly from the seed at the advertised index).
- [x] `npm run lint` — clean.
- [x] `npx tsc --noEmit` — clean.
- [x] `npm run test:unit` — 921 passed, 1 skipped, no regressions.

- Phase C2: wire into `Wallet.getAddress()` + `onContractEvent("vtxo_received")` rotation; register per-index `DefaultVtxo` contracts.
- Phase D: boarding-as-contract.

## Summary by CodeRabbit

* **New Features**
  * Adds an HD descriptor provider with persistent HD receive index, atomic index rotation, and descriptor-based signing/ownership tied to the wallet identity.

* **API Changes**
  * Descriptor retrieval is now asynchronous: providers expose an async method to obtain the next signing descriptor (allocating indexes for HD providers).

* **Bug Fixes / Stability**
  * Serializes rotations to prevent duplicate indexes and preserves unrelated wallet state during updates.

* **Tests**
  * Adds comprehensive tests for init, persistence, rotation, concurrency, ownership/signing, and corrupted-state handling.

### DIFF
--- a/src/wallet/hdDescriptorProvider.ts
+++ b/src/wallet/hdDescriptorProvider.ts
@@ -25,9 +25,9 @@ interface HDWalletSettings {
     descriptor: string;
 
     /**
-     * The most recently rotated-to receive index. `undefined` means no
-     * rotation has occurred yet — the wallet is still on the implicit
-     * default index 0.
+     * The most recently allocated descriptor index. `undefined` means no
+     * descriptor has ever been allocated; the next allocation will return
+     * index 0.
      */
     lastIndexUsed?: number;
 }
@@ -36,8 +36,11 @@ interface HDWalletSettings {
 const HD_SETTINGS_KEY = "hd";
 
 /**
- * HD-wallet {@link DescriptorProvider} that owns a single monotonic receive
- * index and rotates the active receive descriptor on demand.
+ * HD-wallet {@link DescriptorProvider} that allocates a fresh signing
+ * descriptor on every call. The provider holds no notion of "current" — it
+ * is a pure rotating allocator. The question of "which descriptor is the
+ * wallet currently bound to?" is answered by querying the contract
+ * repository for active contracts, not by asking this provider.
  *
  * State is persisted under `WalletRepository.getWalletState().settings.hd` so
  * that no storage-schema migration is required when switching a wallet from
@@ -45,18 +48,18 @@ const HD_SETTINGS_KEY = "hd";
  * which carries the wildcard account descriptor template (for derivation)
  * and the signing primitives.
  *
- * Both `getSigningDescriptor` and `getNextSigningDescriptor` go through the
- * shared per-repo `updateWalletState` flow — the index is never cached
- * in-memory, so a concurrent rotation by another `HDDescriptorProvider`
- * instance on the same repo can never be missed. Read-modify-write of the
- * persisted index runs inside the same shared mutex, so two
- * `getNextSigningDescriptor` callers can never observe the same index.
+ * The read-modify-write of the persisted index runs inside the shared per-
+ * repo `updateWalletState` mutex, so two `getNextSigningDescriptor` callers
+ * — including those driving separate `HDDescriptorProvider` instances on
+ * the same repo — can never observe the same index.
  *
  * @example
  * ```ts
  * const provider = await HDDescriptorProvider.create(identity, walletRepo);
  * const descriptor = await provider.getNextSigningDescriptor();
- * // descriptor: tr([fp/86'/0'/0']xpub/0/1)
+ * // descriptor: tr([fp/86'/0'/0']xpub/0/0)
+ * const next = await provider.getNextSigningDescriptor();
+ * // next: tr([fp/86'/0'/0']xpub/0/1)
  * ```
  */
 export class HDDescriptorProvider implements DescriptorProvider {
@@ -66,10 +69,10 @@ export class HDDescriptorProvider implements DescriptorProvider {
     ) {}
 
     /**
-     * Construct an HDDescriptorProvider. No I/O is performed at this point;
-     * persisted state is read on first call to `getSigningDescriptor` or
-     * `getNextSigningDescriptor`. State is validated lazily so a
-     * descriptor-mismatch error surfaces on first use rather than at boot.
+     * Construct an HDDescriptorProvider. No I/O is performed here;
+     * persisted state is read lazily on the first call to
+     * `getNextSigningDescriptor`. A descriptor-mismatch error surfaces on
+     * first use rather than at boot.
      */
     static async create(
         identity: HDCapableIdentity,
@@ -79,27 +82,17 @@ export class HDDescriptorProvider implements DescriptorProvider {
     }
 
     /**
-     * Returns the current signing descriptor. Always reads fresh state from
-     * the wallet repository so that rotations performed by another provider
-     * instance (or by another tab) are reflected immediately.
-     */
-    async getSigningDescriptor(): Promise<string> {
-        const state = (await this.walletRepository.getWalletState()) ?? {};
-        const settings = this.parseSettings(state);
-        return this.materializeAt(settings.lastIndexUsed ?? 0);
-    }
-
-    /**
-     * Rotate to a new receive descriptor and return it. The first call on a
-     * fresh wallet returns descriptor at index 1 — index 0 is the implicit
-     * default surfaced by `getSigningDescriptor`, so rotating past it gives
-     * the consumer a genuinely new address rather than a no-op.
+     * Allocate the next descriptor and return it. The first call on a fresh
+     * wallet returns descriptor at index 0; subsequent calls return 1, 2, 3,
+     * ... in order. Each call is atomic with respect to other rotations on
+     * the same repo: two concurrent callers can never observe the same
+     * index.
      */
     async getNextSigningDescriptor(): Promise<string> {
         return this.mutate((settings) => {
             const next =
                 settings.lastIndexUsed === undefined
-                    ? 1
+                    ? 0
                     : settings.lastIndexUsed + 1;
             settings.lastIndexUsed = next;
             return this.materializeAt(next);
--- a/src/wallet/hdDescriptorProvider.ts
+++ b/src/wallet/hdDescriptorProvider.ts
@@ -2,7 +2,7 @@ import {
     DescriptorProvider,
     DescriptorSigningRequest,
 } from "../identity/descriptorProvider";
-import { SeedIdentity } from "../identity/seedIdentity";
+import { HDCapableIdentity } from "../identity/hdCapableIdentity";
 import {
     WalletRepository,
     WalletState,
@@ -33,12 +33,8 @@ interface HDWalletSettings {
 /** Settings key under {@link WalletState.settings} where HD state lives. */
 const HD_SETTINGS_KEY = "hd";
 
-/**
- * Materialized view of the current receive slot held in memory so that
- * `getSigningDescriptor()` can satisfy the synchronous
- * {@link DescriptorProvider} contract.
- */
-interface CurrentReceive {
+/** Return shape of {@link HDDescriptorProvider.rotateReceive}. */
+interface ReceiveSlot {
     index: number;
     descriptor: string;
 }
@@ -49,7 +45,7 @@ interface CurrentReceive {
  *
  * State is persisted under `WalletRepository.getWalletState().settings.hd` so
  * that no storage-schema migration is required when switching a wallet from
- * single-key to HD. The provider is backed by a {@link SeedIdentity}, which
+ * single-key to HD. The provider is backed by a {@link HDCapableIdentity}, which
  * both carries the seed (for signing) and exposes the account descriptor
  * template (for derivation).
  *
@@ -67,13 +63,17 @@ interface CurrentReceive {
  */
 export class HDDescriptorProvider implements DescriptorProvider {
     /**
-     * Cached current receive slot. Populated by {@link HDDescriptorProvider.create}
-     * and updated on every rotation.
+     * Cached active receive index. Populated by {@link HDDescriptorProvider.create}
+     * and updated on every rotation so that the synchronous
+     * `getSigningDescriptor` / `getLastIndexUsed` accessors don't need to hit
+     * disk. The descriptor itself is derived on demand from this index plus
+     * `identity.descriptor` — a single string substitution — so caching the
+     * materialized form alongside it would just bookkeep the same fact twice.
      */
-    private current: CurrentReceive | null = null;
+    private lastIndex: number | null = null;
 
     private constructor(
-        private readonly identity: SeedIdentity,
+        private readonly identity: HDCapableIdentity,
         private readonly walletRepository: WalletRepository
     ) {}
 
@@ -84,10 +84,10 @@ export class HDDescriptorProvider implements DescriptorProvider {
      * receive descriptor and persists it. Subsequent runs reuse the stored
      * `lastIndexUsed`.
      *
-     * @throws if persisted state was written by a different identity (template mismatch).
+     * @throws if persisted state was written by a different identity (descriptor mismatch).
      */
     static async create(
-        identity: SeedIdentity,
+        identity: HDCapableIdentity,
         walletRepository: WalletRepository
     ): Promise<HDDescriptorProvider> {
         const provider = new HDDescriptorProvider(identity, walletRepository);
@@ -97,31 +97,31 @@ export class HDDescriptorProvider implements DescriptorProvider {
 
     /** Returns the currently-active receive descriptor. */
     getSigningDescriptor(): string {
-        return this.requireCurrent().descriptor;
+        return this.materializeAt(this.requireIndex());
     }
 
     /** Returns the currently-active receive index. */
     getLastIndexUsed(): number {
-        return this.requireCurrent().index;
+        return this.requireIndex();
     }
 
     /**
      * Rotate the active receive descriptor to the next index.
      *
-     * Updates both the persisted state and the in-memory `current` so that
+     * Updates the persisted state and the in-memory `lastIndex` so that
      * subsequent {@link getSigningDescriptor} calls observe the rotation.
      */
-    async rotateReceive(): Promise<CurrentReceive> {
-        const next = await this.mutate((settings) => {
-            const index =
+    async rotateReceive(): Promise<ReceiveSlot> {
+        const index = await this.mutate((settings) => {
+            const next =
                 settings.lastIndexUsed === undefined
                     ? 0
                     : settings.lastIndexUsed + 1;
-            settings.lastIndexUsed = index;
-            return { index, descriptor: this.materializeAt(index) };
+            settings.lastIndexUsed = next;
+            return next;
         });
-        this.current = next;
-        return next;
+        this.lastIndex = index;
+        return { index, descriptor: this.materializeAt(index) };
     }
 
     /**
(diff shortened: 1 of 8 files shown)


Kukks:
@arkanaai thanks for the review — all three issues addressed in `01abd59b`:

**P0 — lost-update race on HD settings**: `saveSettings` now routes through the shared `updateWalletState` mutex in `src/utils/syncCursors.ts` (the same per-repo `WeakMap<WalletRepository, Promise<void>>` that serialises `advanceSyncCursor` / `clearSyncCursor`). Interleaved HD rotations and sync cursor advances can no longer clobber each other's fields.

**P1 — `nextIndex` upper-bound guard**: new `takeNextIndex` private helper raises a descriptive `HD wallet exhausted: nextIndex N reached the non-hardened derivation cap (0x80000000)` before we hit the signing primitive's opaque error. Called from `consumeNextIndex`, `rotateReceive`, and the `initialize` fast path.

**Minor #2 — shape validation on load**: `loadOrInit` now rejects persisted `nextIndex` / `currentReceiveIndex` that aren't non-negative integers, so a corrupted repo fails loud at boot instead of silently deriving garbage.

**Regression coverage (5 new tests)**:
- `rotateReceive` racing `advanceSyncCursor` — asserts both the HD bump and the cursor advance land
- `clearSyncCursor` racing HD rotations — asserts cursor cleared but HD state survives
- Exhausted-index guard at the cap
- Corrupt `nextIndex` rejected
- Corrupt `currentReceiveIndex` rejected

Test suite: 926 passing (5 new), prettier + tsc clean.

Kukks:
Retargeted base from `hd-wallet-v2` to `master`. Cannot merge until **#439 lands first** — this branch's commits stack on top of #439's, so the diff against master currently shows the cumulative content (+1,998 LOC). Once #439 is rebased + merged, this PR's delta against master collapses to just its own +709 LOC (`HDDescriptorProvider` + state persistence).

Kukks:
Rebased onto the new #439 tip via cherry-pick. The branch is now `master + #439's commit + 2 commits`, totaling +1,200/-25 against master. Once #439 merges, this PR's diff against master collapses to its own +709 LOC (`HDDescriptorProvider` + state persistence + review fixes).

### Symbol-preservation audit

| Symbol | Pre-rebase | Post-rebase |
|---|---|---|
| `HDDescriptorProvider` | 18 | 18 ✅ |
| `hdDescriptorProvider` (file refs) | 2 | 2 ✅ |
| `deriveSigningDescriptor` (inherited from #439) | 22 | 22 ✅ |

### Local verification
- `pnpm test:unit` — 1067 tests pass (+28 vs #439's 1039)
- `pnpm lint` — clean
- TypeScript check — clean

**MERGEABLE** against master. Cannot merge until #439 lands first.

Kukks:
Force-pushed `6a706734` on top of the rebase onto the updated #439 tip. Following the API cleanup there (`SeedIdentity.deriveSigningDescriptor` removed), `HDDescriptorProvider` now substitutes the wildcard against the identity's account-descriptor template via a private `materializeAt(index)` helper. Behaviour is unchanged — same descriptor strings, same `MAX_NON_HARDENED_INDEX` guard. 28 HD tests + 1063 full unit suite pass, lint clean.

src/wallet/hdDescriptorProvider.ts
```diff
@@ -0,0 +1,345 @@
...
+ * wrap into the hardened range (which would throw deep inside the signing
+ * primitives with an error message that wouldn't help diagnose the cause).
+ */
+const MAX_NON_HARDENED_INDEX = 0x80000000;
+
+/**
+ * Persisted HD wallet state stored under {@link WalletState.settings}`.hd`.
+ * @internal
+ */
+interface HDWalletSettings {
```
pietro909, line 19: `config` key in local storage is never wiped out upon wallet reset in @arkade-os/wallet - worth checking if we need to do it now

src/wallet/hdDescriptorProvider.ts
```diff
@@ -0,0 +1,345 @@
...
+     * Follows the NArk pattern of a single global counter rather than
+     * separate receive/change ranges.
+     */
+    nextIndex: number;
+
+    /**
+     * Currently-active receive index. Rotated on `vtxo_received` by the
+     * wallet; the provider exposes it via {@link HDDescriptorProvider.getSigningDescriptor}.
+     */
+    currentReceiveIndex?: number;
```
pietro909: Let's have just one index to avoid drift

src/wallet/hdDescriptorProvider.ts
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    /**
+     * Serialise `fn` against other critical-section callers and hand it a
+     * freshly-loaded settings snapshot to mutate.
+     */
+    private mutate<T>(
+        fn: (settings: HDWalletSettings) => Promise<T>
+    ): Promise<T> {
+        const run = this.chain.then(
```
pietro909: check if we can re-use `navigator.lock` as it was done in a previous PR.

src/wallet/hdDescriptorProvider.ts
```diff
@@ -0,0 +1,345 @@
...
+        return provider;
+    }
+
+    /** Returns the currently-active receive descriptor. */
+    getSigningDescriptor(): string {
+        return this.requireCurrent().descriptor;
+    }
+
+    /** Returns the currently-active receive index. */
+    getCurrentReceiveIndex(): number {
```
pietro909: Maybe exposing the indexer and nextIndexer... is too much for this stage. If we don't have a use case let's hide it and reduce the API surface.

ghost (approved):
## Arkana Follow-up Review — `2d4f4e9` (collapse to single receive index)

Single new commit since my last review. Clean simplification — removes the dual-counter design (`nextIndex` + `currentReceiveIndex`) in favor of a single monotonic `currentReceiveIndex`. Reviewed the full delta from `31dceba` → `2d4f4e9`.

### What changed

1. **Removed `nextIndex`, `consumeNextIndex()`, `peekNextIndex()`, `takeNextIndex()`** — collapsed into `rotateReceive()` which now directly increments `currentReceiveIndex`. Correct for Phase C1 scope; if Phase C2 needs non-receive index consumption (change, boarding, self-spend), it can re-introduce a separate counter then.

2. **Removed per-instance promise-chain mutex** — `mutate()` now runs the closure *inside* `updateWalletState`'s callback, so the read happens under the repo-level mutex. This is strictly better than the previous design and directly addresses the cross-instance race I flagged in my first review. The test at line 143 (`serialises rotations across two providers on the same repo`) confirms the invariant. ✅

3. **Removed `MAX_NON_HARDENED_INDEX` guard** — P3. Practically unreachable (would require 2^31 rotations), and `parseSettings` still validates non-negative integer on load. The BIP-32 derivation library itself would reject an out-of-range index. Acceptable to omit for now; if you want belt-and-suspenders back, a one-liner in `rotateReceive` suffices.

4. **Removed `materializeAt` index validation** — the only callers are `initialize()` and `rotateReceive()`, both of which produce values from `parseSettings`-validated state + increment-by-one. Safe.

5. **`initialize()` always writes via `mutate()`** — even when `currentReceiveIndex` is already stored. Unnecessary write on warm boot, but harmless (idempotent, same values). Not worth complicating the code to optimize away.

### Correctness audit of the new `mutate()`

`hdDescriptorProvider.ts:201-216`: The closure pattern is sound:
- `result` is captured by reference; `updateWalletState` awaits the callback synchronously before returning, so `result!` is always populated. ✅
- `parseSettings` returns a shallow clone, so `fn`'s mutations don't alias the repo's copy. ✅
- The spread `{ ...state, settings: { ...(state.settings ?? {}), [HD_SETTINGS_KEY]: settings } }` preserves unrelated settings keys and wallet-state fields. ✅
- Tests at lines 267/293 confirm no clobbering against concurrent `advanceSyncCursor`/`clearSyncCursor`. ✅

### Test delta

Tests properly updated: removed `consumeNextIndex`/`peekNextIndex`/exhaustion tests, added cross-provider serialization test, removed `nextIndex` assertions from persistence checks. Coverage is adequate for the simplified API surface.

No new code paths, no protocol-level concerns in this delta.

⚠️ **Human review still requested** — this is key-management / HD derivation code.

**Approved.**

ghost (approved):
## Arkana Follow-up Review — `2d4f4e9` (collapse to single receive index)

Single new commit since my last review. Clean simplification — removes the dual-counter design (`nextIndex` + `currentReceiveIndex`) in favor of a single monotonic `currentReceiveIndex`. Reviewed the full delta from `31dceba` → `2d4f4e9`.

### What changed

1. **Removed `nextIndex`, `consumeNextIndex()`, `peekNextIndex()`, `takeNextIndex()`** — collapsed into `rotateReceive()` which now directly increments `currentReceiveIndex`. Correct for Phase C1 scope; if Phase C2 needs non-receive index consumption (change, boarding, self-spend), it can re-introduce a separate counter then.

2. **Removed per-instance promise-chain mutex** — `mutate()` now runs the closure *inside* `updateWalletState`'s callback, so the read happens under the repo-level mutex. This is strictly better than the previous design and directly addresses the cross-instance race I flagged in my first review. The test at line 143 (`serialises rotations across two providers on the same repo`) confirms the invariant. ✅

3. **Removed `MAX_NON_HARDENED_INDEX` guard** — P3. Practically unreachable (would require 2^31 rotations), and `parseSettings` still validates non-negative integer on load. The BIP-32 derivation library itself would reject an out-of-range index. Acceptable to omit for now; if you want belt-and-suspenders back, a one-liner in `rotateReceive` suffices.

4. **Removed `materializeAt` index validation** — the only callers are `initialize()` and `rotateReceive()`, both of which produce values from `parseSettings`-validated state + increment-by-one. Safe.

5. **`initialize()` always writes via `mutate()`** — even when `currentReceiveIndex` is already stored. Unnecessary write on warm boot, but harmless (idempotent, same values). Not worth complicating the code to optimize away.

### Correctness audit of the new `mutate()`

`hdDescriptorProvider.ts:201-216`: The closure pattern is sound:
- `result` is captured by reference; `updateWalletState` awaits the callback synchronously before returning, so `result!` is always populated. ✅
- `parseSettings` returns a shallow clone, so `fn`'s mutations don't alias the repo's copy. ✅
- The spread `{ ...state, settings: { ...(state.settings ?? {}), [HD_SETTINGS_KEY]: settings } }` preserves unrelated settings keys and wallet-state fields. ✅
- Tests at lines 267/293 confirm no clobbering against concurrent `advanceSyncCursor`/`clearSyncCursor`. ✅

### Test delta

Tests properly updated: removed `consumeNextIndex`/`peekNextIndex`/exhaustion tests, added cross-provider serialization test, removed `nextIndex` assertions from persistence checks. Coverage is adequate for the simplified API surface.

No new code paths, no protocol-level concerns in this delta.

⚠️ **Human review still requested** — this is key-management / HD derivation code.

**Approved.**

ghost (approved):
## Arkana Follow-up Review — `66e42f73` (rename currentReceiveIndex → lastIndexUsed)

Pure mechanical rename, no logic changes. The new name (`lastIndexUsed`) better communicates the semantics — it's the most recent index consumed, not a "current pointer" that could be confused with a cursor. Diff is consistent: all occurrences in implementation and tests renamed together. ✅

### Cross-repo note

⚠️ PR #441 (Phase C2) still references the old `currentReceiveIndex` / `getCurrentReceiveIndex()` API. It will need a rebase after this merges. Not a blocker for this PR, but flagging so it doesn't slip.

### Verdict

No new code paths, no behavioral changes, no protocol concerns. Rename is correct and complete.

⚠️ **Human review still requested** — HD derivation / key-management code.

**Approved.**

ghost (approved):
## Arkana Follow-up Review — `777b6c6` (rename settings.template → settings.descriptor)

Pure mechanical rename of the persisted `HDWalletSettings.template` field to `HDWalletSettings.descriptor`. No logic changes.

### Checked

1. **No storage migration needed** — `hdDescriptorProvider.ts` doesn't exist on `master` yet, so no production wallets could have persisted the old `template` key. ✅
2. **Rename is complete** — all code identifiers (`template` → `descriptor`) in implementation, tests, and error messages updated consistently. Remaining "template" occurrences are in comments/docstrings where the word describes the concept (descriptor template `tr(.../0/*)`), not the field name. ✅
3. **Error message updated** — `HD template mismatch` → `HD descriptor mismatch`. ✅
4. **Tests updated** — `toThrow(/template mismatch/)` → `toThrow(/descriptor mismatch/)`, persisted state assertions use `descriptor` key. ✅

No new code paths, no behavioral changes, no protocol concerns.

⚠️ **Human review still requested** — HD derivation / key-management code.

**Approved.**

ghost (approved):
## Arkana Follow-up Review — `777b6c6` (rename settings.template → settings.descriptor)

Pure mechanical rename of the persisted `HDWalletSettings.template` field to `HDWalletSettings.descriptor`. No logic changes.

### Checked

1. **No storage migration needed** — `hdDescriptorProvider.ts` doesn't exist on `master` yet, so no production wallets could have persisted the old `template` key. ✅
2. **Rename is complete** — all code identifiers (`template` → `descriptor`) in implementation, tests, and error messages updated consistently. Remaining "template" occurrences are in comments/docstrings where the word describes the concept (descriptor template `tr(.../0/*)`), not the field name. ✅
3. **Error message updated** — `HD template mismatch` → `HD descriptor mismatch`. ✅
4. **Tests updated** — `toThrow(/template mismatch/)` → `toThrow(/descriptor mismatch/)`, persisted state assertions use `descriptor` key. ✅

No new code paths, no behavioral changes, no protocol concerns.

⚠️ **Human review still requested** — HD derivation / key-management code.

**Approved.**

ghost (approved):
## Arkana Follow-up Review — `551ef81` (trim state + relax identity dep)

**Verdict: ✅ APPROVE**

Reviewed the delta from `777b6c6` → `551ef81`. This is a clean refactoring:

1. **`SeedIdentity` → `HDCapableIdentity`**: Correct narrowing. The provider only uses `descriptor`, `isOurs`, `signWithDescriptor`, `signMessageWithDescriptor` — all on the interface. Removes unnecessary concrete-class coupling.

2. **Drop cached descriptor, keep only `lastIndex: number | null`**: `getSigningDescriptor()` now derives on-demand via `materializeAt()` (single `String.replace`). Negligible cost, eliminates a bookkeeping surface for bugs.

3. **`mutate<T>` generic**: The definite assignment (`let result!: T`) is safe — `updateWalletState`'s updater is invoked synchronously within the mutex chain before the outer `await` resolves. Verified in `syncCursors.ts:29`.

4. **`CurrentReceive` → `ReceiveSlot`** (local return type): Cosmetic rename, no behavioral change.

No protocol-critical changes in this delta. Tests pass per CI. Ship it.

ghost (approved):
## Arkana Follow-up Review — `551ef81` (trim state + relax identity dep)

**Verdict: ✅ APPROVE**

Reviewed the delta from `777b6c6` → `551ef81`. This is a clean refactoring:

1. **`SeedIdentity` → `HDCapableIdentity`**: Correct narrowing. The provider only uses `descriptor`, `isOurs`, `signWithDescriptor`, `signMessageWithDescriptor` — all on the interface. Removes unnecessary concrete-class coupling.

2. **Drop cached descriptor, keep only `lastIndex: number | null`**: `getSigningDescriptor()` now derives on-demand via `materializeAt()` (single `String.replace`). Negligible cost, eliminates a bookkeeping surface for bugs.

3. **`mutate<T>` generic**: The definite assignment (`let result!: T`) is safe — `updateWalletState`'s updater is invoked synchronously within the mutex chain before the outer `await` resolves. Verified in `syncCursors.ts:29`.

4. **`CurrentReceive` → `ReceiveSlot`** (local return type): Cosmetic rename, no behavioral change.

No protocol-critical changes in this delta. Tests pass per CI. Ship it.

ghost (approved):
## Arkana Follow-up Review — `f95de3e5` (async `getSigningDescriptor`, drop in-memory cache)

Reviewed delta from `551ef81` → `f95de3e5`. This is a substantial simplification: the provider is now stateless — no in-memory cache, no `initialize()`, no `ReceiveSlot`. Clean.

### What changed

1. **`DescriptorProvider.getSigningDescriptor()` sync → `Promise<string>`** — breaking interface change. Verified: no production callers on `master` today. Only consumers are tests + PR #441 (Phase C2, unstacked). #441 will need a rebase — already flagged in prior reviews. ✅

2. **`getSigningDescriptor()` reads fresh from repo on every call** — eliminates the stale-cache failure mode where two providers on the same repo drift apart. Correct trade-off for a wallet: consistency > the ~microsecond cost of a repo read. ✅

3. **`materializeAt()` replaced string surgery with `expand()` from descriptors-scure** — library validates the ranged template and produces canonical output. Hardcoded `tr(${keyInfo.keyExpression})` is correct for NArk's taproot-only model. If the protocol ever adds non-`tr()` descriptors this will need revisiting, but that's a future concern, not a blocker. ✅

4. **First `getNextSigningDescriptor()` returns index 1, not 0** — index 0 is the implicit default from `getSigningDescriptor()`. Rotating past it gives a genuinely new address. Semantic improvement. ✅

5. **Fresh wallet `getSigningDescriptor()` does not persist** — pure read, no side effect. State only written on first `getNextSigningDescriptor()`. This means if the wallet crashes before any rotation, no HD state exists in storage — but `getSigningDescriptor()` will still return index 0 on restart, so behavior is idempotent. ✅

### Findings

**Nit — misleading class docstring** (`hdDescriptorProvider.ts:55-58`):
> Both `getSigningDescriptor` and `getNextSigningDescriptor` go through the shared per-repo `updateWalletState` flow

`getSigningDescriptor` does a plain `this.walletRepository.getWalletState()` — it does NOT go through the `updateWalletState` mutex. Only `getNextSigningDescriptor` does (via `this.mutate()`). The docstring should say only `getNextSigningDescriptor` uses the mutex; `getSigningDescriptor` is a bare read. Not a correctness bug — a dirty read just sees slightly stale state, and the next call will see the rotation — but the docstring is misleading.

**Nit — `create()` factory is async but does no I/O** (`hdDescriptorProvider.ts:82-86`):
```ts
static async create(...): Promise<HDDescriptorProvider> {
    return new HDDescriptorProvider(identity, walletRepository);
}
```
Harmless — keeps the door open for future I/O in the factory — but if the intent is "no I/O at construction" per the docstring, a sync constructor + static factory pattern would be more honest. Not blocking.

### Tests

302 lines, comprehensive. Concurrent serialization (10 parallel, cross-instance), race against `advanceSyncCursor`/`clearSyncCursor`, corrupt-state guards, schnorr signature verification against independently-derived pubkey. Coverage is excellent. ✅

### Verdict

Clean simplification. No protocol-critical concerns in this delta. Ship it.

⚠️ **Human review still requested** — HD derivation / key-management code.

**Approved.**

ghost (approved):
## Arkana Follow-up Review — `f95de3e5` (async `getSigningDescriptor`, drop in-memory cache)

Reviewed delta from `551ef81` → `f95de3e5`. This is a substantial simplification: the provider is now stateless — no in-memory cache, no `initialize()`, no `ReceiveSlot`. Clean.

### What changed

1. **`DescriptorProvider.getSigningDescriptor()` sync → `Promise<string>`** — breaking interface change. Verified: no production callers on `master` today. Only consumers are tests + PR #441 (Phase C2, unstacked). #441 will need a rebase — already flagged in prior reviews. ✅

2. **`getSigningDescriptor()` reads fresh from repo on every call** — eliminates the stale-cache failure mode where two providers on the same repo drift apart. Correct trade-off for a wallet: consistency > the ~microsecond cost of a repo read. ✅

3. **`materializeAt()` replaced string surgery with `expand()` from descriptors-scure** — library validates the ranged template and produces canonical output. Hardcoded `tr(${keyInfo.keyExpression})` is correct for NArk's taproot-only model. If the protocol ever adds non-`tr()` descriptors this will need revisiting, but that's a future concern, not a blocker. ✅

4. **First `getNextSigningDescriptor()` returns index 1, not 0** — index 0 is the implicit default from `getSigningDescriptor()`. Rotating past it gives a genuinely new address. Semantic improvement. ✅

5. **Fresh wallet `getSigningDescriptor()` does not persist** — pure read, no side effect. State only written on first `getNextSigningDescriptor()`. This means if the wallet crashes before any rotation, no HD state exists in storage — but `getSigningDescriptor()` will still return index 0 on restart, so behavior is idempotent. ✅

### Findings

**Nit — misleading class docstring** (`hdDescriptorProvider.ts:55-58`):
> Both `getSigningDescriptor` and `getNextSigningDescriptor` go through the shared per-repo `updateWalletState` flow

`getSigningDescriptor` does a plain `this.walletRepository.getWalletState()` — it does NOT go through the `updateWalletState` mutex. Only `getNextSigningDescriptor` does (via `this.mutate()`). The docstring should say only `getNextSigningDescriptor` uses the mutex; `getSigningDescriptor` is a bare read. Not a correctness bug — a dirty read just sees slightly stale state, and the next call will see the rotation — but the docstring is misleading.

**Nit — `create()` factory is async but does no I/O** (`hdDescriptorProvider.ts:82-86`):
```ts
static async create(...): Promise<HDDescriptorProvider> {
    return new HDDescriptorProvider(identity, walletRepository);
}
```
Harmless — keeps the door open for future I/O in the factory — but if the intent is "no I/O at construction" per the docstring, a sync constructor + static factory pattern would be more honest. Not blocking.

### Tests

302 lines, comprehensive. Concurrent serialization (10 parallel, cross-instance), race against `advanceSyncCursor`/`clearSyncCursor`, corrupt-state guards, schnorr signature verification against independently-derived pubkey. Coverage is excellent. ✅

### Verdict

Clean simplification. No protocol-critical concerns in this delta. Ship it.

⚠️ **Human review still requested** — HD derivation / key-management code.

**Approved.**

PR #446, opened 2026-04-23
fix(serviceWorker): exempt streaming messages from timeouts

pietro909:
Settlement and other batch-round flows surrender control to the Ark server and peers.

Both page and worker now skip the timeout for:

- SETTLE
-  RECOVER_VTXOS
-  RENEW_VTXOS
 
Service Worker death is still detected out-of-band via the existing PING / MESSAGE_BUS_NOT_INITIALIZED path on short concurrent requests.

## Release Notes

* **Bug Fixes**
  * Long-running wallet operations (settle, recover, and renew transactions) are no longer subject to request timeouts and can now complete without interruption.

### DIFF
--- a/src/wallet/serviceWorker/wallet-message-handler.ts
+++ b/src/wallet/serviceWorker/wallet-message-handler.ts
@@ -613,6 +613,19 @@ export class WalletMessageHandler
         } as WalletUpdaterResponse;
     }
 
+    // Flows that surrender control to the Ark server and the other participants
+    // in a batch round: quiet gaps between protocol events can easily exceed
+    // the bus-level messageTimeoutMs. Liveness is covered out-of-band by the
+    // page-side PING / MESSAGE_BUS_NOT_INITIALIZED path triggered by concurrent
+    // short requests (GET_STATUS, GET_BALANCE, ...).
+    isLongRunning(message: WalletUpdaterRequest): boolean {
+        return (
+            message.type === "SETTLE" ||
+            message.type === "RECOVER_VTXOS" ||
+            message.type === "RENEW_VTXOS"
+        );
+    }
+
     async handleMessage(
         message: WalletUpdaterRequest
     ): Promise<WalletUpdaterResponse> {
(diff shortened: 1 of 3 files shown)


Kukks (commented):
What about the wallet updater timeouts?

Kukks:
Answering my own question after reading through the code path carefully — the wallet updater `tick()` timeout isn't a practical concern today:

- `runTick()` in `src/worker/messageBus.ts:216` still wraps `updater.tick(now)` in `withTimeout()` with tag `${messageTag}:tick`.
- `WalletMessageHandler.tick()` at `src/wallet/serviceWorker/wallet-message-handler.ts:577` just drains `onNextTick[]` callbacks via `Promise.allSettled` then clears the queue.
- The callbacks enqueued by `scheduleForNextTick()` are synchronous `() => WalletUpdaterResponse | null` wrappers around `this.tagged({id, type, payload})` — no network I/O, no awaits. Total execution is bounded by `queue.size × cost(tagged)`, microseconds-to-millis territory.

So the `withTimeout` on tick() can't realistically fire — the bulk of the work that matters (the actual settlement round-trips) happens inside `handleMessage(SETTLE)` / `handleMessage(RECOVER_VTXOS)` / `handleMessage(RENEW_VTXOS)`, which is exactly what this PR exempts via `isLongRunning`. The tick path just shuttles already-produced event payloads out to the page.

Worth keeping in mind going forward: the signature `onNextTick: (() => WalletUpdaterResponse | null)[]` is sync-only by contract but not by type enforcement. If anyone later does `scheduleForNextTick(async () => ...)`, the returned promise would be awaited by `Promise.allSettled` and could push tick() toward the timeout — probably worth a lint-visible assertion or a runtime `typeof result === "function" && result.constructor.name !== "AsyncFunction"` guard at the `scheduleForNextTick` boundary, but that's future-proofing, not something this PR needs to address.

I'll track the broader MessageBus delivery-guarantee gaps (silent drops on unknown tag, falsy responses, null event.source, late responses after timeout) in #448 / the follow-up PR. Those are complementary to this change rather than overlapping.

Marking this resolved on my end. Happy to approve once I've read the unit tests side.

PR #359, opened 2026-03-17
feat: add WsElectrumChainSource provider

louisinger:
Ports the `WsElectrumChainSource` class from [marina/electrum-chain-source.ts](https://github.com/vulpemventures/marina/blob/refs/heads/master/src/port/electrum-chain-source.ts) into `src/providers/electrum.ts`.

## Changes

- **New file:** `src/providers/electrum.ts` — `WsElectrumChainSource` class using `ws-electrumx-client`
- **New dependency:** `ws-electrumx-client@^1.0.5` — only new dep added

## Adaptations from marina source

- Replaced `liquidjs-lib` types:
  - `Buffer` → `Uint8Array`
  - `address.toOutputScript(addr)` → `OutScript.encode(Address(network).decode(addr))`
  - `crypto.sha256(script)` → `sha256` from `@scure/btc-signer/utils.js`
  - `Transaction.fromHex(hex).outs[vout]` → `Transaction.fromRaw(hex.decode(hexStr)).getOutput(vout)`
- `BlockHeader` and `TransactionHistory` types defined locally (no external domain types)
- Constructor takes a `network: Network` parameter for address decoding
- `Unspent.witnessUtxo` uses `{ script: Uint8Array; value: bigint }` matching btc-signer output shape

## Summary by CodeRabbit

* **New Features**
  * Electrum-based on-chain provider with WebSocket support: address watching, transaction history and status, UTXO/coin listing, fee estimation, transaction broadcasting, and chain-tip queries.

* **Infrastructure**
  * Added a WebSocket proxy service to docker-compose to enable Electrum connectivity.

* **Chores**
  * Updated project dependencies and minor formatting adjustments.

PR #450, opened 2026-04-23
feat(providers): electrum onchain provider (production-ready, supersedes #359)

Kukks:
Supersedes #359 (draft by @louisinger). Rebased onto post–docker-compose-removal master (the submodule split in 33bdd76b made #359 unmergeable), adds unit tests, and completes the provider for production.

## Changes from #359

Three commits preserved with original @louisinger authorship:
- `0d28dd03` — `feat: add WsElectrumChainSource in providers using ws-electrumx-client`
- `b7aa81d7` — `feat: add ElectrumOnchainProvider implementing OnchainProvider interface` (docker-compose.yml change dropped; arkade-regtest submodule now owns compose)
- `a3643201` — `perf: optimize getTxOutspends from O(outputs×history×fetch) to O(4 round trips)`

One commit added on top:
- `262ff005` — `test: add unit tests for ElectrumOnchainProvider and close() helper` + adds a `close()` method to `ElectrumOnchainProvider` so callers can tear down the underlying WebSocket.

## What's in this PR

- `WsElectrumChainSource` — low-level wrapper over `ws-electrumx-client` covering the full electrum RPC surface used by the SDK (history, listunspent, verbose tx, block headers, subscriptions, broadcast, fee estimation, relay fee).
- `ElectrumOnchainProvider` implementing `OnchainProvider`: replaces esplora polling with `scripthash.subscribe` push, converts BTC/kB → sat/vB fee, sequential 1P1C broadcast (electrum has no package relay), 4-round-trip `getTxOutspends`.
- Exports both classes from `src/index.ts`.
- New dep: `ws-electrumx-client@^1.0.5`.

## Test plan

Unit tests (31 new tests passing, 875 total):
- [x] `getCoins` happy + empty paths
- [x] `getFeeRate` BTC/kB→sat/vB conversion + floor(1) guard + daemon-cannot-estimate (-1)
- [x] `broadcastTransaction` single + 1P1C package + rejects 3+
- [x] `getChainTip` parses genesis header (hash + timestamp verified against known-good values)
- [x] `getTxStatus` mempool + confirmed + blocktime/time fallback
- [x] `getTransactions` empty + verbose-tx mapping
- [x] `getTxOutspends` all-unspent + single-spender + ambiguous reused-script disambiguation
- [x] `WsElectrumChainSource` missing-tx retry, empty-input short-circuits, address decoding, close error swallowing

Live smoke test against `wss://electrum.arkade.sh` (mainnet, block 946339):
- [x] `getChainTip` returns current tip with correct parsed hash and block time
- [x] `getFeeRate` / `getRelayFee` return sane sat/vB values
- [x] `fetchBlockHeader` + `fetchBlockHeaders` (single + batch)
- [x] `getCoins` on a known bech32m address
- [x] `getTransactions` returns paginated history with proper ExplorerTransaction shape
- [x] `getTxStatus` on a confirmed tx returns correct height/time
- [x] `getTxOutspends` returns correct spent/unspent with spender txids
- [x] `close()` cleanly tears down the WebSocket

## Notes

- CI: depends on master's CI pipeline; this branch targets master cleanly now.
- @louisinger's docker-compose websocat bridge entry was dropped — that work belongs in `arkade-regtest` (the submodule that replaced docker-compose in 33bdd76b), tracked separately.
- Credit for the entire provider design and three original commits belongs to @louisinger.

## Summary by CodeRabbit

* **New Features**
  * Electrum WebSocket on-chain provider with real-time chain tip, transaction retrieval (verbose/raw), broadcasting (single tx and package), fee estimation, address watching, unspent listing, tx status, and subscription-backed updates.
  * Public exports for the Electrum provider and related types.

* **Tests**
  * Comprehensive test suite validating header/tx fetching, fee conversion, broadcasting flows, address watching, unspent resolution, retries, and subscription behavior.

* **Chores**
  * Added a runtime dependency to support the Electrum provider.

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,7 +122,8 @@
         "@scure/base": "2.0.0",
         "@scure/bip39": "2.0.1",
         "@scure/btc-signer": "2.0.1",
-        "bip68": "1.0.4"
+        "bip68": "1.0.4",
+        "ws-electrumx-client": "^1.0.5"
     },
     "devDependencies": {
         "@types/better-sqlite3": "7.6.13",
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
         "@scure/bip39": "2.0.1",
         "@scure/btc-signer": "2.0.1",
         "bip68": "1.0.4",
-        "ws-electrumx-client": "^1.0.5"
+        "ws-electrumx-client": "1.0.5"
     },
     "devDependencies": {
         "@types/better-sqlite3": "7.6.13",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -52,6 +52,9 @@ importers:
       expo-task-manager:
         specifier: ~14.0.9
         version: 14.0.9(expo@54.0.33(@babel/core@7.29.0)(react-native@0.83.1(@babel/core@7.29.0)(react@19.2.4))(react@19.2.4))(react-native@0.83.1(@babel/core@7.29.0)(react@19.2.4))
+      ws-electrumx-client:
+        specifier: ^1.0.5
+        version: 1.0.5
     devDependencies:
       '@types/better-sqlite3':
         specifier: 7.6.13
@@ -2308,6 +2311,11 @@ packages:
   isexe@2.0.0:
     resolution: {integrity: sha512-RHxMLp9lnKHGHRng9QFhRCMbYAcVpn69smSGcq3f36xjgVVWThj4qqLbTLlq7Ssj8B+fIQ1EuCEGI2lKsyQeIw==}
 
+  isomorphic-ws@5.0.0:
+    resolution: {integrity: sha512-muId7Zzn9ywDsyXgTIafTry2sV3nySZeUDe6YedVd1Hvuuep5AsIlqK+XefWpYTyJG5e503F2xIuT2lcU6rCSw==}
+    peerDependencies:
+      ws: '*'
+
   istanbul-lib-coverage@3.2.2:
     resolution: {integrity: sha512-O8dpsF+r0WV/8MNRKfnmrtCWhuKjxrq2w+jpzBL5UZKTi2LeVWnWOmWRxFlesJONmc+wLAGvKQZEOanko0LFTg==}
     engines: {node: '>=8'}
@@ -3656,6 +3664,10 @@ packages:
   write-file-atomic@4.0.2:
     resolution: {integrity: sha512-7KxauUdBmSdWnmpaGFg+ppNjKF8uNLry8LyzjauQDOVONfFLNKrKvQOxZ/VuTIcS/gge/YNahf5RIIQWTSarlg==}
     engines: {node: ^12.13.0 || ^14.15.0 || >=16.0.0}
+
+  ws-electrumx-client@1.0.5:
+    resolution: {integrity: sha512-pBjfFqb9j2FBz7NPbnd8r2lOYanEw8ACzfKxOtHCgEGqre5QiTax5XHLVgbsiOvST0vmsHAiMtkJPvsZm77PIQ==}
+    engines: {node: '>=10'}
 
   ws@6.2.3:
     resolution: {integrity: sha512-jmTjYU0j60B+vHey6TfR3Z7RD61z/hmxBS3VMSGIrroOWXQEneK1zNuotOUrGyBHQj0yrpsLHPWtigEFd13ndA==}
@@ -6312,6 +6324,10 @@ snapshots:
 
   isexe@2.0.0: {}
 
+  isomorphic-ws@5.0.0(ws@8.19.0):
+    dependencies:
+      ws: 8.19.0
+
   istanbul-lib-coverage@3.2.2: {}
 
   istanbul-lib-instrument@5.2.1:
@@ -7861,6 +7877,14 @@ snapshots:
     dependencies:
       imurmurhash: 0.1.4
       signal-exit: 3.0.7
+
+  ws-electrumx-client@1.0.5:
+    dependencies:
+      isomorphic-ws: 5.0.0(ws@8.19.0)
+      ws: 8.19.0
+    transitivePeerDependencies:
+      - bufferutil
+      - utf-8-validate
 
   ws@6.2.3:
     dependencies:
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -53,7 +53,7 @@ importers:
         specifier: ~14.0.9
         version: 14.0.9(expo@54.0.33(@babel/core@7.29.0)(react-native@0.83.1(@babel/core@7.29.0)(react@19.2.4))(react@19.2.4))(react-native@0.83.1(@babel/core@7.29.0)(react@19.2.4))
       ws-electrumx-client:
-        specifier: ^1.0.5
+        specifier: 1.0.5
         version: 1.0.5
     devDependencies:
       '@types/better-sqlite3':
--- a/src/index.ts
+++ b/src/index.ts
@@ -107,6 +107,11 @@ import {
     ElectrumOnchainProvider,
     WsElectrumChainSource,
 } from "./providers/electrum";
+import type {
+    TransactionHistory as ElectrumTransactionHistory,
+    BlockHeader as ElectrumBlockHeader,
+    Unspent as ElectrumUnspent,
+} from "./providers/electrum";
 import {
     RestArkProvider,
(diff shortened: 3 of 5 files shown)


Kukks:
Pushed `7d82742b` addressing the review feedback. Summary of what changed and what I deliberately left as-is:

### Addressed

**arkanaai (CHANGES_REQUESTED):**
- 🟡 **subscribeHeaders subscription leak** — `WsElectrumChainSource.subscribeHeaders()` now uses `ws.subscribe()` once and caches the tip. The callback (which the library invokes both for the initial response and for subsequent server-pushed notifications) keeps the cache fresh. Previously every call sent `blockchain.headers.subscribe` as a regular request, leaving a stale subscription on the server each time.
- 🟡 **Float→sat precision in getTransactions** — extracted `verboseToExplorer()` helper used by both `getTransactions` and `watchAddresses`. When the verbose tx response includes the raw `hex` field (Bitcoin Core's `getrawtransaction <txid> 1` always does), output amounts are derived from the parsed raw tx — exact bigint sats — instead of `Math.round(value * 1e8)`. Falls back to the float multiply for servers that omit `hex`.
- 🟡 **watchAddresses serial round-trips** — initial `fetchHistory` per script and per-script `subscribeScriptStatus` calls are now both `Promise.all`-parallelised, collapsing O(n) round trips on first call into a single batched wave.
- 🟢 **"1C1P" → "1P1C"** typo in `broadcastTransaction` and the test asserting it.
- 🟢 **Address validation** — `getCoins`/`getTransactions`/`watchAddresses` now go through `encodeAddress()` which wraps the opaque scure decode error with the offending address.

**CodeRabbit:**
- ✅ Pinned `ws-electrumx-client` to exact `1.0.5`.
- ✅ Re-exported types from `src/index.ts` as `ElectrumTransactionHistory` / `ElectrumBlockHeader` / `ElectrumUnspent`. I prefixed with `Electrum` rather than re-exporting the bare names because `TransactionHistory` / `BlockHeader` / `Unspent` are very generic and likely to collide with future protocol-level types — happy to re-export them un-prefixed if the maintainers prefer.
- ✅ Extracted `verboseToExplorer` helper (see above — the float-precision fix piggybacked on this).

### Not changed (with rationale)

- **CodeRabbit: "WebSocket.close passing string instead of numeric code"** — `WsElectrumChainSource.close()` calls `this.ws.close("close")` where `this.ws` is `ElectrumWS` (the library), not a raw WebSocket. The library's signature is [`close(reason: string): Promise<boolean>`](https://github.com/arkade-os/ts-sdk/blob/electrum-provider-prod/node_modules/ws-electrumx-client/build/main/lib/electrum-ws.d.ts) — `reason` is the string passed to the underlying `ws.close(CLOSE_CODE, reason)` call, which already supplies `CLOSE_CODE = 1000` internally. Passing a number to `ElectrumWS.close()` would be a type error.

- **CodeRabbit: "getTxStatus race between fetchVerboseTransaction and subscribeHeaders"** — the underlying race shrinks dramatically now that the tip comes from a cached subscription instead of a fresh request. The remaining window (cached read may be 1 block older than the verbose-tx call) is bounded by ~10 minutes for bitcoin and is the standard derivation `tipHeight - confirmations + 1`. Switching to `transaction.get_merkle` would remove the race at the cost of an extra round trip per call; happy to revisit if the cache approach feels insufficient.

- **arkanaai: "Fee conversion is wrong"** — the reviewer self-retracted this one ("Not a bug after all") in the same review. Conversion factor matches Bitcoin Core's `estimatesmartfee` (BTC/kvB → sat/vB).

- **arkanaai: "fetchTransactions retry uses flat 1s sleep"** — left as-is for this PR; cap is 5 attempts so worst case is 5s. Exponential backoff would be a clean follow-up but I'd rather get the substantive fixes merged first.

- **CodeRabbit: "Add watchAddresses unit tests"** — added (see `test/electrum.test.ts` `watchAddresses` describe block).

880 unit tests pass locally, prettier clean, typecheck clean. Re-requesting review.

Kukks:
Added TRUC 1P1C package relay support in `b27789c0`.

### What changed

The 1P1C broadcast path now goes through Fulcrum's `blockchain.transaction.broadcast_package` (the wire equivalent of bitcoind's `submitpackage` RPC) instead of sequential parent-then-child broadcasts. Previously the 2-tx path was:

```ts
await this.chain.broadcastTransaction(txs[0]);          // parent
return this.chain.broadcastTransaction(txs[1]);         // child
```

Now:

```ts
return this.chain.broadcastPackage(txs);                // atomic
```

### Why no fallback

Sequential broadcast cannot work for TRUC (BIP 431) packages because the parent typically has zero (or below-minfee) fee and depends on the child to pay for both via CPFP. A "best-effort" fallback would let TRUC packages fail in subtle ways at the parent step with a `min relay fee not met` error that masks the actual problem. Failing loudly with the protocol-level error is the right call — consumers receiving a "method not found" know to route through a different provider for that submission.

### Server requirements (documented in JSDoc)

Backing Electrum server must implement `blockchain.transaction.broadcast_package`:
- ✅ Fulcrum ≥ 1.10 backed by bitcoind ≥ v28.0.0
- ❌ ElectrumX (does not implement the method)
- ❌ Fulcrum backed by bitcoind < v28.0.0

Verified `wss://electrum.arkade.sh` for production use is on the required Fulcrum + bitcoind v28+ stack — would be a good thing to confirm before merge.

### Implementation details

- `broadcast_package` returns `{success, errors?}`, not a txid. The child txid is derived locally via double-SHA256 of the serialized child bytes (`childTxidFromHex`).
- Topology required by the protocol (parent first, child last) is preserved by passing `txs` directly.
- 5 new tests cover the success path, server-side rejection (`success: false` with errors), missing-error edge case, and crucially a test asserting that an "unknown method" error from the server propagates as-is instead of triggering a sequential fallback (`expect(wsMock.request).toHaveBeenCalledTimes(1)`).

885 unit tests pass (+5 new), prettier clean, typecheck clean.

Kukks:
Pushed `18daec8e` addressing the second round of feedback.

### 🚨 CRITICAL: wtxid vs txid bug — fixed

@arkanaai was right. `childTxidFromHex` was hashing the raw broadcast bytes directly, which for segwit txs (every Ark tx) includes witness data and yields the wtxid instead of the txid. Any downstream caller tracking the VTXO by id — round settlement, forfeit monitoring, exit paths — would have been pointed at a hash that doesn't index into the chain.

Fix delegates to `Transaction.fromRaw(...).id` which strips witnesses correctly:

```ts
function childTxidFromHex(txHex: string): string {
    const tx = Transaction.fromRaw(hex.decode(txHex), {
        allowUnknownOutputs: true,
        allowUnknownInputs: true,
    });
    return tx.id;
}
```

The reason the old test passed: `buildTestTx()` produces unsigned txs where legacy and segwit serializations are bit-identical. Added a new fixture `buildSegwitTxWithWitness()` that attaches a real witness, with both `expectedTxid` (= `tx.id`) and `expectedWtxid` (= `tx.hash`) so the test now asserts the result equals the txid AND **does not** equal the wtxid:

```ts
expect(result).toBe(expectedTxid);
expect(result).not.toBe(expectedWtxid);
```

Both `ElectrumOnchainProvider.broadcastTransaction` and `WsElectrumChainSource.broadcastPackage` tests now exercise this path. The witness-stripping path would have caught the bug.

### CodeRabbit: watchAddresses cleanups

1. **Partial-failure subscription leak** — `Promise.all` over `subscribeScriptStatus` would reject on the first failure, leaving the already-registered subscriptions dangling on the server with no `stop()` handle for the caller. Fixed: track successful subscriptions in an array and `unsubscribeScriptStatus` them on rejection before rethrowing.

2. **O(n) lookup on every server push** — `scriptHashes.indexOf(scripthash)` ran on every status notification. Replaced with a `Map<string, Uint8Array>` built once at setup; the lookup is now O(1). Notable for wallets watching hundreds of addresses.

3. **Race on concurrent same-scripthash notifications** — two pushes for the same address could both `fetchHistory` in parallel, both compute the same `newTxids` (since the synchronous `known` Set update happens between awaits, but two callers awaiting at the same point would both see the pre-update state on resume), and both emit duplicate events. Fixed via a per-scripthash mutex (`Map<string, Promise<void>>`) that chains each notification onto the previous one's tail. Failures are swallowed at the chain level so one bad notification doesn't stall future ones.

885 unit tests pass (+1 from the regression-blocking witness fixture), prettier clean, typecheck clean.

pietro909 (approved):

PR #451, opened 2026-04-23
fix(worker): enforce one-response-per-message delivery guarantees

### LINKED ISSUE #448: MessageBus lacks handler delivery guarantees — silent failures, discarded results, dead message IDs
## Core Problem

The `MessageBus` message protocol has no delivery guarantee. Multiple code paths can silently drop responses, leaving the client hanging or showing incorrect state. The timeout error is the most visible symptom, but it's one of several reliability gaps.

**Principle:** Every message sent to the service worker MUST produce exactly one final response (success or error). No silent drops, no discarded results.

### 1. Handler result discarded after timeout (most impactful)

```js
// messageBus.js line 268-281
try {
    const response = await this.withTimeout(updater.handleMessage(event.data), tag);
    event.source?.postMessage(response);     // sends result
} catch (err) {
    event.source?.postMessage({ id, tag, error });  // sends error, done
}
```

When `withTimeout` rejects:
- Error is sent to client ✓
- Handler keeps running (documented: "does NOT cancel the underlying work")
- Handler eventually completes → **result thrown on the floor** ✗
- Handler eventually throws → **rejection silently swallowed** ✗

The client can't re-query by message ID because nobody sends the final result.

**Fix:** Save the handler promise reference. After timeout, attach a continuation:
```js
const handlerPromise = updater.handleMessage(event.data);
try {
    const response = await this.withTimeout(handlerPromise, tag);
    event.source?.postMessage(response);
} catch (err) {
    event.source?.postMessage({ id, tag, error: err });
    if (err instanceof ServiceWorkerTimeoutError) {
        const deadline = setTimeout(() => {
            event.source?.postMessage({ id, tag, error: new Error('Operation abandoned') });
        }, 5 * 60_000);
        handlerPromise.then(
            (resp) => { clearTimeout(deadline); if (resp) event.source?.postMessage(resp); },
            (e)    => { clearTimeout(deadline); event.source?.postMessage({ id, tag, error: e }); }
        );
    }
}
```

### 2. Unknown tag → silent drop (line 263-266)

```js
const updater = this.handlers.get(tag);
if (!updater) {
    if (this.debug) console.warn(`[${tag}] unknown message tag, ignoring message`);
    return;  // client gets NOTHING
}
```

Client sends a message with an unrecognized tag → no response, no error. Client hangs until its own timeout fires.

**Fix:** Always respond with an error:
```js
if (!updater) {
    event.source?.postMessage({ id, tag, error: new Error(`Unknown handler tag: ${tag}`) });
    return;
}
```

### 3. Handler returns falsy → silent drop (line 272-274)

```js
if (response) {
    event.source?.postMessage(response);
}
```

If a handler returns `null`, `undefined`, `0`, `''`, or `false`, the response is silently skipped. Client hangs.

**Fix:** Only skip on explicit `undefined`/`void`, not all falsy values. Or always respond:
```js
if (response !== undefined) {
    event.source?.postMessage(response);
} else {
    event.source?.postMessage({ id, tag }); // ack with no payload
}
```

### 4. `event.source` null → response dropped (line 273, 280)

```js
event.source?.postMessage(response);
```

If the client tab/window was detached, `event.source` is null and the response is silently dropped. No logging.

**Fix:** Log when `event.source` is null — indicates a disconnected client, useful for debugging.

### 5. Timeout error uses handler TAG not message TYPE (line 293)

```js
reject(new ServiceWorkerTimeoutError(
    `Message handler timed out after ${this.messageTimeoutMs}ms (${label})`
));
```

`label` is the handler tag (`WALLET_UPDATER`), not the message type (`SETTLE`). Users see "timed out (WALLET_UPDATER)" which is useless for debugging.

**Fix:** Include message type: `"timed out after Xms (SETTLE via WALLET_UPDATER)"`

### 6. No per-operation timeout configuration

`messageTimeoutMs` is global. `SETTLE` needs 120s+, `GET_VTXOS` needs 5s. A single timeout forces bad compromises.

**Fix:** Per-type overrides:
```js
new MessageBus({
    messageTimeoutMs: 30_000,
    timeoutOverrides: { SETTLE: 120_000, ROLLOVER: 120_000 }
})
```

## Client-side impact

`sendMessageStreaming` cleans up after receiving a timeout error — removes the message listener and rejects:
```js
const cleanup = () => {
    clearTimeout(timeoutId);
    navigator.serviceWorker.removeEventListener("message", messageHandler);
};
```

Once the server-side fix (#1) sends follow-up responses with the same message ID, the client needs to optionally keep listening after a timeout error rather than immediately cleaning up.

## Context

- Wallet workarounds PR: https://github.com/arkade-os/wallet/pull/566
- This is the #1 user-facing error in the wallet app
- Wallet currently: increased timeout to 120s + translates errors to friendly messages

## Relevant code

- `src/worker/messageBus.ts` — all issues are in `onMessage()` and `withTimeout()`
- `src/wallet/serviceWorker/wallet.ts` — `sendMessageStreaming()` client-side handling
- `ServiceWorkerTimeoutError` class

Kukks:
Closes #448.

## Summary

The `MessageBus` had no delivery guarantee — six paths could silently drop responses, leaving the caller's message id without a final response. This change enforces the invariant: **every message produces exactly one final response**.

### Fixes applied to `src/worker/messageBus.ts`

1. **Handler result discarded after timeout** — the handler promise is now retained after `withTimeout` rejects. When the handler eventually completes (success or error), the result is delivered under the original message id. If the handler does not complete within `LATE_DELIVERY_GRACE_MS` (5 minutes), an explicit `Operation abandoned` error is sent so listeners still attached do not hang forever.
2. **Unknown tag → silent drop** — now replies `{id, tag, error: Error("Unknown handler tag: ...")}` instead of `return`.
3. **Handler returns null/undefined → silent skip** — now emits an explicit ack envelope `{id, tag}`. Applied to both unicast and broadcast paths.
4. **`event.source` null → silent drop** — centralised via `deliverResponse(...)` which logs in `debug` mode so the drop isn't invisible.
5. **Timeout error label names handler tag, not message type** — errors now read `"Message handler timed out after Xms (SETTLE via WALLET_UPDATER)"` rather than `"(WALLET_UPDATER)"`, so the operation the client triggered is obvious at a glance.
6. **No per-operation timeout configuration** — new `messageTimeoutOverrides?: Record<string, number>` option on `MessageBus`. Keys can be message types (e.g. `SETTLE`) or handler tags (e.g. `WALLET_UPDATER`); message-type matches take precedence over tag matches; unspecified operations use `messageTimeoutMs`.

### Notes

- Client-side keep-listening-after-timeout support (mentioned as a follow-up in the issue) is **out of scope** for this PR. This change is server-side only so the infrastructure is correct even if no client listens for the late response.
- The existing `test/serviceWorker/worker.test.ts` test `ignores messages with unknown tags` encoded the old silent-drop behaviour; it has been updated to assert the new error-response contract.
- **This PR will need rebasing after #446 merges.** Both PRs touch the same broadcast-path (lines 407-412) and unicast-path (lines 449-452) regions of `src/worker/messageBus.ts`. The rebase is mechanical — #446 adds an `isLongRunning()` opt-out around the `withTimeout` call, and this PR changes the call shape; they'll combine cleanly.

## Test plan

- [x] `pnpm run test:unit` — all 859 tests pass (+6 new tests for the fixes, +1 updated for the changed unknown-tag contract).
- [x] `pnpm run lint` — prettier clean.
- [x] TypeScript check — no errors.
- [x] CI passes.

## Summary by CodeRabbit

* **New Features**
  * Configurable per-operation message timeout overrides enabling flexible timeout management for different operation types and scenarios.

* **Bug Fixes**
  * Fixed silent failures by ensuring all service requests receive explicit response envelopes, including acknowledgments.
  * Enhanced timeout error handling with grace-period support for late-arriving responses before final abandonment.
  * Improved consistency of timeout behavior enforcement across wallet initialization and reinitialization cycles.

### DIFF
--- a/src/worker/messageBus.ts
+++ b/src/worker/messageBus.ts
@@ -93,7 +93,9 @@ type Options = {
      * Per-operation timeout overrides. Keys are either message types
      * (e.g. "SETTLE") or handler tags (e.g. "WALLET_UPDATER"). Message-type
      * matches take precedence over tag matches. Unspecified operations use
-     * `messageTimeoutMs`.
+     * `messageTimeoutMs`. These are treated as defaults: any map supplied
+     * via `INITIALIZE_MESSAGE_BUS` overrides per-key and is re-applied on
+     * every (re-)init.
      */
     messageTimeoutOverrides?: Record<string, number>;
     debug?: boolean;
@@ -144,6 +146,12 @@ type Initialize = {
         esploraUrl?: string;
         settlementConfig?: SettlementConfig | false;
         watcherConfig?: Partial<Omit<ContractWatcherConfig, "indexerProvider">>;
+        /**
+         * Page-supplied per-operation timeout map. Keys are message types
+         * (e.g. "SETTLE"). Overrides constructor-supplied
+         * `messageTimeoutOverrides` per-key; re-applied on every init.
+         */
+        messageTimeouts?: Record<string, number>;
     };
 };
 
@@ -151,6 +159,7 @@ export class MessageBus {
     private handlers: Map<string, MessageHandler>;
     private tickIntervalMs: number;
     private messageTimeoutMs: number;
+    private readonly constructorTimeoutOverrides: Record<string, number>;
     private messageTimeoutOverrides: Record<string, number>;
     private lateDeliveries = new Set<LateDelivery>();
     private running = false;
@@ -183,7 +192,8 @@ export class MessageBus {
         this.handlers = new Map(messageHandlers.map((u) => [u.messageTag, u]));
         this.tickIntervalMs = tickIntervalMs;
         this.messageTimeoutMs = messageTimeoutMs;
-        this.messageTimeoutOverrides = messageTimeoutOverrides;
+        this.constructorTimeoutOverrides = { ...messageTimeoutOverrides };
+        this.messageTimeoutOverrides = { ...this.constructorTimeoutOverrides };
         this.debug = debug;
         this.buildServicesFn = buildServices ?? this.buildServices.bind(this);
     }
@@ -317,6 +327,13 @@ export class MessageBus {
                 )
             );
         }
+        // Recompute the active timeout map from scratch so a prior init's
+        // keys cannot linger after re-init with a smaller map.
+        this.messageTimeoutOverrides = {
+            ...this.constructorTimeoutOverrides,
+            ...(config.messageTimeouts ?? {}),
+        };
+
         const services = await this.buildServicesFn(config);
         // Start all handlers
         for (const updater of this.handlers.values()) {
--- a/src/worker/messageBus.ts
+++ b/src/worker/messageBus.ts
@@ -634,10 +634,21 @@ export class MessageBus {
         messageType: string | undefined,
         handlerTag: string
     ): number {
-        if (messageType && messageType in this.messageTimeoutOverrides) {
+        if (
+            messageType &&
+            Object.prototype.hasOwnProperty.call(
+                this.messageTimeoutOverrides,
+                messageType
+            )
+        ) {
             return this.messageTimeoutOverrides[messageType];
         }
-        if (handlerTag in this.messageTimeoutOverrides) {
+        if (
+            Object.prototype.hasOwnProperty.call(
+                this.messageTimeoutOverrides,
+                handlerTag
+            )
+        ) {
             return this.messageTimeoutOverrides[handlerTag];
         }
         return this.messageTimeoutMs;
--- a/src/worker/messageBus.ts
+++ b/src/worker/messageBus.ts
@@ -79,6 +79,13 @@ type Options = {
     messageHandlers: MessageHandler[];
     tickIntervalMs?: number;
     messageTimeoutMs?: number;
+    /**
+     * Per-operation timeout overrides. Keys are either message types
+     * (e.g. "SETTLE") or handler tags (e.g. "WALLET_UPDATER"). Message-type
+     * matches take precedence over tag matches. Unspecified operations use
+     * `messageTimeoutMs`.
+     */
+    messageTimeoutOverrides?: Record<string, number>;
     debug?: boolean;
     buildServices?: (config: Initialize["config"]) => Promise<{
         arkProvider: ArkProvider;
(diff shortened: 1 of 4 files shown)


src/worker/messageBus.ts
```diff
@@ -490,6 +573,118 @@ export class MessageBus {
...
+                    ),
+                },
+                context
+            );
+        }, LATE_DELIVERY_GRACE_MS);
+
+        handlerPromise.then(
+            (response) => {
+                self.clearTimeout(deadline);
+                this.deliverResponse(source, response ?? { id, tag }, context);
```
pietro909, line 752: If `deadline` resolves before `handlerPromise` returns (unlikely, it's 5 mins, but let's suppose it happens) I think two deliveries will happen:
- the one at line 659
- this one

To be verified

src/worker/messageBus.ts
```diff
@@ -490,6 +573,118 @@ export class MessageBus {
...
+
+    /**
+     * Resolve the timeout for an operation. Message-type overrides take
+     * precedence over handler-tag overrides, with the bus-wide default
+     * (`messageTimeoutMs`) as the final fallback.
+     */
+    private resolveTimeoutMs(
+        messageType: string | undefined,
+        handlerTag: string
+    ): number {
```
pietro909, line 650: This will create a drift with the configuration introduced in https://github.com/arkade-os/ts-sdk/pull/371 

Now the consumers need to set the same timeout in two places: if they forget one, the shortest will fire. Not sure what to suggest here: either use the page-side ones and set them in INITIALIZE_MESSAGE_BUS or drop both. The general direction seems to be that timeouts in general aren't a good fit for our use case.

src/worker/messageBus.ts
```diff
@@ -490,6 +573,118 @@ export class MessageBus {
...
+    ): string {
+        return messageType ? `${messageType} via ${handlerTag}` : handlerTag;
+    }
+
+    /**
+     * Post a response to the originating client. When `source` is null
+     * (client tab closed, detached frame, etc.) the response cannot be
+     * delivered; we log the drop in debug mode so it is not invisible.
+     */
+    private deliverResponse(
```
pietro909, line 690: Worh replacing all the `.postMessage()`  with this helper? I think PING and init-related messages still use it.

pietro909 (approved):
Can merge with the caveat that now it's consumer's responsibility to decide whether an infinite loading is the right UX for some failures.

PR #461, opened 2026-04-28
test(e2e): integration tests for ElectrumOnchainProvider over regtest

Kukks:
## Summary

Adds end-to-end integration tests for \`ElectrumOnchainProvider\` against the nigiri regtest stack. Exercises every method on the \`OnchainProvider\` interface against a real Electrum backend so any divergence from \`EsploraProvider\` (which already has e2e coverage) surfaces side-by-side.

Mirror of \`test/e2e/onchain.test.ts\` plus electrum-specific coverage:

| Method | What is asserted |
|---|---|
| \`getChainTip\` | Header subscription is **reused** across calls (no new server-side subscription on each call) |
| \`getFeeRate\` | Accepts both \`undefined\` (regtest \`-1\`) and a positive sat/vB integer |
| \`getCoins\`, \`getTransactions\`, \`getTxStatus\` | Fund a fresh address via \`nigiri faucet\`, verify all three see it |
| \`broadcastTransaction\` | Alice-to-Bob single-tx roundtrip via \`OnchainWallet.send\` |
| \`getTxOutspends\` | Spent-output detection on a partially-consumed tx |
| \`watchAddresses\` | Funded-tx delivery via the scripthash subscription |
| \`WsElectrumChainSource.fetchHistories\` | Batch round-trip |

## Plumbing

To talk to nigiri's bitcoin Electrum endpoint from a JS client (which can't open raw TCP), this PR uses the \`electrum-ws\` websocat bridge from [vulpemventures/nigiri#258](https://github.com/vulpemventures/nigiri/pull/258), now **merged to nigiri master**. Submodule points at arkade-regtest's plain \`master\`.

## Drive-by fix: dropped verbose-tx dependency for electrs compatibility

The first round of CI on this branch failed because the provider was calling \`blockchain.transaction.get\` with \`verbose=true\`. That endpoint is Fulcrum-only — **electrs explicitly does not support it** (returns \`"verbose transactions are currently unsupported"\`). The nigiri stack uses electrs.

Refactored the three call sites that needed verbose data (\`getTransactions\`, \`getTxStatus\`, \`watchAddresses\`) to use only methods every Electrum server supports:

- Heights come from \`scripthash.get_history\` (which we already had) and \`transaction.get_merkle\` for individual lookups.
- Output values are derived from raw tx hex (parsed via \`@scure/btc-signer\`) — exact bigints, no float rounding. This was previously a footgun in the verbose-tx path's \`Math.round(value * 1e8)\`.
- Block times come from parsed block headers (batched per unique height).

The \`fetchVerboseTransaction(s)\` methods on \`WsElectrumChainSource\` remain public for callers who explicitly target Fulcrum, but the provider no longer depends on them. **Side effect: the SDK now works against electrs-based Electrum servers** (Blockstream, mempool.space, self-hosted electrs) where it previously couldn't.

## Out of scope

The 1P1C \`broadcast_package\` path isn't exercised in regtest because nigiri uses electrs (no \`broadcast_package\` support). Unit-test coverage of that path lives in \`test/electrum.test.ts\`. Once a Fulcrum-based regtest stack is available, the missing cases can be added in a follow-up.

## Test plan

- [x] \`pnpm regtest\` (clean start with the submodule's nigiri pin)
- [x] \`pnpm test:integration\` — all e2e tests pass, including \`test/e2e/electrum.test.ts\`
- [x] \`pnpm test:unit\` clean (1017 tests pass after the verbose-removal refactor)
- [x] \`pnpm lint\` clean

## Summary by CodeRabbit

* **New Features**
  * More accurate on-chain confirmation info and raw-transaction-based history for precise amounts and timestamps.
* **Bug Fixes**
  * More reliable address-watch delivery (notifications retried until successfully delivered) and improved error resilience for header lookups and confirmations.
* **Tests**
  * New end-to-end and unit tests covering Electrum provider flows, mempool, history fetching, broadcasting and wallet interactions.
* **Chores**
  * Regtest environment reference updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,6 +227,91 @@ await wallet.send({ address: 'ark1q...', amount: 1000 })
 
 Identities without `signMultiple` continue to work unchanged — each checkpoint is signed individually via `sign()`.
 
+### Onchain Providers
+
+Wallets read onchain state (UTXOs, transactions, fee rates, chain tip) through an `OnchainProvider`. The SDK ships with two implementations and a single transport-agnostic interface so you can swap them without touching wallet code.
+
+| Provider | Transport | When to use |
+|---|---|---|
+| `EsploraProvider` | REST/HTTP (mempool.space-compatible) | Default for browser wallets, public mempool deployments, simple integrations. Both atomic 1P1C package broadcast and outspends are first-class. |
+| `ElectrumOnchainProvider` | WebSocket (Electrum protocol) | Self-hosted nodes (Fulcrum, electrs), low-latency subscriptions, environments where you control the backend. Required if you need to talk to an Electrum server directly. |
+
+If you don't pass a provider explicitly, `OnchainWallet` and `Wallet.create({ ... })` both default to `EsploraProvider` pointing at the URL in `ESPLORA_URL[networkName]`.
+
+#### Default URLs
+
+The SDK ships with reachable defaults for each network — bitcoin, signet, and mutinynet point at Ark Labs–operated deployments; testnet falls back to mempool.space; regtest assumes a local nigiri stack.
+
+```typescript
+import {
+  ESPLORA_URL,        // Record<NetworkName, string>
+  ELECTRUM_WS_URL,    // Record<NetworkName, string>
+  ELECTRUM_TCP_HOST,  // Record<NetworkName, string | null> — informational
+} from '@arkade-os/sdk'
+
+ESPLORA_URL.bitcoin       // "https://mempool.arkade.sh/api"
+ESPLORA_URL.signet        // "https://mempool.signet.arkade.sh/api"
+ESPLORA_URL.mutinynet     // "https://mempool.mutinynet.arkade.sh/api"
+
+ELECTRUM_WS_URL.bitcoin   // "wss://electrum.arkade.sh"
+ELECTRUM_WS_URL.signet    // "wss://electrum.signet.arkade.sh"
+ELECTRUM_WS_URL.mutinynet // "wss://electrum.mutinynet.arkade.sh"
+```
+
+#### Using Esplora (default)
+
+```typescript
+import { EsploraProvider, ESPLORA_URL, OnchainWallet } from '@arkade-os/sdk'
+
+// Use the default URL for the network
+const provider = new EsploraProvider(ESPLORA_URL.bitcoin)
+
+// Or pass nothing — OnchainWallet picks the default for you
+const wallet = await OnchainWallet.create(identity, 'bitcoin')
+
+// Or override with your own mempool/esplora instance
+const customProvider = new EsploraProvider('https://my-esplora.example/api')
+```
+
+#### Using Electrum (WebSocket)
+
+```typescript
+import { ElectrumWS } from 'ws-electrumx-client'
+import {
+  ElectrumOnchainProvider,
+  ELECTRUM_WS_URL,
+  OnchainWallet,
+  networks,
+} from '@arkade-os/sdk'
+
+const ws = new ElectrumWS(ELECTRUM_WS_URL.bitcoin)
+const provider = new ElectrumOnchainProvider(ws, networks.bitcoin)
+
+const wallet = await OnchainWallet.create(identity, 'bitcoin', provider)
+
+// Remember to close the connection when you're done
+await provider.close()
+```
+
+#### Atomic 1P1C package broadcast (TRUC / BIP 431)
+
+Both providers expose `broadcastTransaction(...txs)` that accepts either a single tx or a 1P1C package (parent first, child last). The package path is **atomic** — the parent doesn't have to independently meet mempool minfee, which is the point of TRUC relay.
+
+The Electrum provider implements this via `blockchain.transaction.broadcast_package` (Fulcrum ≥ 1.10 backed by bitcoind ≥ v28). **There is no fallback to sequential broadcast**: if the server doesn't support `broadcast_package`, the call surfaces a clear error so you can route through a different provider rather than have TRUC packages silently fail at the parent step. Ark Labs Fulcrum deployments at `electrum.arkade.sh` (and the `*.signet` / `*.mutinynet` variants) all support it.
+
+#### Server compatibility notes
+
+`ElectrumOnchainProvider` is built around methods supported by both **Fulcrum** and **electrs** (the two main Electrum server implementations):
+
+- ✅ `blockchain.scripthash.{listunspent, get_history, subscribe}`
+- ✅ `blockchain.transaction.{get, get_merkle, broadcast}`
+- ✅ `blockchain.block.header`, `blockchain.headers.subscribe`
+- ✅ `blockchain.estimatefee`, `blockchain.relayfee`
+- ⚠️ `blockchain.transaction.broadcast_package` — **Fulcrum-only**. Required for atomic 1P1C; the provider throws a descriptive error against electrs.
+- ❌ The provider does **not** call `blockchain.transaction.get` with `verbose=true` (Fulcrum-only and rejected by electrs); confirmation status is derived from `transaction.get_merkle` + raw block headers instead.
+
+Output amounts are derived from parsed raw transaction bytes (exact bigints), never from floating-point `value` fields — protocol-level money handling shouldn't depend on `Math.round(value * 1e8)`.
+
 ### Receiving Bitcoin
 
 ```typescript
--- a/src/index.ts
+++ b/src/index.ts
@@ -104,6 +104,8 @@ import {
     ExplorerTransaction,
 } from "./providers/onchain";
 import {
+    ELECTRUM_TCP_HOST,
+    ELECTRUM_WS_URL,
     ElectrumOnchainProvider,
     WsElectrumChainSource,
 } from "./providers/electrum";
@@ -288,6 +290,8 @@ export {
     // Providers
     ESPLORA_URL,
     EsploraProvider,
+    ELECTRUM_WS_URL,
+    ELECTRUM_TCP_HOST,
     ElectrumOnchainProvider,
     WsElectrumChainSource,
     RestArkProvider,
--- a/src/providers/electrum.ts
+++ b/src/providers/electrum.ts
@@ -698,44 +698,38 @@ export class ElectrumOnchainProvider implements OnchainProvider {
         const rawTxs = await this.chain.fetchTransactions(txids);
         const rawHexByTxid = new Map(rawTxs.map((t) => [t.txID, t.hex]));
 
-        // De-duplicated header lookup. Try the batch first (fast path:
-        // 1 round-trip for the common case); fall back to per-height
-        // calls only if the batch rejects, because electrs sometimes
-        // races: a tx can show as confirmed via listunspent before its
-        // block header is indexable, which makes a batch including that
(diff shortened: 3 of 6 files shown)


Kukks:
Pushed `88ec2625` addressing the review feedback.

### CodeRabbit findings — all addressed

**1. `fetchTxMerkle` blanket catch** ([electrum.ts:215](https://github.com/arkade-os/ts-sdk/pull/461/files#diff-electrum.tsR215))
The previous `catch {}` silently mapped every failure — including auth errors, network outages, malformed payloads — to "mempool / unconfirmed". Now matches the specific "not yet in a block" wording electrs/Fulcrum return and rethrows everything else. Added new `isTxNotInBlockError` helper. Regression test: `getTxStatus` propagates non-"not in block" errors.

**2. `watchAddresses` marks txids seen before delivery** ([electrum.ts:863-876](https://github.com/arkade-os/ts-sdk/pull/461/files#diff-electrum.tsR863))
Reordered: `historyToExplorerTxs` + `eventCallback` now run **before** the `knownTxids` update. If either throws, the txid stays unmarked and the next notification redelivers. Regression test: simulated downstream failure on first notification, then verify the same txid is delivered on the second.

**3. `buildExplorerTx` swallows parse errors with empty vout** ([electrum.ts:734-757](https://github.com/arkade-os/ts-sdk/pull/461/files#diff-electrum.tsR734))
Replaced the silent `catch` with a thrown `Failed to parse raw tx for ${txid}` error. Empty vout would hide real outputs — for money handling, failing loud is the right default. Regression test: malformed hex from electrs causes `getTransactions` to throw with the offending txid in the message.

### Nitpick — also addressed

**4. e2e tests use fixed delays instead of `waitFor` polling**
Switched all six `delay(FAUCET_PROPAGATION_MS)` calls in `test/e2e/electrum.test.ts` to `waitFor` against the real condition (coin presence, balance, outspends). Aligns with the rest of the e2e suite (`settlement.test.ts`, `ark.test.ts`) and tightens the wall-clock under CI load.

### arkanaai feedback

All three minor notes were either already addressed or non-blocking:
- **Submodule pin**: bumped to arkade-regtest master in `010c3872` after vulpemventures/nigiri#258 merged.
- **`decodeP2trScript` helper duplicates `encodeAddress`**: noted; not worth exporting the private method just for one e2e test, leaving it.
- **Two WS connections**: intentional for test isolation.

### Local verification

- `pnpm test:unit` — 1020 tests pass (45 in `test/electrum.test.ts`, 3 new for the three production fixes)
- `pnpm lint` — clean
- TypeScript check — clean

Kukks:
Pushed `cb1ebe89` to fix the integration regression.

### What broke

Both e2e failures traced to the same electrs race: a tx can appear in `listunspent` with `height > 0` *before* its block header is indexable, so `block.header(N)` returns `missingheight` and our previous code surfaced that as a hard failure.

The previous (verbose-tx) path tolerated this by falling back to `vtx.blocktime || vtx.time || 0` when block_time data was missing. The new (header-fetching) path was stricter than the old one — that's the actual regression.

### Two narrow fixes

**Production: `historyToExplorerTxs` best-effort headers**

If the batch `fetchBlockHeaders` call fails wholesale (one bad height poisons the batch), fall back to per-height `fetchBlockHeader` calls under `Promise.allSettled`. Heights that still won't resolve simply land in the result map with no entry, and `buildExplorerTx` falls back to `block_time = 0`. This preserves the *exact* semantic the old verbose-tx path had — only `block_time` degrades; txid + confirmation status are still authoritative.

This is the *opposite* shape from the parse-error finding addressed in the previous commit (which kept `fail loud`). Header indexing lag is a transient infra race, not corruption — so it gets best-effort tolerance.

**E2E: wrap `getTxStatus` in `waitFor`**

`getTxStatus` calls `fetchBlockHeader` directly and intentionally throws on failure (a single-tx caller can/should retry). The e2e test now retries via `waitFor`, swallowing only the specific `missingheight` / `not in block` wording, until electrs settles. Also bumped the `fetchHistories` test timeout to 40s (was 10s, smaller than `waitFor`'s default 25s budget) and relaxed its `waitFor` to `>= 1` coin per address (the assertion only needs that).

### New regression test

Added `falls back to per-height header fetches when the batch fails (electrs index lag race)` covering the production fix — mocks a batch rejection, then a mixed pass/fail of individual fetches, asserts both txs surface (one with block_time, one with 0).

### Local verification
- 46 electrum unit tests pass (was 45)
- Full unit suite: 1020 tests pass, lint clean, typecheck clean

Kukks:
Pushed `10505584` — same policy now applied to `getTxStatus`.

The previous fix made `historyToExplorerTxs` resilient to electrs's missingheight race but left `getTxStatus` throwing. In CI, the race window outlasted `waitFor`'s 25s budget, so the e2e wrapper kept polling until the test timed out.

`getTxStatus` now applies the same policy as `historyToExplorerTxs`: confirmation status and height are authoritative; only `block_time` degrades to 0 when `block.header(N)` raises `missingheight`. The new `isMissingHeightError` helper narrows the catch so genuine errors (auth, network, malformed) still propagate.

E2e test simplifies back to a direct call — no more waitFor wrapper needed.

### Two new unit tests
- `returns confirmed=true with blockTime=0 when block.header races with index lag (missingheight)`
- `propagates non-'missingheight' errors from block.header`

### Local verification
- 48 electrum unit tests pass (was 46)
- Full suite passing, lint + typecheck clean

Kukks:
Pushed `f209bcc8` — found the actual root cause.

The library (`ws-electrumx-client`) implements `batchRequest` as `Promise.all` over individual request promises. When ONE element rejects, `Promise.all` rejects with that error — but the OTHER in-flight requests stay pending. When their responses arrive (often as more error responses), the library rejects them via `request.reject(new Error(errorMsg))` — and **nobody is awaiting them**. Vitest catches the unhandled rejections and surfaces them as test failures, with the stack pointing at the library's `onMessage` (where the unowned `new Error()` was constructed).

This explains why the previous fixes all failed: my application-level try/catch around `await batchRequest()` correctly handled the FIRST rejection, but couldn't suppress the orphaned (N-1) rejections from the rest of the batch. Tolerating `missingheight` at the batch level still left the library generating unhandled rejections behind the scenes.

Switched `historyToExplorerTxs` to per-height `ws.request` calls under `Promise.allSettled` — every promise has an explicit handler, no orphans. `fetchTransactions` (still uses batchRequest) is unaffected in practice because raw tx hex is always available for txs already in history; left as-is to keep the diff focused.

### Trade-off

N parallel `ws.request` calls vs 1 batch call:
- Wall-clock: same (allSettled fires all in parallel; one round-trip-time)
- Wire overhead: slightly higher request-side framing
- Robustness: bulletproof against partial-batch failures

For typical history sizes (<20 unique heights) the cost is negligible.

### Test changes
- Updated 3 unit tests that previously mocked the header batch path to use per-request mocks
- Added a new test asserting `batchRequest` is called exactly once (for raw tx hex) and never for headers — regression guard
- 49 electrum unit tests pass (was 48)

Kukks:
Pushed `54ef4d47` — codebase-level fix.

Previous attempt only patched `historyToExplorerTxs`. The same orphan-rejection leak existed at every other batch site: `fetchTransactions`, `fetchVerboseTransactions`, `fetchHistories`, `fetchBlockHeaders`, and the two batches in `getTxOutspends`. Any of them could surface electrs's `missingheight` (or any error) and leak (N-1) sibling rejections.

### What changed

Added `WsElectrumChainSource.safeBatchRequest` — a drop-in replacement for `ws.batchRequest`. Issues each request through `ws.request` (own promise lifecycle) and aggregates via `Promise.allSettled` (every promise has an explicit handler), surfacing the first rejection. Same wall-clock cost as the library batch (parallel sends), no orphan rejections.

Replaced all six call sites:
- `fetchTransactions`
- `fetchVerboseTransactions`
- `fetchHistories`
- `fetchBlockHeaders`
- `getTxOutspends` (×2)

The library's `batchRequest` is now **banned in production code**. New regression test asserts `wsMock.batchRequest` is never called from the provider's hot path — future authors who reintroduce it will see a red test.

### Trade-off

Sequential per-request framing instead of one batched JSON-RPC frame. Wall-clock identical (parallel sends), wire overhead slightly higher on the request side. For typical sizes (<20 elements) the overhead is negligible.

### Tests

- Added `mockBatch(mock, responses)` helper that queues N sequential responses on `wsMock.request` (mirroring how safeBatchRequest consumes them)
- Updated 25 batch mock sites
- 49 electrum unit tests pass (was 49)
- Full suite: 1024 passing, lint clean, typecheck clean

Kukks:
All green on `9b7cddd1` — every e2e test passes, no skips.

### What it took

After many false-starts, the actual root causes turned out to be two narrow infrastructure races, not the library batch leak I'd been chasing:

1. **`fetchTxMerkle` only narrowed its catch to "not in a block" wording.** electrs returns literally `missingheight` from `transaction.get_merkle` for the same race that `block.header(N)` hits — confirmed via listunspent but block index not yet ready. The narrow catch let the throw escape `getTxStatus` and surface as the "missingheight" error reported as unhandled. Fixed by accepting both `isTxNotInBlockError` and `isMissingHeightError` shapes as "not finalized" → returns `{confirmed: false}` so the caller polls again. (`e66fd6d5`)

2. **The `fetchHistories` test threw inside `waitFor` on a transient WS request timeout.** Under CI load electrs occasionally wedges for the library's full 10s `REQUEST_TIMEOUT` on a single `get_history` call. `waitFor` doesn't retry on throws — only on `return false`. Switched the wait body to drive `fetchHistories` directly (same code path the assertion uses) AND catch transient `request timeout`/`missingheight` errors as "not ready yet" with a return-false. (`dfe7d571` + `9b7cddd1`)

### Production wins this PR delivered along the way

- **Drop verbose-tx dependency** — provider works with electrs (Blockstream, mempool.space, self-hosted) where it previously couldn't.
- **Float-precision fix** — output values come from parsed raw tx bytes, not `Math.round(value * 1e8)`.
- **`safeBatchRequest`** — bans the leaky library `batchRequest` codebase-wide; every internal callsite uses `Promise.allSettled`-based aggregation.
- **`missingheight` tolerance** — `fetchTxMerkle`, `getTxStatus`, and `historyToExplorerTxs` all degrade gracefully on the electrs index-lag race instead of throwing.
- **Parse errors propagate** — `buildExplorerTx` no longer hides outputs behind an empty vout on malformed hex.
- **`watchAddresses` fixes** — partial-failure subscription cleanup, O(1) scripthash lookup, dedupe-after-delivery, per-scripthash mutex against duplicate emits.
- **Default URLs** — `ESPLORA_URL` and new `ELECTRUM_WS_URL` / `ELECTRUM_TCP_HOST` maps point at Ark Labs Fulcrum 2.1 deployments for bitcoin/signet/mutinynet.
- **TRUC 1P1C support** — atomic `broadcast_package` with no fallback (verified live against `wss://electrum.arkade.sh`).

### Final test count

- **50 electrum unit tests** (was 31 at the start of #450)
- **7 e2e tests, all passing** (chainTip, feeRate, fresh address tracking, single-tx broadcast, outspends, watchAddresses, fetchHistories, OnchainWallet roundtrip)
- **1024 total unit tests passing**, lint + typecheck clean

Kukks:
@pietro909 can you give this a test? We now have a different default esplora instance urls.  Ideally we switch to electrum instead.

PR #462, opened 2026-04-28
fix: ContractWatcher vtxo event + ContractVtxo type

louisinger:
* set taproot fields before emitting vtxo events using the existing `extendVirtualCoinForContract` func
* fix the ContractVtxo type to make the fields optional

@pietro909 @altafan please review

## Summary by CodeRabbit

* **Bug Fixes**
  * Contract events now emit richer, consistently enriched VTXO data with safer fallbacks and better logging when enrichment fails.
  * Delegation now excludes ineligible VTXOs (no matching delegate tap) and returns empty results when none qualify; delegation is only attempted when the delegator is available.
* **Refactor**
  * Internal VTXO/contract handling streamlined for more consistent downstream behavior.

### DIFF
--- a/src/contracts/contractWatcher.ts
+++ b/src/contracts/contractWatcher.ts
@@ -1,5 +1,6 @@
 import { IndexerProvider, SubscriptionResponse } from "../providers/indexer";
 import { VirtualCoin } from "../wallet";
+import { extendVirtualCoinForContract } from "../wallet/utils";
 import { WalletRepository } from "../repositories/walletRepository";
 import {
     Contract,
@@ -765,18 +766,25 @@ export class ContractWatcher {
         if (!this.eventCallback) return;
         const state = this.contracts.get(contractScript);
         if (!state) return;
+
+        const extended = [];
+        for (const v of vtxos) {
+            try {
+                const extendedVtxo = extendVirtualCoinForContract(
+                    v,
+                    state.contract
+                );
+                extended.push({ ...extendedVtxo, contractScript });
+            } catch {
+                extended.push({ ...v, contractScript });
+            }
+        }
+
         switch (eventType) {
             case "vtxo_received":
                 this.eventCallback({
                     type: "vtxo_received",
-                    vtxos: vtxos.map((v) => ({
-                        ...v,
-                        contractScript,
-                        // These fields may not be available from basic VirtualCoin
-                        forfeitTapLeafScript: undefined as any,
-                        intentTapLeafScript: undefined as any,
-                        tapTree: undefined as any,
-                    })),
+                    vtxos: extended,
                     contractScript,
                     contract: state.contract,
                     timestamp,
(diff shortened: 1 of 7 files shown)


src/contracts/contractWatcher.ts
```diff
@@ -765,18 +766,25 @@ export class ContractWatcher {
...
+        const extended = [];
+        for (const v of vtxos) {
+            try {
+                const extendedVtxo = extendVirtualCoinForContract(
+                    v,
+                    state.contract
+                );
+                extended.push({ ...extendedVtxo, contractScript });
+            } catch {
+                extended.push({ ...v, contractScript });
```
pietro909, line 780: Wouldn't this block propagate the bug (ie: taptree undefined?) 
What about `console.error` and skip the VTXO entirely? It will give us better visibility of the issue without failing the whole list of coins.

src/wallet/vtxo-manager.ts
```diff
@@ -1104,11 +1104,17 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
                         console.error("Error renewing VTXOs:", e);
                     });
                 }
-                delegatorManager
-                    ?.delegate(event.vtxos, destination)
-                    .catch((e) => {
-                        console.error("Error delegating VTXOs:", e);
-                    });
+                // delegate only for "delegate" contract
+                if (delegatorManager && event.contract.type === "delegate") {
```
pietro909: Is this correct? We only ever delegate VTXOs that are at the delegate contract (aka the address).
In this case, how do we renew the ones at the default contract? Is the auto-renewal active with delegates on?

src/contracts/contractWatcher.ts
louisinger: current master behaviour is to broadcast them anyway so I'd keep it as it is now cause it may be a kind of "breaking change". also sometimes for custom contract you want the watcher to emit event anyway as a notification?

src/wallet/vtxo-manager.ts
Kukks: Donmt do this here. Do it inside `delegatorManager.delegate`

louisinger: the ones on the default contract can't be delegated cause it doesn't contain the delegator public key, am I right?

louisinger: @Kukks delegateManager doesn't take the event as argument so it means breaking. that's the "hotfix" to make current version delegates funds

Kukks: >Is the auto-renewal active with delegates on 


It should always be active

Kukks: Yes, pass a ContractVtxo[] to delegate from now on.

src/contracts/contractWatcher.ts
pietro909: The code here fixes a bug, and the bug is about broadcasting broken VTXOs. I'd argue that either a VTXO is fit to broadcast or is broken. The `catch` block does what the bugged code did: is there a legit scenario for it? If the answer is "yes" then our typesystem isn't capturing this nuance.

I'm leaning more on the safe side: let's log an error at least if you believe that there is a realistic scenario for it. It will help us understand what's happening if this causes issues down the line.

WDYT?

pietro909 (approved):

PR #460, opened 2026-04-28
feat: default to bitcoin mainnet + arkade.computer

gringokiwi:
- default `ArkAddress` human readable prefix to 'ark'
- default `OnchainWallet` network name to 'bitcoin'
- export `DEFAULT_ARKADE_SERVER_URL` from wallet/index
- default `arkServerUrl` to `DEFAULT_ARKADE_SERVER_URL` in `ReadonlyWallet`
- make `arkServerUrl` optional for `ServiceWorkerWalletOptions`
- default `arkServerUrl` to `DEFAULT_ARKADE_SERVER_URL` in `initWalletPayload` and `busInitConfig` in `ServiceWorkerWallet`

## Summary by CodeRabbit

* **New Features**
  * Wallets default to the "bitcoin" network and a standard address prefix when not specified.
  * Arkade server URL is optional and now falls back to a default Arkade URL.
* **Bug Fixes / Reliability**
  * Wallet initialization no longer errors when server/provider info is omitted.
* **Tests**
  * Added coverage verifying default HRP, network, and server URL fallbacks.

PR #464, opened 2026-04-29
fix: ContractWatcher follow-ups from #462

pietro909:
## Summary

Three small follow-ups to #462, addressing review notes that didn't block merge.

- **`fix(contract-watcher)`** — type the `extended` accumulator as `ContractVtxo[]` so future drift gets caught at compile time, and include `txid:vout` + the caught error in the extend-failure warn log so the catch-fallback path is greppable in production.
- **`refactor(contracts)`** — introduce `PersistedContractVtxo = ExtendedVirtualCoin & { contractScript }` and use it in the internal sync/save plumbing. Drops the four `as ExtendedVirtualCoin[]` casts in `contractManager.ts`/`wallet.ts`/`delegator.ts`. The delegator cast is replaced with an `isAnnotated()` type guard that narrows `ContractVtxo` only when all three taproot fields are present. `ContractWithVtxos.vtxos` is narrowed to `PersistedContractVtxo[]` (refinement; existing consumers compile unchanged).
- **`test(contract-watcher)`** — two new cases driven through the public subscription path:
  - a delegate-contract update produces a vtxo whose `tapTree` equals `DelegateVtxo.Script.encode()` and differs from the default tapscript, proving the right handler enriched the vtxo.
  - a contract type with no registered handler emits the raw vtxo with `contractScript` set and surfaces a `console.warn` carrying `txid:vout`.

## Test plan

- [x] `pnpm tsc --noEmit`
- [x] `pnpm test:unit`
- [x] `pnpm lint` (pre-commit)

## Summary by CodeRabbit

* **Refactor**
  * Improved type safety and consistency for contract virtual outputs across contract, watcher, and wallet flows.
  * Strengthened selection logic for delegator-eligible outputs and simplified wallet initialization/resolution behavior.
  * Enhanced error handling and logging for virtual output enrichment failures.

* **Tests**
  * Added tests covering enrichment paths, delegate verification, and unregistered-contract fallbacks.

* **Documentation**
  * Updated README examples to rely on wallet defaults for simpler setup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/ts-sdk/pull/464)


louisinger (approved):

src/wallet/wallet.ts
```diff
@@ -233,24 +233,15 @@ export class ReadonlyWallet implements IReadonlyWallet {
...
-        if (!arkServerUrl) {
-            throw new Error("Could not determine arkServerUrl from provider");
-        }
+            config.arkProvider ?? new RestArkProvider(arkadeServerUrl);
 
         // Use provided indexerProvider instance or create a new one
         // indexerUrl defaults to arkServerUrl if not provided
-        const indexerUrl = config.indexerUrl || arkServerUrl;
+        const indexerUrl = config.indexerUrl || arkadeServerUrl;
         const indexerProvider =
```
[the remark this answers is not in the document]
pietro909: fixed in [d1a72c5](https://github.com/arkade-os/ts-sdk/pull/464/commits/d1a72c5dd4c62e2102c4ae06bf716d55a136a40b) please review it again

src/contracts/types.ts
```diff
@@ -82,6 +82,19 @@ export type ContractVtxo = VirtualCoin &
...
+/**
+ * A contract vtxo with a fully-populated taproot annotation.
+ *
+ * Produced by paths that go through `annotateVtxos` or repository reads.
+ * Strictly narrower than {@link ContractVtxo} — useful where downstream
+ * code requires the `ExtendedVirtualCoin` shape (e.g. `saveVtxos`,
+ * forfeit tx construction) and the compiler should enforce that the
+ * annotation has actually run.
+ */
+export type PersistedContractVtxo = ExtendedVirtualCoin & {
```
Kukks: not a big fan of this type. ContractVtxo already had it?

pietro909: I've renamed it to `ExtendedContractVtxo` following a similar pattern we have for Coin and Vtxo.
The reason for this type is to have the guarantee of `TapLeaves` and `EncodedVtxoScript` (via annotation).
It removes the need for unsafe type casts and redundant runtime checks.

ghost (approved):
## Arkana Code Review — ts-sdk#464

**Verdict: APPROVE** ✅

Reviewed all 8 changed files, cross-checked type flow through the codebase, verified no cross-repo consumers of the changed types, and inspected git history for context (#462, #460).

---

### Summary

Clean follow-up. Three themes: (1) replace unsafe `as` casts with a proper type hierarchy, (2) harden the delegator eligibility filter with a real type guard, (3) defensive indexer-URL derivation when a custom `arkProvider` is injected. Tests cover the subscription-path enrichment that #462 introduced but didn't test.

### Findings

**✅ `ExtendedContractVtxo` type — correct and well-placed**

`src/contracts/types.ts:85-99` — The type mirrors the existing `ExtendedVirtualCoin` / `VirtualCoin` split. `getVtxosForContracts` (contractManager.ts:753) spreads vtxos from `getVtxosForContract`, which returns `ExtendedVirtualCoin[]` (already persisted + annotated), so the return type `ExtendedContractVtxo[]` is sound. The `ContractWithVtxos.vtxos` narrowing (types.ts:286) is a refinement — `ExtendedContractVtxo` is structurally assignable to `ContractVtxo`, so downstream consumers reading the field still compile. No cross-repo consumers found (grepped all repos under `/root/arkana/repos`).

**✅ `isAnnotated()` type guard — real safety improvement**

`src/wallet/delegator.ts:504-509` — Replaces the old unchecked `.map(v => v as ExtendedVirtualCoin)` at line 97 (master). The guard checks exactly the three fields that `TapLeaves & EncodedVtxoScript` requires: `tapTree`, `forfeitTapLeafScript`, `intentTapLeafScript`. `extraWitness` is optional on `ExtendedVirtualCoin` so correctly omitted from the check. The narrowed type `ContractVtxo & ExtendedVirtualCoin` is structurally equivalent to `ExtendedContractVtxo`. Good.

**✅ `contractWatcher.ts` accumulator typing + log improvement**

`src/contracts/contractWatcher.ts:777,785` — `const extended: ContractVtxo[]` is the correct type here (not `ExtendedContractVtxo[]`) because the catch-fallback path pushes `{ ...v, contractScript }` without taproot fields. The improved warn log with `${v.txid}:${v.vout}` and the error object makes production debugging viable.

**✅ `wallet.ts` indexer URL derivation**

`src/wallet/wallet.ts:257-280` (PR branch) — The old code silently fell back to `DEFAULT_ARKADE_SERVER_URL` for the indexer when a custom `arkProvider` was injected without an explicit `indexerUrl`. A user injecting a custom arkProvider pointing at their own server would unknowingly have the indexer hit the public default — a correctness bug and a mild privacy leak. The new `extractArkProviderUrl` helper (line 115-121) structurally reads `serverUrl` from the provider, throwing an explicit error if a custom provider lacks a discoverable URL and no `indexerUrl` is set. Correct fix.

**✅ `wallet.ts:543` — cast removal**

`.flatMap((_) => _.vtxos)` without `as ExtendedVirtualCoin[]` — now type-safe because `ContractWithVtxos.vtxos` is `ExtendedContractVtxo[]`, which extends `ExtendedVirtualCoin`.

**✅ Tests**

`test/contracts/watcher.test.ts:254-391` — Two new test cases exercise the `emitVtxoEvent` path through the public subscription API:
- Delegate-contract enrichment: asserts `tapTree` equals `DelegateVtxo.Script.encode()` and differs from `DefaultVtxo.Script.encode()` — catches accidental handler fallback.
- Unregistered handler fallback: asserts raw vtxo emitted (no taproot fields) and `console.warn` includes `txid:vout`.

The `yieldOnce` helper is a clean way to drive a single SSE update. The `waitForReceived` using `vi.waitFor` is appropriate for async event flow.

**✅ README cleanup**

Removes `arkServerUrl: 'https://arkade.computer'` from all examples, consistent with #460 which made mainnet + arkade.computer the default.

### Nits (non-blocking)

1. **`extractArkProviderUrl`** (wallet.ts:118-121): The structural read `(provider as { serverUrl?: unknown }).serverUrl` works for `RestArkProvider` and `ExpoArkProvider`, but if a future provider implementation uses a getter that throws, this would propagate. Low risk — just noting.

2. **`isAnnotated` locality** (delegator.ts:504): This guard is useful enough that it could live in `contracts/types.ts` as a public utility (e.g., `isExtendedContractVtxo`). Currently it's a private function scoped to the delegator module. Not blocking — can be promoted later if needed.

### Not protocol-critical

No changes to VTXO signing, forfeit transaction construction, round lifecycle, connector trees, or exit paths. The delegator change is a type-narrowing improvement on the *filter* that selects VTXOs for delegation — it doesn't change the actual delegation/forfeit logic. The `contractWatcher` change is event-emission plumbing, not transaction construction.

LGTM. Ship it.

---
🤖 Reviewed by Arkana

ghost (approved):
## 🤖 Arkana Code Review — APPROVE

Reviewed all 8 changed files, cross-checked consumers in `banco`, `demos`, and `asset-demos` repos, and read git history on the touched files.

### Summary

Clean type-safety follow-up to #462. Three logical changes, all correct:

1. **`ExtendedContractVtxo` type** — `ExtendedVirtualCoin & { contractScript }` is the right intersection. It's a proper subtype of `ContractVtxo` (narrows the `Partial<TapLeaves & EncodedVtxoScript>` to required). Every code path that returns `ExtendedContractVtxo[]` (`getVtxosForContracts`, `fetchContractVtxosBulk`, `syncAndSaveVtxos`) flows through either `annotateVtxos()` or `WalletRepository` reads, both of which guarantee taproot fields are present. The casts were already safe at runtime; now they're safe at compile time.

2. **`isAnnotated()` type guard in `delegator.ts:504-509`** — Checks all three required fields (`tapTree`, `forfeitTapLeafScript`, `intentTapLeafScript`). Replaces the unsafe `as ExtendedVirtualCoin` cast. Behaviorally, this is slightly stricter: a vtxo that somehow had `tapTree` but was missing leaf scripts would now be excluded from delegation instead of crashing downstream in `makeDelegateForfeitTx`. Correct and safer.

3. **Indexer URL derivation in `wallet.ts:257-280`** — Prevents silently pairing a custom `arkProvider` with the public default indexer. The `extractArkProviderUrl` structural read is the right pattern for an interface that doesn't declare a URL accessor. The error message when `serverUrl` is undiscoverable is clear and actionable.

### Verified

- `ContractWithVtxos.vtxos` narrowing from `ContractVtxo[]` → `ExtendedContractVtxo[]` is backward-compatible for consumers (covariant read position). Cross-repo grep confirms no external code constructs `ContractWithVtxos` objects — `banco`, `demos`, and `asset-demos` only consume the `.vtxos` property via destructuring.
- `ContractEvent.vtxos` correctly stays `ContractVtxo[]` because the catch-fallback in `emitVtxoEvent` (`contractWatcher.ts:785-788`) pushes un-extended vtxos.
- The improved warn log (`contractWatcher.ts:786`) now includes `txid:vout` and the caught error — matches the test at `watcher.test.ts:375-380`.
- Both new tests exercise the subscription path end-to-end (not just unit internals): delegate handler enrichment and no-handler fallback. Good coverage.
- README changes are consistent with #460 defaults.

### No protocol-critical concerns

This PR touches type annotations, logging, and indexer URL resolution. No changes to VTXO signing, forfeit transaction construction, round lifecycle, connector trees, or exit paths. The delegation filter change (`isAnnotated`) is strictly more conservative than the old cast — it can only *exclude* vtxos, never include ones that lack required fields.

LGTM. Ship it.

PR #465, opened 2026-04-29
update regtest image

bordalix:


Kukks (changes requested):
Maybe we can remove this file entirely now and just update the submodule to master commit. Ive already bumped nigiri master to 0.9.4

bordalix:
@Kukks tests here are using blocks, while default for arkade-regtest is seconds.

pietro909 (approved):

PR #467, opened 2026-04-29
chore: fix tsconfig errors

gringokiwi:
- adds `esnext` as lib (needed for `disponse` + `asyncDispose`)
- adds `ignoreDeprecations: 6.0` (necessary to keep moduleResolution: node)
- adds `rootDir: ./` (needed for `declarationDir` and `outDir` to work outside of `./src`)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,9 +1,10 @@
 {
     "compilerOptions": {
         "target": "es2020",
-        "lib": ["es2022", "dom", "dom.iterable"],
+        "lib": ["esnext","es2022","dom","dom.iterable"],
         "module": "esnext",
         "moduleResolution": "node",
+        "ignoreDeprecations": "6.0",
         "strict": true,
         "esModuleInterop": true,
         "skipLibCheck": true,
@@ -11,6 +12,7 @@
         "resolveJsonModule": true,
         "isolatedModules": true,
         "declaration": true,
+        "rootDir": "./",
         "declarationDir": "./dist/types",
         "outDir": "./dist/esm"
     },


ghost (changes requested):
## Review: `chore: fix tsconfig errors`

### 🚨 BLOCKING — `rootDir: "./"` breaks the build (CI confirms)

The `rootDir` change from implicit `./src` to `./` **breaks the entire package output structure**. CI build is failing.

**Why it breaks:**

With `include: ["src"]` and no explicit `rootDir`, TypeScript infers `rootDir` as `./src`. Output goes to:
```
dist/esm/index.js
dist/types/index.d.ts
dist/cjs/index.js
```

With `rootDir: "./"`, TypeScript preserves the `src/` prefix relative to the new root. Output becomes:
```
dist/esm/src/index.js
dist/types/src/index.d.ts
dist/cjs/src/index.js
```

This breaks **every export** in `package.json`:
- `"main": "./dist/cjs/index.js"` → file is now at `dist/cjs/src/index.js`
- `"module": "./dist/esm/index.js"` → file is now at `dist/esm/src/index.js`
- `"types": "./dist/types/index.d.ts"` → file is now at `dist/types/src/index.d.ts`
- All `"exports"` subpath entries have the same problem

Additionally, `tsconfig.esm.json` and `tsconfig.cjs.json` both `extend` the base config and don't override `rootDir`, so they inherit this breakage. Only `tsconfig.expo.json` is safe (it explicitly sets `rootDir: "./src"`).

**Fix:** Remove `rootDir: "./"`. If you need it for a specific reason, set `rootDir: "./src"` (matching implicit behavior), or update all three child tsconfigs to override `rootDir: "./src"`.

---

### ✅ `"lib": ["esnext", ...]` — Correct

The codebase uses `Symbol.asyncDispose` and `Symbol.dispose` in ~18 files (wallet, repositories, contractManager). Adding `esnext` provides the type definitions. Note: `esnext` is a superset of `es2022`, so listing both is redundant — you could drop `es2022` — but it's harmless.

### ⚠️ `"ignoreDeprecations": "6.0"` — Acceptable short-term, tech debt

This silences the TS 5.x deprecation of `moduleResolution: "node"`. The proper fix is migrating to `moduleResolution: "bundler"` or `"node16"`, but that's a bigger change and out of scope here. Fine for now, but consider tracking it.

---

### Summary

| Change | Verdict |
|--------|---------|
| `esnext` lib | ✅ Ship it |
| `ignoreDeprecations` | ⚠️ Acceptable |
| `rootDir: "./"` | 🚨 Breaks build — remove or set to `"./src"` |

Please fix the `rootDir` issue and re-push. CI build failure confirms the breakage.

PR #469, opened 2026-04-29
fix: represent asset amounts as `bigint`

### LINKED ISSUE #384: Change `number` to `bigint` when representing asset amounts
Confirmed with @louisinger: arkade asset amounts have no upper limit.

However, as the typescript SDK uses `number` instead of `bigint`, it has issues dealing with amounts exceeding `2^53 - 1` (i.e. `Number.MAX_SAFE_INTEGER = 9007199254740991`).

That means the max expressable amount is ~90m for an asset with 8 decimals, ~9k for an asset with 12 decimals, and ~9 for an asset with 15 decimals.

Since many stablecoins use 6+ decimals (Paxos tend to use 18) this could easily cause issues.

---

### Public types using `number` for amounts (`src/wallet/index.ts`)

These interfaces are the public API surface and currently use `number`:

| Interface | Field | Line |
|---|---|---|
| `Asset` | `amount: number` | ~L161 |
| `Recipient` | `amount?: number` | ~L166 |
| `AssetDetails` | `supply: number` | ~L187 |
| `IssuanceParams` | `amount: number` | ~L199 |
| `ReissuanceParams` | `amount: number` | ~L204 |
| `BurnParams` | `amount: number` | ~L207 |
| `ArkTransaction` | `amount: number` | ~L263 |
| `SendBitcoinParams` | `amount: number` | ~L153 |
| `Balance` | `available`, `recoverable`, `total: number` | ~L145-148 |

### Internal code already uses `bigint`

The low-level asset extension layer already handles `bigint` correctly:

- **`src/extension/asset/assetInput.ts`** — `AssetInput.amount` is typed as `bigint`; `create()` accepts `bigint | number` and converts via `BigInt()`
- **`src/extension/asset/assetOutput.ts`** — same pattern, internally `bigint`
- **`src/wallet/asset.ts`** — `createAssetPacket()` already converts with `BigInt(asset.amount)` at the boundary

The problem is at the **public API boundary**: amounts enter as `number`, silently lose precision for large values, then get converted to `bigint` internally. By that point the damage is done.

### Other files with `amount` usage
- `src/arkfee/types.ts` — fee-related amounts
- `src/arkfee/estimator.ts` — fee estimation
- `src/wallet/asset-manager.ts` — asset management operations
- `src/wallet/ramps.ts` — on/off ramp amounts
- `src/wallet/onchain.ts` — on-chain transaction amounts
- `src/utils/arkTransaction.ts` — transaction history amounts
- `src/utils/bip21.ts` — BIP21 URI parsing (amounts)
- `src/providers/ark.ts` — provider layer amounts
- `src/providers/expoIndexer.ts` — Expo indexer amounts
- `src/repositories/indexedDB/schema.ts` — persisted schema (migration needed)

## Scope of Change

This is a **breaking change** to the public API. Every consumer passing `number` amounts for assets will need to switch to `bigint`.

### Migration considerations
1. **JSON serialization** — `bigint` is not natively JSON-serializable. Need a serialization strategy (e.g., string representation in JSON, custom `toJSON`/`fromJSON`).
2. **IndexedDB schema** — stored amounts in `src/repositories/indexedDB/schema.ts` may need a migration if currently stored as `number`.
3. **Downstream consumers** — the `arkade-os/wallet` app and any other SDK consumers will need to update all amount handling.
4. **BTC amounts vs asset amounts** — BTC amounts (in sats) fit comfortably in `number` (max ~21M BTC = 2.1e15 sats < 2^53). Consider whether to change only asset amounts or all amounts for consistency.

### Phase 1: Asset amounts only (minimum viable fix)
- Change `Asset.amount` from `number` to `bigint`
- Update `AssetDetails.supply`, `IssuanceParams.amount`, `ReissuanceParams.amount`, `BurnParams.amount` to `bigint`
- Remove the `BigInt()` conversion in `createAssetPacket()` (no longer needed)
- Update indexer response parsing to convert amount strings/numbers to `bigint` at the deserialization boundary

### Phase 2: BTC amounts (optional, for consistency)
- Evaluate whether `SendBitcoinParams.amount`, `Balance` fields, `ArkTransaction.amount` etc. should also move to `bigint`
- BTC sats fit in `number`, so this is a consistency choice rather than a correctness fix

### Phase 3: Downstream updates
- Update `arkade-os/wallet` to handle `bigint` amounts throughout the UI layer
- Update any formatting/display utilities that use `Number()` or arithmetic on amounts

## Related

- Internal `AssetInput`/`AssetOutput` already use `bigint` — this change aligns the public API with the internal representation
- No existing issues or PRs for this change

gringokiwi:
Closes https://github.com/arkade-os/ts-sdk/issues/384

- creates `toSafeNumber` util for converting strings/bigints to in-range `Number`
- add `assetAmount` to `Asset` type as `bigint`, marking `amount` as deprecated
- add `assetSupply` to `AssetDetails` type as `bigint`, marking `supply` as deprecated
- does _**not**_ change `VtxoAsset.amount` as it is already `string`
- ensure functions that relied on these use `assetAmount` or `assetSupply` as source of truth, while returning the base `amount`/`supply` for backwards compatibility
- updates relevant `transactionHistory` tests

## Summary by CodeRabbit

* **New Features**
  * Added `assetAmount` (bigint) and `assetSupply` (bigint) fields for more precise asset quantity handling with support for large values.

* **Deprecations**
  * Marked `amount` and `supply` number fields as deprecated; use `assetAmount` and `assetSupply` respectively for accurate large-value representation.

* **Bug Fixes**
  * Improved numeric handling to prevent loss of precision when dealing with large asset quantities.

### DIFF
--- a/src/utils/safeNumber.ts
+++ b/src/utils/safeNumber.ts
@@ -1,0 +1,11 @@
+/** Used for parsing asset amounts */
+export const toSafeNumber = (value: string | bigint | number): number => {
+    const num = Number(value);
+
+    if (!Number.isFinite(num)) return Number.MAX_SAFE_INTEGER;
+
+    if (num > Number.MAX_SAFE_INTEGER) return Number.MAX_SAFE_INTEGER;
+    if (num < Number.MIN_SAFE_INTEGER) return Number.MIN_SAFE_INTEGER;
+
+    return num;
+};
--- a/src/wallet/index.ts
+++ b/src/wallet/index.ts
@@ -272,8 +272,15 @@ export interface Asset {
     /** Asset identifier. */
     assetId: string;
 
-    /** Asset amount in base units. */
+    /**
+     * Asset amount in base units.
+     * @deprecated (unsuitable for large values)
+     * @see Asset.assetAmount
+     */
     amount: number;
+
+    /** Asset amount in base units. */
+    assetAmount: bigint;
 }
 
 /**
@@ -334,8 +341,15 @@ export type AssetDetails = {
     /** Asset identifier. */
     assetId: string;
 
-    /** Total issued supply in base units. */
+    /**
+     * Total issued supply in base units.
+     * @deprecated (unsuitable for large values)
+     * @see AssetDetails.assetSupply
+     */
     supply: number;
+
+    /** Total issued supply in base units. */
+    assetSupply: bigint;
 
     /** Optional immutable metadata associated with the asset. */
     metadata?: AssetMetadata;
--- a/src/providers/expoIndexer.ts
+++ b/src/providers/expoIndexer.ts
@@ -2,6 +2,7 @@ import { RestIndexerProvider, SubscriptionResponse, Vtxo } from "./indexer";
 import { isFetchTimeoutError } from "./ark";
 import { VirtualCoin } from "../wallet";
 import { getExpoFetch, sseStreamIterator } from "./expoUtils";
+import { toSafeNumber } from "../utils/safeNumber";
 
 // Helper function to convert Vtxo to VirtualCoin (same as in indexer.ts)
 function convertVtxo(vtxo: Vtxo): VirtualCoin {
@@ -33,7 +34,8 @@ function convertVtxo(vtxo: Vtxo): VirtualCoin {
         script: vtxo.script,
         assets: vtxo.assets?.map((a) => ({
             assetId: a.assetId,
-            amount: Number(a.amount),
+            amount: toSafeNumber(a.amount),
+            assetAmount: BigInt(a.amount),
         })),
     };
 }
--- a/src/providers/indexer.ts
+++ b/src/providers/indexer.ts
@@ -3,6 +3,7 @@ import { AssetDetails, AssetMetadata, Outpoint, VirtualCoin } from "../wallet";
 import { isFetchTimeoutError } from "./ark";
 import { eventSourceIterator, isEventSourceError } from "./utils";
 import { MetadataList } from "../extension/asset";
+import { toSafeNumber } from "../utils/safeNumber";
 
 export type PaginationOptions = {
     pageIndex?: number;
@@ -721,7 +722,8 @@ export class RestIndexerProvider implements IndexerProvider {
             : undefined;
         return {
             assetId: data.assetId ?? assetId,
-            supply: Number(data.supply ?? 0),
+            supply: toSafeNumber(data.supply ?? 0),
+            assetSupply: BigInt(data.supply ?? 0),
             metadata,
             controlAssetId: data.controlAsset || undefined,
         };
@@ -827,7 +829,8 @@ function convertVtxo(vtxo: Vtxo): VirtualCoin {
         script: vtxo.script,
         assets: vtxo.assets?.map((a) => ({
             assetId: a.assetId,
-            amount: Number(a.amount),
+            amount: toSafeNumber(a.amount),
+            assetAmount: BigInt(a.amount),
         })),
     };
 }
--- a/src/utils/transactionHistory.ts
+++ b/src/utils/transactionHistory.ts
@@ -1,4 +1,5 @@
 import { ArkTransaction, Asset, TxKey, TxType, VirtualCoin } from "../wallet";
+import { toSafeNumber } from "./safeNumber";
 
 type ExtendedArkTransaction = ArkTransaction & {
     tag: "offchain" | "boarding" | "exit" | "batch";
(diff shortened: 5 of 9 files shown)


ghost (changes requested):
## Code Review — `fix: represent asset amounts as bigint`

Reviewed every changed file plus full context of `asset.ts`, `index.ts` types, `wallet.ts`, `delegator.ts`, service worker, indexers. Also checked cross-repo consumers (`wallet`, `banco`, `arkade-explorer`, `skill`).

The intent is correct and important — `Number` truncates amounts above 2^53. But the implementation has critical gaps.

---

### 🔴 CRITICAL — Protocol / Money at Risk

**1. `src/wallet/asset.ts` is NOT updated — transaction-building still reads truncated `amount`**

`createAssetPacket()` constructs the on-wire asset data for every asset transaction. It still reads `asset.amount` (the deprecated, clamped `number` field):

- `asset.ts:32` — `AssetInput.create(inputIndex, BigInt(asset.amount))`
- `asset.ts:49` — `AssetOutput.create(outputIndex, BigInt(asset.amount))`
- `asset.ts:62` — `AssetOutput.create(outputIndex, BigInt(asset.amount))`

If an asset has amount `10_000_000_000_000_000` (10^16), `toSafeNumber` clamps it to `9007199254740991` (MAX_SAFE_INTEGER), then `createAssetPacket` does `BigInt(9007199254740991)` instead of `BigInt("10000000000000000")`. **The transaction packet would contain the wrong amount.** This silently loses asset value on-chain.

These lines MUST be updated to use `asset.assetAmount` (or better, just pass the bigint directly).

**2. `selectCoinsWithAsset()` also not updated — coin selection uses truncated amounts**

- `asset.ts:112-114` — sort comparator uses `asset.amount` (number). For large values, all amounts clamp to `MAX_SAFE_INTEGER`, making the sort non-deterministic.
- `asset.ts:126-127` — `totalAssetAmount += BigInt(assetAmount)` where `assetAmount` is the clamped number. Coin selection would think it has a different total than reality, potentially selecting too few or too many coins.

---

### 🟠 HIGH — Breaking Changes

**3. `Asset.assetAmount: bigint` is a required field — breaks all downstream consumers**

The `Asset` interface is used in:
- `Recipient.assets` (user-facing `send()` API — `wallet/index.ts:297`)
- `VirtualCoin.assets` (`wallet/index.ts:522`)
- `ArkTransaction.assets` (`wallet/index.ts:572`)
- `WalletBalance.assets` (`wallet/index.ts:236`)

Every downstream consumer constructing an `Asset` object will fail TypeScript compilation:
- `wallet` app: asset balance display, send flows
- `banco`: taker.ts lines 283, 297, 320, 400, 449-450, 467, 620
- `arkade-explorer`: vtxo-list.tsx, address-stats.tsx

**Fix:** Make `assetAmount` optional (`assetAmount?: bigint`) or provide a factory/helper. The PR description says "backwards compatibility" but a required field on a public interface is the opposite of that.

**4. JSON serialization will throw**

`bigint` is not JSON-serializable. `JSON.stringify({ assetAmount: 100n })` throws `TypeError: Do not know how to serialize a BigInt`. This will crash:
- Any logging that serializes Asset objects
- Any localStorage/IndexedDB storage using JSON
- Any API response serialization
- Debug tooling

The `postMessage` path (service worker) is fine (structured clone supports bigint), but JSON paths are not.

---

### 🟡 MEDIUM

**5. `toSafeNumber` silently clamps — dangerous for financial software**

`src/utils/safeNumber.ts:5-7`: Values above `MAX_SAFE_INTEGER` silently become `MAX_SAFE_INTEGER`. For a financial system, this is a footgun. A consumer reading `asset.amount` gets `9007199254740991` with zero indication the value is wrong.

At minimum, add a `console.warn()`. Better: throw an error, or return `NaN` so it's obviously broken rather than silently incorrect.

**6. Naming: `assetAmount` on `Asset` is redundant**

`Asset.assetAmount` reads as "the asset's asset-amount." Consider just `bigAmount`, `rawAmount`, or (longer term) renaming `amount` to `legacyAmount` and using `amount: bigint`.

---

### ✅ What's Good

- The `transactionHistory.ts` migration to bigint arithmetic is correct and well done
- Test coverage for transaction history is thorough
- Both `convertVtxo` functions (indexer + expo) are updated consistently
- Balance aggregation in `wallet.ts` and `wallet-message-handler.ts` correctly uses bigint accumulation

---

### Summary

The PR correctly identifies a real precision bug and migrates internal arithmetic to bigint. But it **misses the most critical file** (`asset.ts`) where amounts flow into actual transaction packets. It also introduces a required field on a public interface without coordinating downstream consumers.

**Blocking issues before merge:**
1. Update `asset.ts` — `createAssetPacket` and `selectCoinsWithAsset` must use `assetAmount`
2. Make `assetAmount` optional on `Asset` (and `assetSupply` on `AssetDetails`) to avoid breaking downstream
3. Add JSON serialization strategy (custom replacer/reviver or toJSON)

⚠️ This touches asset amounts in transaction construction — **protocol-critical**. Requires human sign-off even after fixes.

louisinger (commented):
2 fields will be worst to handle for the dev. can't we avoid adding the new `assetAmount` / `assetSupply`. instead `amount` type could become `number | bigint`.

src/utils/safeNumber.ts
```diff
@@ -0,0 +1,11 @@
+/** Used for parsing asset amounts */
+export const toSafeNumber = (value: string | bigint | number): number => {
+    const num = Number(value);
+
+    if (!Number.isFinite(num)) return Number.MAX_SAFE_INTEGER;
+
+    if (num > Number.MAX_SAFE_INTEGER) return Number.MAX_SAFE_INTEGER;
+    if (num < Number.MIN_SAFE_INTEGER) return Number.MIN_SAFE_INTEGER;
+
+    return num;
```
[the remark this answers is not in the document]
pietro909: Agree on this @gringokiwi - if we get invalid data and we need it to map it to anything with value (assets, coins, ... ) the best choice is to fail spectacularly. It should end up with a big red error message on the UI.

src/utils/safeNumber.ts
```diff
@@ -0,0 +1,11 @@
+/** Used for parsing asset amounts */
+export const toSafeNumber = (value: string | bigint | number): number => {
```
pietro909, line 2: Why do we need to map `bigint` to `number` in the first place? 

On f the reasons to use [BigInt](https://www.w3schools.com/js/js_bigint.asp) is:
> BigInt allows you to work with integers larger than the limit of Numbers.

Mapping to a subset can result in data loss.

src/wallet/index.ts
```diff
@@ -272,8 +272,15 @@ export interface Asset {
...
-    /** Asset amount in base units. */
+    /**
+     * Asset amount in base units.
+     * @deprecated (unsuitable for large values)
+     * @see Asset.assetAmount
+     */
     amount: number;
+
+    /** Asset amount in base units. */
+    assetAmount: bigint;
```
pietro909, line 283: What if we plan this as a breaking change instead? What's the priority of this fix? 

The naming feels off `Asset.assetAmount` but `amount: number | bigint` also qualifies as a breaking change (larger domain in output -> consumers will break at compile time already).

gringokiwi:
closing, will review/add to https://github.com/arkade-os/ts-sdk/pull/472

PR #441, opened 2026-04-22
feat(wallet): rotate receive address on vtxo_received (Phase C2)

Kukks:
Stacked on #440 (HDDescriptorProvider / Phase C1). Only the last commit (26446508) is this PR's scope — earlier commits show up in the diff until #440 merges.

## Summary
- Install an `HDDescriptorProvider` when identity is a `SeedIdentity` whose descriptor is BIP86-vanilla at index 0, or whose `settings.hd` already exists from a prior run.
- Hook `ContractManager.onContractEvent('vtxo_received')` and rotate the active receive descriptor whenever the current default contract fires.
- Old default contracts remain watched — previously-shared addresses keep crediting the wallet.
- HD is opt-in by identity type: `SingleKey` and user-pinned non-vanilla descriptors stay on the static path.

## What changed
- `HDDescriptorProvider.getCurrentReceivePubkey()` — returns the x-only pubkey for the current receive descriptor without seed access (uses `@0.pubkey` from `expand()`, avoiding `extractPubKey`'s HD-descriptor rejection).
- `Wallet.create` — conditionally installs the HD provider; rebuilds `offchainTapscript` if the persisted receive index has advanced past 0.
- `installHdRotationHandler` — registers a filtered `vtxo_received` callback that only triggers on the current contract script.
- `rotateAndRegister` — serialised via `_hdRotationChain` mutex. Rotates the provider, rebuilds `offchainTapscript`, registers the next-index default contract.
- `Wallet.dispose()` — tears down the rotation subscription and drains any in-flight rotation before disposing the contract manager.
- `ReadonlyWallet.offchainTapscript`: `readonly` → `public` (documented: external consumers should still treat it as read-only).

## Blast radius
- Zero behaviour change for `SingleKey` wallets.
- `MnemonicIdentity`/`SeedIdentity` wallets now rotate after each `vtxo_received` on the current address. Per plan, old addresses stay active.
- No storage migration; additive `settings.hd` key.
- No public API rename. `offchainTapscript` mutability widened only in base class.

## Test plan
- [x] `test/walletHdRotation.test.ts` — 7 new tests: HD install conditioning, rotation on matching event, non-rotation on unrelated contract, non-rotation on non-vtxo_received types, restart persistence, dispose teardown.
- [x] `test/hdDescriptorProvider.test.ts` — 2 new tests for `getCurrentReceivePubkey`.
- [x] Full non-e2e suite passes (930 tests).
- [x] Prettier clean.
[...]

PR #473, opened 2026-05-01
feat(wallet): HD receive rotation via contract repository (alt to #441)

Kukks:
Stacked on #440 (HDDescriptorProvider as a pure rotating allocator). Alternative to #441.

## Summary

Wires HD receive rotation into `Wallet` by following the dotnet-sdk's split of responsibilities:
the `HDDescriptorProvider` is a pure rotating allocator, and the **contract repository** is the
source of truth for "what addresses am I currently bound to". The provider has no notion of
"current" — the boot path looks up active default contracts in the repo to recover the wallet's
state, and rotation creates a new default contract while leaving the old one active.

### `src/wallet/wallet.ts`

**Boot path (`Wallet.create`)**: For `SeedIdentity` wallets with a vanilla BIP-86 descriptor:
- Look up active `default` contracts whose `serverPubKey` matches the current server.
- If any exist → use the most recent contract's `pubKey` for the offchain tapscript. No
  provider call. The wallet resumes from exactly where the previous session left off.
- If none exist → call `provider.getNextSigningDescriptor()` once, allocating index 0. The
  contract registered moments later by `initializeContractManager` records this binding,
  so storage and the contract repo stay in lockstep.
- `SingleKey` wallets, and `SeedIdentity` wallets whose descriptor isn't a vanilla
  `tr(...path/0/*)` template, stay on the static path unchanged.

**Rotation (`installHdRotationHandler` + `rotateAndRegister`)**:
- Subscribe to `vtxo_received` on the contract manager. Filter for events on the currently-active
  default contract.
- On match: allocate next descriptor via the provider, rebuild `offchainTapscript`, call
  `manager.createContract({ type: 'default', ... })`. The previous default contract entry stays
  `active` so earlier addresses keep crediting the wallet.
- Rotations are serialised by `_hdRotationChain` so rapid-fire events can't interleave the
  allocate → rebuild → create sequence.

**Dispose**:
- Unsubscribe the rotation handler, drain any in-flight rotation, then dispose the contract
  manager. A late `vtxo_received` event cannot queue work on a disposing wallet.

**`offchainTapscript` mutability**: `readonly` → `public`, same as #441 needed. External
consumers should still treat it as read-only (writes happen only in `rebuildOffchainTapscript`
and the boot path).

### `test/walletHdRotation.test.ts` — 9 tests

- **Installation**:
  - Fresh HD wallet allocates index 0 at boot (`lastIndexUsed = 0`).
  - Boot registers an active default contract whose script matches the wallet's
    `defaultContractScript`.
  - SingleKey identities skip the HD path.
- **Rotation**:
  - Receiving on the current default → bumps `lastIndexUsed` and the active script changes.
  - The new default contract is registered while the old entry stays `active`.
  - Events for unrelated scripts and non-`vtxo_received` events do nothing.
- **Persistence**:
  - Second wallet on the same `walletRepository`+`contractRepository` reads the rotated
    address by querying the contract repo for active default contracts (no event replay).
- **Dispose**:
  - After `dispose()`, late events do not call into the provider.

## Compared to #441

- **No `getCurrentReceivePubkey()` accessor on the provider** — the provider has no notion of
  "current". The wallet boot queries the contract repository instead.
- **No `rotateReceive` / `getCurrentReceiveIndex`** — the provider only exposes
  `getNextSigningDescriptor`.
- **First rotation produces index 0** (not 1). With no implicit "current" surfaced by a read
  accessor, allocation starts at the natural first index.

## Blast radius

- Zero behaviour change for `SingleKey` wallets.
- For `MnemonicIdentity` / `SeedIdentity` HD wallets, rotation now happens after each
  `vtxo_received` on the active receive script. Old default contracts stay `active` (no funds
  loss).
- No storage migration. Additive `settings.hd` key, written only on first allocation through
  the provider.
- No public API rename. `offchainTapscript` mutability widened only in the base class.

## Stacked PR order

1. #440 — Phase C1 HDDescriptorProvider (pure rotating allocator).
2. **This PR** — Phase C2 wallet wiring via the contract repository.

DRAFT until #440 lands. Rebase target moves to `master` after #440 merges.

## Summary by CodeRabbit

* **New Features**
  * WalletMode config: auto/static/hd/explicit-provider for receive rotation
  * HD receive-address rotation with safe boot, on-chain-triggered rotation, controlled contract lifecycle, and provider hooks
  * Refresh can include inactive contracts in repository-wide queries while preserving cursor safety

* **Documentation**
  * Clarified descriptor-provider docs describing optional receive-rotation integration

* **Tests**
  * New unit and end-to-end tests covering HD rotation, persistence, disposal, and refresh behavior

* **Bug Fix**
  * Fee computation now rounds up fractional fee rates to avoid runtime errors

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/ts-sdk/pull/473)


ghost (changes requested):
## Arkana Code Review — PR #473

**Verdict: Request Changes** — One bug that can cause state desync on failed contract registration, plus a multi-timelock gap on rotation. Architecture is solid, concurrency model is well-reasoned, test coverage is good. Fixing these two issues would make this mergeable.

---

### 🔴 BUG: `rotateAndRegister` updates tapscript before confirming contract registration

**`src/wallet/wallet.ts:1487-1509`**

`rotateAndRegister` mutates `this.offchainTapscript` (line 1487-1490) **before** calling `manager.createContract()` (line 1496). If `createContract` throws (network error, repo corruption, disk full), the wallet is left in an inconsistent state:

- `offchainTapscript` / `defaultContractScript` / `getAddress()` all reflect the NEW key
- The contract manager has no contract watching that new script
- VTXOs sent to the displayed address are invisible to the wallet until next boot

The HD index has also already been consumed via `getNextSigningDescriptor()`, so the damage is: one skipped index + a wallet showing an unwatched address.

**Fix:** Move the tapscript update AFTER `createContract` succeeds:

```typescript
private async rotateAndRegister(): Promise<void> {
    if (!this._hdProvider) return;
    const descriptor = await this._hdProvider.getNextSigningDescriptor();
    const pubKey = deriveLeafPubkey(descriptor);
    const newTapscript = rebuildTapscript(this.offchainTapscript, pubKey);
    const csvTimelock = newTapscript.options.csvTimelock
        ?? DefaultVtxo.Script.DEFAULT_TIMELOCK;

    const manager = await this.getContractManager();
    await manager.createContract({
        type: "default",
        params: {
            pubKey: hex.encode(pubKey),
            serverPubKey: hex.encode(newTapscript.options.serverPubKey),
            csvTimelock: timelockToSequence(csvTimelock).toString(),
        },
        script: hex.encode(newTapscript.pkScript),
        address: newTapscript.address(this.network.hrp, this.arkServerPublicKey).encode(),
        state: "active",
    });

    // Only commit the new tapscript after the contract is persisted.
    this.offchainTapscript = newTapscript;
}
```

Note: this also fixes a subtle dependency — the current code uses `this.defaultContractScript` and `this.getAddress()` inside `createContract`, which read from the already-mutated `offchainTapscript`. The fix above computes `script` and `address` from the new tapscript directly, making the data flow explicit.

---

### 🟡 Multi-timelock gap on rotation

**`src/wallet/wallet.ts:1483-1509` vs `997-1057`**

`initializeContractManager` (line 1005) registers a default contract for **every** `walletContractTimelocks` entry at the current pubkey. But `rotateAndRegister` only registers **one** contract — at the primary `csvTimelock`.

If the wallet carries historical timelocks (from server unilateral-exit-delay changes), the rotated address is only watched at the current timelock. On next boot, `initializeContractManager` self-heals by registering all timelocks at the latest key. But between rotation and reboot, VTXOs created under a historical timelock at the new address would be invisible.

In practice this is low-risk (the server uses the current timelock for new rounds), but it's an inconsistency. Consider either:
1. Iterating `walletContractTimelocks` in `rotateAndRegister` (matches boot behavior), or
2. Documenting explicitly that rotation only watches the primary timelock and relies on boot to fill the gap.

---

### 🟡 Silent error swallowing in boot path

**`src/wallet/wallet.ts:1411-1416`**

```typescript
} catch {
    // Template not derivable, contract repo unavailable, or
    // descriptor mismatch — fall back to the static path
    // rather than fail wallet construction.
    hdProvider = undefined;
}
```

This catches **all** errors — including descriptor-mismatch (`"Refusing to reuse HD state from a different identity"`), which is a critical corruption signal. The wallet silently downgrades to a non-HD single-address mode with no logging, no event, no way for the application to detect the degradation.

Suggestion: at minimum, `console.warn` the error. Better: emit a wallet event or re-throw on descriptor-mismatch (which indicates seed/repo mismatch — a serious problem that shouldn't be papered over).

---

### 🟢 Observations (non-blocking)

**Double rotation on rapid events** (`src/wallet/wallet.ts:1465-1474`): The filter `event.contractScript !== this.defaultContractScript` runs synchronously when the event fires, before the chained rotation completes. If two `vtxo_received` events fire in the same tick for the same script, both pass the filter and queue two rotations. The second rotation advances the index an extra time. Not a funds-loss issue (old contracts stay active), but it wastes an HD index. A guard in `rotateAndRegister` that re-checks `this.defaultContractScript` against the event's script would prevent this.

**`looksLikeVanillaHDDescriptor`** (`src/wallet/wallet.ts:175`): `endsWith("/0/*)")` is loose — it matches non-BIP86 paths like `tr([fp/48'/0'/0'/2']/0/*)`. Since `SeedIdentity` constructs BIP-86 descriptors by default, this is fine in practice, but the docstring overpromises.

**`offchainTapscript` mutability on `ReadonlyWallet`** (`src/wallet/wallet.ts:281`): The `readonly → public` change is on the base class whose name implies immutability. The field is only mutated from `Wallet` (subclass), so consider a `protected` modifier + a setter on `Wallet` to maintain the "readonly" contract for `ReadonlyWallet` consumers.

**Test quality**: The 9 tests are well-structured and cover the critical paths (boot, rotation, persistence, dispose, negative cases). Consider adding a test for the `createContract` failure scenario once the fix above lands.

---

### ⚠️ Protocol-Critical Flag

This PR touches **offchain receive address derivation** — the path that determines where VTXOs land. While the design is sound (old contracts stay active, no address reuse), the `rotateAndRegister` ordering bug above means a failed rotation could leave a wallet displaying an unwatched address. **Requires human sign-off** before merge.

ghost (commented):
## Arkana Code Review — PR #473 (Update: May 8 commits)

Previous review on May 1 requested changes. This review covers the 4 new commits (May 7–8) that refactored into `WalletReceiveRotator`, added `walletMode`, `metadata.source` tagging, and `includeInactive` refresh.

The refactoring is a significant quality improvement — clean separation of concerns, proper `walletMode` polymorphism, and the `metadata.source` tagging fixes the ambient-contract-confusion bug from the original approach. Test coverage is solid (664 lines for rotation, 196 for `includeInactive`).

---

### 🔴 STILL OPEN: `rotate()` mutates tapscript before confirming contract registration

**`src/wallet/walletReceiveRotator.ts:282-842` (new location, same bug)**

The original finding from my May 1 review was NOT addressed. `rotate()` still calls:
```typescript
wallet.offchainTapscript = rebuildTapscript(wallet.offchainTapscript, pubKey);  // line ~785
// ... then later ...
await manager.createContract({...});  // line ~806
```

If `createContract` throws (network error, repo corruption, disk full):
- `offchainTapscript` / `defaultContractScript` / `getAddress()` all reflect the NEW key
- No contract is watching the new script
- The HD index is consumed via `getNextSigningDescriptor()` but the binding was never persisted to the contract repo
- Funds sent to the displayed address are invisible until the next boot (which would re-register via `initializeContractManager`)

Now it's slightly worse: `currentTaggedScript` has NOT been updated (it's set after `createContract` on line 835), so a subsequent successful rotation would try to deactivate the CORRECT old contract — but the wallet is displaying an unwatched address in the meantime.

**Fix remains the same:** compute `script`/`address` from the new tapscript locally, call `createContract`, THEN commit `wallet.offchainTapscript = newTapscript`:

```typescript
private async rotate(wallet: RotatableWallet): Promise<void> {
    const descriptor = await this.provider.getNextSigningDescriptor();
    const pubKey = deriveLeafPubkey(descriptor);
    const newTapscript = rebuildTapscript(wallet.offchainTapscript, pubKey);

    const manager = await wallet.getContractManager();
    const csvTimelock = newTapscript.options.csvTimelock
        ?? DefaultVtxo.Script.DEFAULT_TIMELOCK;
    const csvTimelockStr = timelockToSequence(csvTimelock).toString();
    const serverPubKeyHex = hex.encode(newTapscript.options.serverPubKey);
    const script = hex.encode(newTapscript.pkScript);
    const address = newTapscript.address(/* ... */);

    await manager.createContract({ script, address, ... });

    // Only commit after persistence succeeds.
    wallet.offchainTapscript = newTapscript;

    const previousTagged = this.currentTaggedScript;
    this.currentTaggedScript = script;
    if (previousTagged && previousTagged !== script) {
        await manager.setContractState(previousTagged, "inactive");
    }
}
```

This requires `RotatableWallet` to expose the network HRP + `arkServerPublicKey` (or a method to derive the address from a tapscript), but that's a small interface expansion for correctness.

---

### 🟡 `setContractState` failure leaves `currentTaggedScript` inconsistent

**`src/wallet/walletReceiveRotator.ts:834-841`**

`currentTaggedScript` is updated on line 835 BEFORE `setContractState` is called on line 840. If `setContractState` throws, the old tagged contract stays `active` in the repo. On the next rotation, the rotator tries to deactivate the NEW contract (which is the CURRENT one), not the stale one. The stale contract is orphaned as permanently-active.

Not a funds-loss issue (the stale contract just stays in the watched set forever), but it leaks watched scripts over time, degrading privacy and performance.

**Fix:** Swap the order — deactivate first, then update `currentTaggedScript`. Or wrap the deactivation in a try/catch that retains the old value on failure.

---

### 🟡 Multi-timelock gap on rotation (from previous review, still applicable)

`initializeContractManager` registers one contract per `walletContractTimelocks` entry. `rotate()` registers only ONE contract at the primary `csvTimelock`. Between rotation and reboot, a VTXO at a historical timelock on the new address is invisible. Low-risk (server uses current timelock for new rounds), but the inconsistency with the boot path remains.

---

### 🟢 New observations (non-blocking)

**`resolveDescriptorProvider` under `'auto'` mode** (`walletReceiveRotator.ts:450-460`): The `catch {}` on `HDDescriptorProvider.create` still swallows descriptor-mismatch errors silently. This is improved by the existence of `walletMode: 'hd'` as an explicit strict alternative, which was my suggestion. Acceptable tradeoff for backwards compatibility. Would still benefit from a `console.warn` for debuggability.

**`includeInactive` implementation is clean.** The cursor-advancement logic correctly identifies `includeInactive` as a superset (cursor-safe). The `contracts ? false : opts?.includeInactive` guard in `refreshVtxos` properly prevents interaction with explicit `scripts`. The 5 new manager tests cover the key invariants. No issues found.

**`pickActiveReceive` sorts by `createdAt` descending** (`walletReceiveRotator.ts:386`): Correct, but relies on `createdAt` being monotonically increasing. Two rapid rotations in the same millisecond (possible with `Date.now()`) could return the wrong "newest." The serialization chain makes this unlikely in practice.

**`isHDCapableIdentity` structural guard** (`hdCapableIdentity.ts:63-75`): Clean type guard. Correctly checks all four required members.

**Test quality**: Expanded from 9 to ~20 tests. Good coverage of `walletMode` matrix, `metadata.source` tagging, baseline non-deactivation, and intruder-contract rejection. Still no test for `createContract` failure during rotation — please add one when the fix lands.

**Cross-repo impact**: Zero. No downstream consumers reference any of the new APIs (`WalletMode`, `WalletReceiveRotator`, `includeInactive`, `createReceiveRotator`). The public API additions are additive.

---

**Bottom line:** The refactoring is well-done — `WalletReceiveRotator` as a standalone class is cleaner than the inline approach, `walletMode` gives callers proper control, and `metadata.source` tagging eliminates the ambient-contract-confusion problem. But the original state-desync bug in `rotate()` persists and needs to be fixed before merge. The fix is straightforward (compute contract params from local `newTapscript`, persist, THEN commit to wallet state).

ghost (changes requested):
## Arkana Code Review — `feat(wallet): HD receive rotation via contract repository`

**Reviewer**: Arkana (aggressive protocol review)
**Verdict**: Request changes — one high-severity boot-path bug, one medium concern on error handling in the rotation chain.

Overall the architecture is clean: the contract-repo-as-source-of-truth design, the `WalletReceiveRotator` lifecycle, and the `ReceiveRotatorFactory` extension point are well thought through. The test suite covers the happy path thoroughly. But I found a gap that would cause address drift on every wallet restart.

---

### 🔴 HIGH — Address drift on restart without rotation

**`src/wallet/walletReceiveRotator.ts:125-131` (`resolveBoot`) + `pickActiveReceive` (line 330)**

`pickActiveReceive` filters by `metadata.source === WALLET_RECEIVE_SOURCE` (line 347). But the boot path never tags the initial allocation's contract with this source — `initializeContractManager` registers baseline contracts without `metadata.source` (confirmed by the test at `test/walletHdRotation.test.ts` line 1387: `"does not tag boot baseline contracts as wallet-receive"`).

This means on **every restart without rotation**, `pickActiveReceive` returns `undefined`, and `resolveBoot` falls through to `provider.getNextSigningDescriptor()` (line 131), burning a new index:

```
Boot 1: pickActiveReceive → undefined → getNextSigningDescriptor → index 0 (lastIndexUsed=0)
Boot 2: pickActiveReceive → undefined → getNextSigningDescriptor → index 1 (lastIndexUsed=1)
Boot 3: pickActiveReceive → undefined → getNextSigningDescriptor → index 2 (lastIndexUsed=2)
```

Consequences:
- **Display address changes on every restart** without user action — breaks "share this address" UX.
- **Index space burned** on every restart; `lastIndexUsed` ratchets forward permanently.
- **Baseline contracts accumulate** in the repo — one set per restart — growing the watcher's active set indefinitely (perf degradation).
- Not a direct fund-loss vector (old baselines stay `active` so VTXOs at old addresses are still detected), but the UX and state pollution are unacceptable.

**Missing test**: There's no test for "second boot on the same repos WITHOUT rotation preserves the same address." The existing persistence test (`"second wallet on the same repos reads the rotated address"`) only covers the post-rotation case.

**Suggested fix** (pick one):
1. **Tag the boot allocation**: In `resolveBoot`, after allocating the first descriptor, register a tagged display contract immediately (`metadata.source = WALLET_RECEIVE_SOURCE`) so the next boot finds it.
2. **Add a `getCurrentOrNext` method to HDDescriptorProvider**: If `lastIndexUsed` is defined, return the same index without incrementing; only allocate a fresh index when called explicitly for rotation.
3. **Peek before allocating**: In `defaultBoot`, read `lastIndexUsed` from the wallet repo. If defined, re-derive the descriptor at that index without incrementing. Only call `getNextSigningDescriptor` when `lastIndexUsed` is `undefined`.

Option 3 is cleanest because it keeps the provider's `getNextSigningDescriptor` pure (always advances) while making the boot path idempotent.

---

### 🟡 MEDIUM — Silent error swallowing in rotation chain burns an index

**`src/wallet/walletReceiveRotator.ts:157-159`**

```typescript
this.chain = this.chain
    .catch(() => undefined)
    .then(() => this.rotate(wallet));
```

If `rotate()` fails partway — specifically after `getNextSigningDescriptor()` (line 211) but before `createContract()` (line 247) completes — the allocated index is consumed permanently, the wallet's **in-memory** `offchainTapscript` is already swapped to the new pubkey (line 213-216), but no contract is registered for it. The `.catch(() => undefined)` swallows the error silently.

Post-failure state:
- `wallet.defaultContractScript` points to an address with no registered contract.
- The `vtxo_received` handler compares against this new script, so events for the OLD script no longer trigger rotation.
- The watcher doesn't know about the new script, so VTXOs sent to the displayed address are invisible.
- **This is a fund-loss window**: the user sees address N, shares it, someone sends funds, but the wallet never detects them.

On restart, `pickActiveReceive` would fall back to the previous tagged contract (or the baseline), so this is a transient issue within a single session. But the burned index is permanent.

**Suggested fix**: If `createContract` fails, roll back `wallet.offchainTapscript` to its pre-rotation value. Consider logging the error rather than swallowing it. The chain serialization is correct for the happy path, but the error path needs a compensating action.

---

### 🟢 Minor / Style

1. **`src/wallet/wallet.ts:196` — `offchainTapscript` visibility**: `readonly` → `public` on `ReadonlyWallet` is a wider surface than necessary. Since `RotatableWallet` is the write interface, the mutation surface is architecturally contained, but any code with a `ReadonlyWallet` reference can now mutate it at the type level. Consider a `@internal` JSDoc tag at minimum.

2. **`src/wallet/walletReceiveRotator.ts:349` — `pickActiveReceive` sort stability**: `.sort((a, b) => b.createdAt - a.createdAt)` — if two contracts share the same `createdAt` (same millisecond), the sort is non-deterministic. Low probability in production, but could cause flaky test behavior. Add a tiebreaker (e.g., script comparison).

3. **Cross-repo impact**: Minimal. No downstream repos in the org currently import `WalletConfig`, `RefreshVtxosOptions`, or `offchainTapscript`. The new `WalletMode` export and `includeInactive` option are additive. PR #440 (prerequisite) is merged; PR #441 (alternative) is closed. Clean dependency chain.

4. **`includeInactive` cursor logic** (`src/contracts/contractManager.ts`): Correct. The diff properly passes `includeInactive` as a separate option to `syncContracts`, keeping `options.contracts === undefined` so `mustUpdateCursor` stays true for the superset case. The four tests covering cursor advancement / non-advancement are solid.

---

### ⚠️ Protocol-Critical Flag

This PR changes **receive address derivation and rotation** — how the wallet decides where to receive funds. While no VTXOs, signing, forfeit, or exit paths are modified, the receive-address lifecycle directly determines whether incoming funds are detected. **Requesting human review before merge**, per protocol-critical code policy.

---

**TL;DR**: The architecture and test coverage are strong. Fix the boot-path address-drift bug (HIGH) and the rotation-failure error handling (MEDIUM), and this is ready.

src/wallet/index.ts
```diff
@@ -14,8 +15,29 @@ import { DelegatorProvider } from "../providers/delegator";
...
+ *   identity. Throws at `Wallet.create` if the identity isn't HD-capable
+ *   or its descriptor isn't rangeable — no silent fallback.
+ * - A {@link DescriptorProvider} instance: rotate via the supplied
+ *   provider on every incoming VTXO. The wallet does not probe the
+ *   identity; the caller is responsible for ensuring the identity can
+ *   sign for whatever pubkey the provider returns. Errors thrown by the
+ *   provider propagate — there is no silent fallback for an explicit
+ *   provider.
+ */
+export type WalletMode = "auto" | "static" | "hd" | DescriptorProvider;
```
tiero, line 40: mmm seems to me a leaky abstraction to make people be even aware what HD is? What if the wallet is always   in rotate mode, and advanced user may use lower level method to derive "static" address if ever needed?

pietro909: Agree on the HD visibility part: ` "auto" | "static" | "dynamic" `  would communicate the mode without having to know what "hd" means

Kukks:
Pushed `ae1644e0` addressing the overlapping findings from @arkanaai and CodeRabbit. Summary of each fix:

### 1. `rotate()` mutated tapscript before `createContract` succeeded — FIXED
Both reviewers flagged this. The new `rotate()` builds `newTapscript`, `newScript`, and `newAddress` entirely locally, calls `createContract` first, and **only then** assigns `wallet.offchainTapscript = newTapscript`. If `createContract` throws, the wallet keeps displaying the previously-registered address. The HD index is still consumed by `getNextSigningDescriptor()` — that's a separate concern (need a peek/commit primitive on the provider for a true rollback; the simpler "wallet state stays consistent" fix is what we needed for correctness).

### 2. `currentTaggedScript` updated before `setContractState` ran — FIXED
The order is now: capture `previousTagged`, call `setContractState(previousTagged, 'inactive')`, **then** assign `currentTaggedScript = newScript`. If `setContractState` throws, the next rotation will retry deactivating the same orphaned contract instead of racing forward.

### 3. Boot path burned a fresh HD index on every restart without rotation — FIXED
This was the address-drift bug Arkana flagged in the latest review. `HDDescriptorProvider` gains a `getCurrentSigningDescriptor()` peek that re-derives the descriptor at `lastIndexUsed` without advancing. `defaultBoot` now uses it (via a duck-typed `hasPeekableDescriptor` guard): when no tagged display contract exists AND the provider supports peek, it reuses the existing index instead of allocating fresh.

New regression test: `"second boot WITHOUT rotation keeps the same address (no index drift)"` asserts that `lastIndexUsed` stays at `0` and `getAddress()` returns the same value across restarts on the same repo.

### 4. Silent error swallowing in the rotation chain — FIXED
Replaced `.catch(() => undefined)` with `.catch((err) => console.error("WalletReceiveRotator: rotation failed", err))`. The chain still recovers (next rotation can run) but failures are visible to operators instead of dropped.

### New regression tests (2)
- `"second boot WITHOUT rotation keeps the same address (no index drift)"` — locks in fix #3.
- `"createContract failure during rotation does NOT mutate the wallet's displayed tapscript"` — locks in fix #1: spies on `manager.createContract`, makes it reject once, dispatches a `vtxo_received` event, and asserts `defaultContractScript` / `getAddress()` are unchanged.

### Not addressed (yet)
- **Multi-timelock gap on rotation** (Arkana 🟡): `rotate()` still registers one contract at the primary `csvTimelock`, while `initializeContractManager` registers one per `walletContractTimelocks` entry. Self-heals on next boot. Low-risk, but happy to add the iteration if you want consistency.
- **Auto-mode `catch {}` on `HDDescriptorProvider.create`** (Arkana 🟢): the silent fallback in `resolveDescriptorProvider`'s `'auto'` branch still swallows descriptor-mismatch. `walletMode: 'hd'` is the strict alternative. Would add a `console.warn` if you want loud-on-mismatch even in `auto`.
- **`isHDCapableIdentity` partial structural check** (CodeRabbit nit): the guard only validates 4 of ~8 required members. Intentional per docstring, but a `Pick<...>` return type would be more honest. Happy to switch.
- **`offchainTapscript` mutation-during-send concern** (CodeRabbit on `wallet.ts:196-202`): legitimate but wider than this PR. A snapshot-at-start-of-send pattern is the right shape, but it touches send/settle flows. Filing as a follow-up unless you want it in scope.
- **Strict `>` on cursor advancement test** (CodeRabbit on `manager.test.ts:474-476`): trivially tightenable.
- **`deriveLeafPubkey` error includes the full descriptor** (CodeRabbit): low-risk because the descriptor only contains xpubs in practice, but I can redact if you prefer defence-in-depth.

Verification:
- `pnpm exec tsc --noEmit` clean
- `pnpm test:unit` — **1160/1161** pass (was 1158; +2 regressions)
- `pnpm format` clean
- `pnpm lint` clean

Happy to address the "Not addressed" list above — let me know which ones you want in this PR vs. follow-ups.

ghost (changes requested):
## Arkana Code Review — PR #473 (Update: May 11 force-push)

**Reviewer**: Arkana (aggressive protocol review)
**Scope**: Incremental — focusing on changes since last review (4-commit rebase: `6d7ade2..ae1644e`)

---

### What's fixed since last review ✅

The force-push addressed several critical findings from my previous reviews. Credit where due:

1. **Rotation ordering** — `rotate()` now correctly captures `previousTagged` before calling `setContractState`, then updates `currentTaggedScript` last. If `setContractState` throws, the next rotation retries deactivating the same orphaned contract instead of racing forward. (`walletReceiveRotator.ts:255-268`)

2. **Mutation before createContract** — `rotate()` now builds the new tapscript + script + address entirely locally. `wallet.offchainTapscript = newTapscript` only runs AFTER `createContract` succeeds. A regression test (`createContract failure during rotation does NOT mutate the wallet's displayed tapscript`) validates this. Clean fix.

3. **Index drift on restart** — `getCurrentSigningDescriptor()` on `HDDescriptorProvider` peeks the already-allocated index when no tagged display contract exists. No more burning a fresh HD index per cold restart. Validated by the `second boot WITHOUT rotation keeps the same address` test.

4. **Source tagging** — `metadata.source = 'wallet-receive'` disambiguates rotated display contracts from baseline contracts. `pickActiveReceive` filters on the tag, not "any active default contract." Eliminates the false-positive boot-pickup vector.

5. **Error surfacing** — Rotation chain now calls `console.error` on failure instead of silently swallowing.

---

### 🔴 Critical: Rotation/transaction race (still open from prior review)

**File**: `src/wallet/walletReceiveRotator.ts:243` + `src/wallet/wallet.ts:2690-2705`

`rotate()` writes `wallet.offchainTapscript = newTapscript` (walletReceiveRotator.ts:243) without acquiring `_txLock`. Meanwhile, `updateDbAfterOffchainTx` reads `this.offchainTapscript` across 4 separate field accesses (wallet.ts:2690-2705) under `_txLock`.

**Concrete scenario**: A `vtxo_received` event fires while a `send()` is in-flight. The rotation runs on the microtask queue (Promise chain, no lock), and writes `wallet.offchainTapscript` between the `forfeit()` read (line 2690) and the `pkScript` read (line 2705). The change VTXO is stamped with **mixed scripts** — forfeit from the OLD tapscript, pkScript from the NEW one. The VTXO becomes unspendable or bound to the wrong contract.

**This is a fund-loss vector.** The rotation event is triggered by VTXO reception, which is exactly when the wallet might also be in the middle of a send.

**Fix options** (pick one):
- **(A)** Snapshot `this.offchainTapscript` into a local const at the top of `updateDbAfterOffchainTx` and use the snapshot for all 4 reads. Cheapest fix, no lock contention.
- **(B)** Have `rotate()` acquire `_txLock` so it serializes against send/settle. Semantically cleaner but adds latency to rotation.

Option A is strongly preferred — it's a 1-line change and makes the read side immune to any future mutation.

---

### 🟡 Medium: Descriptor leaked in error message

**File**: `src/wallet/walletReceiveRotator.ts:283-286` (`deriveLeafPubkey`)

```typescript
throw new Error(
    `Cannot derive leaf pubkey from descriptor "${descriptor}": ...`
);
```

The full descriptor string is interpolated into the error. Even though descriptors are normally xpub-based, a misconfigured caller could pass an xprv. Error messages surface in logs, crash reporters, and Sentry. Redact or truncate:

```typescript
throw new Error(
    `Cannot derive leaf pubkey from descriptor (length=${descriptor.length}): ...`
);
```

---

### 🟡 Medium: Test cursor assertion is non-strict

**File**: `test/contracts/manager.test.ts:475`

```typescript
expect((stateAfter?.lastSyncTime ?? 0) >= SEEDED_CURSOR).toBe(true);
```

This passes even if the cursor didn't advance (equal case). The whole point of the test is proving the cursor **moved forward**. Use `>` or `toBeGreaterThan`:

```typescript
expect(stateAfter?.lastSyncTime ?? 0).toBeGreaterThan(SEEDED_CURSOR);
```

---

### 🟢 Observations (non-blocking)

1. **`isHDCapableIdentity` type guard** (hdCapableIdentity.ts:64-74) — Still checks only 4 members but narrows to the full `HDCapableIdentity` type. Callers that destructure the remaining Identity members (`sign`, `xOnlyPublicKey`, etc.) would hit runtime errors on a partial implementation. Low risk in practice since only `MnemonicIdentity`/`SeedIdentity` pass the guard, but the type annotation overpromises. Consider narrowing the return type or adding a note.

2. **`WalletMode` design** — Clean. The `'auto'` silent-fallback vs `'hd'` loud-failure distinction is well-thought-out. The escape hatch of passing a raw `DescriptorProvider` covers HSM/external-signer use cases without framework coupling.

3. **`includeInactive` cursor logic** — Correct. The `mustUpdateCursor` check (`options.contracts === undefined && options.window === undefined`) ensures `includeInactive` (which widens scope) is cursor-safe while `scripts` (which narrows scope) is not.

4. **Test coverage** — Significantly improved. The `createContract failure` regression test directly validates the prior ordering fix. The `second boot WITHOUT rotation` test catches index drift. The `second rotation deactivates previous tagged` test validates the retire chain.

---

### Verdict

The architecture is sound and the prior review findings are substantively addressed. The **rotation/transaction race** remains the sole blocker — it's a real fund-loss vector on any wallet that receives VTXOs while sending. Fix that (option A is a 1-line snapshot), fix the test assertion, and this is ready.

⚠️ **Protocol-critical**: This PR touches VTXO lifecycle, contract registration, and the receive-address derivation path. Requesting **human review** per protocol rules, even after the race fix lands.

ghost (changes requested):
## Arkana Code Review — PR #473 (Update: May 11 force-push)

**Reviewer**: Arkana (aggressive protocol review)
**Scope**: Incremental — focusing on changes since last review (4-commit rebase: `6d7ade2..ae1644e`)

---

### What's fixed since last review ✅

The force-push addressed several critical findings from my previous reviews. Credit where due:

1. **Rotation ordering** — `rotate()` now correctly captures `previousTagged` before calling `setContractState`, then updates `currentTaggedScript` last. If `setContractState` throws, the next rotation retries deactivating the same orphaned contract instead of racing forward. (`walletReceiveRotator.ts:255-268`)

2. **Mutation before createContract** — `rotate()` now builds the new tapscript + script + address entirely locally. `wallet.offchainTapscript = newTapscript` only runs AFTER `createContract` succeeds. A regression test (`createContract failure during rotation does NOT mutate the wallet's displayed tapscript`) validates this. Clean fix.

3. **Index drift on restart** — `getCurrentSigningDescriptor()` on `HDDescriptorProvider` peeks the already-allocated index when no tagged display contract exists. No more burning a fresh HD index per cold restart. Validated by the `second boot WITHOUT rotation keeps the same address` test.

4. **Source tagging** — `metadata.source = 'wallet-receive'` disambiguates rotated display contracts from baseline contracts. `pickActiveReceive` filters on the tag, not "any active default contract." Eliminates the false-positive boot-pickup vector.

5. **Error surfacing** — Rotation chain now calls `console.error` on failure instead of silently swallowing.

---

### 🔴 Critical: Rotation/transaction race (still open from prior review)

**File**: `src/wallet/walletReceiveRotator.ts:243` + `src/wallet/wallet.ts:2690-2705`

`rotate()` writes `wallet.offchainTapscript = newTapscript` (walletReceiveRotator.ts:243) without acquiring `_txLock`. Meanwhile, `updateDbAfterOffchainTx` reads `this.offchainTapscript` across 4 separate field accesses (wallet.ts:2690-2705) under `_txLock`.

**Concrete scenario**: A `vtxo_received` event fires while a `send()` is in-flight. The rotation runs on the microtask queue (Promise chain, no lock), and writes `wallet.offchainTapscript` between the `forfeit()` read (line 2690) and the `pkScript` read (line 2705). The change VTXO is stamped with **mixed scripts** — forfeit from the OLD tapscript, pkScript from the NEW one. The VTXO becomes unspendable or bound to the wrong contract.

**This is a fund-loss vector.** The rotation event is triggered by VTXO reception, which is exactly when the wallet might also be in the middle of a send.

**Fix options** (pick one):
- **(A)** Snapshot `this.offchainTapscript` into a local const at the top of `updateDbAfterOffchainTx` and use the snapshot for all 4 reads. Cheapest fix, no lock contention.
- **(B)** Have `rotate()` acquire `_txLock` so it serializes against send/settle. Semantically cleaner but adds latency to rotation.

Option A is strongly preferred — it's a 1-line change and makes the read side immune to any future mutation.

---

### 🟡 Medium: Descriptor leaked in error message

**File**: `src/wallet/walletReceiveRotator.ts:283-286` (`deriveLeafPubkey`)

```typescript
throw new Error(
    `Cannot derive leaf pubkey from descriptor "${descriptor}": ...`
);
```

The full descriptor string is interpolated into the error. Even though descriptors are normally xpub-based, a misconfigured caller could pass an xprv. Error messages surface in logs, crash reporters, and Sentry. Redact or truncate:

```typescript
throw new Error(
    `Cannot derive leaf pubkey from descriptor (length=${descriptor.length}): ...`
);
```

---

### 🟡 Medium: Test cursor assertion is non-strict

**File**: `test/contracts/manager.test.ts:475`

```typescript
expect((stateAfter?.lastSyncTime ?? 0) >= SEEDED_CURSOR).toBe(true);
```

This passes even if the cursor didn't advance (equal case). The whole point of the test is proving the cursor **moved forward**. Use `>` or `toBeGreaterThan`:

```typescript
expect(stateAfter?.lastSyncTime ?? 0).toBeGreaterThan(SEEDED_CURSOR);
```

---

### 🟢 Observations (non-blocking)

1. **`isHDCapableIdentity` type guard** (hdCapableIdentity.ts:64-74) — Still checks only 4 members but narrows to the full `HDCapableIdentity` type. Callers that destructure the remaining Identity members (`sign`, `xOnlyPublicKey`, etc.) would hit runtime errors on a partial implementation. Low risk in practice since only `MnemonicIdentity`/`SeedIdentity` pass the guard, but the type annotation overpromises. Consider narrowing the return type or adding a note.

2. **`WalletMode` design** — Clean. The `'auto'` silent-fallback vs `'hd'` loud-failure distinction is well-thought-out. The escape hatch of passing a raw `DescriptorProvider` covers HSM/external-signer use cases without framework coupling.

3. **`includeInactive` cursor logic** — Correct. The `mustUpdateCursor` check (`options.contracts === undefined && options.window === undefined`) ensures `includeInactive` (which widens scope) is cursor-safe while `scripts` (which narrows scope) is not.

4. **Test coverage** — Significantly improved. The `createContract failure` regression test directly validates the prior ordering fix. The `second boot WITHOUT rotation` test catches index drift. The `second rotation deactivates previous tagged` test validates the retire chain.

5. **Cross-repo impact** — Verified: no downstream consumers (banco, demos, asset-demos) import `WalletConfig`, `WalletMode`, `DescriptorProvider`, or call `refreshVtxos`. All reads of `offchainTapscript` are read-only. The `readonly → public` change is backwards-compatible. **No cross-repo breakage.**

---

### Verdict

The architecture is sound and the prior review findings are substantively addressed. The **rotation/transaction race** remains the sole blocker — it's a real fund-loss vector on any wallet that receives VTXOs while sending. Fix that (option A is a 1-line snapshot), fix the test assertion, and this is ready.

⚠️ **Protocol-critical**: This PR touches VTXO lifecycle, contract registration, and the receive-address derivation path. Requesting **human review** per protocol rules, even after the race fix lands.

ghost (changes requested):
## Arkana Code Review — PR #473 (Update: May 12 `15eb9056`)

**Reviewer**: Arkana (aggressive protocol review)
**Scope**: Delta since last review (`ae1644e0` → `15eb9056`)

---

### New in `15eb9056` (CodeRabbit auto-fixes)

**✅ Cross-family contract restore** — `pickActiveReceive` now accepts `expectedType` and `resolveBoot` passes the correct family derived from the wallet's tapscript shape. Correct fix, no issues.

**✅ Narrower error swallowing** — `resolveBoot` catch block now only swallows errors matching `"wildcard descriptor"` under `auto` mode; everything else rethrows. Correct fix.

---

### Outstanding findings (still open from prior reviews)

#### 🔴 BLOCKER: Rotation/transaction race on `offchainTapscript`

**Files**: `src/wallet/wallet.ts:2690-2705`, `src/wallet/walletReceiveRotator.ts:448`

`updateDbAfterOffchainTx` reads `this.offchainTapscript` four times without snapshotting:
- L2690: `.forfeit()`
- L2691: `.forfeit()`
- L2694: `.encode()`
- L2705: `.pkScript`

Meanwhile `rotate()` mutates `wallet.offchainTapscript` at L448. If a `vtxo_received` event fires mid-`updateDbAfterOffchainTx`, the change VTXO gets a Frankenstein tapscript — old forfeit leaf + new pkScript. The VTXO becomes unspendable.

**Fix**: Snapshot at the top of `updateDbAfterOffchainTx`:
```typescript
const tapscript = this.offchainTapscript; // snapshot — rotation may swap this mid-method
```
Then use `tapscript` for all 4 reads. One-line fix.

#### 🟡 Descriptor leaked in error message

**File**: `src/wallet/walletReceiveRotator.ts:480`

```typescript
`Cannot derive leaf pubkey from descriptor "${descriptor}": ...`
```

The full descriptor string is interpolated. A misconfigured caller passing an xprv leaks private key material into logs/crash reporters. Redact:
```typescript
`Cannot derive leaf pubkey from descriptor (length=${descriptor.length}): ...`
```

#### 🟡 Test cursor assertion is non-strict

**File**: `test/contracts/manager.test.ts:476`

```typescript
expect((stateAfter?.lastSyncTime ?? 0) >= SEEDED_CURSOR).toBe(true);
```

Passes even when the cursor doesn't move (equal case). The test's purpose is proving the cursor **advanced**. Use:
```typescript
expect(stateAfter?.lastSyncTime ?? 0).toBeGreaterThan(SEEDED_CURSOR);
```

#### 🟡 Multi-timelock gap on rotation

**File**: `src/wallet/walletReceiveRotator.ts:388-440`

`rotate()` registers one contract at the primary `csvTimelock`. `initializeContractManager` registers one per `walletContractTimelocks` entry. After rotation, only the primary timelock has the new pubkey; other timeslots point at stale keys until next boot. Self-heals on restart but creates a window where funds sent to non-primary-timelock rotated addresses aren't watched.

Low risk (most wallets have one timelock), but worth noting for multi-timelock configurations.

---

### Verdict

The `15eb9056` fixes are correct and well-scoped. The **rotation/transaction race** remains the sole blocker — it's a real fund-loss vector. The other three are medium/low priority but should land before merge.

⚠️ **Protocol-critical**: This PR touches VTXO lifecycle, contract registration, and receive-address derivation. **Human review required** per protocol rules.

ghost (changes requested):
## Arkana Code Review — PR #473 (Update: May 12 `15eb9056`)

**Reviewer**: Arkana (aggressive protocol review)
**Scope**: Delta since last review (`ae1644e0` → `15eb9056`)

---

### New in `15eb9056` (CodeRabbit auto-fixes)

**✅ Cross-family contract restore** — `pickActiveReceive` now accepts `expectedType` and `resolveBoot` passes the correct family derived from the wallet's tapscript shape. Correct fix, no issues.

**✅ Narrower error swallowing** — `resolveBoot` catch block now only swallows errors matching `"wildcard descriptor"` under `auto` mode; everything else rethrows. Correct fix.

---

### Outstanding findings (still open from prior reviews)

#### 🔴 BLOCKER: Rotation/transaction race on `offchainTapscript`

**Files**: `src/wallet/wallet.ts:2690-2705`, `src/wallet/walletReceiveRotator.ts:448`

`updateDbAfterOffchainTx` reads `this.offchainTapscript` four times without snapshotting:
- L2690: `.forfeit()`
- L2691: `.forfeit()`
- L2694: `.encode()`
- L2705: `.pkScript`

Meanwhile `rotate()` mutates `wallet.offchainTapscript` at L448. If a `vtxo_received` event fires mid-`updateDbAfterOffchainTx`, the change VTXO gets a Frankenstein tapscript — old forfeit leaf + new pkScript. The VTXO becomes unspendable.

**Fix**: Snapshot at the top of `updateDbAfterOffchainTx`:
```typescript
const tapscript = this.offchainTapscript; // snapshot — rotation may swap this mid-method
```
Then use `tapscript` for all 4 reads. One-line fix.

#### 🟡 Descriptor leaked in error message

**File**: `src/wallet/walletReceiveRotator.ts:480`

```typescript
`Cannot derive leaf pubkey from descriptor "${descriptor}": ...`
```

The full descriptor string is interpolated. A misconfigured caller passing an xprv leaks private key material into logs/crash reporters. Redact:
```typescript
`Cannot derive leaf pubkey from descriptor (length=${descriptor.length}): ...`
```

#### 🟡 Test cursor assertion is non-strict

**File**: `test/contracts/manager.test.ts:476`

```typescript
expect((stateAfter?.lastSyncTime ?? 0) >= SEEDED_CURSOR).toBe(true);
```

Passes even when the cursor doesn't move (equal case). The test's purpose is proving the cursor **advanced**. Use:
```typescript
expect(stateAfter?.lastSyncTime ?? 0).toBeGreaterThan(SEEDED_CURSOR);
```

#### 🟡 Multi-timelock gap on rotation

**File**: `src/wallet/walletReceiveRotator.ts:388-440`

`rotate()` registers one contract at the primary `csvTimelock`. `initializeContractManager` registers one per `walletContractTimelocks` entry. After rotation, only the primary timelock has the new pubkey; other timeslots point at stale keys until next boot. Self-heals on restart but creates a window where funds sent to non-primary-timelock rotated addresses aren't watched.

Low risk (most wallets have one timelock), but worth noting for multi-timelock configurations.

---

### Verdict

The `15eb9056` fixes are correct and well-scoped. The **rotation/transaction race** remains the sole blocker — it's a real fund-loss vector. The other three are medium/low priority but should land before merge.

⚠️ **Protocol-critical**: This PR touches VTXO lifecycle, contract registration, and receive-address derivation. **Human review required** per protocol rules.

Kukks:
Rebased onto master (`ab0c67e4`) and pushed `203fa4ab` addressing the May 11 Arkana review.

### Fixes
- **🔴 Rotation/transaction race**: `updateDbAfterOffchainTx` now snapshots `this.offchainTapscript` AND the display address synchronously at function entry, before any `await`. The five reads (forfeit ×2, encode, pkScript, primaryAddr) all consume the snapshot so a concurrent `vtxo_received` → `rotate()` between awaits can no longer mix scripts. Option (A) from your recommendation — local snapshot, no extra lock contention.

- **🟡 Descriptor leak in error**: `deriveLeafPubkey` redacts to `(length=N)` instead of interpolating the descriptor. Defends against accidental xprv exposure in logs / Sentry.

- **🟡 Cursor assertion non-strict**: `expect(... >= SEEDED_CURSOR).toBe(true)` → `expect(...).toBeGreaterThan(SEEDED_CURSOR)`. The previous form passed even on the no-op (cursor unchanged) case, which defeated the test.

### Rebase notes
One conflict in `src/contracts/contractManager.ts:syncContracts` — master had changed its return type from `Map<string, ContractVtxo[]>` to `Map<string, ExtendedContractVtxo[]>` while this branch added the `includeInactive?: boolean` field. Kept both.

### Verification
- `pnpm exec tsc --noEmit` clean
- `pnpm test:unit` — **1162/1163** pass (was 1160; +2 from master's new tests)
- `pnpm format` clean

Integration CI rerunning on the rebased tip — the previous run's `should unroll` insufficient-funds failure was on the pre-rebase commit, and master's own CI is green on the new tip, so I'm not adding any extra investigation there until the fresh run reports back.

### Still in the "non-blocking" pile from prior reviews
- Multi-timelock gap on rotation (Arkana 🟡)
- `isHDCapableIdentity` partial structural check (CodeRabbit / Arkana 🟢)
- `auto`-mode silent fallback warn-on-mismatch (Arkana 🟢)
- `pickActiveReceive` `createdAt` tiebreaker (Arkana 🟢)
- `offchainTapscript` `readonly→public` `@internal` JSDoc (Arkana 🟢)

Happy to address any of these in this PR if you want, otherwise filing as follow-ups.

ghost (approved):
## Arkana Code Review — PR #473 (Update: May 12 `203fa4a`)

**Reviewer**: Arkana (aggressive protocol review)
**Scope**: Delta since last review (`15eb9056` → `203fa4a`)

---

### Previous findings — all resolved ✅

#### 🔴 → ✅ Rotation/transaction race on `offchainTapscript`

**File**: `src/wallet/wallet.ts:2638-2649`

Correctly fixed. `updateDbAfterOffchainTx` now snapshots both `offchainTapscript` and `primaryAddress` synchronously at function entry (before any `await`), and all downstream reads use the snapshot. The comment explains the invariant clearly. The Frankenstein-tapscript vector is closed.

#### 🟡 → ✅ Descriptor leaked in error message

**File**: `src/wallet/walletReceiveRotator.ts:479-484`

Now emits `(length=${descriptor.length})` instead of the raw descriptor. Good.

#### 🟡 → ✅ Test cursor assertion

**File**: `test/contracts/manager.test.ts:474-476`

Now uses `toBeGreaterThan(SEEDED_CURSOR)` — strict forward check.

---

### New changes in this delta — reviewed

**✅ `ExtendedContractVtxo` type** (`src/contracts/types.ts:85-101`): Correctly mirrors the `ExtendedVirtualCoin` / `VirtualCoin` split. `ContractVtxo` keeps `Partial<TapLeaves>` for raw indexer data; `ExtendedContractVtxo` narrows to required after annotation. `extraWitness` is inherited from `ExtendedVirtualCoin`. Return types in `contractManager.ts` updated consistently across all methods (`getVtxosForContracts`, `refreshContracts`, `fetchContractVtxosBulk`, etc.). No breakage.

**✅ `isAnnotated` type guard** (`src/wallet/delegator.ts:505-511`): Runtime narrowing before passing `ContractVtxo[]` to `makeDelegateForfeitTx`. Correct — filters out unannotated vtxos that would blow up on `tapTree` access.

**✅ `extractArkProviderUrl` + indexer resolution** (`src/wallet/wallet.ts:119-124, 269-286`): Clean provider URL fallback chain: explicit `indexerUrl` > derived from `arkProvider.serverUrl` > `arkadeServerUrl`. Throws when a custom `arkProvider` has no discoverable URL and no explicit `indexerUrl` — prevents silent cross-server pairing. Good defensive behavior.

**✅ Watcher enrichment tests** (`test/contracts/watcher.test.ts:255-390`): Two new test cases: (1) delegate contract vtxos get correct tapscript, not default fallback; (2) unknown handler type falls back to raw vtxo + console.warn. Good coverage for the `contractWatcher.ts` catch-path improvement.

**✅ `VirtualCoin` field reorder + JSDoc** (`src/wallet/index.ts:546-568`): Fields reordered for logical grouping; JSDoc improved. No structural change — all fields identical.

**✅ ESM import fixes** (various files + `scripts/add-extensions.js`): `.js` extensions added to `@scure/btc-signer` sub-path imports. Build script extended to also fix `.d.ts` declaration files. Build order corrected (add-extensions runs after all tsc passes). Mechanical, correct.

**✅ README cleanup**: Removed hardcoded `arkServerUrl` from examples — consistent with `DEFAULT_ARKADE_SERVER_URL` default.

---

### Still noted (unchanged, low-risk)

**🟡 Multi-timelock gap on rotation** (`walletReceiveRotator.ts:388-440`): `rotate()` still registers one contract at the primary `csvTimelock` only. Self-heals on next boot. Acknowledged as low-risk in the PR thread — tracking for future follow-up.

---

### Verdict

All three prior findings fixed correctly. New code (type narrowing, provider resolution, watcher tests) is clean and well-structured. The rotation/transaction race — the only blocker — is closed.

**Approving the code.** ⚠️ **Protocol-critical**: This PR touches VTXO lifecycle, contract registration, and receive-address derivation. **Human review still required** per protocol rules before merge.

Kukks:
On the "multi-timelock gap on rotation" item — that was misframed as low-risk-self-heals. It's an actual design choice the boot path wasn't honoring. Fixed in `ca2fa035`.

### The design (now explicit in code)

- **Index 0** = identity's x-only pubkey = the **permanent baseline**. The multi-timelock matrix (default + delegate × every `walletContractTimelocks` entry) lives here. Always active, never tagged.
- **Index 1, 2, …** = rotated display addresses. Intentionally single-timelock-single-pubkey at the current arkd delay. Registered by `rotate()`. Tagged with `metadata.source = WALLET_RECEIVE_SOURCE`.

### What the bug was

`initializeContractManager` built the matrix at `this.offchainTapscript.options.pubKey` — which on a rotated wallet is the *rotated* pubkey, not the identity pubkey. So every boot after rotation registered a full matrix at the rotated pubkey too, which is exactly what the design says NOT to do (rotated pubkeys are display-only, not baseline).

There was nothing to "self-heal" — there was something to NOT re-register.

### The fix

`initializeContractManager` now uses `await this.identity.xOnlyPublicKey()` for the matrix's `pubKey`. The matrix stays bound to index 0 across all boots regardless of rotation state. Static wallets are unaffected (identity pubkey == display pubkey, no change).

### Regression test

`does NOT re-register the multi-timelock matrix at a rotated pubkey on reboot`:

1. Boot fresh wallet, rotate once (so the next boot has a tagged display contract).
2. Boot a second wallet on the same repos.
3. Assert: the rotated pubkey appears in **exactly one** contract (the tagged display), never duplicated across timelocks. The matrix is at index 0, unchanged.

`pnpm test:unit` — 1163/1164 pass.

ghost (approved):
## Arkana Code Review — PR #473 (Update: May 12 `ca2fa03`)

**Reviewer**: Arkana (aggressive protocol review)  
**Scope**: Delta since last review (`203fa4a` → `ca2fa03`)

### Changes reviewed

Single commit `ca2fa03` — **fix(wallet): bind baseline matrix to index 0, not the rotated pubkey**

**What it does:** In `initializeContractManager`, the baseline multi-timelock contract matrix (default + delegate × every `walletContractTimelocks` entry) was previously using `this.offchainTapscript.options.pubKey`. After HD rotation + reboot, that field holds the *rotated* pubkey, so the matrix would silently shift away from index 0 — breaking the "index-0 stays active forever" invariant.

The fix introduces `const baselinePubkey = await this.identity.xOnlyPublicKey()` (which always returns `indexZero.pubkey` on HD identities) and uses it for both DefaultVtxo and DelegateVtxo script construction.

### Verdict

✅ **Correct fix.** Verified:
- `ReadonlyDescriptorIdentity.xOnlyPublicKey()` returns `this.indexZero.pubkey` — always index 0 regardless of HD state (`seedIdentity.ts:505-508`).
- `serverPubKey` and `delegatePubKey` still read from `this.offchainTapscript.options` — correct, these are server-side params unaffected by rotation.
- The comment block (`wallet.ts:937-949`) accurately describes the design invariant.
- New test (`walletHdRotation.test.ts:498-562`) covers the exact regression: rotate → reboot → assert matrix stays at index 0, rotated pubkey only appears in the single tagged display contract, and contract count doesn't grow.

### No open findings

All prior flagged items from my May 11/12 reviews have been addressed. No new issues in this delta.

🤖 Arkana

ghost (approved):
## Arkana Code Review — PR #473 (Update: May 12 `ca2fa03`)

**Reviewer**: Arkana (aggressive protocol review)  
**Scope**: Delta since last review (`203fa4a` → `ca2fa03`)

### Changes reviewed

Single commit `ca2fa03` — **fix(wallet): bind baseline matrix to index 0, not the rotated pubkey**

**What it does:** In `initializeContractManager`, the baseline multi-timelock contract matrix (default + delegate × every `walletContractTimelocks` entry) was previously using `this.offchainTapscript.options.pubKey`. After HD rotation + reboot, that field holds the *rotated* pubkey, so the matrix would silently shift away from index 0 — breaking the "index-0 stays active forever" invariant.

The fix introduces `const baselinePubkey = await this.identity.xOnlyPublicKey()` (which always returns `indexZero.pubkey` on HD identities) and uses it for both DefaultVtxo and DelegateVtxo script construction.

### Verdict

✅ **Correct fix.** Verified:
- `ReadonlyDescriptorIdentity.xOnlyPublicKey()` returns `this.indexZero.pubkey` — always index 0 regardless of HD state (`seedIdentity.ts:505-508`).
- `serverPubKey` and `delegatePubKey` still read from `this.offchainTapscript.options` — correct, these are server-side params unaffected by rotation.
- The comment block (`wallet.ts:937-949`) accurately describes the design invariant.
- New test (`walletHdRotation.test.ts:498-562`) covers the exact regression: rotate → reboot → assert matrix stays at index 0, rotated pubkey only appears in the single tagged display contract, and contract count doesn't grow.

### No open findings

All prior flagged items from my May 11/12 reviews have been addressed. No new issues in this delta.

⚠️ **Reminder**: This PR touches protocol-critical VTXO contract registration. Human approval required before merge.

🤖 Arkana

ghost (approved):
## Arkana Code Review — PR #473 (Update: May 12 `afb6def`)

**Reviewer**: Arkana (aggressive protocol review)  
**Scope**: Delta since last review (`ca2fa03` → `afb6def`)

### Changes reviewed

Single commit `afb6def` — **fix(unroll): Math.ceil feeRate before BigInt conversion**

**The bug:** `prepareUnrollTransaction` called `BigInt(feeRate)` where `feeRate` comes from `OnchainProvider.getFeeRate()`. Esplora / bitcoind regtest can return fractional sat/vB (e.g. `1.006`), causing `BigInt(1.006)` to throw `RangeError`. This is a real crash in the unilateral exit path — protocol-critical.

**The fix:** `BigInt(Math.ceil(feeRate))` — rounds up before conversion. Correct direction (overpay by < 1 sat/vB, never underpay). `Wallet.MIN_FEE_RATE = 1` is already an integer, so the floor guard doesn't bypass the ceil.

### Verified

- ✅ Only call site for `.fee(BigInt(feeRate))` in the codebase — no other instances of this bug pattern.
- ✅ `onchain.ts` handles `feeRate` differently (stays as `number`, uses `Math.ceil(feeRate * vsize)`) — not affected.
- ✅ Comment accurately describes the failure mode.

### No findings

Clean, minimal, correct fix for a crash in the unilateral exit path.

⚠️ **Reminder**: This PR touches protocol-critical code (unilateral exit / unroll path). Human approval required before merge.

🤖 Arkana

ghost (approved):
## Arkana Code Review — PR #473 (Update: May 12 `afb6def`)

**Reviewer**: Arkana (aggressive protocol review)  
**Scope**: Delta since last review (`ca2fa03` → `afb6def`)

### Changes reviewed

Single commit `afb6def` — **fix(unroll): Math.ceil feeRate before BigInt conversion**

**The bug:** `prepareUnrollTransaction` called `BigInt(feeRate)` where `feeRate` comes from `OnchainProvider.getFeeRate()`. Esplora / bitcoind regtest can return fractional sat/vB (e.g. `1.006`), causing `BigInt(1.006)` to throw `RangeError`. This is a real crash in the unilateral exit path — protocol-critical.

**The fix:** `BigInt(Math.ceil(feeRate))` — rounds up before conversion. Correct direction (overpay by < 1 sat/vB, never underpay). `Wallet.MIN_FEE_RATE = 1` is already an integer, so the floor guard doesn't bypass the ceil.

### Verified

- ✅ Only call site for `.fee(BigInt(feeRate))` in the codebase — no other instances of this bug pattern.
- ✅ `onchain.ts` handles `feeRate` differently (stays as `number`, uses `Math.ceil(feeRate * vsize)`) — not affected.
- ✅ Comment accurately describes the failure mode.

### No findings

Clean, minimal, correct fix for a crash in the unilateral exit path.

⚠️ **Reminder**: This PR touches protocol-critical code (unilateral exit / unroll path). Human approval required before merge.

🤖 Arkana

pietro909 (approved):

pietro909:
tested in the Wallet @Kukks: rotation works per-receive, DB updates. Let's go!

EDIT: I've used https://github.com/arkade-os/wallet/pull/545 which I also updated with latest master branch fyi

image

PR #475, opened 2026-05-04
docs: add CLAUDE.md and AGENTS.md repo guides

Kukks:
## Summary

Adds three repo-guide files for AI agents working in this repo, and a fully populated changelog for the 0.4.x line:

- **`CLAUDE.md`** — canonical agent doc. Focused on the recurring workflow we want agents to follow consistently:
  - **Version bumps**: how to find the diff since the last release commit, what to verify before bumping, and that bumps must always go through `pnpm release` (never hand-edit `package.json`).
  - **Changelog discipline**: explicit section ordering (`Breaking Changes` → `Features` → `Bug Fixes` → `Performance` → `Observability` → `Internal`), a per-fix entry style with a bolded headline + root-cause paragraph, and an example entry. Modeled on the changelog style used in [`arklabshq/btcpay-arkade`](https://github.com/ArkLabsHQ/btcpay-arkade/blob/master/CHANGELOG.md), which reads well a year later because each entry names the symptom and the cause, not just the file.
  - PR/commit conventions and a brief directory map for orientation.
- **`AGENTS.md`** — one-line pointer to `CLAUDE.md` so non-Claude agents (Codex, Gemini, etc.) land on the same canonical doc.
- **`CHANGELOG.md`** — backfilled for **0.4.0 → 0.4.23** in the format `CLAUDE.md` defines. Each entry leads with a bolded one-line headline and explains the root cause / capability with affected files referenced by name. Pre-0.4 history is left in `git log` (those releases predate this style and backfilling them would mostly be guesswork — called out in the file's header).

## Notes for reviewers

- `CLAUDE.md` was previously listed under the `# AI` block in `.gitignore`. Removed that single line so the new file is tracked. `TASKS.md`, `*.agents.md`, and `REVIEW.md` remain ignored as agent scratch files.
- Source for the changelog entries: squash-commit bodies on `master` between each release tag (`git log v0.4.{n-1}..v0.4.{n}`), with cross-references to the originating PRs.
- No code changes; docs only.

## Test plan

- [x] `pnpm lint` passes (ran via husky pre-commit hook on all three commits)
- [x] `pnpm test:unit` passes (1087 passed / 1 skipped)
- [ ] CI green on PR

PR #477, opened 2026-05-05
fix(wallet): reconcile VTXO_ALREADY_SPENT by outpoint, not full re-scan

Kukks:
## Summary

When the server rejects an intent with \`VTXO_ALREADY_SPENT\`, the wallet's local cache disagrees with the server's authoritative view of that one VTXO. The previous recovery path was a full \`refreshVtxos()\` sweep — but that uses a cursor-derived \`?after=\` filter on \`created_at\`, so a VTXO **created before the cursor and spent recently** can never surface there. The local cache stays stale, the renewal selector keeps picking the same stuck VTXO, and the loop repeats every \`pollIntervalMs\`.

Captured in the wild as a 60-second cycle:

- \`POST /v1/batch/registerIntent → 400 VTXO_ALREADY_SPENT (input \`6400…:0\` already spent)\`
- Wallet calls \`maybeRefreshAfterVtxoSpent\` → broad \`refreshVtxos()\`
- Same VTXO returned next cycle. Forever.

## Fix

Add a surgical recovery path that targets the offending outpoint instead of doing a full re-scan.

### \`IContractManager.refreshOutpoints(outpoints)\`

Queries the indexer by outpoint (no cursor filter), annotates with the owning contract's tapscripts, upserts into the wallet repository at that contract's address. No cursor change, no full re-scan, no traffic for vtxos we already know about.

Outpoints not owned by any tracked contract are silently dropped.

### \`maybeRefreshAfterVtxoSpent(spentOutpoint?)\`

The server attaches \`metadata.vtxo_outpoint\` to its ArkError envelope for \`VTXO_ALREADY_SPENT\`. The wallet now extracts that via \`maybeArkError\` and routes the recovery through \`refreshOutpoints([outpoint])\`. Falls back to the broader \`refreshVtxos()\` only when no outpoint metadata is available (e.g. tests using plain \`new Error('VTXO_ALREADY_SPENT')\`).

Both call sites that catch \`VTXO_ALREADY_SPENT\` in \`vtxo-manager.ts\` (event-driven renewal + periodic settle) wire up the parsed outpoint.

### Service-worker proxy

New \`REFRESH_OUTPOINTS\` message + handler so wallets running behind a service worker get the same precise recovery.

## Tests

**\`test/contracts/manager.test.ts\`** (3 new):
- happy path: queries indexer by outpoint, upserts into repo with the right contract address, the \`isSpent\` flag is visible afterward
- silent skip when the indexer returns a script we don't track
- no-op for empty input

**\`test/vtxo-manager.test.ts\`** (2 new):
- structured ArkError carrying \`vtxo_outpoint\` → \`refreshOutpoints([{txid, vout}])\` is called, \`refreshVtxos\` is NOT
- plain \`Error('VTXO_ALREADY_SPENT')\` (no metadata) → falls back to \`refreshVtxos\` (existing behaviour preserved)

## Test plan

- [x] 5 new tests; all pass
- [x] Full unit suite: 1092 tests pass (5 new on top of 1087 baseline)
- [x] Lint clean
- [x] Typecheck clean

## Related

- Companion to #476 (\`refreshVtxos\` cursor short-circuit) — that PR fixes the wasteful traffic; this one fixes the underlying \"local cache never gets the truth\" loop. Both can be merged independently in either order.

## Summary by CodeRabbit

* **New Features**
  * Added targeted outpoint refresh capability and service-worker messaging to request/acknowledge outpoint refreshes with a 30s timeout.
* **Bug Fixes**
  * VTXO stale/spent errors now trigger focused outpoint reconciliation instead of always performing a full refresh.
* **Tests**
  * New tests cover outpoint refresh behavior, skipping unknown scripts, and periodic-settle error handling (including parsing/fallback cases).

### DIFF
--- a/src/contracts/contractManager.ts
+++ b/src/contracts/contractManager.ts
@@ -14,7 +14,7 @@ import {
 } from "./types";
 import { ContractWatcher, ContractWatcherConfig } from "./contractWatcher";
 import { contractHandlers } from "./handlers";
-import { ExtendedVirtualCoin, VirtualCoin } from "../wallet";
+import { ExtendedVirtualCoin, Outpoint, VirtualCoin } from "../wallet";
 import { extendVirtualCoinForContract } from "../wallet/utils";
 import { ContractFilter, ContractRepository } from "../repositories";
 import {
@@ -130,6 +130,22 @@ export interface IContractManager extends Disposable {
      * With options, narrows the refresh to specific scripts and/or a time window.
      */
     refreshVtxos(opts?: RefreshVtxosOptions): Promise<void>;
+
+    /**
+     * Reconcile specific outpoints with the indexer's authoritative state and
+     * upsert the result into the wallet repository.
+     *
+     * The cursor-derived delta sync filters by `created_at`, so a VTXO that
+     * was created before the cursor but spent recently won't surface in a
+     * standard `refreshVtxos()` call. This method is the surgical recovery
+     * path for that case: when something hands us a stale outpoint (e.g. the
+     * server returns `VTXO_ALREADY_SPENT` with a `vtxo_outpoint` in its
+     * error metadata), call this to pull the latest state and unblock the
+     * caller — no full re-scan, no cursor change.
+     *
+     * Outpoints not owned by any tracked contract are silently dropped.
+     */
+    refreshOutpoints(outpoints: Outpoint[]): Promise<void>;
 
     /**
      * Whether the underlying watcher is currently active.
@@ -636,6 +652,38 @@ export class ContractManager implements IContractManager {
                 ? { after: opts?.after, before: opts?.before }
                 : undefined,
         });
+    }
+
+    async refreshOutpoints(outpoints: Outpoint[]): Promise<void> {
+        if (outpoints.length === 0) return;
+
+        const { vtxos } = await this.config.indexerProvider.getVtxos({
+            outpoints,
+        });
+        if (vtxos.length === 0) return;
+
+        // Filter to outputs whose script we own. Map them to their owning
+        // contract so we can write through to the right per-address entry
+        // in the wallet repository.
+        const scripts = Array.from(new Set(vtxos.map((v) => v.script)));
+        const contracts = await this.config.contractRepository.getContracts({
+            script: scripts,
+        });
+        const scriptToContract = new Map(contracts.map((c) => [c.script, c]));
+        const owned = vtxos.filter((v) => scriptToContract.has(v.script));
+        if (owned.length === 0) return;
+
+        const annotated = await this.annotateVtxos(owned);
+        const byAddress = new Map<string, ExtendedVirtualCoin[]>();
+        for (const vtxo of annotated) {
+            const address = scriptToContract.get(vtxo.script)!.address;
+            const arr = byAddress.get(address) ?? [];
+            arr.push(vtxo);
+            byAddress.set(address, arr);
+        }
+        for (const [address, addressVtxos] of byAddress) {
+            await this.config.walletRepository.saveVtxos(address, addressVtxos);
+        }
     }
 
     /**
--- a/src/wallet/serviceWorker/wallet-message-handler.ts
+++ b/src/wallet/serviceWorker/wallet-message-handler.ts
@@ -267,6 +267,16 @@ export type ResponseRefreshVtxos = ResponseEnvelope & {
     type: "REFRESH_VTXOS_SUCCESS";
 };
 
+export type RequestRefreshOutpoints = RequestEnvelope & {
+    type: "REFRESH_OUTPOINTS";
+    payload: {
+        outpoints: { txid: string; vout: number }[];
+    };
+};
+export type ResponseRefreshOutpoints = ResponseEnvelope & {
+    type: "REFRESH_OUTPOINTS_SUCCESS";
+};
+
 export type RequestGetAllSpendingPaths = RequestEnvelope & {
     type: "GET_ALL_SPENDING_PATHS";
     payload: { options: GetAllSpendingPathsOptions };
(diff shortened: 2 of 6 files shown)


ghost (approved):
## Arkana Code Review — APPROVE ✅

Reviewed the full diff + surrounding source for `contractManager.ts`, `vtxo-manager.ts`, `wallet-message-handler.ts`, `wallet.ts`, and both test files.

### Summary

Clean, well-scoped fix for a real production bug. The cursor-based `refreshVtxos()` fundamentally cannot reconcile a VTXO that was created before the cursor but spent recently — adding a surgical outpoint-scoped query is the correct fix.

### Verified

1. **`refreshOutpoints` correctness** — Queries indexer by outpoint (no cursor filter), filters to scripts owned by tracked contracts, annotates with tapscripts via `annotateVtxos`, upserts into wallet repo grouped by address. `saveVtxos` does merge-by-key (`txid:vout`), so the spent flag propagates correctly without clobbering other VTXOs.

2. **`extractSpentOutpoint` parsing** — Validates: `maybeArkError` → name match → string metadata → split on `:` → integer ≥ 0. Robust against malformed payloads, returns `undefined` for anything unexpected, which correctly triggers the fallback path.

3. **Fallback preserved** — When no outpoint metadata is available (plain `Error`, old server version, etc.), the code falls back to the broader `refreshVtxos()`. Both paths exercised in tests.

4. **Service worker proxy** — New `REFRESH_OUTPOINTS` message type + handler + client-side method. Cases use `return`, no fall-through risk.

5. **Cross-repo impact** — `IContractManager` gains a new method, but only `ContractManager` and the service worker proxy implement it. No downstream SDK consumers of this TypeScript interface. No breaking change.

6. **Test coverage** — 5 new tests cover: happy path upsert, unowned script filtering, empty input no-op, structured ArkError → `refreshOutpoints`, plain error → `refreshVtxos` fallback. Good.

### One observation (non-blocking)

**Throttle window silently drops subsequent outpoints** (`vtxo-manager.ts:1147-1161`): If two *different* VTXOs are stale within the same cooldown window, only the first outpoint's `refreshOutpoints` runs — the second call returns the first's promise and its outpoint is never passed through. Self-healing on the next poll cycle, so not a bug, but worth knowing. If this ever becomes a problem, accumulating outpoints into a `Set` during the cooldown and flushing them all at once would be a clean fix.

### Not protocol-critical

This is cache reconciliation — no changes to signing, spending, forfeit paths, or VTXO lifecycle. Trust model unchanged (wallet already trusts indexer responses in `refreshVtxos`).

LGTM. Ship it.

src/contracts/contractManager.ts
```diff
@@ -638,6 +654,38 @@ export class ContractManager implements IContractManager {
...
+            script: scripts,
+        });
+        const scriptToContract = new Map(contracts.map((c) => [c.script, c]));
+        const owned = vtxos.filter((v) => scriptToContract.has(v.script));
+        if (owned.length === 0) return;
+
+        const annotated = await this.annotateVtxos(owned);
+        const byAddress = new Map<string, ExtendedVirtualCoin[]>();
+        for (const vtxo of annotated) {
+            const address = scriptToContract.get(vtxo.script)!.address;
```
pietro909, line 683: It's an edge case, but:
- `contractRepository.getContracts({script: scripts})` doesn't guarantee it returns all the scripts in `script` (line 669)
- if it doesn't, this line will fail with a generic `cannot read property address of undefined`

Better to not use the `!` operator and assume the Map may not contain all the scripts.

pietro909: Ok, I see line 673 actually filters them.

pietro909 (approved):

ghost (approved):
## Arkana Code Review — Re-review after merge-from-master ✅

Only new commit since my previous approval is `2a0b20c` — a merge of `master` into the feature branch, pulling in:

- `#481` (script-gating for persisted VTXOs via `vtxoOwnership.ts`)
- `#479` (unilateral exit bundle)
- `#475` (repo docs)
- `v0.4.24` release

**Merge is clean.** Verified:

1. **No conflicts in `contractManager.ts`** — `refreshOutpoints()` sits cleanly alongside the new `warnAndFilterVtxosForScript` call sites from #481. The method already does its own ownership filtering at `scriptToContract.has(v.script)` before writing, which is consistent with #481's approach.

2. **Service worker handler** — `REFRESH_OUTPOINTS` case at line 893 correctly placed after `REFRESH_VTXOS`, each with `return` (no fallthrough). No merge artifacts.

3. **`vtxo-manager.ts`** — `extractSpentOutpoint` and both `maybeRefreshAfterVtxoSpent` call sites unchanged by merge. Parsing logic remains sound.

4. **No new code, no new risk.** Previous approval stands.

LGTM — still approving.

ghost (approved):
## Arkana Code Review — Re-review after merge-from-master ✅

Only new commit since my previous approval is `2a0b20c` — a merge of `master` into the feature branch, pulling in:

- `#481` (script-gating for persisted VTXOs via `vtxoOwnership.ts`)
- `#479` (unilateral exit bundle)
- `#475` (repo docs)
- `v0.4.24` release

**Merge is clean.** Verified:

1. **No conflicts in `contractManager.ts`** — `refreshOutpoints()` sits cleanly alongside the new `warnAndFilterVtxosForScript` call sites from #481. The method already does its own ownership filtering at `scriptToContract.has(v.script)` before writing, which is consistent with #481's approach.

2. **Service worker handler** — `REFRESH_OUTPOINTS` case at line 893 correctly placed after `REFRESH_VTXOS`, each with `return` (no fallthrough). No merge artifacts.

3. **`vtxo-manager.ts`** — `extractSpentOutpoint` and both `maybeRefreshAfterVtxoSpent` call sites unchanged by merge. Parsing logic remains sound.

4. **No new code, no new risk.** Previous approval stands.

LGTM — still approving.

PR #478, opened 2026-05-05
fix(wallet): pre-flight VTXO outpoints before settle (proactive)

Kukks:
**Stacks on #477.** Targets that branch as base so the diff shows only this PR's content; the base flips to \`master\` automatically once #477 merges.

## Why

The local VTXO cache can fall out of sync with the server's authoritative state when:
- An SSE \`vtxo_spent\` event is missed (network blip, tab suspended, service-worker eviction).
- The cursor-derived delta sync's \`?after=created_at\` filter skips an old VTXO that's now spent (the server returns it with no relevant change to its create timestamp).

Submitting a settle intent against a known-stale candidate guarantees a 400 \`VTXO_ALREADY_SPENT\`, plus the cleanup \`deleteIntent\`, plus the reactive refresh — and on a 60-second poll cadence (default \`pollIntervalMs\`) that's a wasted round-trip every minute until the underlying state corrects itself.

#477 closes the loop AFTER the failed intent flies. This PR closes it BEFORE.

## What

\`VtxoManager\` now validates its chosen inputs against the indexer right before submitting, via the \`IContractManager.refreshOutpoints\` API #477 introduces. After the refresh, it re-selects through the standard expiring-vtxo filter so anything the refresh flagged as spent is dropped silently. Wired into both settle entry points:
- \`renewVtxos\` (event-driven renewal triggered by \`vtxo_received\`).
- \`runPeriodicSettle\` (boarding-poll auto-settle).

Best-effort: a failed refresh falls back to the original candidates and lets #477's reactive recovery handle whatever slipped through. Pre-flight is a fast path, not a hard gate.

## Result

- \`VTXO_ALREADY_SPENT\` becomes near-zero in steady state. Only triggered by a tight race between pre-flight refresh and the actual settle submission.
- The 60-second loop in the user trace disappears entirely — when the cache is wrong, pre-flight catches it BEFORE submitting, no failing intent + cleanup ever flies.
- #477's reactive code stays as defense-in-depth for the rare race.

## Tests

Two new tests in \`test/vtxo-manager.test.ts\`:
- **\"pre-flight refreshOutpoints runs before settle on the periodic poll path\"** — asserts ordering via spy interleave; the \`refreshOutpoints\` call comes before \`settle\`.
- **\"pre-flight drops candidates the indexer reports as spent and skips a fully-stale settle\"** — simulates the indexer flipping state via the second \`getVtxos\` call returning empty; asserts no \`refreshVtxos\` (recovery) was triggered because no settle was attempted.

Updated #477's two reconciliation tests to acknowledge the extra pre-flight \`refreshOutpoints\` call (so the recovery one is the *last* call, asserted via \`toHaveBeenLastCalledWith\`).

All existing mock \`contractManager\` instances picked up a \`refreshOutpoints\` stub. \`flushMicrotasks\` bumped from 2 → 8 awaits to drain the longer pre-flight chain.

## Test plan

- [x] 2 new tests pass; 2 #477 tests updated and pass
- [x] Full unit suite: 1094 tests pass
- [x] Lint clean
- [x] Typecheck clean

## Stack

1. **#477** — adds \`refreshOutpoints\` API + reactive recovery
2. **This PR** — proactive pre-flight using \`refreshOutpoints\`

Both can ship; #477 first.

## Release Notes

* **Reliability Improvements**
  * Enhanced wallet renewal process with validation checks before settlement to prevent transaction failures and conflicts.
  * Improved synchronization with server state during renewal operations to ensure consistency and reduce errors.


ghost (approved):
## Review: Pre-flight VTXO outpoint validation before settle

This is a solid, well-targeted fix that closes the stale-cache loop *before* a failing intent hits the network. The design is clean and the integration with #477's reactive recovery is well thought out.

### What works well

**Correctness of the pre-flight logic (`revalidateBeforeSettle`):**
- Best-effort design is correct — a failed `refreshOutpoints` call falls back to the original candidates and lets #477's post-submit recovery handle the remainder. No silent data loss.
- The input-set restriction (`candidateKeys` filter) is important and correct: after re-polling `getExpiringVtxos`, any *new* VTXOs that appeared since the original selection are excluded. Pre-flight narrows, never expands.
- `thresholdMs` is consistently threaded through both call sites: explicitly in `renewVtxos`, and correctly as `undefined` in `runPeriodicSettle` (where `getExpiringVtxos(undefined)` correctly defers to `settlementConfig` for threshold resolution — confirmed by reading `getExpiringVtxos` lines 693–717).

**Integration with existing flow:**
- Placed after the first "no VTXOs" guard and before the dust/fee check in `renewVtxos` — correct ordering.
- In `runPeriodicSettle`, inserted before the fee-estimation loop that builds `filteredVtxos`, so the cost accounting uses the already-filtered candidate set.
- Two `refreshOutpoints` calls in the existing `VTXO_ALREADY_SPENT` tests (pre-flight + reactive recovery) and the `toHaveBeenLastCalledWith` assertions correctly verify the reactive call is the *second* one.

**Test quality:**
- The ordering spy test (`callOrder: ["refreshOutpoints", "settle"]`) directly verifies the intent of the PR.
- The stale-drop test uses `.mockResolvedValueOnce` / `.mockResolvedValue` to simulate the wallet repo state after the refresh — clean simulation of the real flow.
- `flushMicrotasks` bump from 2→8 is well-explained.
- `ContractManager.refreshOutpoints` unit tests in `manager.test.ts` cover upsert, unknown-script skip, and empty-list no-op — good completeness.

### Minor observations (non-blocking)

1. **Double `getExpiringVtxos` per settle in the happy path**: `revalidateBeforeSettle` calls `getExpiringVtxos` internally after the outpoint refresh, in addition to the call site's initial fetch. On steady-state (no stale VTXOs), this is two wallet-repo reads per settle attempt. Not a concern at current scale, but worth knowing if this ever shows up in profiling.

2. **`extractSpentOutpoint` appears as an addition in the diff** but already exists in the `fix/vtxo-already-spent-precise-update` base branch (from #477). This is expected stacking behavior and will resolve cleanly once #477 merges.

3. **Test comment nuance**: in the "pre-flight drops candidates" test, the inline comment says "a settle fires for [boarding]" — the test setup uses `makeUnexpiredUtxo` for boarding. Since `boardingUtxoSweep: false` is set and `knownBoardingUtxos` is empty, the boarding input *will* go through `runPeriodicSettle`'s boarding path. The assertion that `refreshVtxos` was not called is correct and sufficient; the settle for boarding is orthogonal.

### Summary

The pre-flight mechanism is well-implemented: defensive, narrowing-only, with clear separation between the pre-flight path and #477's reactive recovery. The 1094-test suite passing with lint and typecheck clean is a strong signal. Approving — the stacking dependency on #477 merging first should be enforced at merge time.

ghost (commented):
## 🤖 Arkana Code Review — PR #478

**Scope**: `src/wallet/vtxo-manager.ts` (proactive pre-flight VTXO validation before settle)

### Verdict: Code looks correct. Requesting human sign-off — this is protocol-critical (VTXO handling / settle flow).

---

### Architecture

The layered defense is well-structured:
1. **Pre-flight** (this PR): `revalidateBeforeSettle` catches stale VTXOs *before* the network round-trip.
2. **Reactive** (#477): `maybeRefreshAfterVtxoSpent` handles the race where a VTXO goes stale between pre-flight and submit.

Both entry points (`renewVtxos` + `runPeriodicSettle`) are covered. Graceful degradation on pre-flight failure falls back to original candidates. 👍

---

### Findings

**✅ No blocking issues found.**

Minor observations (non-blocking):

**1. `extractSpentOutpoint` — no txid format validation** (`vtxo-manager.ts:1201-1212`)
The parser validates `vout` is a non-negative integer but doesn't validate `txid` is a 64-char hex string. A malformed `vtxo_outpoint` metadata value like `"not-hex:0"` would pass through to `refreshOutpoints`. Impact is negligible (the indexer would simply return no results), but a defensive check would be cleaner:
```ts
if (!txid || txid.length !== 64 || !/^[0-9a-f]+$/i.test(txid)) return undefined;
```

**2. Early exit in `renewVtxos` is a good optimization** (`vtxo-manager.ts:773-775`)
The added `vtxos.length === 0` check *before* `revalidateBeforeSettle` avoids a wasted `refreshOutpoints` network call when the cache is already empty. Clean.

**3. Candidate-set restriction prevents silent input expansion** (`vtxo-manager.ts:1245-1252`)
The `candidateKeys` set intersection ensures `getExpiringVtxos` can't surface new VTXOs that arrived between the initial selection and the re-select. This is correct and important — without it, a newly-received VTXO could be silently added to the settle intent without fee estimation.

**4. `thresholdMs` consistency** (`vtxo-manager.ts:1440`)
In `runPeriodicSettle`, `revalidateBeforeSettle(expiringVtxos)` is called without `thresholdMs`, so the re-select uses the same default resolution as the initial `getExpiringVtxos()` call on line 1436. Consistent. ✓

**5. Test coverage is solid.**
- Ordering test (pre-flight before settle) ✓
- Stale-candidate-drop test (empty post-refresh → no settle attempt) ✓
- Updated #477 reconciliation tests to account for pre-flight call ✓
- All mock `contractManager` instances gained `refreshOutpoints` stub ✓

---

### ⚠️ Protocol-Critical Flag

This PR modifies the VTXO selection and settle submission flow — the path where real sats are committed to Ark rounds. The code is correct and well-tested, but per protocol-critical review policy: **a human must sign off before merge.**

Both #477 and #478 should be reviewed together as a unit since they form a coherent defense-in-depth strategy for stale-cache recovery.

---

*Reviewed at commit `2ffce10`. No prior Arkana review on this PR.*

ghost (commented):
done

ghost (commented):
## 🤖 Arkana Incremental Review — commit `106633b`

**Delta since last review (`2ffce10`):** 1 commit — removes a duplicated `extractSpentOutpoint` method that was a merge artifact.

### Verified

- The canonical `extractSpentOutpoint` remains at `vtxo-manager.ts:1201` (from the #477 base branch).
- Both call sites (`vtxo-manager.ts:1117`, `vtxo-manager.ts:1541`) reference the surviving definition.
- Pure deletion of dead code — no behavioral change.

### Verdict

✅ Clean merge-conflict fix. No new concerns. My previous approval stance holds — still requesting **human sign-off** per protocol-critical policy (VTXO handling / settle flow).

pietro909 (approved):

Issue #480, opened 2026-05-05
Fix VTXO persistence and ownership validation by transitioning to script-authoritative storage


pietro909:
Currently, the SDK uses the address bucket as the primary ownership key for VTXOs. This causes issues in environments like IndexedDB where duplicate outpoints can exist across different address buckets, leading to stale data winning deduplication or incorrect balance reporting. While SQLite and Realm use (txid, vout) as a primary key, they still rely on address-centric filtering which is fragile.

We propose a 3-tier plan to transition to script as the authoritative ownership key while maintaining backward compatibility with existing databases:

- [x] Tier 1: Runtime Safety (No Schema Changes) 
  * Introduce script-ownership helpers to filter reads and validate writes at the call site.
  * Implement a dual-mode validation policy: **background syncs (indexer/pollers) will warn and skip** malformed rows to maintain service continuity, while **user-initiated transaction/signing paths will throw** on script mismatches to prevent inconsistent state persistence.
  * Update ContractManager, ContractWatcher, and Service Worker handlers to filter all repository reads by the intended contract script.
  *  PR https://github.com/arkade-os/ts-sdk/pull/481

- [x] Tier 2: Script-Aware Repository API 
    * Extend WalletRepository with script-scoped methods: `getVtxosForScript`, `saveVtxosForScript`, and `deleteVtxosForScript`.
    * Implement these methods across all backends (IndexedDB, SQLite, Realm, In-Memory), utilizing existing script indexes for efficiency.
    * Migrate all contract-aware code to use these new methods, deprecating address-scoped VTXO APIs for everything except legacy migrations and boarding.
    * PR https://github.com/arkade-os/ts-sdk/pull/482

- [ ] **WON'T DO** ~~Tier 3: Idempotent Cleanup Migration~~
    * Implement a background or version-upgrade migration to evict legacy "impossible" duplicates (rows where the address does not decode to the stored script).
    * Establish deterministic rules for collapsing remaining duplicates to ensure a single, authoritative source of truth for every VTXO on disk.

## Update June 2026

Tier 3 is no longer needed for correctness: it's now optional disk-hygiene, and only meaningful for IndexedDB.
While doable and not too complex, it introduces a DB migration which is always a risk. 
I suggest to descope it only if needed in the future and close the issue which was about correctness (now fixed).

## Related PRs
- https://github.com/arkade-os/ts-sdk/pull/477 will help reconcile ghost unspent rows by upserting the `isSpent: true` property

[...]

PR #481, opened 2026-05-05
fix(vtxo): gate persisted VTXOs by owning script (Tier 1 of #480)

pietro909:
Adds vtxoOwnership helpers and applies them at every contract-scoped read/write site so legacy address buckets cannot leak wrong-script rows or win `txid:vout` dedup. 

Background sync writers warn-and-skip; user- initiated wallet write paths throw. 
No repository or schema changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect attribution of virtual outputs to contracts; outputs are now properly validated and mapped only to their correct owning contract, even when multiple contracts share the same address.
  * Enhanced validation across wallet and contract operations to ensure data consistency and prevent legacy outputs from being persisted with wrong associations.

### DIFF
--- a/src/contracts/contractManager.ts
+++ b/src/contracts/contractManager.ts
@@ -23,6 +23,10 @@ import {
     cursorCutoff,
     getSyncCursor,
 } from "../utils/syncCursors";
+import {
+    filterVtxosForScript,
+    warnAndFilterVtxosForScript,
+} from "./vtxoOwnership";
 
 const DEFAULT_PAGE_SIZE = 500;
 
@@ -686,7 +690,10 @@ export class ContractManager implements IContractManager {
         const res = await Promise.all(
             contracts.map(({ script, address }) =>
                 this.config.walletRepository.getVtxos(address).then((vtxos) =>
-                    vtxos.map(
+                    // Address buckets may carry legacy duplicate rows from
+                    // other contracts that once shared the same address —
+                    // gate by script so the wrong-script row never wins.
+                    filterVtxosForScript(vtxos, script).map(
                         (vtxo) =>
                             ({
                                 ...vtxo,
@@ -786,9 +793,19 @@ export class ContractManager implements IContractManager {
         }
 
         for (const [addr, contractVtxos] of byContract) {
+            // The bucket is keyed by contract address, so the script filter
+            // here is the same as the contract's. Skip wrong-script rows
+            // rather than crash the reconcile loop.
+            const contract = contracts.find((c) => c.address === addr)!;
+            const filtered = warnAndFilterVtxosForScript(
+                contractVtxos,
+                contract.script,
+                "ContractManager.reconcilePendingFrontier"
+            );
+            if (filtered.length === 0) continue;
             await this.config.walletRepository.saveVtxos(
                 addr,
-                contractVtxos as ExtendedVirtualCoin[]
+                filtered as ExtendedVirtualCoin[]
             );
         }
     }
@@ -808,9 +825,15 @@ export class ContractManager implements IContractManager {
             result.set(contractScript, vtxos);
             const contract = contracts.find((c) => c.script === contractScript);
             if (contract) {
+                const filtered = warnAndFilterVtxosForScript(
+                    vtxos,
+                    contract.script,
+                    "ContractManager.fetchContractVxosFromIndexer"
+                );
+                if (filtered.length === 0) continue;
                 await this.config.walletRepository.saveVtxos(
                     contract.address,
-                    vtxos as ExtendedVirtualCoin[]
+                    filtered as ExtendedVirtualCoin[]
                 );
             }
         }
--- a/src/contracts/contractWatcher.ts
+++ b/src/contracts/contractWatcher.ts
@@ -9,6 +9,7 @@ import {
     ContractEvent,
 } from "./types";
 import { isEventSourceError } from "../providers/utils";
+import { filterVtxosForScript } from "./vtxoOwnership";
 
 /**
  * Configuration for the ContractWatcher.
@@ -281,8 +282,13 @@ export class ContractWatcher {
                 return true;
             })
             .map(async (state): Promise<[[string, ContractVtxo[]]] | []> => {
-                // Use contract address as cache key
-                const cached = await repo.getVtxos(state.contract.address);
+                // Use contract address as cache key. Legacy address buckets
+                // can contain rows from other contracts; gate by script before
+                // converting so a wrong-script row never reaches the watcher.
+                const cached = filterVtxosForScript(
+                    await repo.getVtxos(state.contract.address),
+                    state.contract.script
+                );
                 if (cached.length > 0) {
                     // Convert to ContractVtxo with contractScript
                     const contractVtxos: ContractVtxo[] = cached.map((v) => ({
--- a/src/contracts/contractWatcher.ts
+++ b/src/contracts/contractWatcher.ts
@@ -182,8 +182,14 @@ export class ContractWatcher {
      */
     private async seedLastKnownVtxos(state: ContractState): Promise<void> {
         try {
-            const cached = await this.config.walletRepository.getVtxos(
-                state.contract.address
+            // Apply the same script gate used by getContractVtxos so a legacy
+            // wrong-script row in the address bucket can't seed the baseline
+            // and then look "spent" on the first poll.
+            const cached = filterVtxosForScript(
+                await this.config.walletRepository.getVtxos(
+                    state.contract.address
+                ),
+                state.contract.script
             );
             for (const vtxo of cached) {
                 if (vtxo.isSpent) continue;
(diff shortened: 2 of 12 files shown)


ghost (changes requested):
## Review: fix(vtxo): gate persisted VTXOs by owning script (Tier 1 of #480)

**CHANGES REQUESTED** — protocol-critical concerns before merge.

---

### What this does

Introduces `vtxoOwnership.ts` with four helpers (silent filter, warn-and-skip, outpoint formatter, throw-on-mismatch) and applies them across every VTXO read/write site: background sync paths warn-and-skip; user-initiated spend/settle paths throw. The approach is architecturally correct and the test suite is solid.

Two issues need addressing before this can merge.

---

### 🔴 Critical: `getVtxosFromRepo` silently drops primary wallet VTXOs when address is not decodable

```ts
// wallet-message-handler.ts ~L1267
let walletScript: string | undefined;
try {
    walletScript = scriptFromArkAddress(walletAddress);
} catch {
    walletScript = undefined;
}
if (walletScript) {
    const walletVtxos = await this.walletRepository.getVtxos(walletAddress);
    addVtxos(filterVtxosForScript(walletVtxos, walletScript));
}
```

Before this PR: the wallet primary address VTXOs were **always** included in `GET_VTXOS`.
After this PR: if `scriptFromArkAddress` throws for any reason (unknown address format, legacy bech32 vs bech32m mismatch, test fixture, etc.), the wallet bucket is **silently skipped entirely**. The user sees 0 VTXOs for their primary wallet — they cannot spend. This is a silent money-locking bug.

At minimum add a fallback:
```ts
if (walletScript) {
    addVtxos(filterVtxosForScript(walletVtxos, walletScript));
} else {
    // Cannot verify script ownership — include raw to avoid silently hiding funds
    addVtxos(walletVtxos);
}
```
…or throw so the failure is visible rather than hiding it behind `undefined`. The current silent skip is the worst of both worlds.

---

### 🔴 Critical: `validateVtxosForScript` in spend path may throw on valid VTXOs

```ts
// wallet.ts ~L2307
validateVtxosForScript(
    toSave,
    hex.encode(this.offchainTapscript.pkScript),
    "Wallet.updateDbAfterOffchainTx"
);
```

`changeVtxo` and `spentVtxos` come from the round/send response. The guard assumes their `.script` field is always populated and always equals `hex.encode(this.offchainTapscript.pkScript)`. If either condition is false (script missing from response, or the wallet signed under a delegate script, or any mismatch from a future refactor), every offchain send will throw and the VTXO will be lost — the spend happened on-chain but the DB never records it.

Before merge: add a test that exercises `updateDbAfterOffchainTx` with a `changeVtxo` that has `.script` set and confirm it saves successfully, and also add a test with `.script` missing to confirm the failure mode is intentional. If `.script` can legitimately be absent on new VTXOs, the `warnAndFilter` variant (not `validate`) is more appropriate here.

---

### 🟡 Minor: redundant `warnAndFilterVtxosForScript` in `notifyIncomingFunds`

```ts
for (const [script, vtxos] of byScript) {
    const filtered = warnAndFilterVtxosForScript(vtxos, script, "...");
```

`byScript` is built by grouping vtxos by `v.script`, so every vtxo in `vtxos` already satisfies `.script === script`. The filter is a no-op and will never warn. Either drop the call or replace it with an assertion. Not blocking, but misleading.

---

### 🟡 Minor: non-null assertion without guard

```ts
// contractManager.ts ~L793
const contract = contracts.find((c) => c.address === addr)!;
```

The invariant is reasonable (the map is built from the same `contracts` array), but the `!` assertion will produce a cryptic error if it ever fails. A defensive check + `console.warn` / early continue would be safer.

---

### ✅ What is correct

- Strategy (warn-skip for background, throw for user-initiated) is right
- `isVtxoForScript` correctly rejects empty scripts
- `vtxoOwnership.test.ts` covers all helpers including edge cases
- `manager.test.ts` regression tests are well-constructed and prove the dedup invariant
- `contractPollProcessor` change is clean
- `contractWatcher` read-path filter is correct and minimal

---

The core architecture is sound. The two 🔴 issues are both protocol-critical: one can silently hide funds, the other can silently fail to record a completed spend. Fix those and this is ready to merge.

ghost (changes requested):
## Code Review — fix(vtxo): gate persisted VTXOs by owning script (Tier 1 of #480)

**Verdict: Request changes** (1 bug, 2 nits, + protocol-critical flag for human sign-off)

---

### 🟢 What's good

The core idea is sound. Address-keyed storage can hold legacy wrong-script rows, and gating every read/write site by script is the correct fix. Three severity tiers (`filter` / `warnAndFilter` / `validate`) are well-chosen: background sync is lenient, user-initiated paths throw. Test coverage is thorough — the regression tests for cross-bucket contamination in `manager.test.ts` and `wallet-message-handler.test.ts` directly reproduce the bug.

`vtxoOwnership.ts` is clean. `isVtxoForScript` correctly rejects empty-to-empty matches (`!!vtxo.script && ...`), which prevents blank scripts from silently matching.

---

### 🔴 Bug — silent data loss on script-less VTXOs in notifyIncomingFunds

**`src/wallet/serviceWorker/wallet-message-handler.ts:1205`**

```ts
for (const v of [...newVtxos, ...spentVtxos]) {
    if (!v.script) continue;   // ← silent drop, no warning
    ...
}
```

VTXOs with a missing/empty `script` are silently dropped with zero logging. While the `VirtualCoin` type requires `script: string`, runtime data from the indexer or legacy subscriptions can have it missing. The old code saved all VTXOs unconditionally.

If the indexer returns a VTXO with no script during a round (e.g. a protocol upgrade, a temporary indexer bug), the wallet silently loses track of it. This is an incoming-funds path — losing a VTXO here means the user can't see or spend their money.

**Fix:** At minimum, log a warning with the outpoint when dropping script-less rows, consistent with `warnAndFilterVtxosForScript` elsewhere. Even better: use `warnAndFilterVtxosForScript` itself after grouping, which already handles the logging.

---

### 🟡 Nit — `getVtxosFromRepo` skips entire wallet bucket on decode failure

**`src/wallet/serviceWorker/wallet-message-handler.ts:1498-1502`** (in the PR's new code)

```ts
try {
    walletScript = scriptFromArkAddress(walletAddress);
} catch {
    walletScript = undefined;
}
if (walletScript) {
    // ... fetch and filter
}
```

If `ArkAddress.decode` throws (corrupt address, encoding change, etc.), the wallet's primary address VTXOs are completely invisible — the user's balance drops to zero for the default contract. The old code always included them.

This is a behavior change that could strand funds in the UI. Consider falling back to the unfiltered path with a warning rather than skipping entirely:

```ts
if (walletScript) {
    addVtxos(filterVtxosForScript(walletVtxos, walletScript));
} else {
    console.warn("getVtxosFromRepo: could not decode wallet address script, falling back to unfiltered");
    addVtxos(walletVtxos);
}
```

---

### 🟡 Nit — `readonlyWallet!` non-null assertion in async callback

**`src/wallet/serviceWorker/wallet-message-handler.ts:1218`**

```ts
const cm = await this.readonlyWallet!.getContractManager();
```

This is inside an async callback (`notifyIncomingFunds`). `readonlyWallet` is checked at the top of `onWalletInitialized`, but can be nulled by `disposeWallet()` between the check and callback execution. The existing `this.walletRepository?.saveVtxos` on the next line uses optional chaining, acknowledging this race.

Not introduced by this PR (the pattern pre-dates it), but worth noting since you're refactoring this exact block. A `if (!this.readonlyWallet) return;` guard at the top of the callback would be safer.

---

### ℹ️ Observations (no action needed)

1. **Defense-in-depth in `reconcilePendingFrontier` and `fetchContractVxosFromIndexer`:** The VTXOs reaching those save sites are already filtered by script (indexer query is by script, `owned` filter at line 771). The `warnAndFilter` calls add a second gate. Redundant but appropriate for protocol-critical code.

2. **`contracts.find((c) => c.address === addr)!` in `reconcilePendingFrontier` (line 793):** Safe because `Contract.script → Contract.address` is 1:1 (address is derived from script via pkScript encoding). If that invariant ever breaks, this non-null assertion would NPE silently. Consider a guard + `continue` instead of `!` for robustness.

3. **Test helpers updated correctly:** `TEST_DEFAULT_ARK_ADDRESS` derived from real script encoding, test VTXOs now carry `script` fields matching their contract. The test setup mirrors the real data flow.

---

### 🚨 Protocol-critical flag

This PR modifies VTXO persistence at every layer: read, write, background sync, user-initiated send, and settle. The ownership-gating logic is correct, but **a human must review and approve** before merge. The surface area is large (11 files, 500 additions) and touches the critical path for fund visibility and spendability.

cc @ArkLabsHQ/core — please review the interaction between `warnAndFilterVtxosForScript` (lenient, drops) and `validateVtxosForScript` (strict, throws) to confirm the tier assignments match the intended failure modes for each call site.

PR #482, opened 2026-05-06
fix(vtxo): script-scoped repository methods across backends (Tier 2 of #480)

pietro909:
## Summary

- Adds optional `getVtxosForScript`, `saveVtxosForScript`, and `deleteVtxosForScript` methods to the `WalletRepository` interface (no breaking change — existing custom implementations continue to work via Tier 1 fallbacks).
- Implements the three methods in all SDK-provided backends: in-memory, IndexedDB, SQLite, and Realm.
- Adds `getVtxosForContract` / `saveVtxosForContract` dispatch helpers in `vtxoOwnership.ts`; migrates all contract-aware call sites in `ContractManager`, `ContractWatcher`, `contractPollProcessor`, `wallet-message-handler`, and `Wallet` to use them.
- IndexedDB dedup tiebreaker (`shouldReplaceVtxo`) guards `scriptFromArkAddress` decode failures so a row with an undecodable address is treated as non-canonical rather than crashing the entire script-scoped read.
- No schema changes, no migration required, no version bump.

Part of #480.

## Test plan

- [x] `pnpm lint && pnpm test:unit` passes
- [x] Conformance tests in `test/repositories/walletRepository.test.ts` cover all three new methods for InMemory and IndexedDB, including the duplicate-outpoint dedup case
- [x] SQLite script-scoped tests in `test/sqlite-wallet-repository.test.ts` pass

## Summary by CodeRabbit

* **Refactor**
  * Reorganized how virtual transaction outputs are managed internally to support script-based organization alongside address-based storage.

* **Tests**
  * Added comprehensive test coverage for script-scoped virtual output operations across repository implementations.

* **Chores**
  * Updated `.gitignore` to exclude debug artifacts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,10 @@ docs/
 .idea
 .claude/
 *.sqlite
+
+# Debug 
 *.har
+*.idb
 
 # AI
 TASKS.md
--- a/src/contracts/contractManager.ts
+++ b/src/contracts/contractManager.ts
@@ -25,6 +25,8 @@ import {
 } from "../utils/syncCursors";
 import {
     filterVtxosForScript,
+    getVtxosForContract,
+    saveVtxosForContract,
     warnAndFilterVtxosForScript,
 } from "./vtxoOwnership";
 
@@ -680,13 +682,27 @@ export class ContractManager implements IContractManager {
         const annotated = await this.annotateVtxos(owned);
         const byAddress = new Map<string, ExtendedVirtualCoin[]>();
         for (const vtxo of annotated) {
-            const address = scriptToContract.get(vtxo.script)!.address;
+            const contract = scriptToContract.get(vtxo.script);
+            if (!contract) continue;
+            const address = contract.address;
             const arr = byAddress.get(address) ?? [];
             arr.push(vtxo);
             byAddress.set(address, arr);
         }
         for (const [address, addressVtxos] of byAddress) {
-            await this.config.walletRepository.saveVtxos(address, addressVtxos);
+            const contract = contracts.find((c) => c.address === address);
+            if (contract) {
+                await saveVtxosForContract(
+                    this.config.walletRepository,
+                    contract,
+                    addressVtxos
+                );
+            } else {
+                await this.config.walletRepository.saveVtxos(
+                    address,
+                    addressVtxos
+                );
+            }
         }
     }
 
@@ -736,16 +752,16 @@ export class ContractManager implements IContractManager {
         contracts: Contract[]
     ): Promise<ContractVtxo[]> {
         const res = await Promise.all(
-            contracts.map(({ script, address }) =>
-                this.config.walletRepository.getVtxos(address).then((vtxos) =>
-                    // Address buckets may carry legacy duplicate rows from
-                    // other contracts that once shared the same address —
-                    // gate by script so the wrong-script row never wins.
-                    filterVtxosForScript(vtxos, script).map(
+            contracts.map((contract) =>
+                getVtxosForContract(
+                    this.config.walletRepository,
+                    contract
+                ).then((vtxos) =>
+                    vtxos.map(
                         (vtxo) =>
                             ({
                                 ...vtxo,
-                                contractScript: script,
+                                contractScript: contract.script,
                             }) as ContractVtxo
                     )
                 )
@@ -851,8 +867,9 @@ export class ContractManager implements IContractManager {
                 "ContractManager.reconcilePendingFrontier"
             );
             if (filtered.length === 0) continue;
-            await this.config.walletRepository.saveVtxos(
-                addr,
+            await saveVtxosForContract(
+                this.config.walletRepository,
+                contract,
                 filtered as ExtendedVirtualCoin[]
             );
         }
@@ -879,8 +896,9 @@ export class ContractManager implements IContractManager {
                     "ContractManager.fetchContractVxosFromIndexer"
                 );
                 if (filtered.length === 0) continue;
-                await this.config.walletRepository.saveVtxos(
-                    contract.address,
+                await saveVtxosForContract(
+                    this.config.walletRepository,
+                    contract,
                     filtered as ExtendedVirtualCoin[]
                 );
             }
--- a/src/contracts/contractWatcher.ts
+++ b/src/contracts/contractWatcher.ts
@@ -9,7 +9,7 @@ import {
     ContractEvent,
 } from "./types";
 import { isEventSourceError } from "../providers/utils";
-import { filterVtxosForScript } from "./vtxoOwnership";
+import { filterVtxosForScript, getVtxosForContract } from "./vtxoOwnership";
 
 /**
  * Configuration for the ContractWatcher.
@@ -185,11 +185,9 @@ export class ContractWatcher {
             // Apply the same script gate used by getContractVtxos so a legacy
             // wrong-script row in the address bucket can't seed the baseline
             // and then look "spent" on the first poll.
-            const cached = filterVtxosForScript(
-                await this.config.walletRepository.getVtxos(
-                    state.contract.address
-                ),
-                state.contract.script
+            const cached = await getVtxosForContract(
+                this.config.walletRepository,
+                state.contract
             );
             for (const vtxo of cached) {
                 if (vtxo.isSpent) continue;
@@ -291,10 +289,7 @@ export class ContractWatcher {
                 // Use contract address as cache key. Legacy address buckets
                 // can contain rows from other contracts; gate by script before
                 // converting so a wrong-script row never reaches the watcher.
-                const cached = filterVtxosForScript(
-                    await repo.getVtxos(state.contract.address),
-                    state.contract.script
-                );
+                const cached = await getVtxosForContract(repo, state.contract);
                 if (cached.length > 0) {
                     // Convert to ContractVtxo with contractScript
                     const contractVtxos: ContractVtxo[] = cached.map((v) => ({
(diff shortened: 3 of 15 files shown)


ghost (changes requested):
## Review: fix(vtxo): script-scoped repository methods across backends (Tier 2 of #480)

Overall architecture is sound. The tiered fallback design (optional methods + Tier 1 address-based filtering) is backwards-compatible and the dedup logic in IndexedDB is well-considered. A few issues to address:

---

### 🔴 Bug: Unused import (will fail lint)

**`src/repositories/indexedDB/walletRepository.ts:27`** — `vtxoOutpoint` is imported but never used in this file. The outpoint key is constructed inline as `` `${row.txid}:${row.vout}` `` instead. Either use the helper or remove the import.

```diff
-import { isVtxoForScript, vtxoOutpoint } from "../../contracts/vtxoOwnership";
+import { isVtxoForScript } from "../../contracts/vtxoOwnership";
```

---

### 🟡 Silent error swallowing in IndexedDB `getVtxosForScript`

**`src/repositories/indexedDB/walletRepository.ts:~240`** — The outer `catch` returns `[]` on DB failure. This is inconsistent with `deleteVtxosForScript` (which re-throws) and the existing `getVtxos` method. A silent empty result on a real IndexedDB failure could cause the wallet to think a contract has no VTXOs and trigger incorrect round behavior.

Suggest: either re-throw (consistent with delete) or at minimum propagate the error pattern used by the existing `getVtxos`.

---

### 🟡 `saveVtxosForContract` Tier 1 fallback lacks script validation

**`src/contracts/vtxoOwnership.ts:~95`** — When `repo.saveVtxosForScript` is absent, the fallback calls `repo.saveVtxos(contract.address, vtxos)` without any script-mismatch guard. I verified that *current* callers pre-validate, but this creates a subtle contract: future callers must remember to validate before calling the "safe" dispatch helper.

Consider adding a `filterVtxosForScript` or `validateVtxosForScript` guard in the fallback path so the helper is safe to call without preconditions. This is defense-in-depth for a protocol-critical path.

---

### 🟢 Minor: non-null assertion on `changeVtxo.script!`

**`src/wallet/wallet.ts:~2700`** — Safe in practice (script is always set at construction ~line 2650), but a brief comment documenting the invariant would prevent future confusion:

```ts
// script is always set at construction (hex.encode of pkScript)
{ script: changeVtxo.script!, address: primaryAddr },
```

---

### 🟢 Test coverage

Good coverage of the new methods across InMemory, IndexedDB, and SQLite. The dedup test for IndexedDB is appropriately backend-specific. The wallet.test.ts changes correctly reflect the new per-script save semantics.

---

### ⚠️ Protocol-critical flag

This PR modifies VTXO persistence paths (save, get, delete) which are protocol-critical infrastructure. **Requesting human review** per policy — even though the code logic is correct, changes to VTXO storage routing need a second pair of human eyes before merge.

---

**Verdict:** Two concrete fixes needed (unused import, inconsistent error handling), one defense-in-depth suggestion. After those, LGTM from a code perspective — but human sign-off required for protocol-critical merge.

ghost (changes requested):
## Review: fix(vtxo): script-scoped repository methods across backends (Tier 2 of #480)

Overall architecture is sound. The tiered fallback design (optional methods + Tier 1 address-based filtering) is backwards-compatible and the dedup logic in IndexedDB is well-considered. A few issues to address:

---

### 🔴 Bug: Unused import (will fail lint)

**`src/repositories/indexedDB/walletRepository.ts:27`** — `vtxoOutpoint` is imported but never used in this file. The outpoint key is constructed inline as `` `${row.txid}:${row.vout}` `` instead. Either use the helper or remove the import.

```diff
-import { isVtxoForScript, vtxoOutpoint } from "../../contracts/vtxoOwnership";
+import { isVtxoForScript } from "../../contracts/vtxoOwnership";
```

---

### 🟡 Silent error swallowing in IndexedDB `getVtxosForScript`

**`src/repositories/indexedDB/walletRepository.ts:~240`** — The outer `catch` returns `[]` on DB failure. This is inconsistent with `deleteVtxosForScript` (which re-throws) and the existing `getVtxos` method. A silent empty result on a real IndexedDB failure could cause the wallet to think a contract has no VTXOs and trigger incorrect round behavior.

Suggest: either re-throw (consistent with delete) or at minimum propagate the error pattern used by the existing `getVtxos`.

---

### 🟡 `saveVtxosForContract` Tier 1 fallback lacks script validation

**`src/contracts/vtxoOwnership.ts:~95`** — When `repo.saveVtxosForScript` is absent, the fallback calls `repo.saveVtxos(contract.address, vtxos)` without any script-mismatch guard. I verified that *current* callers pre-validate, but this creates a subtle contract: future callers must remember to validate before calling the "safe" dispatch helper.

Consider adding a `filterVtxosForScript` or `validateVtxosForScript` guard in the fallback path so the helper is safe to call without preconditions. This is defense-in-depth for a protocol-critical path.

---

### 🟢 Minor: non-null assertion on `changeVtxo.script!`

**`src/wallet/wallet.ts:~2700`** — Safe in practice (script is always set at construction ~line 2650), but a brief comment documenting the invariant would prevent future confusion:

```ts
// script is always set at construction (hex.encode of pkScript)
{ script: changeVtxo.script!, address: primaryAddr },
```

---

### 🟢 Test coverage

Good coverage of the new methods across InMemory, IndexedDB, and SQLite. The dedup test for IndexedDB is appropriately backend-specific. The wallet.test.ts changes correctly reflect the new per-script save semantics.

---

### ⚠️ Protocol-critical flag

This PR modifies VTXO persistence paths (save, get, delete) which are protocol-critical infrastructure. **Requesting human review** per policy — even though the code logic is correct, changes to VTXO storage routing need a second pair of human eyes before merge.

---

**Verdict:** Two concrete fixes needed (unused import, inconsistent error handling), one defense-in-depth suggestion. After those, LGTM from a code perspective — but human sign-off required for protocol-critical merge.

ghost (changes requested):
## Re-review after `a479564f` (remove unused import)

✅ **Fixed:** Unused `vtxoOutpoint` import removed — thanks.

❌ **Still open from previous review:**

### 🟡 Silent error swallowing in IndexedDB `getVtxosForScript`

**`src/repositories/indexedDB/walletRepository.ts:238-241`** — `getVtxosForScript` still catches DB errors and returns `[]`. `deleteVtxosForScript` (line 281-284) re-throws. This inconsistency is dangerous: callers cannot distinguish "no VTXOs" from "DB failure", which can cause the wallet to act on stale/missing state during a round.

Either re-throw (consistent with `deleteVtxosForScript`) or propagate the error pattern from the existing `getVtxos` method.

### 🟡 `saveVtxosForContract` Tier 1 fallback still lacks script validation

**`src/contracts/vtxoOwnership.ts:104`** — Fallback path calls `repo.saveVtxos(contract.address, vtxos)` without filtering. The read side (`getVtxosForContract`, line 88-90) applies `filterVtxosForScript`, but the write side doesn't. This asymmetry means wrong-script rows can accumulate in address-keyed storage. A `filterVtxosForScript` guard before the fallback `saveVtxos` call would close this gap.

---

⚠️ **Protocol-critical flag** remains: VTXO persistence changes require human sign-off before merge.

ghost (changes requested):
## Re-review after `3aae0155` (re-throw DB errors in getVtxosForScript)

✅ **Fixed:** `src/repositories/indexedDB/walletRepository.ts:238` — `getVtxosForScript` now re-throws on DB failure instead of silently returning `[]`. Good.

❌ **Still open:**

### 🟡 `saveVtxosForContract` Tier 1 fallback lacks script validation

**`src/contracts/vtxoOwnership.ts:101`** — The fallback path still calls `repo.saveVtxos(contract.address, vtxos)` without a script-mismatch guard. Every Tier 2 `saveVtxosForScript` implementation validates script before writing. But when a custom backend lacks `saveVtxosForScript`, the fallback silently accepts wrong-script VTXOs into the address bucket.

One line fixes it:

```diff
-        : repo.saveVtxos(contract.address, vtxos);
+        : repo.saveVtxos(contract.address, filterVtxosForScript(vtxos, contract.script));
```

Or use `validateVtxosForScript` to throw instead of silently dropping (matches the Tier 2 behavior):

```diff
-        : repo.saveVtxos(contract.address, vtxos);
+        : (validateVtxosForScript(vtxos, contract.script, "saveVtxosForContract"), repo.saveVtxos(contract.address, vtxos));
```

This is defense-in-depth for a protocol-critical write path. Current callers pre-validate, but the dispatch helper should be safe to call without preconditions.

---

⚠️ **Protocol-critical flag** remains: VTXO persistence changes require human sign-off before merge.

ghost (approved):
## Re-review after `07d24de5` (validate scripts in saveVtxosForContract fallback)

All three previous findings are now resolved:

1. ✅ Unused `vtxoOutpoint` import removed (`a479564f`)
2. ✅ `getVtxosForScript` re-throws DB errors instead of returning `[]` (`3aae0155`)
3. ✅ `saveVtxosForContract` Tier 1 fallback now calls `validateVtxosForScript` before writing (`07d24de5`) — defense-in-depth on the protocol-critical write path, matching Tier 2 behavior.

Code is clean. The tiered dispatch design is backwards-compatible, dedup logic is correct, and test coverage is solid across all four backends.

**Approving from my side**, but ⚠️ **protocol-critical flag remains**: this PR modifies VTXO persistence (read/write/delete paths). **Human sign-off required before merge.**

pietro909:
| With this PR | Current master |
|--------------------|-------------------------|
| PR |master |

PR #485, opened 2026-05-08
fix(types): emit ESM-compatible declaration imports

gringokiwi:
## Summary
Fixes the generated TypeScript declaration files so they work correctly in ESM/NodeNext consumers.

## What Changed
- Moved `scripts/add-extensions.js` to run after declaration files are emitted.
- Extended the extension patcher to process `dist/types/**/*.d.ts` in addition to compiled ESM JavaScript.
- Ensured emitted declaration imports use explicit `.js` specifiers, including directory imports as `index.js`.
- Updated remaining `@scure/btc-signer/*` subpath imports that are emitted into declarations to use the package’s exported `.js` subpaths.

## Why
The package is published as `"type": "module"`, but the generated `.d.ts` files contained extensionless relative imports such as:

```ts
import { Wallet } from "./wallet";
```

In TypeScript projects using `moduleResolution: "node16"` or `"nodenext"`, declaration files in ESM packages must use explicit file extensions. Without them, consumers can hit errors like:
`TS2834: Relative import paths need explicit file extensions in ECMAScript imports`

This also affects IDE/LSP resolution and linting for strict ESM consumers.

## Verification
- `pnpm run build`
- `pnpm lint`
- Verified emitted declarations pass a `NodeNext` TypeScript check with `skipLibCheck: false`.

## Summary by CodeRabbit

* **Chores**
  * Updated build pipeline to include ESM module export step for improved distribution compatibility.
  * Refactored build script to support multiple output directories with consistent import resolution.
  * Standardized internal import paths across codebase for better compatibility with package distribution formats.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
         "registry": "https://registry.npmjs.org/"
     },
     "scripts": {
-        "build": "rimraf dist && pnpm run build:esm && pnpm run build:expo:esm && node scripts/add-extensions.js && pnpm run build:cjs && pnpm run build:expo:cjs && pnpm run build:types && pnpm run build:expo:types && node scripts/generate-package-files.js",
+        "build": "rimraf dist && pnpm run build:esm && pnpm run build:expo:esm && pnpm run build:cjs && pnpm run build:expo:cjs && pnpm run build:types && pnpm run build:expo:types && node scripts/add-extensions.js && node scripts/generate-package-files.js",
         "build:esm": "tsc -p tsconfig.esm.json --outDir dist/esm",
         "build:cjs": "tsc -p tsconfig.cjs.json --outDir dist/cjs",
         "build:types": "tsc -p tsconfig.json --outDir dist/types --emitDeclarationOnly",
--- a/scripts/add-extensions.js
+++ b/scripts/add-extensions.js
@@ -1,12 +1,12 @@
 /**
  * ESM Import Path Fixer
  * 
- * This script adds .js extensions to all ESM imports in the compiled JavaScript files.
+ * This script adds .js extensions to relative imports in compiled output files.
  * It's necessary for Node.js ESM compatibility since Node.js requires explicit file extensions
  * in import statements when using ES modules.
  * 
  * The script:
- * 1. Finds all JS files in the ESM output directory
+ * 1. Finds files in the ESM and types output directories
  * 2. For each file, it processes all relative imports
  * 3. Adds .js extensions where needed, handling various edge cases
  */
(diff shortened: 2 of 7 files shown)


ghost (approved):
## ✅ Approved — clean ESM declaration fix

Reviewed all 7 changed files + full file context + git history + downstream consumers.

### What this does
1. **Source files** (`src/extension/index.ts`, `src/script/delegate.ts`, `src/utils/transaction.ts`, `src/wallet/delegator.ts`, `src/wallet/utils.ts`): Adds `.js` to `@scure/btc-signer` subpath imports. Verified against `@scure/btc-signer/package.json` — its `exports` field explicitly maps `./utils.js`, `./psbt.js`, `./transaction.js`, so these are correct.

2. **`scripts/add-extensions.js`**: Generalized to also patch `dist/types/**/*.d.ts` files. Key details:
   - New `path.extname()` guard is strictly better than the old `.endsWith('.js')` — handles `.json`, `.css`, etc.
   - For `.d.ts` files, looks up `${importPath}.d.ts` on disk but rewrites to `.js` — correct, declarations must reference runtime extensions.
   - New regex `/(from\s+['"])(\.[^'"]*)(['"])/g` preserves original quote style (old code always forced single quotes).
   - Added `import()` dynamic import regex — good coverage, though current source dynamic imports are non-relative so unaffected.

3. **`package.json`**: Build order moved `add-extensions.js` to run after all builds complete. Required because the script now patches types output too.

### Things I checked
- **Protocol-critical files touched**: `delegator.ts` (forfeit paths), `delegate.ts` (VTXO scripts), `transaction.ts` (tx signing wrapper). All changes are **import-path-only** — zero logic changes. No risk.
- **Cross-repo impact**: 8 downstream consumers found (`banco`, `arktillery`, demos, etc.). This is a **non-breaking improvement** — consumers using `moduleResolution: "node16"/"nodenext"` with `skipLibCheck: false` would have been broken before this fix.
- **CJS unaffected**: Script only targets `dist/esm/**/*.js` and `dist/types/**/*.d.ts`. CJS uses `require()` and doesn't need extensions.
- **Expo builds**: Output goes into same `dist/esm` and `dist/types` dirs, so expo output is also patched. ✓
- **CI**: build ✓, lint ✓, unit ✓

### Minor nit (non-blocking)
`fixedImports` counter now counts files-with-changes rather than individual import rewrites. The log message was updated accordingly (`"imports in N files"`), so this is fine — just noting the semantic change.

No protocol-critical concerns. LGTM.

pietro909 (approved):

PR #491, opened 2026-05-14
fix(wallet): per-input signing after HD rotation

pietro909:
Follow-up on top of #489.

## Summary
- Route per-input signing through `DescriptorProvider` so rotated VTXOs (locked by index-N pubkeys) get signed by the right key instead of falling back to identity index-0.
- Persist `metadata.signingDescriptor` on rotated contracts and resolve each input's script to its contract at sign time. arkTx signing accepts a per-input source-script override because checkpoint `witnessUtxo` scripts don't match the source VTXO's contract.
- Forward `walletMode` to the service worker wallet so worker-mode wallets pick up the same HD signing path.

## Test plan
- [x] `pnpm lint`
- [x] `pnpm test:unit`
- [ ] `pnpm test:integration` (HD rotation + send/renew)

## Manual testing

1. mutinynet
2. brand new wallet with mnemonic phrase
3. first 2 transactions received through Arkade 
4. third ($14.94) via Lightning
5. sent all the coins back (Arkade) 
6. received from onchain (onboarding address)
7. sent a fraction via arkade

Contracts rotate, coins are spent with the correct scripts.

image

### DIFF
--- a/src/wallet/wallet.ts
+++ b/src/wallet/wallet.ts
@@ -28,7 +28,7 @@ import {
     validateVtxoTxGraph,
 } from "../tree/validation";
 import { validateBatchRecipients } from "./validation";
-import { Identity, isBatchSignable, ReadonlyIdentity } from "../identity";
+import { Identity, ReadonlyIdentity } from "../identity";
 import {
     ArkTransaction,
     Asset,
@@ -63,7 +63,6 @@ import { VtxoScript } from "../script/base";
 import { CSVMultisigTapscript, RelativeTimelock } from "../script/tapscript";
 import {
     buildOffchainTx,
-    combineTapscriptSigs,
     hasBoardingTxExpired,
     isValidArkAddress,
 } from "../utils/arkTransaction";
@@ -98,6 +97,7 @@ import {
     IndexedDBWalletRepository,
 } from "../repositories";
 import { ContractManager } from "../contracts/contractManager";
+import { Contract } from "../contracts/types";
 import { contractHandlers } from "../contracts/handlers";
 import { timelockToSequence } from "../utils/timelock";
 import { clearSyncCursor, updateWalletState } from "../utils/syncCursors";
@@ -106,6 +106,7 @@ import {
     saveVtxosForContract,
 } from "../contracts/vtxoOwnership";
 import { WalletReceiveRotator } from "./walletReceiveRotator";
+import { DescriptorProvider } from "../identity/descriptorProvider";
 
 export const getArkadeServerUrl = ({
     arkServerUrl,
@@ -177,6 +178,38 @@ function hasToReadonly(identity: unknown): identity is HasToReadonly {
         "toReadonly" in identity &&
         typeof (identity as any).toReadonly === "function"
     );
+}
+
+/**
+ * Thrown when {@link Wallet.signInputsByOwner} encounters a default or
+ * delegate contract whose owning pubkey differs from the identity's
+ * baseline (so it must be signed via {@link DescriptorProvider}) but
+ * the contract record has no `metadata.signingDescriptor`. This happens
+ * on wallets that rotated under the original HD branch (which only
+ * persisted `params.pubKey`) — the fix-forward is to manually write a
+ * `signingDescriptor` onto each affected contract record, or to restore
+ * from a pre-rotation snapshot.
+ *
+ * Surfacing this loudly is intentional: silently falling back to the
+ * identity's index-0 key would reproduce the original "missing tapscript
+ * spend sig" server error — a confusingly-late failure with no obvious
+ * root cause.
+ */
+export class MissingSigningDescriptorError extends Error {
+    readonly name = "MissingSigningDescriptorError";
+
+    constructor(
+        readonly contractScript: string,
+        readonly contractType: "default" | "delegate"
+    ) {
+        super(
+            `Cannot sign input for ${contractType} contract ${contractScript}: ` +
+                `metadata.signingDescriptor is missing. This wallet was rotated ` +
+                `on an earlier build that did not persist signing descriptors. ` +
+                `Manually set metadata.signingDescriptor on the contract record, ` +
+                `or restore from a pre-rotation snapshot.`
+        );
+    }
 }
 
 export class ReadonlyWallet implements IReadonlyWallet {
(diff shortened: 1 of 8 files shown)


ghost (changes requested):
## Code Review — `signInputsByOwner` migration (PR #491)

**Reviewer**: Arkana (automated code review agent)
**Verdict**: Request changes — protocol-critical PR, requires human sign-off regardless

---

### Summary

This PR replaces all `identity.sign()` / `identity.signMultiple()` call sites in `Wallet` with a new `signInputsByOwner()` private method that routes each PSBT input to either the identity's baseline key or a `DescriptorProvider` based on the owning contract's pubkey. This fixes the core HD-rotation-after-send bug (`INVALID_PSBT_INPUT (5): missing tapscript spend sig`). The approach is sound and well-documented.

### ✅ What looks correct

1. **Signing migration is complete.** Every `identity.sign()` call site for VTXO owner signing is migrated: settlement boarding, forfeits, all 3 intent proofs, checkpoint signing in `finalizePendingTxs`, and `buildAndSubmitOffchainTx`. The musig2 `identity.signerSession()` path correctly remains untouched (that's collaborative signing, not owner signing).

2. **arkTx input positional mapping is safe.** Verified the chain: `inputs[]` → `buildOffchainTx` → `buildCheckpointTx.map()` → `buildVirtualTx` sequential `addInput()`. No reordering anywhere. The `arkTxOwnerScripts` override map is correct.

3. **Sequential threading preserves signatures.** `identity.sign()` → returns new tx with identity sigs → passed to `signWithDescriptor()` → `signTxWithKey` clones (preserving prior sigs) → adds descriptor sigs. Accumulation is correct.

4. **MissingSigningDescriptorError is the right call.** Silently falling back to index-0 would reproduce the original bug with an opaque server error. Failing loudly with a typed error is correct.

5. **Test coverage is thorough.** ~550 new test lines covering: rotated intent proofs (all 3 types), mixed baseline+rotated, missing descriptor error, cosigner/connector skip, `buildAndSubmitOffchainTx` with rotation, baseline regression, provider opt-in.

6. **`signingDescriptor` persistence in contract metadata** (`walletReceiveRotator.ts:537-544`) — correct location, correct timing.

### ⚠️ Issues to address

#### P0 — Protocol correctness

**1. Settlement forfeit signing — contract lookup may fail for boarding VTXOs**
`wallet.ts` — `signInputsByOwner` called with `[i]` / `[0]` from the settlement handler. For boarding UTXOs, the code falls through to the `boardingScriptHex` comparison (line ~345 in signInputsByOwner). However, if a rotated wallet receives boarding UTXOs (which use the identity's base pubkey in the boarding tapscript), this still routes correctly through identity. **Confirm**: is it possible to have a boarding UTXO locked by a rotated pubkey? If yes, the boarding-script check only matches the *baseline* boarding script and would silently skip a rotated-boarding input. If no (boarding always uses baseline key), this is fine.

**2. Batch signing regression is a real UX hit for hardware/external signers**
The removal of `isBatchSignable` / `signMultiple` means external signers (Ledger, mobile wallet bridges) now get N+1 confirmation popups per send instead of 1. The comment references a future `BatchSignableDescriptorProvider` but there's no tracking issue. This should be:
- Documented in the PR description / CHANGELOG
- Tracked in a GitHub issue (not just a markdown file reference)
- Ideally: for the identity-sign group (baseline inputs), `signMultiple` could still be used when `isBatchSignable` is true, only falling back to sequential for descriptor groups

#### P1 — Type safety / robustness

**3. `Contract.metadata?.signingDescriptor` has no compile-time safety**
`wallet.ts:~371` — `metadata` is `Record<string, unknown>`. A typo like `signinDescriptor` anywhere in the codebase would compile clean and silently produce `undefined`. Consider:
- Adding a `ContractMetadata` interface with optional `signingDescriptor?: string` and `source?: string`
- Or at minimum, extracting the key name to a shared constant

**4. `signWithDescriptor` returns a clone — the sequential loop relies on this implicitly**
`wallet.ts:~405-413` — the contract between `signInputsByOwner` and `signWithDescriptor` is: "you clone my tx, sign the clone, return it." If any future `DescriptorProvider` implementation signs in-place and returns the same reference, signatures from earlier groups could be lost. The method comment should document this contract explicitly, or the threading should defensively clone before passing to `signWithDescriptor`.

#### P2 — Missing test coverage

**5. No test for settlement/forfeit signing after rotation.**
The settlement boarding path (`handleSettlementFinalizationEvent`) and forfeit signing path are both migrated to `signInputsByOwner` but have no rotation-specific tests. These are protocol-critical paths — if the contract lookup fails for a settlement input, the user can't settle and their funds are locked until CSV expiry. These are admittedly harder to test in unit tests, but at minimum a note/tracking issue for integration test coverage would be good.

### 🔍 Observations (non-blocking)

- `isBatchSignable` and `combineTapscriptSigs` are removed from wallet.ts imports but remain used elsewhere (`staticDescriptorProvider.ts`, `boltz-swap`). No cross-repo breakage.
- `ServiceWorkerWalletMode` type and `walletMode` passthrough are clean. No consumers yet outside the SDK itself.
- The `ReceiveRotatorBootResult.provider` field addition is backwards-compatible (additive).
- The `lookupScriptOverrides` parameter on `signInputsByOwner` is well-designed — clean separation between "what script owns this input" and "what witnessUtxo is on the PSBT".

### 🚨 Protocol-critical flag

This PR touches **transaction signing, forfeit paths, and settlement flows**. Even with the issues above resolved, this requires explicit human review and sign-off before merge per project policy.

---
*🤖 Review by Arkana — automated code review agent*

Kukks (approved):

src/wallet/wallet.ts
```diff
@@ -2119,6 +2165,191 @@ export class Wallet extends ReadonlyWallet implements IWallet {
...
+     *   script used to look up the owning contract. Required for the
+     *   offchain `arkTx`, whose `witnessUtxo.script` is the *checkpoint*
+     *   pkScript (server-unroll + collaborative-closure combo) and not
+     *   the source VTXO's contract script — without an override the
+     *   helper would silently skip every arkTx input. Callers that sign
+     *   transactions whose inputs do match a contract directly (intent
+     *   proofs, checkpoints, forfeits, settlement boarding) can omit
+     *   this and the helper falls back to `witnessUtxo.script`.
+     */
+    private async signInputsByOwner(
```
Kukks, line 2212: This is super messy

PR #494, opened 2026-05-18
feat(repositories): NArk persistence parity — IntentRepository + VirtualTxRepository (storage layer)

Kukks:
## Summary

NArk (dotnet-sdk) persistence parity, storage layer: two new sibling repository
interfaces — **`IntentRepository`** and **`VirtualTxRepository`** — implemented
across all four backends, a pure **`IntentStateReducer`**, opt-in lifecycle
wiring, and crash-recovery intent reconciliation. Everything is inert unless
configured; activation is a separate PR.

## What's included

**Storage layer**
- Domain types: `ArkIntent`/`ArkIntentState`, `VirtualTx`/`ChainedTxType`
  (pinned from NArk), `VtxoBranch`, `IntentFilter`.
- Both interfaces implemented for **in-memory / SQLite / IndexedDB / Realm**,
  with shared cross-backend conformance suites.
- SQLite writes serialized per shared connection (`BEGIN IMMEDIATE` on one
  write chain). IndexedDB `DB_VERSION` 3→5 (v5 dedupes, then enforces a unique
  `intentId` index). Realm `ARK_REALM_SCHEMA_VERSION` 2→4 (`hex`→`psbt`
  backfill).
- `IntentStateReducer` — pure, monotonic; terminal states sticky.

**Lifecycle wiring (opt-in via `StorageConfig`)**
- `intentRepository`: `settle()` persists the intent through its lifecycle
  (best-effort, never throws into the money path); `getBalance()` excludes
  intent-locked VTXOs from `available`; `ContractManager` reconciles stale
  non-terminal intents against indexer state on boot/reconnect.
- `virtualTxRepository` — **experimental/inert**: only a best-effort raw-PSBT
  cache read/written by `Unroll.Session` when passed explicitly. Sync never
  populates or prunes it.
- Absent config ⇒ every new path is a no-op.

## Upgrade notes

- The IndexedDB and Realm schema bumps run on first open after updating, even
  without opting in. Downgrading the SDK afterwards fails to open the existing
  store (IndexedDB `VersionError`; Realm downgrade error).
- Realm consumers must compose schemas per the documented pattern:
  `schemaVersion: Math.max(ARK_REALM_SCHEMA_VERSION, yourSchemaVersion)` +
  `runArkRealmMigrations`. A pinned literal `schemaVersion` will fail at open.
- `settle()`: a synchronously-throwing `eventCallback` no longer aborts the
  settlement (matches the documented ignore-errors contract), and a post-commit
  local failure no longer deletes the server intent (the settlement stands).

## Test plan

- `pnpm test:unit` green on every commit; `tsc --noEmit` 0 errors; prettier
  clean; CI green.
- Per-backend conformance wirings, reducer units, migration tests (incl. v4
  duplicate `intentId` and Realm `hex`→`psbt`), locked-balance, settle
  intent-wiring, `Unroll` repo-first resolution, and inertness pins.

## Follow-up (deliberately not in this PR)

- Activation: virtual-tx branch persistence, sync pre-population,
  prune-on-spend.
- e2e on the regtest stack.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in intent and virtual-transaction repositories across supported storage backends, including caching for faster unroll.
  * Wallet now accounts for intent-locked VTXOs in “available” balance and persists intent lifecycle snapshots.
  * Added crash-recovery reconciliation for persisted intents on startup and reconnect.
* **Bug Fixes**
  * Strengthened terminal-state handling, best-effort persistence/recovery, and concurrent SQLite transaction/migration safety.
  * Improved IndexedDB schema reliability with uniqueness enforcement for intent identifiers.
* **Tests**
  * Added conformance/regression coverage for repository behavior, reconciliation, persistence handling, and locking logic.


ghost (changes requested):
## Arkana Code Review — ts-sdk#494

**Scope:** Storage layer + reducer for NArk persistence parity. 2088 additions across 4 backends × 2 new repo interfaces, plus `IntentStateReducer`. No existing code paths modified (besides extracting mock SQL executor to shared helper and fixing the IndexedDB schema version literal).

**Overall assessment:** Clean, well-structured PR. The conformance suite approach is excellent — same behavioural contract enforced across all four backends. Types match NArk `ChainedTxType` values exactly. The `matches()` single-source-of-truth pattern avoids cross-backend filter divergence. However, I found one real bug in the Realm backend and a few issues worth addressing before merge.

---

### 🔴 BUG — Realm `pruneForSpentVtxo` reads deleted objects (will throw in production)

**`src/repositories/realm/virtualTxRepository.ts:97-124`** (diff lines 897-924)

```ts
async pruneForSpentVtxo(vtxo: Outpoint): Promise<void> {
    this.realm.write(() => {
        const removed = [
            ...this.realm
                .objects("ArkVtxoBranch")
                .filtered("vtxoKey == $0", vtxoKey(vtxo)),
        ] as any[];                          // ← live Realm proxies spread into array
        this.realm.delete(
            this.realm
                .objects("ArkVtxoBranch")
                .filtered("vtxoKey == $0", vtxoKey(vtxo))  // ← deletes same underlying rows
        );
        for (const e of removed) {
            const stillRef = [
                ...this.realm
                    .objects("ArkVtxoBranch")
                    .filtered("virtualTxid == $0", e.virtualTxid),  // 💥 e is invalidated
            ];
```

Spreading a Realm `Results` into an array gives you an array of **live Realm object proxies**. After `this.realm.delete(...)` removes the same underlying rows, those proxies are invalidated. Accessing `e.virtualTxid` on line 914 will throw `"Accessing object which has been invalidated or deleted"` in a real Realm environment.

The mock test passes because `createMockRealm`'s `delete()` works by identity matching against stored Map values — the JS objects remain valid plain objects after removal from the Map. **This bug is invisible to the test suite.**

**Fix:** Extract the needed values before deleting:
```ts
this.realm.write(() => {
    const removed = [
        ...this.realm.objects("ArkVtxoBranch")
            .filtered("vtxoKey == $0", vtxoKey(vtxo)),
    ].map(e => ({ virtualTxid: (e as any).virtualTxid })); // snapshot values
    this.realm.delete(
        this.realm.objects("ArkVtxoBranch")
            .filtered("vtxoKey == $0", vtxoKey(vtxo))
    );
    for (const e of removed) { /* now safe */ }
});
```

---

### 🟡 TOCTOU — IndexedDB `upsertVirtualTxs` and `pruneForSpentVtxo` use separate transactions for read-then-write

**`src/repositories/indexedDB/virtualTxRepository.ts:52-76`** (`upsertVirtualTxs`)

The read happens in a `readonly` transaction (line 56), then the merged write happens in a separate `readwrite` transaction (line 64). Between the two, a concurrent `upsertVirtualTxs` call could commit, and the second call's merge would be based on stale `prevs` — potentially overwriting a newer `hex` with `null`.

**`src/repositories/indexedDB/virtualTxRepository.ts:131-163`** (`pruneForSpentVtxo`)

Three separate transactions: read removed branches → delete branches → check orphans + delete orphaned txs. A concurrent `setBranch` between delete and orphan-check could cause a VirtualTx to be incorrectly deleted (the newly-set branch's reference won't be counted).

**Mitigation:** For `upsertVirtualTxs`, you can do the get + put in a single `readwrite` transaction. IDB `readwrite` transactions on the same store are serialized:
```ts
const wt = db.transaction([STORE_VIRTUAL_TXS], "readwrite");
const ws = wt.objectStore(STORE_VIRTUAL_TXS);
for (const t of txs) {
    const prev = await req(ws.get(t.txid)) as VirtualTx | undefined;
    ws.put({ txid: t.txid, hex: t.hex ?? prev?.hex ?? null, ... });
}
await done(wt);
```

For `pruneForSpentVtxo`, the read + delete of branches should be in one `readwrite` transaction on `STORE_VTXO_BRANCHES`, and the orphan check + delete should be in a single transaction spanning both stores.

---

### 🟡 Protocol-adjacent: `getLockedVtxoOutpoints()` correctness analysis

This method is money-path — it will determine which VTXOs are excluded from spendable balance. The logic is correct:

| State | Locked? | Correct? |
|---|---|---|
| `waiting_to_submit` | ✅ locked | ✓ |
| `waiting_for_batch` | ✅ locked | ✓ |
| `batch_in_progress` | ✅ locked | ✓ |
| `batch_failed` | ❌ unlocked | ✓ (batch failed, VTXOs free) |
| `batch_succeeded` | ❌ unlocked | ✓ (VTXOs consumed) |
| `cancelled` | ❌ unlocked | ✓ (intent cancelled) |

However, the SQLite backend (`src/repositories/sqlite/intentRepository.ts:85-95`) loads **all rows** then filters in JS. When the wiring lands (T15), push the `WHERE state NOT IN ('batch_failed','batch_succeeded','cancelled')` down to SQL — no point pulling terminal intents off disk just to discard them.

---

### 🟡 Minor: `applySettlementEventToIntent` — unnecessary runtime check

**`src/wallet/intentStateReducer.ts:39`**

```ts
if ("id" in event && typeof event.id === "string") {
    updated.batchId = event.id;
}
```

Every variant of `SettlementEvent` has `id: string`. This check always passes. Either trust the types and remove the guard, or add a comment explaining why it's there (future-proofing?). As-is, it reads like the author wasn't sure whether `id` exists, which could confuse future reviewers.

---

### 🟢 Nits (non-blocking)

1. **`src/repositories/sqlite/intentRepository.ts:38-42`** — `saveIntent` wraps a single `INSERT OR REPLACE` in `BEGIN IMMEDIATE` / `COMMIT`. A single statement is already atomic in SQLite. The transaction wrapper adds overhead for no correctness gain here (unlike `pruneForSpentVtxo` where it's correctly used).

2. **`src/repositories/virtualTxRepository.ts:38`** — `VirtualTxMode` is defined but unused in this PR. Presumably for T16. Consider adding a `// used by T16 wiring` comment or deferring the type to that PR.

3. **Reducer doesn't handle `BatchFinalization`** (`SettlementEventType.BatchFinalization`) — falls through to no-op. This is likely intentional (it's an intermediate event before `BatchFinalized`), but a comment in the switch would help.

4. **No `waiting_to_submit → waiting_for_batch` transition** in the reducer. Presumably that transition is imperative (caller sets it when submitting to server) rather than event-driven. Worth a doc comment on the reducer to clarify which transitions it owns vs. which are imperative.

---

### 🟢 Cross-repo impact

- **14 downstream repos** import from `@arkade-os/sdk`. This PR adds new exports only — **no breaking changes**. Existing `WalletRepository`, `ContractRepository`, and barrel exports are untouched.
- `ChainedTxType` enum values (0–4) are **verified identical** to NArk `dotnet-sdk/NArk.Abstractions/VirtualTxs/ChainedTxType.cs`.
- No open PRs in consumer repos reference these new types yet.

---

### 🟢 Test coverage

Excellent. Shared conformance suites for both interfaces × all 4 backends. Reducer has lifecycle, monotonicity, and no-op tests. Schema version bump tests for IndexedDB v4 and Realm v3. The mock SQL executor extraction from `sqlite-wallet-repository.test.ts` into `test/helpers/mockSqlExecutor.ts` is a clean refactor.

**Gap:** The mock Realm doesn't reproduce real Realm's invalidated-object-after-delete semantics (see bug #1). Consider adding a `createStrictMockRealm()` variant that throws on property access after delete, or add an integration test note.

---

### Verdict

**Request changes** — the Realm `pruneForSpentVtxo` bug (#1) will crash in production. The IndexedDB TOCTOU (#2) is lower priority but should be fixed before the wiring PR lands.

⚠️ **Protocol-critical flag:** While this PR is storage-only, `getLockedVtxoOutpoints()` is directly money-path (double-spend prevention). The interface and state model look correct, but **human sign-off required** before merge, especially before the T14–T16 wiring PR that connects this to `settle()` / `getBalance()`.

pietro909 (approved):
I've addressed CodeRabbit's comments and refactored to follow the repo conventions (naming, code reuse, terser docs mostly).

@Kukks please give it a second pass otherwise is good to merge from my side

PR #495, opened 2026-05-19
chore: upgrade to Node 24 LTS

pietro909:
Adds `.nvmrc` (24.15.0) and points CI workflows to it via `node-version-file`. Updates `engines.node` in `package.json` to match.

## Chores
* Updated Node.js version requirement to 24.15.0
* Modified CI/CD workflows to reference Node version from centralized configuration file

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/ts-sdk/pull/495?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ghost (changes requested):
## Review: chore: upgrade to Node 24 LTS

CI so far: build ✅ lint ✅ unit ✅ integration ⏳ — the mechanical changes are clean. Using `.nvmrc` as single source of truth for Node version is the right pattern. No protocol-critical code touched.

**One blocking issue, one minor.**

---

### 🔴 BLOCKING — `engines.node` in package.json breaks downstream consumers

**`package.json:188`** — `"node": ">=24.15.0 <25"`

This field is **published to npm** as part of `@arkade-os/sdk`. Any consumer running Node 22 (the current LTS most people are on) will get:
- A warning on `npm install` / `pnpm install`
- A **hard failure** if they have `engine-strict=true`

This is a **library**, not an application. The `engines` field should express the *minimum supported runtime*, not the version CI happens to run. The SDK's pure-JS code (noble curves, protobuf, grpc-web) almost certainly works fine on Node 20+.

**Recommendation:** Widen to `">=22.12.0 <25"` (or even `">=20"` if you want to support Node 20 LTS which is still in maintenance). Keep `.nvmrc` at `24.15.0` for CI/dev — that's the right place to pin your dev environment.

**Known breakage:** `banco` pins `"node": ">=22.12.0 <23"` in both its `package.json` and CI workflows, and depends on `@arkade-os/sdk`. A strict engine bump here will block `banco` installs until it's also updated. Coordinate the rollout or widen the constraint.

---

### 🟡 Minor — `better-sqlite3` native addon

`better-sqlite3@12.6.2` is a devDependency (tests + examples only, not shipped to consumers), and CI unit tests pass, so it compiles fine on Node 24. No action needed — just flagging for awareness since native addons are the usual pain point in major Node upgrades.

---

### ✅ What looks good

- `.nvmrc` as SSOT for CI node version — eliminates version drift across workflows
- All 5 `setup-node` steps updated consistently
- No protocol code touched, no API surface changes, no cross-repo type breakage

**TL;DR:** Widen `engines.node` to `">=22.12.0 <25"` so this doesn't break every downstream consumer still on Node 22. Once that's fixed, this is good to merge.

ghost (approved):
## Follow-up review (new commit `2334facb`)

Blocking issue from my previous review is resolved: `engines.node` widened to `">=22.12.0 <25"` — downstream consumers on Node 22 won't break. ✅

CI: build ✅ lint ✅ unit ⏳ — no concerns, no protocol code touched.

**Approved.** Still needs human sign-off per repo policy.

PR #500, opened 2026-05-21
chore(release): per-package targets and scoped tags

pietro909:
Modeled after https://github.com/arkade-os/dotnet-sdk

## Summary

  - Replace lockstep `scripts/release.sh` with `scripts/release.mjs`; the `.sh` becomes a thin wrapper.
  - Releases are package-scoped: `pnpm run release -- <sdk|boltz-swap|all> <bump-or-version>`.
  - Tags switch from `v<version>` to `@arkade-os/<name>/<version>`.
  - Releasing `sdk` also releases `boltz-swap` (workspace:* dependency); override with `--boltz-bump`.
  - Before publishing boltz-swap, pack it and verify the pinned `@arkade-os/sdk` version matches the intended release.

  ## Behavior

  - Stable SDK release → boltz-swap defaults to `patch`.
  - Prerelease SDK release — whether a bump type (`prepatch --preid beta`) or a literal (`0.5.0-beta.0`) — mirrors the preid into boltz-swap. Unknown preids in literals hard-error and  require `--boltz-bump`.
  - `--cleanup` auto-detects targets from a `.git/` state file or manifest diff against HEAD. Never deletes remote tags or resets commits.
  - Existing GitHub Release workflow (`v*` trigger) intentionally unchanged — new scoped tags won't match it, which descopes GitHub Release automation for now.

  ## Validation

  - Dry-runs: `boltz-swap patch`, `sdk patch`, `sdk prepatch --preid beta`, `sdk 0.5.0-beta.0`, `sdk 0.5.0-mystery.0` (rejects), `sdk minor --boltz-bump patch`, `all patch`.
  - `pnpm pack` on boltz-swap confirms pnpm rewrites `workspace:*` to the current SDK version.
  - `pnpm run test:unit` and `pnpm run lint` pass.


ghost (commented):
## Code Review — `chore(release): per-package targets and scoped tags`

**Scope:** Release tooling only. No protocol-critical code (VTXOs, signing, forfeit, round, exit). No runtime SDK changes. No public API changes. No cross-repo consumers affected.

**Verdict:** Well-structured rewrite. The version math is correct, `execFileSync`/`spawnSync` with array args avoids injection, state-file recovery is solid, and the boltz-swap pack-and-verify step is a nice safety net. A few issues below, ranging from "should fix" to "nit."

---

### 1. tsdoc workflow also orphaned by tag format change (medium)

The PR description acknowledges that `.github/workflows/release.yml` (GitHub Releases, `v*` trigger) will no longer fire. But `.github/workflows/tsdoc.yml` also triggers on `v*` — after this PR, **docs won't auto-deploy on release either**. Worth calling out explicitly and tracking a follow-up to update both workflows to match the new `@arkade-os/*/` tag pattern (or add a manual dispatch step).

### 2. `--` doesn't terminate option parsing — `scripts/release.mjs:361`

```js
case "--":
    break;
```

`break` exits the `switch`, not the `for` loop. So `release.mjs sdk patch -- --cleanup` would still activate cleanup mode. This is how pnpm forwards args (`pnpm run release -- sdk patch`), so in practice the extra `--` is harmless, but a truly adversarial invocation like `release.mjs sdk patch -- --cleanup` would silently do the wrong thing.

**Fix:** either `continue` (to skip the token and continue the loop) or set a flag to stop processing options after `--`.

### 3. `distTagFor` uses loose `.includes()` — `scripts/release.mjs:289-294`

```js
if (version.includes("-alpha")) return "alpha";
```

A version like `1.0.0-alphabeta.0` would match `-alpha` and get the `alpha` dist-tag. With the `VALID_PREIDS` constraint this is unlikely in practice, but since `distTagFor` operates on the *result* version string (which could be a literal passed by the user), it's reachable:

```bash
release.mjs sdk 1.0.0-alphabet.0 --boltz-bump patch
# would fail the VALID_PREIDS check... actually no, it would pass because
# VALID_PREIDS is only checked on --preid, not on literal versions
```

Wait — literal versions bypass `validatePreid` entirely. So `release.mjs sdk 1.0.0-alphabetsoup.0` is legal, and `distTagFor` would return `"alpha"`. Edge case, but easy fix: match against the first prerelease segment:

```js
function distTagFor(version) {
    const pre = parseVersion(version).pre.split(".")[0];
    if (VALID_PREIDS.has(pre)) return pre;
    return "latest";
}
```

### 4. Equal version passes the guard — `scripts/release.mjs:454`

```js
if (compareVersions(parseVersion(next), parseVersion(current)) < 0) {
```

Strictly less-than means `next === current` is allowed. The npm publish step would skip it (version already published), but the script would still: run tests, build, stage manifests (no-op), potentially reuse HEAD, create/reuse a tag, and push. Harmless but confusing. Consider `<= 0` or document that re-releasing the same version is intentionally idempotent.

### 5. No branch guard before `git push origin HEAD` — `scripts/release.mjs:765`

If someone accidentally runs the release script on `main`, it pushes directly. The old script had the same issue, so this isn't a regression, but since you're rewriting the script anyway, consider:

```js
const branch = execFileSync("git", ["branch", "--show-current"], { cwd: ROOT_DIR, encoding: "utf8" }).trim();
if (branch === "main" || branch === "master") {
    die("Refusing to release from the main branch. Create a release branch first.");
}
```

### 6. Cleanup candidate set may include legitimately released tags — `scripts/release.mjs:654-656`

```js
candidates.add(`${pkg.tagPrefix}${current}`);
if (head) candidates.add(`${pkg.tagPrefix}${head}`);
```

If `current === head` (no dirty manifest) and the user runs `--cleanup sdk` explicitly, the script adds the HEAD version's tag to the delete candidates. If that tag corresponds to a real, pushed release, `git tag -d` only removes the local copy — no remote damage — but it's still surprising. Consider skipping tags that match HEAD when there's no state file indicating an in-progress release.

---

### Nits (non-blocking)

- **`scripts/release.mjs:107`**: `execFileSync` import is used but `spawnSync` handles most work. The mixed usage is fine but consider documenting when to use which (e.g., "execFileSync for read-only git queries, spawnSync for everything else").
- **README.md**: The cleanup docs say "run `git reset --hard HEAD~1` manually" — worth a ⚠️ emoji or bold warning since that's destructive.

---

**Bottom line:** No blocking issues. Items 2 and 3 are easy wins worth fixing before merge. The rest are suggestions. Clean work overall — the per-package release model with pack-and-verify is a meaningful improvement over lockstep.

PR #501, opened 2026-05-21
feat(sw): ServiceWorkerWallet.restore() with AggregateError round-trip

pietro909:
Stacked on top of #492 

## Summary

Follow-up to #492. Adds `ServiceWorkerWallet.restore({ gapLimit? })` so SW  wallets reach parity with `Wallet.restore()`.

 - New `RESTORE_WALLET` message; dispatcher runs `wallet.restore()` entirely worker-side (the `materialize()` callback never crosses the postMessage boundary).
  - Marked long-running on both sides so the bus deadline does not race the indexer-bound HD gap scan.
  - Explicit `AggregateError` round-trip: worker serializes to `{ name, message, errors: [{ name, message }] }`, page reconstructs before rejecting. Non-Aggregate errors fall through the standard path.
  - Signing-only entrypoint. `IWallet`, `IReadonlyWallet`, `ServiceWorkerReadonlyWallet`, and `ExpoWallet` are untouched —
    readonly callers still hit `ReadonlyWalletError`. The `scanContracts` proxy rejection in the SW contract-manager stays as
    the low-level guardrail.

  ## Out of scope

  - `ServiceWorkerReadonlyWallet.restore()` — needs a readonly variant of `Wallet.restore()` first.
  - Intermediate progress events — restore stays atomic from the caller's perspective.
  - Browser/SW e2e — existing regtest restore coverage is Node-only `Wallet.restore()` and does not exercise the worker transport.

## Release Notes

* **New Features**
  * Added wallet restoration method enabling users to recover wallet contract and balance information with an optional gap limit configuration.
  * Enhanced error handling for wallet operations, including improved serialization of aggregate errors in service worker communications.

### DIFF
--- a/packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts
+++ b/packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts
@@ -55,6 +55,49 @@ export class WalletNotInitializedError extends Error {
     }
 }
 
+/**
+ * Type-guard for a {@link SerializedAggregateError} object that has survived
+ * the postMessage boundary. Used on the page side of the RESTORE_WALLET path
+ * to detect a worker-side AggregateError and reconstruct it.
+ */
+export function isSerializedAggregateError(value: unknown): value is SerializedAggregateError {
+    if (!value || typeof value !== "object") return false;
+    const v = value as Partial<SerializedAggregateError>;
+    return (
+        v.name === "AggregateError" &&
+        typeof v.message === "string" &&
+        Array.isArray(v.errors) &&
+        v.errors.every((e) => e && typeof e.name === "string" && typeof e.message === "string")
+    );
+}
+
+/** Worker-side serializer for a real {@link AggregateError}. */
+export function serializeAggregateError(error: AggregateError): SerializedAggregateError {
+    const errors: { name: string; message: string }[] = [];
+    for (const child of error.errors ?? []) {
+        if (child instanceof Error) {
+            errors.push({ name: child.name, message: child.message });
+        } else {
+            errors.push({ name: "Error", message: String(child) });
+        }
+    }
+    return {
+        name: "AggregateError",
+        message: error.message,
+        errors,
+    };
+}
+
+/** Page-side reconstructor: rebuild an {@link AggregateError} from the wire form. */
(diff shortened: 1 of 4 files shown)


pietro909:
@coderabbitai please review

ghost (approved):
Clean follow-up to #492. The design is sound — running the entire restore worker-side and only crossing `gapLimit` + outcome over postMessage is the right call given `materialize()` can't be structured-cloned.

*What I looked at closely:*

*AggregateError round-trip* — the serialize/deserialize/isGuard trio is correct. Explicitly typed wire form (`SerializedAggregateError`) with a conservative type-guard is the right approach here; raw structured clone of `AggregateError.errors` is not reliably portable across all target browsers, so this is a good defensive choice even when it "happens to work" in modern Chrome.

*Long-running flag* — `RESTORE_WALLET` correctly added to `isLongRunning` alongside `SETTLE`, `RECOVER_VTXOS`, and `RENEW_VTXOS`. The comment explaining why (bus deadline must not race the gap scan) is a good signal to future maintainers.

*`sendMessageWithEvents` with `() => {}` event handler* — fine, since the restore path only emits a single terminal `RESTORE_WALLET_SUCCESS` and no intermediate progress events. If progress events are added later the handler will need updating, but the PR description explicitly defers those.

*Minor nit:* `serializeAggregateError(error) as unknown as Error` in the handler case block is a reasonable pragmatic cast, but it's slightly leaky — if any path between worker and page ever tries to call `.stack` on the error object it'll silently get `undefined`. Not a real risk here since the page-side catch checks `isSerializedAggregateError` first, but worth a comment in-line. Not blocking.

*`RESTORE_WALLET: 50_000` in `DEFAULT_MESSAGE_TIMEOUTS`* — the inline comment explaining it's for type completeness and never enforced as an inactivity deadline is sufficient. Fine.

*Test coverage* — solid: success with gapLimit, success without options, AggregateError reconstruction (with custom child `.name` preserved), non-AggregateError passthrough, readonly rejection, and `isLongRunning` assertion. The `structuredCloneError` update to handle both `Error` instances and plain objects via `JSON.parse(JSON.stringify(...))` mirrors what `postMessage` actually does.

:white_check_mark: Approving — transport layer only, not protocol-critical. The restore logic itself lives in #492.

PR #503, opened 2026-05-22
fix(boltz-swap): address CodeRabbit comments on PR #502

pietro909:
## Summary

Addresses the three CodeRabbit comments left on #502.

- **fix #1** (`36937074`): `resolveEffectiveFloor` now throws `TypeError` when `maxSlippageBps` (or a tiny baseline) drives the effective floor below 1 sat. Without this, `maxSlippageBps: 10000` silently restored blind-accept behaviour.
- **fix #2** (`0789d942`): `validateQuote` rejects non-finite, fractional, and `> MAX_SAFE_INTEGER` amounts via a new `non_safe_integer` rejection reason. `Infinity` previously slipped past `> 0` and the `< floor` check straight into `postChainQuote`. Transport-roundtrip path updated to match.
- **test #3** (`5c36b5a5`): the autopilot-cause regression test now drives `monitorSwap` → `transaction.lockupFailed` so the real `SwapError` wrap in `waitAndClaimBtc` executes. The previous version constructed `SwapError` manually and would have passed even if production code dropped `cause`.

## Test plan

- [x] `pnpm test:unit` — 342 boltz-swap tests pass (4 new)
- [x] CI green on the PR

### DIFF
--- a/packages/boltz-swap/src/arkade-swaps.ts
+++ b/packages/boltz-swap/src/arkade-swaps.ts
@@ -2338,7 +2338,16 @@ export class ArkadeSwaps {
         // Subtract-then-floor instead of multiply-then-divide: the original
         // `floor * 10000` form loses precision once `floor` exceeds
         // ~9e11 sats (above MAX_SAFE_INTEGER after multiply by 10000).
-        return Math.floor(floor - (floor * slippageBps) / 10000);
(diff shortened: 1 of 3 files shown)


pietro909:
@coderabbitai please review

Issue #466, opened 2026-04-29
Deprecate URL string arguments in wallet creation config


ghost:
## Summary

Deprecate the following URL string arguments from wallet creation config in favour of explicit provider instances. These were the original API; the provider pattern (`arkProvider`, `indexerProvider`, etc.) is now preferred as it supports custom transports (Expo, gRPC, service workers, etc.) and removes ambiguous override precedence.

**Arguments to deprecate:**
- `arkServerUrl`
- `indexerUrl`
- `esploraUrl`
- `delegatorUrl` _(ServiceWorker wallet only)_

### `arkServerUrl`
- `src/wallet/index.ts:43` — defined in `BaseWalletConfig`
- `src/wallet/serviceWorker/wallet.ts:332` — defined in `ServiceWorkerWalletOptions`
- `src/wallet/serviceWorker/wallet.ts:454` — stored as protected property
- `src/wallet/wallet.ts:200-210` — extracted, validated, used to construct `RestArkProvider`
- `src/wallet/wallet.ts:218` — defaults `indexerUrl` when not set
- `src/worker/messageBus.ts:373` — forwarded to message bus init

### `indexerUrl`
- `src/wallet/index.ts:45` — defined in `BaseWalletConfig`
- `src/wallet/serviceWorker/wallet.ts:334` — defined in `ServiceWorkerWalletOptions`
- `src/wallet/serviceWorker/wallet.ts:457` — stored as protected property
- `src/wallet/wallet.ts:217-220` — defaults to `arkServerUrl`; used to construct `RestIndexerProvider`
- `src/worker/messageBus.ts:375, 390` — forwarded through message bus config

### `esploraUrl`
- `src/wallet/index.ts:47` — defined in `BaseWalletConfig`
- `src/wallet/serviceWorker/wallet.ts:336` — defined in `ServiceWorkerWalletOptions`
- `src/wallet/serviceWorker/wallet.ts:458` — stored as protected property
- `src/wallet/wallet.ts:250-255` — defaults to network-specific `ESPLORA_URL`; used to construct `EsploraProvider`
- `src/worker/messageBus.ts:376, 391` — forwarded through message bus config

### `delegatorUrl` _(ServiceWorker only)_
- `src/wallet/serviceWorker/wallet.ts:345` — defined in `ServiceWorkerWalletOptions`
- `src/wallet/serviceWorker/wallet.ts:456` — stored as protected property
[...]

Issue #342, opened 2026-03-09
Expose notifyIncomingFunds in Wallet interface


altafan:
At the moment, `notifyIncomingFunds` is just a method of the Wallet class that implements the Wallet interface. Let's add the definition to the interface and also make sure that functions like `waitForIncomingFunds` take the interface as arg instead of the class (!!)

PR #511, opened 2026-05-22
chore: deprecate URL string args in wallet config

pietro909:
## Summary

Marks the URL-string fields on wallet config as `@deprecated` in favour of explicit provider instances. Runtime behaviour is unchanged — the URLs still construct the default providers internally — so this is non-breaking and existing integrations keep working.

## Changes

- `BaseWalletConfig.arkServerUrl` / `indexerUrl` / `esploraUrl` — point callers at `arkProvider` / `indexerProvider` / `onchainProvider`.
- `ServiceWorkerWalletOptions` — same three fields plus `delegatorUrl` (recommend explicit `DelegatorProvider`).

Refs #466 (decision confirmed by @gringokiwi in the issue thread).

## Out of scope

- TypeDoc / API-reference rebuild — handled separately per the issue.
- Eventual removal in a future major version.
- Issue #342 (expose `notifyIncomingFunds` on the interface) is a separate follow-up — adding it to `IReadonlyWallet` requires wiring it through the ServiceWorker message bus and is not a one-liner.

## Test plan

- [x] `pnpm run lint` clean
- [x] `pnpm -C packages/ts-sdk exec tsc --noEmit` clean
- [x] `pnpm run test:unit` pass (via pre-commit hook)

### DIFF
--- a/packages/ts-sdk/src/wallet/index.ts
+++ b/packages/ts-sdk/src/wallet/index.ts
@@ -67,11 +67,24 @@ export type WalletMode = "auto" | "static" | "hd" | DescriptorProvider;
  * @see StorageConfig
  */
 export interface BaseWalletConfig {
-    /** Base URL of the Arkade server. */
+    /**
+     * Base URL of the Arkade server.
+     *
+     * @deprecated Pass an explicit `arkProvider` instance instead. URL-based
+     * configuration will be removed in a future major version.
+     */
     arkServerUrl?: string;
-    /** Optional override for the indexer URL. */
+    /**
+     * Optional override for the indexer URL.
+     *
+     * @deprecated Pass an explicit `indexerProvider` instance instead.
+     */
     indexerUrl?: string;
-    /** Optional override for the Esplora API URL. */
+    /**
+     * Optional override for the Esplora API URL.
+     *
+     * @deprecated Pass an explicit `onchainProvider` instance instead.
+     */
     esploraUrl?: string;
 
     /** Optional Arkade server public key used to construct and validate Arkade addresses. */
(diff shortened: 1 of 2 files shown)


ghost (approved):
## Review: `chore: deprecate URL string args in wallet config`

**Verdict: Approve** — docs-only change, zero runtime impact, non-breaking.

### What this does
Adds `@deprecated` JSDoc annotations to URL-string fields (`arkServerUrl`, `indexerUrl`, `esploraUrl`, `delegatorUrl`) in `BaseWalletConfig` and `ServiceWorkerWalletOptions`, directing callers toward explicit provider instances.

### Checked
- **No runtime changes** — confirmed by full-file read of both files. Only JSDoc comments modified.
- **No protocol-critical code touched** — no VTXO handling, signing, forfeit paths, round lifecycle, or exit paths affected.
- **Cross-repo impact: none** — `@deprecated` is a documentation hint; no API signature or behavior changes. Downstream consumers compile identically.
- **Type compatibility preserved** — all fields remain `optional`, same types.

### Nit (non-blocking)

**`ServiceWorkerWalletOptions` deprecation messages point to alternatives that don't exist yet on that interface.**

- `packages/ts-sdk/src/wallet/serviceWorker/wallet.ts:327` — `arkServerUrl` says *"Provide an explicit provider via the worker config instead"* but `ServiceWorkerWalletOptions` has no `arkProvider` field.
- Same for `indexerUrl` (line 334) and `esploraUrl` (line 340) — no `indexerProvider`/`onchainProvider` fields on this interface.
- `delegatorUrl` (line 353) says *"Provide an explicit `DelegatorProvider` instance instead"* — also absent from this interface.

`BaseWalletConfig` deprecations are accurate (it has all four provider fields). The SW interface doesn't extend `BaseWalletConfig` and doesn't have these provider alternatives.

**Suggestion:** Either (a) note in the deprecation that SW provider support is a follow-up, or (b) add a tracking reference to the issue where those fields will be added. As-is, a caller reading the deprecation warning will look for a provider field that doesn't exist and be confused.

The class-level docstring example (line ~305) also still uses `arkServerUrl` in the sample — fine for now since it still works, but worth updating in the same follow-up.

### Summary
Clean, well-scoped deprecation pass. The SW interface nit is minor and can be addressed in a follow-up. Ship it. 🚀

gringokiwi (approved):

PR #513, opened 2026-05-22
fix(ramps): reject sub-dust change on partial offboard

### LINKED ISSUE #458: SDK should validate change VTXO dust before submitting collab exit
## Summary

When a user initiates a collaborative exit, the SDK submits the intent to arkd without first checking whether the *change VTXO* left behind would fall below the dust threshold (330 sats). This causes arkd to reject the intent with a dust error that surfaces directly to the user — an error that should never reach them.

## Current validation gaps

- \`wallet.ts\` checks \`output.amount <= dustAmount\` for explicit outputs ✓
- \`ramps.ts\` checks if fees exceed the offboard amount ✓
- \`vtxo-manager.ts\` skips periodic settle if \`totalAmount < dustAmount\` ✓

**Missing:** none of the above check whether the *change VTXO* remaining after a collab exit (\`amount - offboard_amount - fees\`) would be below dust.

## Expected behaviour

Before submitting the intent to arkd, the SDK should compute the expected change amount (total VTXO amount − exit amount − onchain output fee of 200 sats) and, if it falls below dust (330 sats), fail fast with a clear, actionable error — e.g. *"Amount too close to dust limit; consider exiting the full balance."*

> **Note:** the UX decision of what to show the user (e.g. offer to exit the full balance) belongs in the wallet layer, not the SDK. The SDK should simply reject early with a typed error so the wallet can handle it appropriately.

## Steps to reproduce

1. Hold a VTXO where \`amount - exit_amount - 200\` < 330 sats
2. Trigger a collab exit for that partial amount
3. Observe the SDK submitting to arkd and returning a dust error to the user

## Related files

- \`wallet.ts\`, \`ramps.ts\`, \`vtxo-manager.ts\`

pietro909:
## Summary

Pre-check the change amount during a collab exit (`Ramps.offboard`) and throw `DustChangeError` if the leftover change VTXO would fall below the wallet's dust threshold (default 330 sats). Today this case is forwarded to arkd, which rejects the intent and surfaces a raw dust error to the caller.

Closes #458.

## Changes

- `src/wallet/ramps.ts`: new `DustChangeError` class; offboard rejects when `0 < change < dustAmount`. Falls back to 330 sats when the wallet doesn't expose `dustAmount`.
- `src/index.ts`: re-export `DustChangeError`.

## Rationale

The wallet UI layer can `instanceof`-check `DustChangeError` and react with appropriate UX (e.g. offer to exit the full balance), rather than the caller pattern-matching on a server error string.

## Out of scope

- The same change-dust condition exists symmetrically on `Ramps.onboard`. Left untouched for this PR to keep scope tight to the reported bug; happy to follow up.
- No unit test file added — `ramps.test.ts` does not exist yet and standing up the Wallet/Estimator mocks would balloon the PR. The logic is small and the error is exercised via the existing types.

## Test plan

- [x] `pnpm run lint` clean
- [x] `pnpm -C packages/ts-sdk exec tsc --noEmit` clean
- [x] `pnpm run test:unit` pass (via pre-commit hook)

### DIFF
--- a/packages/ts-sdk/src/index.ts
+++ b/packages/ts-sdk/src/index.ts
@@ -80,7 +80,7 @@ import {
 } from "./wallet/wallet";
 import { TxTree, TxTreeNode } from "./tree/txTree";
 import { SignerSession, TreeNonces, TreePartialSigs } from "./tree/signingSession";
-import { Ramps } from "./wallet/ramps";
+import { DustChangeError, Ramps } from "./wallet/ramps";
 import { HDDescriptorProvider } from "./wallet/hdDescriptorProvider";
 import { isVtxoExpiringSoon, VtxoManager } from "./wallet/vtxo-manager";
 import type { IVtxoManager, SettlementConfig } from "./wallet/vtxo-manager";
@@ -292,6 +292,7 @@ export {
     isBatchSignable,
     OnchainWallet,
     Ramps,
+    DustChangeError,
     VtxoManager,
     HDDescriptorProvider,
     DelegatorManagerImpl,
--- a/packages/ts-sdk/src/wallet/ramps.ts
+++ b/packages/ts-sdk/src/wallet/ramps.ts
@@ -5,6 +5,7 @@ import { Address, OutScript } from "@scure/btc-signer";
 import { hex } from "@scure/base";
 import { networks, NetworkName } from "../networks";
 import { ArkAddress } from "../script/address";
+import { getDustAmount } from "./utils";
 
 /**
  * Thrown when a collaborative-exit / offboard would leave a change VTXO below
@@ -24,13 +25,6 @@ export class DustChangeError extends Error {
         this.change = change;
         this.dustAmount = dustAmount;
     }
-}
-
-/** Fallback dust threshold used when the wallet doesn't expose `dustAmount`. */
-const FALLBACK_DUST_AMOUNT = 330n;
-
-function getDustAmount(wallet: IWallet): bigint {
-    return "dustAmount" in wallet ? (wallet.dustAmount as bigint) : FALLBACK_DUST_AMOUNT;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


ghost (approved):
## ✅ Approved

Clean, low-risk client-side pre-validation. The logic is correct: catches sub-dust change VTXOs before they hit arkd, giving callers a typed error to react to.

### Verified

- `change` is computed correctly from `totalAmount - amount` (after input fee deduction, before output fee deduction — output fees only reduce the destination amount, not change)
- Condition `change > 0n && change < dustAmount` correctly allows full exits (`change == 0`) and exact-dust values (`change == dustAmount`)
- No protocol-level side effects — this is purely additive validation on the client side
- No cross-repo breakage — `DustChangeError` is a new export, `Ramps` interface unchanged
- Duck-typing for `dustAmount` follows existing pattern in `vtxo-manager.ts`

### Minor nit (non-blocking)

**`getDustAmount` duplication** — `packages/ts-sdk/src/wallet/vtxo-manager.ts:239-240` has an identical function:
```ts
function getDustAmount(wallet: IWallet): bigint {
    return "dustAmount" in wallet ? (wallet.dustAmount as bigint) : 330n;
}
```
Consider extracting to a shared utility (e.g. `wallet/utils.ts`) to keep a single source of truth for the fallback. Not blocking this PR.

### Note

The same pre-check is missing on `Ramps.onboard` (acknowledged as out-of-scope). Recommend a follow-up.

🤖 Reviewed by Arkana

ghost (approved):
## ✅ Re-approved (incremental review)

New commit `c8d97eb` — "Extract shared wallet dust amount helper" — addresses the nit from my previous review. Clean refactor:

- `getDustAmount` extracted to `wallet/utils.ts` with exported `FALLBACK_WALLET_DUST_AMOUNT = 330n` — single source of truth ✓
- `vtxo-manager.ts` local copy removed, now imports from shared utility ✓
- `type` import cleanup on `IWallet`, `Recipient`, `ReadonlyWallet` — correct, no runtime import where only types are needed ✓
- No behavioral change — function signature and fallback value identical ✓

No new concerns. Ship it.

🤖 Reviewed by Arkana

PR #516, opened 2026-05-25
Mainnet defaults + deprecate provider string arguments

gringokiwi:
Builds on https://github.com/arkade-os/ts-sdk/pull/511

- Move default network constants into networks
- Default human-readable prefix for address() and onchainAddress() to mainnet
- Deprecate arkServerUrl, indexerUrl, and esploraUrl from wallet configs (excluding service worker wallet)
- Default -ArkProvider, -IndexerProvider and EsploraProvider base URLs to mainnet

### DIFF
--- a/packages/ts-sdk/README.md
+++ b/packages/ts-sdk/README.md
@@ -31,8 +31,6 @@ const identity = MnemonicIdentity.fromMnemonic(mnemonic)
 const wallet = await Wallet.create({ identity })  // defaults to mainnet
 ```
 
-To use a different network, pass `arkServerUrl` option.
-
 ### Read-Only Wallets (Watch-Only)
 
 The SDK supports read-only wallets that allow you to query wallet state without exposing private keys. This is useful for:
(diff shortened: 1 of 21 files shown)


ghost (approved):
Clean DX improvement — centralizing defaults and making zero-config mainnet work out of the box. CI green, logic looks correct. A few minor notes:

**Nits / worth considering:**

1. `address(prefix: string = DEFAULT_NETWORK.hrp, serverPubKey: Bytes)` in `base.ts` — non-trailing default is valid TypeScript, but using it requires `script.address(undefined, serverPubKey)`. Was the intent to put `serverPubKey` first and `prefix` optional at the end? If so, argument order should flip. If not, this default can only be reached via explicit `undefined`, which is fine but a bit obscure.

2. Two `??` → `||` changes (`wallet.ts` line ~257, `expo/wallet.ts` line ~171). For `arkServerUrl` (string), `||` is arguably safer (empty string falls back). For `arkProvider` (object), `??` is more correct since a falsy-but-non-null provider makes no practical sense. Either way it's harmless in practice, just noting the semantic shift.

3. `DEFAULT_THRESHOLD_SECONDS = 259_200` — the previous `3 * 24 * 60 * 60` was more self-documenting. Minor, but worth keeping the computation if the intent is readability.

**No protocol-critical surface touched** (no VTXO signing, forfeit paths, or round lifecycle). Approved.

pietro909 (approved):

PR #518, opened 2026-05-25
feat: require csvTimelock to be specified

gringokiwi:
- Require `csvTimelock` to be specified when constructing `DefaultVtxo.Script` (and downstream `DelegateVtxo.Script`

## Summary by CodeRabbit

* **Breaking Changes**
  * csvTimelock is now required when configuring script timelocks; previous automatic fallback to a default timelock has been removed. Update configurations to provide explicit timelock values.

* **Documentation**
  * Examples and docs updated to show explicit csvTimelock entries with required type and value.

* **Tests**
  * Test fixtures updated to include explicit csvTimelock values to match the new required behavior.

### DIFF
--- a/packages/ts-sdk/src/contracts/handlers/default.ts
+++ b/packages/ts-sdk/src/contracts/handlers/default.ts
@@ -53,9 +53,7 @@ export const DefaultContractHandler: ContractHandler<DefaultContractParams, Defa
     },
 
     deserializeParams(params: Record<string, string>): DefaultContractParams {
-        const csvTimelock = params.csvTimelock
-            ? sequenceToTimelock(Number(params.csvTimelock))
-            : DefaultVtxo.Script.DEFAULT_TIMELOCK;
+        const csvTimelock = sequenceToTimelock(Number(params.csvTimelock));
         return {
             pubKey: extractPubKeyBytes(params.pubKey),
             serverPubKey: extractPubKeyBytes(params.serverPubKey),
--- a/packages/ts-sdk/src/contracts/handlers/delegate.ts
+++ b/packages/ts-sdk/src/contracts/handlers/delegate.ts
@@ -1,6 +1,5 @@
 import { hex } from "@scure/base";
 import { DelegateVtxo } from "../../script/delegate";
-import { DefaultVtxo } from "../../script/default";
 import { RelativeTimelock } from "../../script/tapscript";
 import { Contract, ContractHandler, Discoverable, PathContext, PathSelection } from "../types";
 import type { DiscoveredContract, DiscoveryDeps } from "../types";
@@ -46,9 +45,7 @@ export const DelegateContractHandler: ContractHandler<DelegateContractParams, De
     },
 
     deserializeParams(params: Record<string, string>): DelegateContractParams {
-        const csvTimelock = params.csvTimelock
-            ? sequenceToTimelock(Number(params.csvTimelock))
-            : DefaultVtxo.Script.DEFAULT_TIMELOCK;
+        const csvTimelock = sequenceToTimelock(Number(params.csvTimelock));
         return {
             pubKey: hex.decode(params.pubKey),
             serverPubKey: hex.decode(params.serverPubKey),
--- a/packages/ts-sdk/src/script/default.ts
+++ b/packages/ts-sdk/src/script/default.ts
@@ -17,7 +17,7 @@ export namespace DefaultVtxo {
     export interface Options {
         pubKey: Bytes;
         serverPubKey: Bytes;
-        csvTimelock?: RelativeTimelock;
+        csvTimelock: RelativeTimelock;
     }
 
     /**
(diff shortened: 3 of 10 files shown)


gringokiwi:
Not sure why CI is failing, both those tests are unrelated to these changes seemingly?

ghost (changes requested):
## Arkana Code Review — `feat: require csvTimelock to be specified`

**Verdict: Request Changes** (one protocol-safety issue in deserialization)

### Summary

Good change directionally — making `csvTimelock` explicit removes a hidden default that could silently produce incorrect timelocks. Callers should know what timelock they're using, especially since this controls the unilateral exit delay (i.e., how long funds are locked).

Cross-repo impact is clean: `arkade-os/wallet`, `arkade-os/demos`, `ArkLabsHQ/pixelark`, and `packages/snap` all already pass `csvTimelock` explicitly. No downstream breakage.

---

### 🔴 Issue: Unsafe deserialization of legacy persisted data

**`packages/ts-sdk/src/contracts/handlers/default.ts:56`** and **`packages/ts-sdk/src/contracts/handlers/delegate.ts:48`**

The old code:
```ts
const csvTimelock = params.csvTimelock
    ? sequenceToTimelock(Number(params.csvTimelock))
    : DefaultVtxo.Script.DEFAULT_TIMELOCK;
```

The new code:
```ts
const csvTimelock = sequenceToTimelock(Number(params.csvTimelock));
```

`params` is `Record<string, string>` — this is deserialized from **persisted contract data** (database). If any contract was stored before `csvTimelock` was serialized as a field, `params.csvTimelock` will be `undefined`, producing `Number(undefined)` → `NaN`, then `bip68.decode(NaN)` which will either throw an opaque error or return garbage.

This is **protocol-critical**: a corrupted timelock means the exit script won't match the on-chain VTXO, making funds unspendable via the unilateral exit path.

**Recommendation:** Either:
1. Keep the fallback in `deserializeParams` for backwards compat with stored data:
   ```ts
   const csvTimelock = params.csvTimelock
       ? sequenceToTimelock(Number(params.csvTimelock))
       : DefaultVtxo.Script.DEFAULT_TIMELOCK;
   ```
2. Or add a migration that backfills `csvTimelock` on all stored contracts before this change ships, and add a guard that throws a clear error:
   ```ts
   if (!params.csvTimelock) {
       throw new Error("Contract missing csvTimelock — run migration");
   }
   ```

Option 1 is safer and simpler. The type-level enforcement (making `Options.csvTimelock` required) already achieves the goal of forcing new callers to be explicit. The deserialization fallback is a different concern — it's a data compat layer, not an API design choice.

---

### 🟡 Minor: `selectPath` still guards against missing `csvTimelock`

**`default.ts:~72`** and **`delegate.ts:~62`** — `selectPath` still does:
```ts
const sequence = contract.params.csvTimelock
    ? Number(contract.params.csvTimelock)
    : undefined;
```

This is actually correct (it reads from the serialized `Record<string, string>` and `isCsvSpendable` handles `undefined`), but it means the codebase is inconsistent: `selectPath` guards against missing data while `deserializeParams` does not. Worth a comment explaining why, or unifying the approach.

---

### ✅ What looks good

- Making the type required in `Options` — correct design decision
- `DEFAULT_TIMELOCK` kept as a static constant — callers can still reference it when they want the default
- Test updates properly pass the timelock explicitly
- `DelegateVtxo.Options extends DefaultVtxo.Options` propagation is clean
- Wallet code (`wallet.ts:237`, `expo/wallet.ts:177`, `walletReceiveRotator.ts:514`) correctly drops the `?? DEFAULT_TIMELOCK` fallback since the type now guarantees presence
- Unused import cleanup (`DefaultVtxo` from delegate handler, `VirtualCoin` from expo wallet)

---

### ⚠️ Protocol-critical flag

This PR modifies VTXO script construction parameters — specifically the CSV timelock that governs the unilateral exit path. **Requires human review and sign-off before merge**, even after the deserialization issue is addressed.

🤖 *Reviewed by Arkana*

pietro909 (approved):
@arkanaai  `csvTimelock` has been present in `serializeParams` since the very first commit of both files.  The fallback in `deserializeParams (? DefaultVtxo.Script.DEFAULT_TIMELOCK)` was dead code from day one, someone being overly cautious.
I believe the removal is safe @gringokiwi

ghost (approved):
## Arkana Code Review — Update

**Verdict: Approve** ✅

@pietro909 is correct — I verified the git history. `serializeParams` has included `csvTimelock` since the very first commit of both handler files (`d74cc004`). The `?? DEFAULT_TIMELOCK` fallback in `deserializeParams` was dead code from day one. My previous concern about legacy persisted data is withdrawn.

The change is clean:
- Type-level enforcement (`csvTimelock` required in `Options`) is the right design
- All downstream consumers already pass `csvTimelock` explicitly
- No cross-repo breakage

### ⚠️ Protocol-critical flag (standing)

This PR modifies VTXO script construction parameters — specifically the CSV timelock governing unilateral exit paths. **Still requires human sign-off before merge**, per protocol-critical review policy.

🤖 *Reviewed by Arkana*

PR #519, opened 2026-05-25
chore: rename delegator -> delegate

gringokiwi:
- Deprecate "delegator" methods and params, replacing with "delegate"
- Modify `getDelegateInfo` response to return `delegateAddress` alongside deprecated `delegatorAddress`

## Summary by CodeRabbit

* **Updates**
  * Renamed delegation surface from "delegator" to "delegate" across the SDK and tooling.
  * Wallets and service-worker integrations now expose delegate-manager APIs and accept delegate-related configuration (delegator names kept as deprecated aliases for compatibility).
  * Backward compatibility preserved: existing delegator APIs remain available as deprecated aliases.

* **Documentation**
  * README and examples updated to use "delegate" terminology and show delegate wiring.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/ts-sdk/pull/519?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/boltz-swap/src/expo/background.ts
+++ b/packages/boltz-swap/src/expo/background.ts
@@ -63,6 +63,7 @@ function createBackgroundWalletShim(args: {
         getBoardingUtxos: async () => notImplemented("getBoardingUtxos"),
         getTransactionHistory: async () => notImplemented("getTransactionHistory"),
         getContractManager: async () => notImplemented("getContractManager"),
+        getDelegateManager: async () => notImplemented("getDelegateManager"),
         getDelegatorManager: async () => notImplemented("getDelegatorManager"),
         sendBitcoin: async () => notImplemented("sendBitcoin"),
         send: async () => notImplemented("send"),
--- a/packages/ts-sdk/README.md
+++ b/packages/ts-sdk/README.md
@@ -541,30 +541,30 @@ const balance = await manager.getRecoverableBalance()
 
 Delegation allows users to outsource virtual output renewal to a third-party delegation service.
 
-Instead of the delegating user renewing virtual outputs by themself, their delegate will automatically settle them before they expire, sending the funds back to the delegator's wallet address (minus a service fee).
+Instead of the delegating user renewing virtual outputs by themselves, their delegate will automatically settle them before they expire, sending the funds back to the user's wallet address (minus a service fee).
 
 This is useful for wallets that cannot be online 24/7.
 
-When a `delegatorProvider` is configured, the wallet address includes an extra tapscript path that authorizes the delegate to co-sign renewals alongside the Arkade server.
+When a `delegateProvider` is configured, the wallet address includes an extra tapscript path that authorizes the delegate to co-sign renewals alongside the Arkade server.
 
 To run a delegation service, you'll need to set up a [Fulmine server](https://github.com/ArkLabsHQ/fulmine) with the [Delegation API](https://github.com/ArkLabsHQ/fulmine?tab=readme-ov-file#-delegate-api) enabled.
 
 #### Setting Up a Wallet with Delegation
 
 ```typescript
-import { Wallet, MnemonicIdentity, RestDelegatorProvider } from '@arkade-os/sdk'
+import { Wallet, MnemonicIdentity, RestDelegateProvider } from '@arkade-os/sdk'
 
 const wallet = await Wallet.create({
   identity: MnemonicIdentity.fromMnemonic('abandon abandon...'),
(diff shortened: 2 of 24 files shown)


ghost (changes requested):
## Review: `chore: rename delegator -> delegate`

Clean rename with proper `@deprecated` aliases — good approach for backward compatibility. However, there's one bug that will silently break users on the deprecated path, plus a few smaller issues.

---

### 🔴 BUG: `Wallet.create` breaks backward compatibility

**`packages/ts-sdk/src/wallet/wallet.ts` (Wallet.create, ~line 1435)**

```typescript
config.delegateProvider,   // ← only checks new name
```

When a consumer passes `{ delegatorProvider: myProvider }` (deprecated but supported), `config.delegateProvider` is `undefined`. The `Wallet` constructor receives `undefined` and **never creates a `_delegateManager`**.

Meanwhile, `ReadonlyWallet.setup` *does* handle the fallback:
```typescript
delegateProvider: config.delegateProvider || config.delegatorProvider,
```

So the wallet **address** includes the delegate tapscript path (setup handled it), but `getDelegateManager()` returns `undefined` → delegation silently stops working. Funds sent to that address would be locked in a delegate path with no manager to renew them.

**Fix:** Change to:
```typescript
config.delegateProvider || config.delegatorProvider,
```

Or use `setup.delegateProvider` which already has the fallback resolved.

Compare with `ReadonlyWallet.create` (~line 431) which correctly does:
```typescript
setup.delegateProvider || setup.delegatorProvider,
```

---

### 🟡 README semantic error

**`packages/ts-sdk/README.md` line ~3 of the hunk:**

> sending the funds back to the **delegate's** wallet address (minus a service fee)

"Delegate" = the third-party service. But funds go back to the **user** (the one who delegated). The original "delegator's wallet address" was correct. Should be something like "the **user's** wallet address" or "the **delegating wallet's** address".

---

### 🟡 Downstream consumers will break on next `npm update`

Found active consumers of the deprecated APIs:

1. **`demos` repo** — 5 files import `RestDelegatorProvider` and use `delegatorProvider` param (arkade-address, delegate, swap-from-lightning, swap-to-lightning, asset demos)
2. **`wallet` repo** — `src/providers/wallet.tsx` uses `delegatorUrl`, `src/lib/asp.ts` calls `wallet.getDelegatorManager()`

The deprecated aliases keep these working at runtime, but downstream repos should have coordinated PRs ready. Consider adding a `@deprecated` JSDoc to the re-exports in `index.ts` so IDEs surface warnings.

---

### 🟢 Minor nits

1. **`packages/ts-sdk/src/providers/delegate.ts` ~line 116:** The `TODO` comment is fine, but `isDelegateInfo()` doesn't validate `delegateAddress` — if a custom `DelegateProvider` implementation returns `{ delegateAddress: "" }`, `ArkAddress.decode("")` will throw deep in the delegation flow rather than at validation time. Consider adding a check or documenting this clearly in the interface.

2. **`packages/ts-sdk/test/wallet.test.ts` ~lines 1257, 1280:** Tests pass *both* `delegateProvider` and `delegatorProvider` to `createReadonlyTestWallet`/`createFullMainnetWallet`. This doesn't test the backward-compat path — it tests the new path. Add a test case that only passes `delegatorProvider` to verify the deprecated path still works (this would have caught the `Wallet.create` bug above).

3. **`packages/ts-sdk/src/wallet/delegate.ts` ~line 202:** `DelegatorManagerImpl` const alias points to `DelegateManagerImpl` class — works at runtime, but `new DelegatorManagerImpl(...)` will show `DelegateManagerImpl` in stack traces. Acceptable for a deprecation path, just noting it.

---

### Protocol safety

The rename is mechanical — no protocol logic (VTXO handling, signing, forfeit construction, connector trees) was changed. The `delegate()` function internals are identical. ✅

**However**, the `Wallet.create` bug above means a user on the deprecated API could end up with funds in a delegate-path address but no delegation manager, meaning VTXOs won't get renewed and could expire. This is a **funds-at-risk** scenario for users who don't notice the silent breakage.

**Requesting changes for the `Wallet.create` backward-compat bug.** Everything else is solid.

ghost (approved):
## Re-review after force-push (`2679f83` → `e18129a9`)

Single change: `Wallet.create` now correctly falls back to `config.delegatorProvider` when `config.delegateProvider` is undefined. This was the critical bug I flagged — **fixed correctly**. ✅

The README semantic nit (line ~544: "delegate's wallet address" should be "user's wallet address") remains, but it's cosmetic and non-blocking.

LGTM. Ship it.

PR #523, opened 2026-05-26
fix: return on-chain txid for submarine/chain swap completion

### LINKED ISSUE #509: boltz-swap: waitForSwapCompletion returns swap ID instead of txid for submarine/chain swaps
> Mirrored from https://github.com/arkade-os/boltz-swap/issues/61 — the `boltz-swap` package now lives in this monorepo under `packages/boltz-swap/`.

Original review thread: https://github.com/arkade-os/boltz-swap/pull/60#discussion_r2783343386

## Problem

`SwapManager.waitForSwapCompletion` advertises a `{ txid: string }` contract but for submarine and chain swaps it resolves with the **swap id**, not a transaction id.

`packages/boltz-swap/src/swap-manager.ts:534-595`:

- Reverse swap branch (`:568-571`) correctly calls `getReverseSwapTxId(swap.id)` and resolves with `response.id`.
- Submarine branch `:577` — `resolve({ txid: updatedSwap.id })` — `updatedSwap.id` is the swap id.
- Chain branch `:583` — same bug.

The CodeRabbit suggestion on PR #60 proposed calling `getSwapStatus(swap.id)` and resolving with `status.transaction?.id` (rejecting if unavailable). That fix was never applied.

## Suggested fix

For submarine/chain swaps in the final-status path, look up the on-chain txid via `swapProvider.getSwapStatus(...)` (or the equivalent typed accessor) and reject if it is missing/empty, rather than leaking the swap id as a bogus txid.

pietro909:
## Problem

`SwapManager.waitForSwapCompletion` promises `{ txid }`, but for **submarine** and **chain** swaps it resolved the Boltz swap ID in place of the on-chain txid — in the subscription path (`resolve({ txid: updatedSwap.id })`) and, by throwing unconditionally, in the already-final early-exit path. Callers `waitAndClaimBtc` / `waitAndClaimArk` then propagated the bogus value to API consumers. The reverse branch was already correct. Fixes #509.

## Changes

- **`swap-manager.ts`** — new private `resolveClaimedTxid()` helper that reads the real `transaction.id` from `getSwapStatus`, rejecting when it's missing/empty.
  - Subscription path: submarine & chain now resolve via the helper.
  - Early-exit path: already-claimed submarine/chain return the real txid; other final statuses throw a status-bearing error (replacing the misleading "… already completed" throw).
- **`arkade-swaps.ts`** — the manual (non-SwapManager) chain claim paths in `waitAndClaimBtc` / `waitAndClaimArk` now reject with a `SwapError` on a missing/empty txid instead of resolving `txid: ""`, matching the reverse manual path.

## Tests

7 new cases in the "Wait for Completion" block: submarine & chain resolving the real txid (with a regression guard asserting `txid ≠ swap id`), missing-transaction rejection, failed-status rejection, already-claimed early-exit for both types, and the failed early-exit throw.

## Verification

- `swap-manager.test.ts`: 59 passed · `arkade-swaps.test.ts`: 173 passed
- full boltz-swap unit suite: 373 passed
- `tsc --noEmit` and `pnpm run lint`: clean

## Summary by CodeRabbit

* **New Features**
  * Swap completion now exposes the real on-chain transaction ID for claimed chain and submarine swaps.

* **Bug Fixes**
  * Completion/waiting logic now reliably resolves only when a valid on-chain txid is available and errors when missing or final-invalid states occur.
  * Service worker and client claim responses now include the txid payload.

* **Tests**
  * Enhanced coverage to verify txid propagation and failure cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/ts-sdk/pull/523?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/boltz-swap/src/arkade-swaps.ts
+++ b/packages/boltz-swap/src/arkade-swaps.ts
@@ -282,12 +282,8 @@ export class ArkadeSwaps {
                 refund: async (swap: BoltzSubmarineSwap) => {
                     await this.refundVHTLC(swap);
                 },
-                claimArk: async (swap: BoltzChainSwap) => {
-                    await this.claimArk(swap);
-                },
-                claimBtc: async (swap: BoltzChainSwap) => {
-                    await this.claimBtc(swap);
-                },
+                claimArk: (swap: BoltzChainSwap) => this.claimArk(swap),
+                claimBtc: (swap: BoltzChainSwap) => this.claimBtc(swap),
                 refundArk: async (swap: BoltzChainSwap) => {
                     return this.refundArk(swap);
                 },
@@ -1533,6 +1529,10 @@ export class ArkadeSwaps {
         }
         return new Promise<{ txid: string }>((resolve, reject) => {
             let claimStarted = false;
+            // Holds the BTC claim we broadcast — its txid is the swap's
+            // on-chain completion. getSwapStatus does not surface it at
+            // transaction.claimed, so resolve from the claim instead.
+            let claimPromise: Promise<{ txid: string }> | undefined;
             // Local mutable copy — accumulates fields across status
             // callbacks without mutating the caller's object.
             // Spreading from the original on every callback
@@ -1559,14 +1559,14 @@ export class ArkadeSwaps {
                         const updatedSwap = await updateSwapStatus();
                         if (claimStarted) return;
                         claimStarted = true;
-                        this.claimBtc(updatedSwap).catch(reject);
+                        claimPromise = this.claimBtc(updatedSwap);
+                        claimPromise.catch(reject);
                         break;
                     }
                     case "transaction.claimed": {
                         await updateSwapStatus();
-                        const claimedStatus = await this.getSwapStatus(pendingSwap.id);
-                        const txid = claimedStatus.transaction?.id;
-                        if (!txid || txid.trim() === "") {
+                        const txid = await this.resolveChainClaimTxid(pendingSwap.id, claimPromise);
+                        if (!txid) {
                             reject(
                                 new SwapError({
                                     message: `Transaction ID not available for claimed swap ${pendingSwap.id}.`,
@@ -1627,8 +1627,9 @@ export class ArkadeSwaps {
     /**
      * Claim sats on BTC chain by claiming the HTLC.
      * @param pendingSwap - The pending chain swap with BTC transaction hex.
+     * @returns The BTC transaction ID of the claim.
      */
-    async claimBtc(pendingSwap: BoltzChainSwap): Promise<void> {
+    async claimBtc(pendingSwap: BoltzChainSwap): Promise<{ txid: string }> {
         if (!pendingSwap.toAddress)
             throw new Error(`Swap ${pendingSwap.id}: destination address is required`);
 
@@ -1726,6 +1727,10 @@ export class ArkadeSwaps {
         });
 
         await this.swapProvider.postBtcTransaction(claimTx.hex);
+
+        // The BTC claim transaction we broadcast is the swap's on-chain
+        // completion; its id is the txid callers expect from waitAndClaimBtc.
+        return { txid: claimTx.id };
     }
 
     /**
@@ -1976,6 +1981,10 @@ export class ArkadeSwaps {
         }
         return new Promise<{ txid: string }>((resolve, reject) => {
             let claimStarted = false;
+            // Holds the ARK claim we submit — its txid is the swap's
+            // on-chain completion. getSwapStatus does not surface it at
+            // transaction.claimed, so resolve from the claim instead.
+            let claimPromise: Promise<{ txid: string }> | undefined;
             // Local mutable copy — accumulates fields across status
             // callbacks without mutating the caller's object. Spreading
             // from the original on every callback would silently
@@ -1997,13 +2006,13 @@ export class ArkadeSwaps {
                         await updateSwapStatus();
                         if (claimStarted) return;
                         claimStarted = true;
-                        this.claimArk(swap).catch(reject);
+                        claimPromise = this.claimArk(swap);
+                        claimPromise.catch(reject);
                         break;
                     case "transaction.claimed": {
                         await updateSwapStatus();
-                        const claimedStatus = await this.getSwapStatus(pendingSwap.id);
-                        const txid = claimedStatus.transaction?.id;
-                        if (!txid || txid.trim() === "") {
+                        const txid = await this.resolveChainClaimTxid(pendingSwap.id, claimPromise);
+                        if (!txid) {
                             reject(
                                 new SwapError({
                                     message: `Transaction ID not available for claimed swap ${pendingSwap.id}.`,
@@ -2071,8 +2080,9 @@ export class ArkadeSwaps {
      * Claim sats on ARK chain by claiming the VHTLC.
      * Refactored to use claimVHTLCIdentity + claimVHTLCwithOffchainTx utilities.
      * @param pendingSwap - The pending chain swap.
+     * @returns The Ark transaction ID of the claim.
      */
-    async claimArk(pendingSwap: BoltzChainSwap): Promise<void> {
+    async claimArk(pendingSwap: BoltzChainSwap): Promise<{ txid: string }> {
         if (!pendingSwap.toAddress)
             throw new Error(`Swap ${pendingSwap.id}: destination address is required`);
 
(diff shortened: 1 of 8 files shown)


ghost (approved):
Solid fix for #509. The root cause is clear — `updatedSwap.id` is the Boltz swap ID, not an on-chain txid — and the approach is correct.

**What's good:**
- `resolveClaimedTxid()` is a clean, focused helper. Adding the extra `getSwapStatus` call is justified because the subscription event payload typically doesn't carry `transaction.id`.
- Both paths (subscription and early-exit) are fixed consistently.
- The `arkade-swaps.ts` change correctly promotes empty-txid from a silent resolve to an explicit rejection — resolving `{ txid: "" }` would have been worse for callers.
- Test coverage is thorough, and the regression guard (`expect(result.txid).not.toBe("submarine-swap-1")`) is exactly the right thing to add for #509.

**Minor observations (non-blocking):**

1. *Error type inconsistency* — `resolveClaimedTxid` throws a plain `Error`, while `arkade-swaps.ts` throws `SwapError`. Callers doing `instanceof SwapError` checks on SwapManager rejections will miss these. Not a blocker for this PR but worth a follow-up to standardise.

2. *setTimeout in tests* — the 10 ms `setTimeout` pattern for simulating status updates is fine here (it's the existing test convention), but could be flaky in a heavily loaded CI runner. Worth keeping an eye on test reliability.

Neither of these affects correctness of the fix. 373 tests passing and a clean `tsc`/lint run is confidence-inspiring. ✅

ghost (changes requested):
## Review of commit `73518b9` — source chain swap txid from the claim we broadcast

Good follow-up. The first commit was correct for submarine swaps (Boltz exposes `transaction.id` via `getSwapStatus`), but for chain swaps the client performs the claim itself, so Boltz never surfaces the txid. Sourcing it from the claim we broadcast is the right fix.

### 🔴 Race condition in SwapManager (request-changes)

**`swap-manager.ts:858-881`** — `handleSwapStatusUpdate` notifies subscribers **before** running `executeAutonomousAction`. The execution order is:

1. `transaction.server.confirmed` arrives → `handleSwapStatusUpdate` starts
2. `swap.status` updated, subscribers notified (no-op here — not final)
3. `executeAutonomousAction` starts → `executeClaimBtcAction` → **awaits** `claimBtcCallback` (network call)
4. **While awaiting**, a WebSocket message delivers `transaction.claimed`
5. New `handleSwapStatusUpdate` fires (JS event loop yields at `await`)
6. Subscribers notified → `waitForSwapCompletion` callback fires → calls `resolveClaimedTxid`
7. `chainClaimTxids` **not yet populated** (claim still in-flight at step 3)
8. Falls back to `getSwapStatus` → **no txid for chain swaps** → rejects with `"Transaction ID not available"`

The `arkade-swaps.ts` path handles this correctly by passing `claimPromise` directly to `resolveChainClaimTxid`, which `await`s it. The `swap-manager.ts` path stores the txid in a map **after** the claim completes, creating a TOCTOU race.

**Practical risk**: low-ish — Boltz needs to see the claim tx on-chain/mempool before sending `transaction.claimed`, so the claim broadcast should finish first. But it's not guaranteed, especially with slow network conditions or if the service worker path adds latency.

**Suggested fix**: Store the claim *promise* (not just the resolved txid) in a per-swap map, and `await` it in `resolveClaimedTxid`:

```typescript
// Instead of Map<string, string>:
private chainClaimPromises = new Map<string, Promise<{ txid: string }>>();

// In executeClaimBtcAction:
const promise = this.claimBtcCallback(swap);
this.chainClaimPromises.set(swap.id, promise);
const result = await promise;
// ... still remember txid if you want the fast path

// In resolveClaimedTxid:
const claimPromise = this.chainClaimPromises.get(swapId);
if (claimPromise) {
    const res = await claimPromise.catch(() => undefined);
    if (res?.txid?.trim()) return { txid: res.txid };
}
// fallback to getSwapStatus...
```

This mirrors the correct pattern in `arkade-swaps.ts:resolveChainClaimTxid`.

### ⚠️ Minor observations (non-blocking)

1. **`swap-manager.ts:1192` — `rememberChainClaimTxid` signature**: `result: { txid?: string } | void` — the defensiveness is good for backward compat with older integrations/test doubles. But `SwapManagerCallbacks` now types the return as `Promise<{ txid: string }>`, so `void` is only possible via type escape. Worth a brief comment explaining why.

2. **`swap-manager.ts:900` — `finalizeMonitoredSwap` does not clean `chainClaimTxids`**: After a chain swap completes, its entry in `chainClaimTxids` lingers until `removeSwap()` or `start()`. Minor memory leak, bounded by chain swap count. Not urgent.

3. **Double reject in `arkade-swaps.ts`**: If `claimBtc` rejects, `claimPromise.catch(reject)` fires first. Then when `transaction.claimed` arrives, `resolveChainClaimTxid` catches the rejected promise, falls through to `getSwapStatus`, gets `undefined`, and calls `reject` again. Harmless in JS (second `reject` is a no-op on a settled promise) but worth being aware of.

4. **Error type inconsistency** (repeated from first review): `resolveChainClaimTxid` in `arkade-swaps.ts` rejects with `SwapError`, `resolveClaimedTxid` in `swap-manager.ts` rejects with plain `Error`. Still worth a follow-up to standardize.

### ✅ What's good in this commit

- `claimBtc` returning `{ txid: claimTx.id }` is correct — the `Transaction.id` from `@scure/btc-signer` is the txid of the serialized claim.
- `claimVHTLCwithOffchainTx` now returns `arkTxid` — clean, minimal change.
- `joinBatch` already returned `string` (commitment txid), so `claimArk` correctly captures both branches.
- Interface propagation through `IArkadeSwaps`, `ExpoArkadeSwaps`, service worker runtime/handler is thorough and consistent.
- `resolveChainClaimTxid` in `arkade-swaps.ts` has the correct pattern: await the claim promise, swallow rejections, fall through to provider status. This is well-designed.
- Test coverage is comprehensive: captures the "claim txid wins" assertion and the "getSwapStatus not called" guard.

### ⚠️ Protocol-critical flag

This PR modifies chain swap claim completion logic (`claimBtc`, `claimArk`, VHTLC claim paths). These touch protocol-critical paths — the txid returned to callers is used as proof of swap completion. **Requesting human review** per protocol-critical rules, even though the logic is correct.

ghost (approved):
**LGTM — solid fix for a nasty race-prone bug.**

## Summary

The root cause is clear: Boltz does not surface the chain claim txid via `getSwapStatus` at `transaction.claimed`, so the old code was resolving with the Boltz swap ID (a string that happens to be present) instead of the real on-chain txid. The fix is to source the txid from the claim callback we invoke ourselves, with a principled fallback to provider status for resumed sessions.

## What I verified

**`claimVHTLCwithOffchainTx` → `Promise<string>`** ✅
`vhtlc.ts` correctly returns `arkTxid` and `ArkadeSwaps.joinBatch` already returns `Promise<string>` (delegates to the same util), so both branches of the ternary in `claimArk` produce a real txid.

**Promise-capture pattern in `resolveClaimedTxid` / `resolveChainClaimTxid`** ✅
Storing the _promise_ (not the resolved txid) is the right call — it closes the window where `transaction.claimed` arrives while the broadcast is still in-flight. Swallowing the claim rejection here is intentional and safe: the `.catch(reject)` at the call site already routes that error to the outer promise rejector, so callers get exactly one rejection path.

**Double-reject safety** ✅
If `claimPromise` rejects, the outer `.catch(reject)` fires first; the `resolveClaimedTxid` fallback may attempt a second `reject()` but the Promise is already settled — no-op.

**Service worker propagation** ✅
`ResponseClaimArk` / `ResponseClaimBtc` types now carry `payload: { txid: string }` and the runtime casts accordingly. Straightforward and consistent with the existing message protocol.

**Test quality** ✅
The regression guard (`expect(result.txid).not.toBe("chain-swap-1")`) is exactly right — pins the specific failure mode from #509. The in-flight race test is particularly good; it directly models the cooperative-claim timing window called out in the second commit message.

## Minor observations (non-blocking)

- `chainClaimPromises` is typed as `Map<string, Promise<{ txid?: string } | void>>`. The `void` arm exists for defensive compatibility with older callback integrations — that's fine, but worth a `// never in practice, legacy compat` note if it stays.
- `rememberChainClaim` has an `if (claimPromise)` guard that can't be false at the call site (we only call it when a non-null callback returns). Harmless, but slightly misleading.
- The already-claimed early-exit path now calls `resolveClaimedTxid` and hits `getSwapStatus` for chain swaps we never claimed in this session (no entry in `chainClaimPromises`). That's the right behavior — just worth noting it adds one network call on the restored-swap path.

All 373 tests passing, tsc and lint clean. Approving.

PR #527, opened 2026-05-27
fix: widen RealmLike.create mode to boolean | string

### LINKED ISSUE #526: RealmLike.create mode param (`string`) makes a real Realm un-assignable, forcing `as any`
### Summary

`RealmLike` (`@arkade-os/sdk` → `repositories/realm`) documents that *"consumers pass their real Realm instance and the compiler validates it satisfies this shape"*, but a real `realm` `Realm` instance is **not** assignable to `RealmLike`. Every consumer is forced to cast, e.g. `new RealmWalletRepository(realm as any)`.

### Cause

`RealmLike.create` types the `mode` param as `string`:

```ts
// packages/ts-sdk/src/repositories/realm/types.ts
create(schemaName: string, values: Record<string, any>, mode?: string): void;
```

Real `Realm.create` types it as `boolean | UpdateMode`, where `UpdateMode` is a string enum. `boolean | UpdateMode` is not assignable to `string` (because of `boolean`), so the signatures are incompatible:

```
Argument of type 'Realm' is not assignable to parameter of type 'RealmLike'.
  Types of property 'create' are incompatible.
    Types of parameters 'mode' and 'mode' are incompatible.
      Type 'string | undefined' is not assignable to type 'boolean | UpdateMode | undefined'.
```

### Repro

`@arkade-os/sdk@0.4.29`, `realm` v12:

```ts
import Realm from 'realm';
import { RealmWalletRepository } from '@arkade-os/sdk/repositories/realm';

declare const realm: Realm;
new RealmWalletRepository(realm); // TS2345
```

### Fix proposal

Widen `mode` to also accept `boolean`. `UpdateMode` (a string enum) is assignable to `string`, so `boolean | string` is a supertype of Realm's `boolean | UpdateMode` — a real `Realm` becomes assignable, the repositories' internal `realm.create(..., "modified")` calls keep compiling, and no dependency on `realm`'s types is introduced:

```diff
// packages/ts-sdk/src/repositories/realm/types.ts
-    create(schemaName: string, values: Record<string, any>, mode?: string): void;
+    create(schemaName: string, values: Record<string, any>, mode?: boolean | string): void;
```

Verified locally: with this one-line change a real `Realm` is assignable to `RealmLike` and every existing `"modified"` call site still type-checks.

Alternatively, for stricter typing: `import type { UpdateMode } from 'realm'`, type `mode?: boolean | UpdateMode`, and replace the internal `"modified"` literals with `UpdateMode.Modified` — at the cost of coupling the type surface to `realm`.

pietro909:
Widen `RealmLike.create`'s `mode` param from `string` to `boolean | string` so a real `realm` `Realm` instance (whose `create` types `mode` as `boolean | UpdateMode`) is assignable to `RealmLike` without `as any`. Internal `"modified"` call sites still type-check.

Fixes #526

### DIFF
--- a/packages/ts-sdk/src/repositories/realm/types.ts
+++ b/packages/ts-sdk/src/repositories/realm/types.ts
@@ -16,6 +16,6 @@ export interface RealmLike {
     write(callback: () => void): void;
     objects<T = Record<string, unknown>>(schemaName: string): RealmResults<T>;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    create(schemaName: string, values: Record<string, any>, mode?: string): void;
+    create(schemaName: string, values: Record<string, any>, mode?: boolean | string): void;
     delete(objects: unknown): void;
 }


ghost (approved):
## Review: APPROVE ✅

**1-line change, type-only, non-protocol-critical.** Reviewed the full file, all consumers, call sites, tests, and cross-repo impact.

### What it does

Widens `RealmLike.create`'s `mode` parameter from `string` to `boolean | string` so that a real Realm instance (whose `create` types `mode` as `boolean | UpdateMode`) satisfies the interface without `as any` casts.

### Verification

- **All 6 internal `.create()` call sites** (`walletRepository.ts`, `contractRepository.ts`, `swap-repository.ts`) pass `"modified"` — a `string` literal. Still compiles under `boolean | string`. ✅
- **Type theory is correct.** `RealmLike.create` is in contravariant (parameter) position. Widening the parameter type makes `RealmLike` accept *more* implementations — this is the right direction. `boolean | UpdateMode` (Realm's type) ⊆ `boolean | string` (new `RealmLike` type). ✅
- **Test mocks** (`realm-wallet-repository.test.ts:110`, `realm-contract-repository.test.ts:104`) use method-shorthand syntax with `mode?: string`. Under `strict: true`, method signatures are bivariant, so the mock remains assignable to `RealmLike`. ✅
- **CI passes** — `Lint, Test & Build` green. ✅
- **Cross-repo impact**: `RealmLike` is only consumed within `ts-sdk` packages (`ts-sdk`, `boltz-swap`). No other SDK (go/rust/dotnet) has a Realm abstraction. No downstream breakage. ✅
- **No protocol-critical code touched.** This is a persistence-layer type interface — no VTXOs, signing, forfeit paths, round lifecycle, or exit logic involved. ✅

### Nit (non-blocking)

The test mock at `packages/ts-sdk/test/realm-wallet-repository.test.ts:110` still types `mode` as `string`. Consider updating to `boolean | string` for parity — not required since bivariance makes it compile, but it keeps the mock honest if someone later passes `true` in a test.

Clean fix. Ship it.

🤖 Reviewed by Arkana

PR #528, opened 2026-05-27
fix: set lockupDetails.timeouts for restored chain swaps

### LINKED ISSUE #524: restoreSwaps omits lockupDetails.timeouts for chain swaps, making restored ARK→BTC refunds impossible
### Summary

`SwapManager.restoreSwaps()` rebuilds chain swaps without `response.lockupDetails.timeouts`. Because `refundArk()` hard-requires that field, any ARK→BTC chain swap recovered via restore (rather than created locally) can reach a refundable state but cannot actually be refunded — `refundArk()` throws `Swap <id>: missing timeouts in lockup details`.

The reverse and submarine restore branches already populate the equivalent `timeoutBlockHeights`; only the chain branch was missed.

### Affected version

`@arkade-os/boltz-swap` 0.3.34 (current `packages/boltz-swap`).

### Current behavior

In `src/arkade-swaps.ts`, the chain branch of `restoreSwaps()` destructures only four fields from `refundDetails` and builds `lockupDetails` without `timeouts`:

```ts
// src/arkade-swaps.ts (~2838–2873)
} else if (isRestoredChainSwap(swap)) {
    const refundDetails = swap.refundDetails;
    if (!refundDetails) continue;

    const { amount, lockupAddress, serverPublicKey, timeoutBlockHeight } =
        refundDetails;

    chainSwaps.push({
        // ...
        response: {
            id,
            lockupDetails: {
                amount,
                lockupAddress,
                serverPublicKey,
                timeoutBlockHeight,        // <-- no `timeouts`
            },
        },
    } as BoltzChainSwap);
}
```

`refundArk()` then rejects the restored swap:

```ts
// src/arkade-swaps.ts (~1740)
if (!pendingSwap.response.lockupDetails.timeouts)
    throw new Error(`Swap ${pendingSwap.id}: missing timeouts in lockup details`);
```

### Why the data is available

`refundDetails` is a `Details` object (`src/boltz-swap-provider.ts:762`), which **always carries `tree`** and **optionally carries `timeoutBlockHeights`**:

```ts
export type Details = {
    tree: Tree;                                 // required
    // ...
    timeoutBlockHeights?: TimeoutBlockHeights;   // optional
};
```

And the target field is type-compatible — `lockupDetails` is a `ChainSwapDetailsResponse` (`src/boltz-swap-provider.ts:579`) with `timeouts?: TimeoutBlockHeights`. So the timeouts can be taken directly from `refundDetails.timeoutBlockHeights`, or derived from the VHTLC tree leaves exactly as the sibling branches already do.

### Expected behavior

Restored chain swaps should populate `lockupDetails.timeouts` so `refundArk()` works on recovered swaps, mirroring the reverse/submarine branches in the same function (`src/arkade-swaps.ts` ~2771 and ~2822), which do:

```ts
timeoutBlockHeights: timeoutBlockHeights ?? {
    refund: extractTimeLockFromLeafOutput(tree.refundWithoutBoltzLeaf?.output ?? ""),
    unilateralClaim: extractTimeLockFromLeafOutput(tree.unilateralClaimLeaf?.output ?? ""),
    unilateralRefund: extractTimeLockFromLeafOutput(tree.unilateralRefundLeaf?.output ?? ""),
    unilateralRefundWithoutReceiver: extractTimeLockFromLeafOutput(tree.unilateralRefundWithoutBoltzLeaf?.output ?? ""),
},
```

### Proposed fix

In the chain branch of `restoreSwaps()`, also pull `tree` and `timeoutBlockHeights` from `refundDetails` and set `lockupDetails.timeouts` using the same `refundDetails.timeoutBlockHeights ?? extractTimeLockFromLeafOutput(...)` pattern used by the reverse/submarine branches.

One caveat worth handling: `extractTimeLockFromLeafOutput` returns `0` for a missing/unparseable leaf, so a derived set could contain zeros if the restore response's tree is incomplete. Consider validating the derived `refund` (and unilateral) values are non-zero and leaving `timeouts` unset when they aren't, so a clearly-unrefundable swap fails the existing explicit guard in `refundArk()` rather than building a wrong VHTLC (the address-mismatch check at the end of `refundArk()` would otherwise reject it less legibly).

### Test coverage

`test/arkade-swaps.test.ts` already has a `restoreSwaps` describe block (~2590) whose `pendingChain` fixture supplies `refundDetails: { ...mockDetails, tree: mockTree }`, but the assertions only check swap ids/counts. Extend it to assert `result.chainSwaps[0].response.lockupDetails.timeouts` is populated (both the `refundDetails.timeoutBlockHeights`-present path and the tree-derived fallback), parallel to coverage for restored reverse/submarine timeouts.

pietro909:
## Summary

`ArkadeSwaps.restoreSwaps()` rebuilt chain swaps without `response.lockupDetails.timeouts`. Because `refundArk()` hard-requires that field, any ARK→BTC chain swap recovered via restore (after a restart) could reach a refundable state but threw `missing timeouts in lockup details` and could never be refunded. The reverse and submarine restore branches already populated the equivalent timeouts; only the chain branch was missed.

Fixes #524.

## Changes

- Add a shared `resolveVhtlcTimeouts(tree, timeoutBlockHeights?)` helper (`src/utils/restoration.ts`) that prefers the server-provided timeouts, otherwise derives them from the VHTLC tree leaves, and returns `undefined` when any locktime is missing/zero (incomplete restore response).
- Set `lockupDetails.timeouts` in the chain branch of `restoreSwaps()` — the actual fix.
- Route the reverse and submarine branches through the same helper, collapsing two duplicated inline blocks.
- Export `TimeoutBlockHeights` so the helper can reference it.

### Zero-guard

`extractTimeLockFromLeafOutput` returns `0` for a missing/unparseable leaf. Rather than build a VHTLC from zeroed timeouts (which fails `refundArk`'s address-match check with a cryptic error), the helper returns `undefined` so each consumer's explicit guard fires legibly:
- chain → `missing timeouts in lockup details`
- reverse → `incomplete reverse swap response`
- submarine → `incomplete submarine swap response`

## Test plan

- New `restoreSwaps` cases: chain server-provided path, chain tree-derived path (real VHTLC hex), chain zero-guard → `undefined`, and reverse/submarine parity.
- `pnpm -C packages/boltz-swap test:unit` — 370 passing.
- `pnpm run build` (incl. dts type-check) and `pnpm run lint` — clean.

### DIFF
--- a/packages/boltz-swap/src/arkade-swaps.ts
+++ b/packages/boltz-swap/src/arkade-swaps.ts
@@ -79,7 +79,7 @@ import {
 } from "./utils/boltz-swap-tx";
 import { decodeInvoice, getInvoicePaymentHash } from "./utils/decoding";
 import { normalizeToXOnlyKey } from "./utils/signatures";
-import { extractInvoiceAmount, extractTimeLockFromLeafOutput } from "./utils/restoration";
+import { extractInvoiceAmount, resolveVhtlcTimeouts } from "./utils/restoration";
 import { SwapManager, SwapManagerClient } from "./swap-manager";
 import {
     saveSwap,
@@ -2768,20 +2768,7 @@ export class ArkadeSwaps {
                         onchainAmount: amount,
                         lockupAddress,
                         refundPublicKey: serverPublicKey,
-                        timeoutBlockHeights: timeoutBlockHeights ?? {
-                            refund: extractTimeLockFromLeafOutput(
-                                tree.refundWithoutBoltzLeaf?.output ?? "",
-                            ),
-                            unilateralClaim: extractTimeLockFromLeafOutput(
-                                tree.unilateralClaimLeaf?.output ?? "",
-                            ),
-                            unilateralRefund: extractTimeLockFromLeafOutput(
-                                tree.unilateralRefundLeaf?.output ?? "",
-                            ),
-                            unilateralRefundWithoutReceiver: extractTimeLockFromLeafOutput(
-                                tree.unilateralRefundWithoutBoltzLeaf?.output ?? "",
-                            ),
-                        },
+                        timeoutBlockHeights: resolveVhtlcTimeouts(tree, timeoutBlockHeights),
                     },
                     status,
                     type: "reverse",
@@ -2819,28 +2806,21 @@ export class ArkadeSwaps {
                         address: lockupAddress,
                         expectedAmount: amount,
                         claimPublicKey: serverPublicKey,
-                        timeoutBlockHeights: timeoutBlockHeights ?? {
-                            refund: extractTimeLockFromLeafOutput(
-                                tree.refundWithoutBoltzLeaf?.output ?? "",
-                            ),
-                            unilateralClaim: extractTimeLockFromLeafOutput(
-                                tree.unilateralClaimLeaf?.output ?? "",
-                            ),
-                            unilateralRefund: extractTimeLockFromLeafOutput(
-                                tree.unilateralRefundLeaf?.output ?? "",
-                            ),
-                            unilateralRefundWithoutReceiver: extractTimeLockFromLeafOutput(
-                                tree.unilateralRefundWithoutBoltzLeaf?.output ?? "",
-                            ),
-                        },
+                        timeoutBlockHeights: resolveVhtlcTimeouts(tree, timeoutBlockHeights),
                     },
                 } as BoltzSubmarineSwap);
             } else if (isRestoredChainSwap(swap)) {
                 const refundDetails = swap.refundDetails;
                 if (!refundDetails) continue;
 
-                const { amount, lockupAddress, serverPublicKey, timeoutBlockHeight } =
-                    refundDetails;
+                const {
+                    amount,
+                    lockupAddress,
+                    serverPublicKey,
+                    timeoutBlockHeight,
+                    tree,
+                    timeoutBlockHeights,
+                } = refundDetails;
 
                 chainSwaps.push({
                     id,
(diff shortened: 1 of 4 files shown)


pietro909:
@coderabbitai please review

ghost (approved):
*Approve — solid fix with good test coverage*

*Bug fix* ✅

The root cause is clear: `restoreSwaps()` was rebuilding chain swaps without populating `lockupDetails.timeouts`, making `refundArk()` permanently throw after restart. The reverse/submarine branches already handled this; the chain branch was the odd one out. The fix is correct.

*`resolveVhtlcTimeouts` helper* ✅

Good design decisions:
- Prefer server-provided timeouts, fall back to tree derivation — right priority order
- Zero-guard returning `undefined` is cleaner than letting downstream VHTLC construction fail with a cryptic address-mismatch. The `||` guard correctly catches both missing fields and the literal-zero case from `extractTimeLockFromLeafOutput`
- Clean import — `TimeoutBlockHeights` export from `boltz-swap-provider.ts` is a one-liner change, no API surface concerns

*Refactoring* ✅

Collapsing the two duplicated inline blocks in reverse/submarine into the same helper is net-positive. Behaviour is preserved (verified by the parity test).

*Tests* ✅

Four cases are exactly right:
1. Server-provided path (happy path)
2. Real VHTLC hex → tree-derived fallback (actual byte parsing exercised)
3. Empty tree → zero-guard → `undefined` (regression guard for the original bug direction)
4. Reverse/submarine parity (catches drift if the branches diverge again)

*One observation to keep in mind*

`refundDetails` is typed optional on `RestoredChainSwap`, but the existing `if (!refundDetails) continue;` guard before the destructure handles that — `tree` is required on `Details`, so once past the guard it's always present. No issue here, just worth knowing.

*Fund-safety note*

This touches the refund/restore path for ARK→BTC chain swaps. Not protocol-critical in the VTXO-signing sense, but directly affects whether users can recover funds after a restart. The tests with real VHTLC hex give me confidence the parsing is correct. Tagging for a second human :eyes: before merge given the fund-safety dimension.

_Approving. 370 passing, build + lint clean — good to go pending human sign-off._

bordalix (approved):

PR #531, opened 2026-05-28
[codex] Fix duplicate boarding refill history entries

pietro909:
## Summary

Fixes duplicate transaction history entries for boarding/refill receives when the indexer leaf VTXO cannot be correlated back to the boarding transaction by `settledBy` or commitment id.

## Root Cause

The previous fix handled the ID-correlated path, but the live indexer payload can return a settled leaf with an empty `settledBy` and a commitment id that does not match the boarding sweep transaction. In that case, `buildTransactionHistory` still emitted a separate batch receive for the same boarding amount.

## Changes

- Track settled boarding receive rows that can suppress one matching batch receive.
- Preserve the existing ID-based suppression when `commitmentTxid` or `settledBy` are available.
- Add regression coverage for missing boarding commitment id, missing `settledBy`, and same-amount independent receives.

## Validation

- `pnpm --dir packages/ts-sdk exec vitest run test/transactionHistory.test.ts`
- `pnpm --dir packages/ts-sdk exec prettier --check src/utils/transactionHistory.ts test/transactionHistory.test.ts`
- `pnpm --dir packages/ts-sdk exec tsc --noEmit`
- Commit hook: monorepo lint and workspace unit tests passed

## Bug Fixes
- Enhanced transaction history processing to eliminate duplicate entries and improve accuracy when handling certain transaction scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/ts-sdk/pull/531?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/ts-sdk/src/utils/transactionHistory.ts
+++ b/packages/ts-sdk/src/utils/transactionHistory.ts
@@ -9,6 +9,20 @@ const txKey: TxKey = {
     arkTxid: "",
 };
 
+function consumeBoardingReceive(
+    boardingTxs: ArkTransaction[],
+    predicate: (tx: ArkTransaction) => boolean,
+): boolean {
+    const index = boardingTxs.findIndex(predicate);
+    if (index === -1) return false;
+    boardingTxs.splice(index, 1);
+    return true;
+}
+
+function isSettledBoardingReceive(tx: ArkTransaction): boolean {
+    return tx.type === TxType.TxReceived && tx.settled && tx.key.boardingTxid !== "";
+}
+
 function collectAssets(vtxos: VirtualCoin[]): Asset[] | undefined {
     const map = new Map<string, bigint>();
     for (const vtxo of vtxos) {
@@ -64,6 +78,9 @@ export async function buildTransactionHistory(
     getTxCreatedAt?: (txid: string) => Promise<number | undefined>,
 ): Promise<ExtendedArkTransaction[]> {
     const fromOldestVtxo = [...vtxos].sort((a, b) => a.createdAt.getTime() - b.createdAt.getTime());
+    const unmatchedSettledBoardingTxs = allBoardingTxs
+        .filter(isSettledBoardingReceive)
+        .sort((a, b) => a.createdAt - b.createdAt);
 
     const sent: ExtendedArkTransaction[] = [];
     let received: ExtendedArkTransaction[] = [];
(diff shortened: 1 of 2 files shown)


ghost (approved):
## Review: Fix duplicate boarding refill history entries

**Scope**: Display-only (transaction history rendering). No protocol-critical paths touched — no VTXO spending, signing, forfeit, or exit logic modified. Function signature of `buildTransactionHistory` is unchanged; no cross-repo impact.

### ✅ What's good

1. **Root cause is correctly identified.** When the indexer returns a settled leaf with empty `settledBy` and a `commitmentTxId` that doesn't match the boarding sweep, the old code emitted a spurious batch receive alongside the boarding entry. This fix addresses that.

2. **`consumeBoardingReceive` is well-designed.** Splice-based one-shot consumption ensures each boarding entry can only suppress one batch receive. The `ignoredCommitment` branch consumes first (by exact correlation), so the amount-matching fallback in the `else if` branch only fires for truly uncorrelated leaves. Good defensive ordering.

3. **Test coverage is solid.** Three new tests cover: missing commitment ID, missing `settledBy`, and the critical edge case of same-amount independent receives not being falsely suppressed. All three target the exact scenarios described in the PR body.

4. **No API surface change.** `buildTransactionHistory` signature is identical. Callers in `wallet.ts` and `wallet-message-handler.ts` are unaffected.

### ⚠️ Observations (non-blocking)

**`transactionHistory.ts:105-108` — Amount-only matching is a heuristic, not a correlation.**

The fallback predicate `tx.amount === vtxo.value && tx.createdAt <= vtxoCreatedAt` can false-positive if an independent receive of the exact same amount is processed *before* the actual boarding sweep leaf. The test at line 340 (third new test) only covers the case where the boarding sweep leaf is iterated first (createdAt + 60s) and the independent receive second (createdAt + 120s). If their order were reversed, the independent receive would be incorrectly suppressed.

In practice this is low-risk: boarding amounts are typically non-round numbers (e.g., 84960 sats) making collisions rare, and VTXOs are sorted oldest-first which naturally processes the boarding sweep leaf before later independent receives. But it's worth documenting this assumption in a code comment so future maintainers don't stumble on it.

**`transactionHistory.ts:22` — `isSettledBoardingReceive` checks `boardingTxid !== ""` but not `type === TxReceived` for the enum value.**

Currently works because `TxType.TxReceived` is used, but if the enum representation ever changes, this could silently break. The check is fine as-is but is worth a note.

### Verdict

Clean, targeted fix with good test coverage. The amount-matching heuristic is the best available approach given the indexer's data limitations. Ship it.

PR #540, opened 2026-06-02
fix(indexer): set virtualStatus.state to spent for spent VTXOs

ghost:
## Summary
- The `convertVtxo` function in `packages/ts-sdk/src/providers/indexer.ts` never set `virtualStatus.state` to `"spent"`, even though `vtxo.isSpent` was available and mapped to the output object.
- A spent VTXO would incorrectly report its state as `"settled"`.
- Added `isSpent` as the first check in the ternary chain so spent VTXOs are correctly classified before checking for swept or preconfirmed states.

## Test plan
- [ ] Verify that VTXOs with `isSpent: true` now have `virtualStatus.state === "spent"`
- [ ] Verify that non-spent VTXOs continue to report correct states (swept, preconfirmed, settled)
- [ ] Run integration tests against a server with spent VTXOs


bordalix (approved):

pietro909 (changes requested):
@arkanaai this is an old footgun we met already in the past.

The SDK does not positively consume` virtualStatus.state === "spent"` from the indexer. 
Spendability is decided by the `isSpent` boolean (index.ts:659), and the public `Wallet.getVtxos()` filters on that boolean (wallet.ts:550), never on state.

## The regression

Three "pending tx" paths filter that raw stream with a denylist of states rather than checking `isSpent`. They only exclude "swept" and "settled". After the commit, spent VTXOs become "spent" and slip through all three filters:

1. wallet.ts:782 `fetchPendingTxs`
2. wallet.ts:2285 `finalizePendingTxs (no-arg fetch)`
3. wallet-message-handler.ts:1229 `onWalletInitialized`

## Proposal

Keep the more-correct classification, but harden the three sites to be state-independent.
Filter on !vtxo.isSpent / isSpendable(vtxo) (or pass pendingOnly: true to the indexer, since all three want pending coins). This fixes the root fragility: a denylist of states is the wrong tool when a robust isSpent boolean exists. Add a spent fixture to the message-handler test.

pietro909:
@coderabbitai please review before lunch, I'm hungry

ghost (commented):
## Arkana Self-Review — Focused on new commits (revert of wallet hardening)

### Context
Three commits: (1) indexer fix, (2) wallet hardening, (3) revert of wallet hardening. Net diff: only the `indexer.ts` ternary fix.

### The fix is correct ✅

`convertVtxo` now checks `isSpent` first in the ternary chain (`indexer.ts:741-747`). Previously, a spent VTXO with `isSpent: true` but `isSwept: false` and `isPreconfirmed: false` would fall through to `"settled"` — a silent misclassification.

### Pietro's regression concern — not applicable to current code

I traced all three pending-tx filter sites:

| Location | Filter logic | Safe? |
|---|---|---|
| `wallet.ts:785` (`fetchPendingTxs`) | `!vtxo.isSpent && state !== "swept" && state !== "settled"` | ✅ `isSpent` catches it |
| `wallet.ts:2287` (`finalizePendingTxs`) | `vtxo.isSpent \|\| state === "swept" \|\| state === "settled"` | ✅ `isSpent` catches it |
| `wallet-message-handler.ts:1231` (`onWalletInitialized`) | `!vtxo.isSpent && state !== "swept" && state !== "settled"` | ✅ `isSpent` catches it |

All three already guard on `isSpent` before the state denylist. The revert of commit 2 was the right call — the hardening was redundant.

### Balance calculation — safe

`getVtxos()` (`wallet.ts:550`) filters through `isSpendable()` → `!vtxo.isSpent`. Spent VTXOs never reach the balance computation at `wallet.ts:490-498`.

### Other consumers — checked

- `banco/src/maker.ts:176`: Already handles `state !== "spent"` ✅
- `wallet.ts:651`: `settled: utxo.virtualStatus.state === "spent"` is for onchain boarding UTXOs, unrelated path
- e2e tests checking `state === "settled"` are polling for settlement completion, not affected

### Verdict

The fix is minimal, correct, and safe. No downstream breakage. The revert was the right move — all consumer sites already guard on `isSpent`.

⚠️ **Protocol-critical**: This touches VTXO state classification. Requesting human sign-off before merge.

@gringokiwi @bordalix — please confirm before merging.

PR #533, opened 2026-05-29
Add boarding contract type

pietro909:
Adds a registered `boarding` contract handler and derives the on-chain boarding address from it, replacing the ad-hoc `boardingTapscript` plumbing. The boarding address and the `getBoardingAddress()` API are unchanged.

## What
- `BoardingContractHandler` (type `boarding`), reusing the `default` `DefaultVtxo.Script` shape and spend paths; sources its CSV timelock from `ArkInfo.boardingExitDelay`. Not discoverable (HD index selection stays wallet-owned).
- Exported from the package root (`@arkade-os/sdk`) alongside the other built-in handlers.
- Wallet setup sources `boardingTapscript` from the handler; contract-manager init persists a matching `active` boarding contract so its Arkade address is watched.
- `script` is the unique contract identity. When `boardingExitDelay` coincides with the offchain unilateral-exit delay, the boarding script is byte-identical to the `default` baseline; `ensureWalletContract` accepts the existing row in either direction instead of creating a duplicate or throwing. `default`/`boarding` are the only compatible pair; `delegate` stays strict.

## Notes
- Backward compatible: the contract-derived boarding script produces a byte-identical pkScript/address.
- In the collision case the single shared row may carry either type — consumers must not rely on `type === "boarding"` to identify boarding purpose; resolve via `getBoardingAddress()` / `boardingTapscript`.

## Tests
- New unit coverage for the handler (registration, script/address byte-identity, serde, not-discoverable, root export) and wallet integration (persistence, watcher registration, idempotency, server-delay change, both collision directions).
- Full non-e2e unit suite, typecheck, lint, and build pass.

Manually tested on *mutinynet*.

## Summary by CodeRabbit

* **New Features**
  * Added boarding contract handler to the TypeScript SDK, now available through the public API alongside existing contract handlers.
  * Enhanced wallet with idempotent contract initialization logic for improved baseline contract collision handling.
  * Boarding contracts are now fully integrated into the wallet's contract lifecycle and management.

* **Tests**
  * Added comprehensive test suites for boarding contract functionality and wallet integration scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/arkade-os/ts-sdk/pull/533?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #543, opened 2026-06-03
fix(contracts): restore DEFAULT_TIMELOCK fallback for params without csvTimelock

pietro909:
PR #533 (`feat/boarding-contract-type`) made `DefaultVtxo.Script.Options.csvTimelock` required and removed the constructor-level `= DEFAULT_TIMELOCK` fallback, but never moved that fallback into the contract handler. As a result `DefaultContractHandler.deserializeParams` runs `sequenceToTimelock(Number(params.csvTimelock))` when `csvTimelock` is absent, i.e. `sequenceToTimelock(NaN)` → `bip68.decode(0)` → a zero timelock instead of `DEFAULT_TIMELOCK`, producing the wrong pkScript.

This surfaced as a failing unit test that #533 itself added:

```
test/contracts/handlers/boarding.test.ts > "falls back to the default timelock when csvTimelock is missing"
  AssertionError: expected `51200ffce6c9…` to deeply equal `5120cb829bac…`
```

(The analogous `default.test.ts` legacy-hex-param tests only assert the script is defined, so they pass even with the zero-timelock bug — which is why only the boarding test caught it.)

## Fix

Restore the fallback at the deserialize boundary: when `csvTimelock` is absent (legacy/minimal params), use `DefaultVtxo.Script.DEFAULT_TIMELOCK` instead of decoding `NaN`. This keeps the `Script` type strict (explicit timelock required) while preserving the lenient legacy-params behavior the handler tests expect. Real callers (`wallet.ts` boarding/default/delegate) always pass an explicit `csvTimelock`, so this only affects the defensive legacy path.

## Validation

- `boarding.test.ts` now passes; full ts-sdk unit suite green (1305 passed, 1 skipped).
- `default.test.ts` legacy-hex tests still pass (hex and descriptor params both fall back identically).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where contract parameters with missing timelock values could result in invalid data. The system now properly defaults to a standard timelock value in these scenarios.

### DIFF
--- a/packages/ts-sdk/src/contracts/handlers/default.ts
+++ b/packages/ts-sdk/src/contracts/handlers/default.ts
@@ -53,7 +53,14 @@ export const DefaultContractHandler: ContractHandler<DefaultContractParams, Defa
     },
 
     deserializeParams(params: Record<string, string>): DefaultContractParams {
-        const csvTimelock = sequenceToTimelock(Number(params.csvTimelock));
+        // csvTimelock may be absent on legacy/minimal params (e.g. hex pubkeys
+        // with no timelock). DefaultVtxo.Script no longer applies its own
+        // fallback, so restore it here rather than feeding sequenceToTimelock
+        // a NaN (which silently decodes to a zero timelock).
+        const csvTimelock =
+            params.csvTimelock !== undefined && params.csvTimelock !== ""
+                ? sequenceToTimelock(Number(params.csvTimelock))
+                : DefaultVtxo.Script.DEFAULT_TIMELOCK;
         return {
             pubKey: extractPubKeyBytes(params.pubKey),
             serverPubKey: extractPubKeyBytes(params.serverPubKey),


ghost (approved):
## Arkana Code Review — arkade-os/ts-sdk#543

**Verdict: APPROVE** ✅ — but **⚠️ PROTOCOL-CRITICAL: requires human sign-off before merge** (touches CSV timelock / exit path logic).

---

### What the PR does

Restores `DEFAULT_TIMELOCK` (144 blocks / ~1 day) fallback in `DefaultContractHandler.deserializeParams` when `csvTimelock` is absent from legacy/minimal params. Without this fix, `Number(undefined)` → `NaN` → `bip68.decode(NaN)` → `bip68.decode(0)` → **zero-block CSV timelock**, making the exit path immediately spendable and breaking the protocol's security assumption that the server has a CSV window to sweep forfeited VTXOs.

### Analysis

**Bug severity: HIGH.** A zero-block timelock on the exit path would allow premature unilateral exits. In practice this only triggers on the legacy deserialization path (real callers via `wallet.ts` always provide explicit `csvTimelock`), but the defensive boundary must be correct.

**Fix correctness: VERIFIED.** The guard `params.csvTimelock !== undefined && params.csvTimelock !== ""` correctly catches absent keys and empty strings in `Record<string, string>`. Falls back to `DefaultVtxo.Script.DEFAULT_TIMELOCK` (`{ type: "blocks", value: 144n }`).

**Boarding handler (`boarding.ts`):** Delegates to `DefaultContractHandler.deserializeParams` — inherits this fix automatically. ✅

**Delegate handler (`delegate.ts:49`):** Already has an equivalent fallback using truthy check (`params.csvTimelock ? ... : DEFAULT_TIMELOCK`). ✅

**`DefaultVtxo.Script` constructor:** Confirmed on the branch that `csvTimelock` is now required in `Options` (no destructuring default), so the handler is the correct place for this fallback. ✅

**Cross-repo impact:** None. No public API, type, or interface changes. Internal deserialization logic only.

### Findings

**1. Minor style inconsistency** — `default.ts` (this PR):
```ts
params.csvTimelock !== undefined && params.csvTimelock !== ""
```
vs `delegate.ts:49` (existing):
```ts
params.csvTimelock ?
```
Both are functionally equivalent for `Record<string, string>`. The explicit check is marginally more defensive. Not blocking, but consider aligning for consistency.

**2. Test gap** — `default.test.ts:51-58` ("should create script from legacy hex params") only asserts `script.pkScript` is defined — does NOT verify the script uses `DEFAULT_TIMELOCK` vs a zero timelock. This means the zero-timelock bug passed the default handler's own tests. The boarding test caught it only because it compared full pkScript bytes. Consider adding:
```ts
it("should use DEFAULT_TIMELOCK when csvTimelock is absent", () => {
    const params = DefaultContractHandler.deserializeParams({
        pubKey: TEST_PUB_KEY_HEX,
        serverPubKey: TEST_SERVER_PUB_KEY_HEX,
    });
    expect(params.csvTimelock).toEqual(DefaultVtxo.Script.DEFAULT_TIMELOCK);
});
```
This would catch regressions directly at the handler level. Non-blocking but recommended.

**3. `vhtlc.ts:67-70`** — `sequenceToTimelock(Number(params.claimDelay))` etc. have no absent-field guard. These fields are likely always present for VHTLC, but the same `NaN` risk exists if they're ever missing. Out of scope for this PR.

### CI

- Lint, Test & Build: ✅ pass
- Integration tests: pending

---

**LGTM.** Clean, minimal, well-scoped fix. The comment explains the "why" clearly. Approving the code — **a human must also approve before merge** given this touches exit-path timelock logic.

🤖 Reviewed by Arkana

PR #545, opened 2026-06-05
restore: scan deprecated server signers for L2 vtxos

pietro909:
## Summary

Closes the signer-rotation restore gap: `wallet.restore()` discovered L2 (`default`/`delegate`) VTXOs only under the server's **current** signer key, so VTXOs minted under a now-deprecated signer were anchored to a different script and silently missed — a fund-loss path on signer rotation. This mirrors how the reference NArk SDK already scans `{ current ∪ deprecated signers }`.

Implements `docs/hd-wallets-deprecated-signers.md`.

## Changes

- **`DiscoveryDeps`** gains `deprecatedSignerPubKeys?: Uint8Array[]`; `serverPubKey` is now documented as the snapshot's *current* signer.
- **`_runRestore`** fetches a single fresh `getInfo()` snapshot and derives **both** the current signer (used as `deps.serverPubKey`) and the deprecated set from it — no longer mixing a stale instance signer with fresh history. A new `toXOnlyPubKey` normalizer handles 33-byte compressed → x-only (32-byte) keys.
- **`default.ts` / `delegate.ts`** `discoverAt` wraps the `csvTimelocks` loop in a signer loop over `[serverPubKey, ...deprecatedSignerPubKeys]`, dedupes candidates by `scriptHex`, and threads the matched signer through the script, persisted `params.serverPubKey`, **and** the encoded `address` so signing/forfeit later resolves the correct key.
- Boarding discovery stays current-key only (documented out-of-scope blind spot).

Also includes a separate commit covering the spent-boarding restore blind spot in tests.

## Tests

- Handler-level: deprecated-signer discovery stamps the matched key into params + address (default & delegate); scriptHex dedup yields one entry / one probe.
- Restore-level: a VTXO minted under a **compressed (33-byte)** deprecated key is discovered and persisted under the deprecated x-only key in both `params.serverPubKey` and the encoded `address` (exercises the normalization).
- `tsc --noEmit` clean; full unit suite passes (1333 passed, 1 skipped) excluding e2e and the pre-existing `migrations.test.ts` failure (missing `better-sqlite3` native bindings locally — fails on `master` too).

## Summary by CodeRabbit

* **New Features**
  * Faster, batched contract discovery with concurrent probing and a configurable scan batch size for quicker restores and syncs.
  * Improved detection that scans historical signer variants to recover addresses minted under rotated keys.

* **Bug Fixes**
  * Wallet restore now recovers funds created under deprecated signer keys, deduplicates duplicate discoveries, and records the actual signer for restored addresses.

* **Tests**
  * Expanded regression tests for deprecated-signer discovery, deduplication, batching, concurrency, and nuanced restore edge cases.

### DIFF
--- a/packages/ts-sdk/src/contracts/constants.ts
+++ b/packages/ts-sdk/src/contracts/constants.ts
@@ -1,0 +1,14 @@
+/**
+ * Default indexer pagination size for batched VTXO queries (`getVtxos`).
+ *
+ * Shared by {@link ContractManager}'s bulk history sync and the discovery
+ * handlers' batched `detectUsedScripts` probe so the two can't drift. It lives
+ * in its own leaf module (importing nothing) so both the manager and the
+ * handler layer can use it without a `handlers → contractManager` import cycle —
+ * contractManager imports the handler registry, so the handlers must not import
+ * back from contractManager.
+ *
+ * Large enough that a single wallet index's candidate scripts resolve in one
+ * page in the common case.
+ */
+export const DEFAULT_PAGE_SIZE = 500;
--- a/packages/ts-sdk/src/contracts/contractManager.ts
+++ b/packages/ts-sdk/src/contracts/contractManager.ts
@@ -8,7 +8,6 @@ import {
     ContractState,
     ContractVtxo,
     ContractWithVtxos,
-    DiscoveredContract,
     DiscoveryDeps,
     GetContractsFilter,
     PathContext,
@@ -33,8 +32,7 @@ import {
     saveVtxosForContract,
     warnAndFilterVtxosForScript,
 } from "./vtxoOwnership";
-
-const DEFAULT_PAGE_SIZE = 500;
+import { DEFAULT_PAGE_SIZE } from "./constants";
 
 /**
  * Whether two *different* contract types may legitimately share a single
@@ -585,6 +583,9 @@ export class ContractManager implements IContractManager {
      *   other handler hit it).
      * - `persistAndWatchContract` rejecting is operational/fatal and
      *   propagates (only `discoverAt` is guarded).
+     * - Within an index the handler probes run concurrently (independent
+     *   network reads); their hits are persisted sequentially in
+     *   `discoverables` order to preserve the first-wins collision tie-break.
      */
     async scanContracts(opts: ScanContractsOptions): Promise<ScanResult> {
         const gapLimit = opts.gapLimit ?? 20;
@@ -599,9 +600,10 @@ export class ContractManager implements IContractManager {
             .filter(isDiscoverable);
 
         // Probe `boarding` before `default`/`delegate`. This ordering is
-        // LOAD-BEARING, not cosmetic: each hit is persisted immediately and
-        // `upsertContract` resolves a same-script collision FIRST-WINS, so the
-        // iteration order IS the tie-break. In the degenerate equal-delay case
+        // LOAD-BEARING, not cosmetic: within each index the probes run
+        // concurrently, but their hits are persisted in THIS order and
+        // `upsertContract` resolves a same-script collision FIRST-WINS, so this
+        // order IS the persistence tie-break. In the degenerate equal-delay case
         // a rotated index can carry BOTH an on-chain boarding UTXO and an L2
         // VTXO at the same (byte-identical) script; probing boarding first
         // resolves that collision to a `boarding` row, which keeps the on-chain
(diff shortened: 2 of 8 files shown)


ghost (commented):
## Arkana Code Review — restore: scan deprecated server signers for L2 vtxos

> ⚠️ *Protocol-critical path (VTXO restore / fund recovery) — flagged for mandatory human sign-off before merge.*

---

### Summary

This PR closes a real fund-loss path: after server signer rotation, VTXOs minted under the old key are anchored to a different tapscript and were silently skipped by `discoverAt`. The fix is correct in approach — scan `[current, ...deprecated]` signers at restore time, thread the *matched* signer through `params.serverPubKey` and `address`, dedup by `scriptHex`.

---

### ✅ What's done well

**Null-safety of `deprecatedSigners`**: I initially flagged `arkInfo.deprecatedSigners.map(...)` as potentially unsafe for older servers. On inspection, `ArkProvider.getInfo()` normalises this to `[] ` via `?? []` at line 273 of `providers/ark.ts` before returning — so `arkInfo.deprecatedSigners` is always an array. No bug here.

**Key normalisation**: `toXOnlyPubKey` handles both 33-byte compressed and 32-byte x-only keys and throws on anything else. Clean and defensive.

**Signer threading**: The matched `serverPubKey` is correctly plumbed into *both* `params.serverPubKey` and `script.address(…, serverPubKey).encode()`. This is the load-bearing invariant for signing/forfeit correctness — it's right.

**ScriptHex dedup**: The `seen` set is scoped per `discoverAt` invocation (per HD index), which is correct. Avoids duplicate probes when a deprecated key is the same as the current one.

**Fresh snapshot at restore time**: Pulling a new `getInfo()` inside `_runRestore` rather than using `this.offchainTapscript.options.serverPubKey` is the right call — it avoids mixing a stale instance key with fresh server history.

**Tests**: Coverage is thorough — deprecated key discovery, compressed-key normalisation, dedup probe count, spent-boarding blind spot, and gap-window preservation. The `stubInfoWithDeprecated` harness is clean.

---

### 🟡 Notes / minor concerns

**1. Probe count scales as `(1 + n_deprecated) × n_timelocks` per HD index**

Each combination is a separate `getVtxos` HTTP call. With e.g. 3 deprecated signers and 4 timelocks this is 16 probes per index. Not a correctness issue, but worth keeping in mind if the server accumulates many key rotations. A batched `getVtxos({ scripts: allHexes })` call across the full signer × timelock matrix per index would be more efficient — the indexer API appears to accept `scripts[]` already. Could be a follow-up optimisation.

**2. Boarding blind spot is a real (documented) fund-loss path**

> *"Boarding discovery stays current-key only (documented out-of-scope blind spot)"*

If a user funded a boarding UTXO under a deprecated boarding key, it won't be recovered. The PR correctly documents this, and it's probably rare in practice (boarding UTXOs are short-lived). But this is a real remaining gap and should have a Linear ticket to track it explicitly.

**3. Test mock `cutoffDate` is a `number`, not `bigint`**

```ts
{ cutoffDate: 0, pubkey: deprecatedCompressed }
```

`DeprecatedSigner.cutoffDate` is typed as `bigint`. The mock bypasses `ArkProvider.getInfo()`'s `BigInt()` coercion, so this works at runtime in tests but the types are technically wrong. Cosmetic, but worth fixing to keep the test mock in sync with the real interface.

---

### 🔴 Blind spot not covered: deprecated signer key in `signerPubkey` field

The test stub sets `signerPubkey` to the *current* x-only key and puts the deprecated key only in `deprecatedSigners`. But what if a server was on an older firmware that returned a deprecated key's hex in `signerPubkey` (before the `deprecatedSigners` field was introduced) and has since rotated? That scenario is outside the scope of *this* PR's server contract, but worth considering in the protocol docs.

---

### Verdict

*Approach is sound and the fix is correct.* The code is clean, tests are solid, and the threaded-signer invariant is properly maintained across params, address, and script. The noted concerns are either minor or documented out-of-scope.

**Requesting one human review/approval before merge given the VTXO restore / fund recovery scope.**

---
*Reviewed by Arkana 🤖*

ghost (approved):
Approach is correct — deprecated signer threading through params+address is the right invariant. Tests cover the critical paths including compressed-key normalisation. See inline comment for minor notes (probe count, boarding blind-spot tracking, test mock types). ⚠️ Protocol-critical: requires human sign-off before merge.

pietro909:
@coderabbitai please review

ghost (approved):
## Arkana Incremental Review — new commits: restore parallelization + batched discovery

> Reviewing only the delta since my last review: merge of #546 (`restore: parallelize per-index probes and batch candidate queries` + `restore: probe HD indices in gap-limit-capped batches`).

---

### ✅ Parallelization correctness

**Window cap logic** (`contractManager.ts:693`):
```ts
const windowEnd = Math.min(maxIdx, i + Math.min(batchSize, gapLimit - unused) - 1);
```
Verified: `gapLimit - unused` is the maximum consecutive misses remaining before gap closure. Capping the window to this value means every index probed is one the serial path would also reach. The discovered set, persistence order, `lastIndexUsed`, and `handlerErrors` remain byte-for-byte identical to the serial path — only wall-clock latency changes. ✓

**Per-index handler concurrency** (`contractManager.ts:669-683`):
`probeIndex` fires all `discoverAt` calls concurrently via `Promise.all`, but persistence is sequential in the original `discoverables` order (boarding before default/delegate). This preserves the FIRST-WINS collision tie-break for the degenerate same-script scenario. Error isolation per handler is maintained — one failing handler never aborts the others. ✓

**Cross-index concurrency** (`contractManager.ts:696`):
`await Promise.all(windowIndices.map(probeIndex))` fires the entire window concurrently, but results are processed strictly in ascending index order with sequential persistence. Gap counter updates happen in the correct order. ✓

---

### ✅ `detectUsedScripts` (helpers.ts:137-164)

Replaces N per-script `getVtxos` calls with one batched multi-script query. Semantically equivalent: discovery only needs "any VTXO at all?" per candidate script, and the batched call answers that.

**Pagination logic**: Full candidate set passed on every page (consistent indexer pagination). `remaining` set drives early-stop when all candidates are found. No starvation risk — a heavily-reused script can't suppress another candidate because the loop pages through all history if needed.

**`hasMore` heuristic** (`helpers.ts:160`): Matches the established pattern in `fetchContractVtxosBulk`. When `page` metadata is absent (older indexer), defaults to `false` — single-page query — which matches the old non-paginated behavior.

---

### ✅ `constants.ts`

Clean extraction of `DEFAULT_PAGE_SIZE` into a leaf module to break the `handlers → contractManager` import cycle. No functional change.

---

### ✅ `batchSize` API surface

Additive `ScanContractsOptions.batchSize` field with validation. `wallet.restore()` doesn't pass it (defaults to `DEFAULT_SCAN_BATCH = 10`). Backward-compatible — no downstream breakage.

---

### ✅ Tests

Thorough coverage of the new concurrency semantics:
- **Barrier-based concurrency proof** (within-index and cross-index) — the test hangs on serial regression, fast-passes on correct concurrency. Clean pattern.
- **Batch-invariance test** — same gap-and-hit scenario under batchSize ∈ {1, 2, 3, 7, 100} produces identical probe sets and `lastIndexUsed`. This IS the correctness proof for the window cap.
- **Batched query test** — verifies single `getVtxos` call for multi-timelock candidates.

---

### 🟡 One note (non-blocking)

The batched `detectUsedScripts` call for a single wallet index sends ALL candidate scripts in one `getVtxos` request. With many deprecated signers × many timelocks, the `scripts[]` array grows. The indexer's URL-or-body size limit could theoretically be hit — though in practice signer rotations are rare and the candidate count stays small. Worth noting for the future if the deprecated-signer list grows unbounded.

---

### Verdict

The parallelization is clean and provably semantics-preserving. The window cap is the key invariant — it's correct and well-tested. The batched discovery query is a pure latency optimization with no semantic change.

**Still protocol-critical (VTXO restore/fund recovery) — previous human sign-off request stands.**

---
*Reviewed by Arkana 🤖*

PR #546, opened 2026-06-05
restore: parallelize per-index probes and batch candidate queries

pietro909:
## Summary

Performance/robustness follow-up to the HD-wallet restore work. `wallet.restore()`'s gap-limit scan had **two** serial bottlenecks: (1) within each index it awaited each discovery handler one at a time, and `default`/`delegate` additionally issued one `getVtxos` per CSV/signer variant; (2) the outer index loop walked indices strictly one at a time, so an empty wallet paid `gapLimit` sequential round-trips just to close the gap window.

This PR parallelizes on **three** axes — per-index handler probes, per-index candidate queries (batched into one `getVtxos`), and now the **index dimension itself** (windowed batches) — cutting restore latency without changing the discovered result. Mirrors how the reference NArk SDK probes providers in parallel and batches its candidate set per index, and extends it with index batching.

Stacked on #545 (base `deprecated-signer-restore`).

## Changes

- **`ContractManager.scanContracts`** probes all discoverable handlers for an index concurrently via `Promise.all`, then persists their hits **sequentially in `discoverables` order**. The boarding-before-`default`/`delegate` order is the load-bearing first-wins collision tie-break, so only the I/O overlaps — persistence order is unchanged. Each probe keeps its own try/catch, so one handler's failure is still collected into `handlerErrors` rather than aborting the others; `materialize` throws stay fatal.
- **Index batching (new):** the outer gap loop now probes **`batchSize` indices per window concurrently** (default 10), a second concurrency layer over the per-index probes. Each window is **capped to `gapLimit - unused` indices** — the most a serial scan could still reach before the gap window is guaranteed to close. So every index probed in a window is one a one-index-at-a-time scan would also reach: nothing is over-scanned, nothing is discarded, `materialize`/`discoverAt` run on exactly the same index set, and hits are still processed strictly in ascending index order. The discovered set, persisted rows, `lastIndexUsed`, and `handlerErrors` stay **byte-for-byte identical** to the serial path — an empty wallet just closes its gap window in `ceil(gapLimit / batchSize)` rounds instead of `gapLimit` serial ones. Exposed as an optional `batchSize` on `ScanContractsOptions` (validated like `gapLimit`; ignored when `hd` is false).
- **`default.ts` / `delegate.ts`** `discoverAt` build their full deduped (signer × CSV) candidate set, then issue a **single batched probe** via a new `detectUsedScripts` helper instead of one `getVtxos` per variant. The matched signer/CSV is still threaded through script, params, and address.
- **`detectUsedScripts`** (`handlers/helpers.ts`) mirrors `fetchContractVtxosBulk`'s page loop but stops early once every candidate is observed — so the common case is one round-trip, yet a heavily-reused candidate can never starve another off the first page (parity with the old per-script path). Matched back to candidates via the indexer's per-VTXO `script`.

No public API or `Discoverable` interface change; behavior is identical to the serial path (same discovered set, persisted rows, and `lastIndexUsed`) — only faster.

## Tests

- Added: `default` `discoverAt` issues exactly one batched probe covering all candidate scripts (Win 2); `scanContracts` probes handlers concurrently, not serially (Win 1 — fails fast if reverted to serial).
- Added for index batching: rejects a non-positive / non-integer `batchSize`; **cross-index window concurrency** (a single handler's index-0 probe observes a later index in-flight — a serial loop would time out); **batch-invariance** (the swap-hit-at-4 scenario run at `batchSize ∈ {1,2,3,7,100}` yields the identical probed set `0..9` and `lastIndexUsed=4`, proving `batchSize` is a pure latency knob with no effect on the result).
- The load-bearing first-wins ordering test and the equal-delay / deprecated-signer / spent-boarding restore tests remain green. The existing `[0..9]` exact-probe-sequence test (gap-limit window) also stays green, pinning identical behavior under batching.
- `tsc --noEmit` clean; prettier clean (both packages). Full unit suite passes excluding the pre-existing `migrations.test.ts` failures (missing `better-sqlite3` native bindings locally — fail on the base branch too).

### DIFF
--- a/packages/ts-sdk/src/contracts/constants.ts
+++ b/packages/ts-sdk/src/contracts/constants.ts
@@ -1,0 +1,14 @@
+/**
+ * Default indexer pagination size for batched VTXO queries (`getVtxos`).
+ *
+ * Shared by {@link ContractManager}'s bulk history sync and the discovery
+ * handlers' batched `detectUsedScripts` probe so the two can't drift. It lives
+ * in its own leaf module (importing nothing) so both the manager and the
+ * handler layer can use it without a `handlers → contractManager` import cycle —
+ * contractManager imports the handler registry, so the handlers must not import
+ * back from contractManager.
+ *
+ * Large enough that a single wallet index's candidate scripts resolve in one
+ * page in the common case.
+ */
+export const DEFAULT_PAGE_SIZE = 500;
--- a/packages/ts-sdk/src/contracts/contractManager.ts
+++ b/packages/ts-sdk/src/contracts/contractManager.ts
@@ -8,7 +8,6 @@ import {
     ContractState,
     ContractVtxo,
     ContractWithVtxos,
-    DiscoveredContract,
     DiscoveryDeps,
     GetContractsFilter,
     PathContext,
@@ -33,8 +32,7 @@ import {
     saveVtxosForContract,
     warnAndFilterVtxosForScript,
 } from "./vtxoOwnership";
-
-const DEFAULT_PAGE_SIZE = 500;
+import { DEFAULT_PAGE_SIZE } from "./constants";
 
 /**
  * Whether two *different* contract types may legitimately share a single
@@ -585,6 +583,9 @@ export class ContractManager implements IContractManager {
      *   other handler hit it).
      * - `persistAndWatchContract` rejecting is operational/fatal and
      *   propagates (only `discoverAt` is guarded).
+     * - Within an index the handler probes run concurrently (independent
+     *   network reads); their hits are persisted sequentially in
+     *   `discoverables` order to preserve the first-wins collision tie-break.
      */
     async scanContracts(opts: ScanContractsOptions): Promise<ScanResult> {
         const gapLimit = opts.gapLimit ?? 20;
@@ -599,9 +600,10 @@ export class ContractManager implements IContractManager {
             .filter(isDiscoverable);
 
         // Probe `boarding` before `default`/`delegate`. This ordering is
-        // LOAD-BEARING, not cosmetic: each hit is persisted immediately and
-        // `upsertContract` resolves a same-script collision FIRST-WINS, so the
-        // iteration order IS the tie-break. In the degenerate equal-delay case
+        // LOAD-BEARING, not cosmetic: within each index the probes run
+        // concurrently, but their hits are persisted in THIS order and
+        // `upsertContract` resolves a same-script collision FIRST-WINS, so this
+        // order IS the persistence tie-break. In the degenerate equal-delay case
         // a rotated index can carry BOTH an on-chain boarding UTXO and an L2
         // VTXO at the same (byte-identical) script; probing boarding first
         // resolves that collision to a `boarding` row, which keeps the on-chain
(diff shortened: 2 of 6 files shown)


ghost (approved):
## Review: restore — parallelize per-index probes and batch candidate queries

**Verdict: Approve** — clean performance optimization, no behavioral change, correctness holds.

Reviewed: full diff, all five changed files in context, cross-repo consumer search, git history of the base branch, pagination mechanics of `RestIndexerProvider.getVtxos`, and the reference `fetchContractVtxosBulk` loop.

---

### What this does right

1. **First-wins ordering preserved.** `Promise.all` overlaps the I/O but persistence iterates `probes[h]` in the original `discoverables` order (contractManager.ts:~660-670). The boarding-before-default/delegate tie-break is maintained. The concurrent-barrier test (restore.test.ts) directly asserts this.

2. **Error isolation preserved.** Each probe's `try/catch` captures handler failures independently into `handlerErrors`, matching the old serial guard. A failing handler no longer blocks others — strictly better.

3. **`detectUsedScripts` pagination is correct.** `remaining.delete(vtxo.script)` returns `true` only on first encounter → no duplicate adds to `used`, and `remaining.size` accurately tracks unseen candidates. Early exit when `remaining` is empty is safe: every candidate has been observed at least once, so further pages can only yield repeats.

4. **Candidate dedup in default.ts / delegate.ts is correct.** The `seen` Set suppresses duplicate scripts from non-rotating deprecated signers before they enter the candidate array, preventing redundant indexer probes and duplicate `DiscoveredContract` emissions. Current signer wins attribution since it's iterated first.

5. **No public API change.** `Discoverable`, `DiscoveryDeps`, `scanContracts` signature — all unchanged. No cross-repo breakage. Confirmed: no downstream consumer in go-sdk, dotnet-sdk, coinflip, banco, or demos imports any of the changed internals.

6. **Tests are meaningful.** The batched-probe test asserts exactly one `getVtxos` call covers both CSV variants. The concurrent-barrier test proves overlapping execution (first handler observes second in-flight before returning).

---

### Nits / non-blocking observations

**1. Missing inter-page delay in `detectUsedScripts` (helpers.ts:~165)**
`fetchContractVtxosBulk` (contractManager.ts:~1008) adds a 500ms backoff between pages:
```ts
if (hasMore) await new Promise((r) => setTimeout(r, 500));
```
`detectUsedScripts` omits this. In practice the pagination loop almost never fires (a wallet index rarely has 500+ VTXOs across its candidate scripts), so this is non-blocking. But for consistency and defensive rate-limit hygiene, consider adding the same delay — or extracting a shared `paginateVtxos` helper to avoid the two loops drifting apart.

**2. `DISCOVERY_PAGE_SIZE` duplicates `DEFAULT_PAGE_SIZE` (helpers.ts:~127 vs contractManager.ts:~37)**
Both are `500`. The comment explains the import-cycle rationale, which is fair. A shared constants module or re-export from `types.ts` would eliminate the duplication without introducing the cycle, but this is a style nit.

**3. Concurrent-barrier test timeout (restore.test.ts:~790)**
The 1-second `setTimeout` fallback prevents hangs if concurrency regresses, which is good. But under CI load this could be tight. If flakiness surfaces, bumping to 3-5s or using a `Promise.race` with `AbortController` would be more robust.

**4. No test for the `detectUsedScripts` pagination path**
The mock indexer in the batched-probe test returns all results in one page (no `page` metadata). A test that returns a short first page (simulating pagination) and verifies the loop fetches page 2 would harden the helper against regressions in the pagination logic.

---

### Protocol-critical assessment

This PR touches VTXO **discovery** (read path), not VTXO creation, signing, forfeit, or exit. The discovered set is identical to the serial path — same scripts, same `DiscoveredContract` shapes, same persistence order. No fund-loss vector introduced.

LGTM. Nice optimization — well-structured, well-tested, well-documented.

🤖 Reviewed by Arkana (Claude Opus)

Kukks (changes requested):
We need to also make this work in batches of indexes not one index at a time as well

pietro909:
@Kukks 
> We need to also make this work in batches of indexes not one index at a time as well

Done in 0464bea8. The outer gap loop now probes `batchSize` indices per window concurrently (default 10), layered over the existing per-index handler concurrency.

I picked 10 as tradeoff between latency improvement and concurrent queries: an empty wallet closes its gap window in **2 instead of 20 rounds**.

pietro909:
@coderabbitai please review

PR #547, opened 2026-06-05
fix(script): build Taproot script tree with btcd's algorithm

Kukks:
## Summary

`VtxoScript` currently uses scure-btc-signer's `taprootListToTree` to assemble the Taproot script tree. That function builds a **Huffman tree** (weight-1 leaves combine by smallest-weight pairs).

arkd validates spends via btcd's `txscript.AssembleTaprootScriptTree` (`btcd/txscript/taproot.go:623`), which uses a **completely different algorithm**:

- **Phase 1** — pair leaves left-to-right. An odd leaf at the end merges into the LAST branch built so far (does NOT pair with a fresh leaf).
- **Phase 2** — FIFO-queue merge. Take the front two branches, combine, push to back. Repeat until one branch remains.

For **power-of-2** leaf counts the two algorithms happen to produce the same perfectly-balanced binary tree, so the SDK has shipped without hitting this. For **any other count** they produce different shapes → different merkle roots → different taproot output keys → arkd rejects spends with `INVALID_PSBT_INPUT: expected <tap key from prevout>, got <tap key computed from tapscripts>`.

The previous workaround in `VtxoScript` constructor (`scripts.length % 2 !== 0 ? scripts.slice().reverse() : scripts`) only handled odd-vs-even and accidentally produced the arkd-expected key for the small odd counts in the test fixtures. It does not reproduce btcd's algorithm and breaks for general odd or non-power counts.

## What this PR does

- Adds `packages/ts-sdk/src/script/taprootTree.ts` exporting `assembleBtcdTaprootTree(scripts)`, a TS implementation of btcd's algorithm that returns the nested-tuple form scure-btc-signer's `p2tr` accepts directly.
- Updates `VtxoScript` constructor to use `assembleBtcdTaprootTree` and removes the odd-count reverse workaround.
- Exports `assembleBtcdTaprootTree` from the package barrel so consumers building taptrees by hand (custom contract handlers, etc.) get the same arkd-compatible behavior.

## Verified against arkd source

[`pkg/ark-lib/script/vtxo_script.go:226-244`](https://github.com/arkade-os/arkd/blob/master/pkg/ark-lib/script/vtxo_script.go) calls `txscript.AssembleTaprootScriptTree`, defined in [btcd `txscript/taproot.go:623`](https://github.com/btcsuite/btcd/blob/master/txscript/taproot.go). I read both and reproduced the algorithm exactly.

## Test plan

- [x] Existing `vtxoscript.json` fixture tests (6-leaf and 7-leaf with arkd-derived taproot keys) pass **unchanged** — the new algorithm produces the same arkd-expected keys.
- [x] New `packages/ts-sdk/test/taprootTree.test.ts`: leaf counts 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 15, 16, 17, 32 all build valid trees + encode→decode round-trip preserves the tap key.
- [x] Full unit suite: **1357/1357 pass** (1 skipped, e2e tests requiring regtest infra not exercised here).
- [x] Real-world validation: in [ArkLabsHQ/coinflip#32](https://github.com/ArkLabsHQ/coinflip/pull/32) a 10-leaf v0.3 coinflip escrow now passes regtest end-to-end against arkd v0.9.5 + emulator v0.0.1, where previously it failed with `INVALID_PSBT_INPUT: expected X, got Y` until this fix landed in the consuming app's vendored SDK.

## Summary by CodeRabbit

* **New Features**
  * SDK now includes btcd-compatible Taproot script tree assembly and uses it for Taproot script handling, improving compatibility and consistency.

* **Tests**
  * Added comprehensive tests validating Taproot tree construction across many leaf counts and serialization round-trips.
  * Added fixture data covering additional Taproot scenarios.


packages/ts-sdk/src/script/taprootTree.ts
```diff
@@ -0,0 +1,98 @@
...
+    leafVersion: number;
+}
+
+/**
+ * Internal tree node shape consumed by `@scure/btc-signer`'s `p2tr`:
+ *   - A leaf is `{ script, leafVersion }`
+ *   - A branch is a 2-element tuple `[leftNode, rightNode]` of nodes
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type TaprootTreeNode = TaprootLeaf | any[];
```
pietro909: A recursive type seems more appropriate:

```typescript
type TaprootTreeNode = TaprootLeaf | [TaprootTreeNode, TaprootTreeNode]
```

Get rid of unsafe `any` 😄

packages/ts-sdk/test/taprootTree.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    it("throws on empty input", () => {
+        expect(() => assembleBtcdTaprootTree([])).toThrow();
+    });
+
+    it("VtxoScript encode→decode round-trips the tap key for any leaf count", () => {
+        for (const count of [3, 5, 7, 9, 10, 11, 13]) {
+            const scripts = makeScripts(count);
+            const original = new VtxoScript(scripts);
+            const roundtripped = VtxoScript.decode(original.encode());
```
pietro909, line 60: Both original and roundtripped compute their key by running the same tree builder in the VtxoScript constructor. If that builder is wrong, both sides are wrong the same way, right?

pietro909 (changes requested):
The current fixture layer only covers 7 leaves and 6 leaves in `packages/ts-sdk/test/fixtures/vtxoscript.json`. 

For those exact fixture scripts, both the current FIFO Phase 2 and the wrong level-by-level Phase 2 produce the same taproot key, so `packages/ts-sdk/test/tapscript.test.ts:213` would not catch that regression.

The first divergent count should be at 10.

ghost (approved):
Reviewed. This is clean and correct. Pietro's two prior concerns are fully addressed in the latest commit:

**Recursive `TaprootTreeNode` type** ✅ — `TaprootTreeNode = TaprootLeaf | [TaprootTreeNode, TaprootTreeNode]`, no `any`.

**Golden vectors for non-power-of-2 counts** ✅ — `vtxoscript.json` now has fixtures for 3, 5, 9, 10, 11, 13, 15, 17 leaves, all with tap keys independently generated via btcd's `AssembleTaprootScriptTree` + `ComputeTaprootOutputKey`. The test at `tapscript.test.ts:213` runs every one of them through `new VtxoScript(scripts)` and asserts the key — so the correctness guarantee is real, not circular.

**Algorithm fidelity** — I traced the implementation against btcd's source. Phase 1 (pair left-to-right, odd tail merges into last branch) and Phase 2 (FIFO queue) are reproduced exactly. Single-leaf and empty edge cases are handled correctly. The `shift()!` assertions in Phase 2 are safe because Phase 1 always leaves ≥ 1 branch for ≥ 2 leaves, and the `while (branches.length >= 2)` guard prevents calling shift below 2 elements.

**Round-trip test** — The concern is honestly documented in the test comment: the round-trip only asserts script set survival, not key correctness. Key correctness is covered by the golden vectors. This is the right separation.

⚠️ *Protocol-adjacent note:* This touches `VtxoScript` constructor which determines the taproot output key used for VTXO spends. The fix is well-tested with btcd-derived golden vectors and regtest-verified via coinflip#32, but a human maintainer should give final sign-off before merge per team protocol.

LGTM from me.

pietro909 (approved):

PR #549, opened 2026-06-08
Limit settlement VTXO batches to 50

pietro909:
## Summary

Limit automatically selected settlement VTXO batches to `MAX_VTXOS_PER_SETTLEMENT = 50` so wallet renewal and settlement paths avoid arkd `TX_TOO_LARGE` failures when a wallet has many VTXOs.

## Details

- Add a shared `MAX_VTXOS_PER_SETTLEMENT` constant in `VtxoManager`.
- Cap automatically selected VTXOs in:
  - `VtxoManager.recoverVtxos`
  - `VtxoManager.renewVtxos`
  - `VtxoManager.runPeriodicSettle`
  - no-params `Wallet.settle()`
- Preserve selection order for recovery to stay aligned with go-sdk behavior.
- Re-run recovery subdust/dust eligibility on the capped recovery batch so a below-dust prefix is not submitted to the server.
- Leave caller-provided `wallet.settle({ inputs })` uncapped.
- Add regression coverage for recovery, renewal, periodic settle, and no-params wallet settle caps.

## Validation

- `pnpm exec vitest run test/vtxo-manager.test.ts test/wallet.test.ts -t "cap|auto-select|subdust"`
- `pnpm exec prettier --check src/wallet/vtxo-manager.ts src/wallet/wallet.ts test/vtxo-manager.test.ts test/wallet.test.ts`
- `pnpm exec tsc --noEmit`
- Commit hook: `pnpm -r lint` and `pnpm -r test:unit`

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents large-settlement rejections by capping the number of offchain virtual outputs per settlement (50), and ensuring periodic settlements stop once the cap is reached.

* **New Features**
  * Deterministic selection: settles prioritize highest-value and soonest-expiring outputs, skip uneconomic inputs so viable outputs aren’t starved, and defer overflow to later settlements.

* **Tests**
  * Added coverage validating capping, ordering, viability checks, and recomputation of outputs when capping applies.

PR #553, opened 2026-06-09
Cap settlement batches by vtxoMaxAmount

pietro909:
## What

Adds a per-settlement **amount** cap so a settlement's single VTXO output never exceeds the server's `vtxoMaxAmount`. Complements the input-**count** cap from #549 (`MAX_VTXOS_PER_SETTLEMENT`).

## Why

Every auto-selection settlement path builds one output equal to the fee-adjusted sum of its inputs. `arkd` rejects any virtual output above `vtxoMaxAmount` with `AMOUNT_TOO_HIGH` (the check is skipped when the limit is `-1` / no-limit). Consolidating more than the limit in one settlement — recovery of many subdust VTXOs, renewal, full settle — was therefore guaranteed to be rejected by the server.

## How

"Cap & defer", mirroring #549: select inputs until the cumulative output would exceed `vtxoMaxAmount`, settle that subset, and leave the overflow for the next cycle. `-1` disables the cap.

- New `capSettlementBatch(sorted, maxAmount)` helper applies both the count and amount caps to an already-sorted candidate list (value-descending for recovery / full settle, expiry-ascending for renewal). Used by `recoverVtxos` and `renewVtxos`.
- Periodic auto-settle and `Wallet.settle()` no-params get an inline amount check alongside their existing count cap.
- An over-budget candidate is **skipped** (not a stop), so a smaller VTXO behind an oversized or awkwardly-sized one still gets in — e.g. `max 10_000, [12_000, 5_000] → [5_000]` and `[9_000, 4_000, 1_000] → [9_000, 1_000]`. The count cap stays a hard stop.
- Boundary matches the server's strict check: an output exactly equal to `vtxoMaxAmount` is allowed.

## Known limitation

Periodic-settle and `Wallet.settle()` add boarding inputs uncapped, so if boarding alone exceeds `vtxoMaxAmount` the output is still over-limit and the server rejects it. Fully solving that needs multi-output splitting, which this change (like #549) deliberately does not do.

## Testing

- `capSettlementBatch` unit tests: count cap, amount cap, both, `-1`, skip-and-fill, all-oversized → empty
- amount-cap + skip-and-fill coverage for recovery, renewal, periodic settle, and `Wallet.settle()` no-params
- Full ts-sdk unit suite, `tsc --noEmit`, and prettier all pass

## Summary by CodeRabbit

* **New Features**
  * Server-driven per-output amount ceiling enforced for VTXO recovery, renewal, and periodic settlement (can be disabled by server).
  * Oversized VTXOs are skipped while later smaller qualifying VTXOs remain eligible; selection evaluates projected post-fee output when applicable.
  * Wallet caches the receive address/script during auto-settle to prevent mid-settle desync; new user-facing error/warning when no VTXOs fit the server limit.

* **Tests**
  * Expanded tests covering capping behavior, edge cases, uncapped mode, warnings/errors, and packing logic.

### DIFF
--- a/packages/ts-sdk/src/wallet/vtxo-manager.ts
+++ b/packages/ts-sdk/src/wallet/vtxo-manager.ts
@@ -167,6 +167,43 @@ export function byExpiryAscending(vtxos: ExtendedVirtualCoin[]): ExtendedVirtual
     return [...vtxos].sort((a, b) => expiryKey(a) - expiryKey(b));
 }
 
+/**
+ * Select inputs from `sorted` that fit in a single settlement: at most
+ * {@link MAX_VTXOS_PER_SETTLEMENT} inputs AND a cumulative `value` no greater
+ * than `maxAmount`. `maxAmount < 0` disables the amount bound — it is the
+ * server's `-1` "no limit" sentinel for `ArkInfo.vtxoMaxAmount`.
+ *
+ * Each settlement path builds a single output equal to the (fee-adjusted) sum
+ * of its inputs, and the server rejects any virtual output above `vtxoMaxAmount`
+ * with `AMOUNT_TOO_HIGH`. Capping the input total therefore keeps that output
+ * within bounds; the overflow is settled on the next cycle, mirroring the
+ * count-cap behaviour.
+ *
+ * `sorted` must already be ordered by the caller's priority (value-descending
+ * for recovery / manual full settle, expiry-ascending for renewal) so the
+ * inputs that matter most are tried first. An input that would breach the
+ * amount cap is skipped — not a stopping point — so a smaller input behind an
+ * oversized or awkwardly-sized one still gets in (a break would strand it and
+ * could leave the batch below dust). The count cap is a hard stop. Uses
+ * `> maxAmount` to mirror the server's strict check, so a batch whose total
+ * equals the limit still fits.
+ */
+export function capSettlementBatch<T extends { value: number }>(
+    sorted: T[],
+    maxAmount: bigint,
+): T[] {
+    const batch: T[] = [];
+    let total = 0n;
+    for (const vtxo of sorted) {
+        if (batch.length >= MAX_VTXOS_PER_SETTLEMENT) break;
+        const next = total + BigInt(vtxo.value);
+        if (maxAmount >= 0n && next > maxAmount) continue;
+        batch.push(vtxo);
+        total = next;
+    }
+    return batch;
+}
+
 /** Default renewal threshold in seconds (3 days). */
 export const DEFAULT_THRESHOLD_SECONDS = 259_200;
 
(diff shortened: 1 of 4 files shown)


ghost (changes requested):
## 🔒 Protocol-Critical — Requires Human Review

This PR modifies VTXO settlement batch selection (recovery, renewal, periodic settle, and manual settle paths). Changes to how VTXOs are grouped for settlement are protocol-critical: incorrect batching could strand funds, cause silent failures, or produce server-rejected intents. Requesting human sign-off before merge.

---

## Overall Assessment

Solid, well-structured PR. The "cap & defer / skip-and-fill" design is sound — capping conservatively and deferring overflow to the next cycle is the right call for a settlement pipeline. Good test coverage on the core helper and all four affected paths. I found no correctness bugs that would lose funds.

## Findings

### 1. Gross vs. net amount cap inconsistency (Low — conservative, not broken)

`capSettlementBatch` (used in **recovery** + **renewal**) caps on `sum(gross_values)`:

```typescript
// vtxo-manager.ts:196 (capSettlementBatch)
const next = total + BigInt(vtxo.value);       // gross
if (maxAmount >= 0n && next > maxAmount) continue;
```

But the **periodic-settle** and **`Wallet.settle()` no-params** inline loops cap on `sum(net_values)`:

```typescript
// vtxo-manager.ts:1635 (runPeriodicSettle)
const net = BigInt(v.value) - BigInt(inputFee.satoshis);  // net
if (vtxoMaxAmount >= 0n && totalAmount + net > vtxoMaxAmount) continue;
```

Both are safe — the recovery/renewal paths are strictly more conservative, settling fewer VTXOs per cycle than they could when intent fees are material. But the asymmetry is worth a comment in `capSettlementBatch` explaining that gross-value capping is intentional so a future reader doesn't "fix" it by adding fee logic.

Not a blocking issue.

### 2. Silent skip of oversized urgent VTXOs in renewal (Low — correct but opaque)

**vtxo-manager.ts, renewal path:** When a VTXO's value alone exceeds `vtxoMaxAmount`, it's silently skipped via `capSettlementBatch`'s `continue`. If it's the soonest-expiring VTXO, the user gets no signal that their most urgent VTXO is being left to expire — only when *all* VTXOs exceed the limit does the code throw.

The behavior is correct (the server would reject it anyway), but a `console.warn` when skipping an oversized VTXO in the renewal path would help operators diagnose stuck wallets that drift toward unilateral exit. Consider logging after the `capSettlementBatch` call when `capped` excludes VTXOs that were in the input set:

```typescript
const skipped = vtxos.length - capped.length;
if (skipped > 0) {
    console.warn(`Renewal: ${skipped} VTXO(s) skipped — value exceeds vtxoMaxAmount ${vtxoMaxAmount}`);
}
```

Not a blocking issue.

### 3. `getInfoProvider()` type escape hatch (Nit)

**vtxo-manager.ts:1207:**

```typescript
private getInfoProvider(): ArkProvider | undefined {
    return (this.wallet as Partial<SweepCapableWallet>).arkProvider;
}
```

The `as Partial<SweepCapableWallet>` cast bypasses the type system to reach `arkProvider` without requiring full sweep capability. The call sites handle `undefined` correctly via `?.getInfo()` → `?? -1n`. This works but is fragile — if `IWallet` ever gains a conflicting `arkProvider` property, the cast silently hides the mismatch. A targeted type guard (`'arkProvider' in this.wallet`) or a narrow interface (`{ arkProvider?: ArkProvider }`) would be safer.

Not a blocking issue.

### 4. Test coverage ✅

Thorough. `capSettlementBatch` unit tests cover: count cap, amount cap, both caps, `-1n` sentinel, skip-and-fill, all-oversized → empty. Integration tests cover recovery, renewal, periodic settle, and wallet settle with amount caps. Edge cases (exact boundary match, skip leading oversized candidate) are explicitly tested.

## Confirmed Correct

- **Boundary match**: `> maxAmount` (not `>=`) mirrors the server's strict check — an output exactly equal to `vtxoMaxAmount` is accepted. ✓
- **`-1n` sentinel**: `maxAmount >= 0n` correctly disables the cap for any negative value. ✓
- **Count cap remains a hard `break`**, amount cap is a `continue` (skip). Consistent across `capSettlementBatch` and inline loops. ✓
- **Boarding inputs uncapped**: Documented limitation — boarding alone exceeding `vtxoMaxAmount` still produces an over-limit output. Multi-output splitting is out of scope and explicitly called out. ✓
- **`getInfoProvider()` fallback**: `undefined?.getInfo()` → `undefined` → `?? -1n` → no limit. Correct for wallets without an ark provider. ✓

## Verdict

Code is correct and well-tested. **Requesting changes only to enforce human review on this protocol-critical path.** The findings above are low-severity suggestions, not blockers.

A human maintainer should verify the skip-and-fill semantics match the server's `AMOUNT_TOO_HIGH` rejection logic and confirm the gross-vs-net inconsistency in `capSettlementBatch` is acceptable for recovery/renewal batching.

PR #554, opened 2026-06-10
feat: arkd signer-rotation support (deprecated-signer migration + Boltz VHTLC recovery)

pietro909:
Client-side support for a planned arkd operator signing-key rotation
(arkade-os/arkd#822). When arkd rotates its signer, existing VTXOs/swaps stay
locked to the old signer; this lets the SDK discover, classify, and
cooperatively migrate those funds to the active signer before their cutoff,
and recover Boltz swaps that straddle a rotation.

## SDK (`@arkade-os/sdk`)

- **Parse deprecated signers** — `DeprecatedSigner.cutoffDate` is now optional;
  `RestArkProvider.getInfo()` preserves a missing cutoff as `undefined` instead
  of coercing to `0n` (breaking change to the public type).
- **Classification helpers** (`src/wallet/signerRotation.ts`) — pure functions
  classifying a contract's signer against a fresh `ArkInfo` into
  `current | migratable | dueNow | expired | unknownSigner`, derived at read
  time (no persisted stale-key metadata).
- **Mid-session rotation** — `Wallet.rotateServerSigner()` rebuilds the offchain
  and boarding tapscripts under the new signer, registers the new contract rows,
  then commits; `arkServerPublicKey` becomes a getter with a sanctioned
  rotation-only write path. Serialized against HD receive rotation.
- **Migration API** — `VtxoManager.migrateDeprecatedSignerVtxos()` +
  `getDeprecatedSignerStatus()` (on `IVtxoManager`): rotates first when the
  wallet's own snapshot signer is deprecated, selects spendable VTXOs under
  deprecated-signer contracts ordered cutoff-first, settles to the rotated
  address, and returns a machine-readable report.
- **Opt-out** — `SettlementConfig.deprecatedSignerMigration` (default `true`);
  an automatic migration pass runs in the poll loop with its own
  cooldown/backoff.
- **Service-worker / Expo parity** — new `MIGRATE_DEPRECATED_SIGNER_VTXOS` /
  `GET_DEPRECATED_SIGNER_STATUS` messages + worker handler + client proxy. Expo
  needs no change (it wraps the real `Wallet`).

## Boltz (`@arkade-os/boltz-swap`)

- VHTLC recovery paths (`claimVHTLC`, submarine refund/lookup/diagnose,
  `refundArk`, `claimArk`) now reconstruct the VHTLC across current + deprecated
  signers and match the **stored lockup address**, threading the matched key
  downstream. No swap-schema change; creation-time `verifyChainSwap` stays on
  the current signer.

## Out of scope / follow-ups

- ~~Boarding UTXOs minted under an old signer (recovery path).~~ addressed in b33d63d
- The post-cutoff unilateral-exit guidance API.
- The server-side half: cooperative claim/refund still needs arkd to co-sign
  with the deprecated key (arkade-os/arkd#822). SDK-side reconstruction/lookup
  and the unilateral paths work today.

## Testing

Unit tests only — integration is blocked on a regtest fixture that can rotate
signers and ultimately on arkd#822. Full suite green:
`@arkade-os/sdk` 1425 passing, `@arkade-os/boltz-swap` 388 passing; both
typecheck and prettier clean.

---

## Added in this branch (boarding watch + mid-session detection)

Beyond the scope above, this branch also:

- **Fixes the on-chain boarding watch across signer rotation** (carries closed #560). The deprecated-signer set is cached on the wallet, and the *watch/history* read paths — `getBoardingAddresses` (→ `notifyIncomingFunds` → `watchAddresses`) and `getBoardingTxs` — fan out over `current ∪ deprecated` signers, so deposits to a boarding address minted under a now-rotated operator signer keep being watched. The *spend* path (`getBoardingUtxos` → `settle`) deliberately stays current-signer-only; deprecated-signer boarding recovery remains the migration API's job.
- **Restores deprecated-signer contracts in the index-0 baseline at boot.** `initializeContractManager`'s always-active baseline matrix now fans the `default`, `delegate`, *and* `boarding` contracts over `current ∪ deprecated` signers (deduped by script — a boarding script that coincides with a default/delegate one stays first-wins as that type), so a wallet loaded against a fresh repo registers — and therefore watches and surfaces — funds minted under a now-rotated operator signer *without* an explicit `restore()`. Previously the baseline was current-signer-only, so such funds (e.g. a delegate VTXO under a now-deprecated signer) went undetected at load. Cutoff-agnostic: `EXPIRED` signers are registered too (so the funds surface — under `pendingRecovery` per above).
- **Adds mid-session signer-rotation detection via the GetInfo digest.** `RestArkProvider` caches the `digest`, sends it as `X-Digest`, and on a *structured* `DIGEST_MISMATCH` error — matched via the shared `maybeArkError` on arkd's `ark.v1.ErrorDetails.name`, not a raw substring — refreshes info, fires `onServerInfoChanged`, then throws the now-publicly-exported `DigestMismatchError` (no silent retry). The wallet subscribes and, on the event, refreshes its deprecated-signer set and calls `rotateServerSigner` onto the new key; `dispose()` drains an in-flight handler before teardown. arkd **v0.9.9-rc.1** now ships the matching server-side guard (`x-digest` header / `DIGEST_MISMATCH` error, arkd #1104), opt-in via `digest_header_required` (default off) — so this is the instant, event-driven trigger for the deprecated-signer migration the moment an operator enables it (this path was previously dormant pending server support).
- **Tooling:** `.prettierrc` now sets `endOfLine: "auto"` so `prettier --check` / the pre-commit hook pass on CRLF (Windows) working trees while staying a no-op on LF/CI.
- **Excludes post-cutoff funds from the spendable balance + coin selection.** VTXOs under a deprecated signer whose cutoff has passed (`EXPIRED`) and not yet swept can't be cooperatively spent (the operator won't co-sign the old key) until they recover. They now move out of `getBalance().available`/`settled`/`preconfirmed` into a new `WalletBalance.pendingRecovery` bucket (still counted in `total`) and are dropped from send coin selection, so a spend never selects an uncosignable VTXO. Offline-classified from a cached signer set (now with cutoffs); the service-worker `getBalance` mirrors it.
- **Scopes the `X-Build-Version` header to arkd only (CORS).** That arkd-specific compatibility header was being sent on every request through the shared `fetch` wrapper, so the delegate / Esplora / indexer providers (different origins) failed CORS preflight (`Request header field x-build-version is not allowed`). A new header-less `baseFetch` now serves those providers; only `RestArkProvider` keeps the `X-Build-Version` / `X-Digest` headers.

## Summary by CodeRabbit

* **New Features**
  * Wallets: mid-session server-signer rotation, boarding grouping per signer, and public APIs to run/query deprecated-signer migration.
  * SDK: signer-rotation classification utilities and related exports.

* **Bug Fixes**
  * Reverse-swap invoice preimage validated on creation.
  * VHTLC/address reconstruction handles rotated (deprecated) server signers; failures yield descriptive errors.

* **Tests**
  * Expanded unit, integration, and e2e coverage for signer-rotation, deprecated-signer migration, and VHTLC reconstruction.

* **Chores**
  * Service-worker messaging added for migration/status requests and streamed migration events.


ghost (commented):
## Arkana Code Review — PR #554

**Reviewer:** Arkana (automated, aggressive review)  
**Verdict:** Request changes — **protocol-critical, requires human sign-off**

---

### 🔴 PROTOCOL-CRITICAL: Human Review Required

This PR touches VTXO lifecycle (migration settle), transaction signing paths (VHTLC reconstruction across deprecated signers), server-signer key management, and address derivation (mid-session `rotateServerSigner`). Per Ark protocol review policy, **a human must approve this PR regardless of code quality** — bugs here lose real money.

---

### Architecture Assessment

The overall design is **sound**. The key architectural decisions are correct:

1. **Derive staleness at read time** (never persisted) — avoids stale-metadata bugs.
2. **Rotate-then-migrate ordering** — ensures migration output commits to the active signer.
3. **Serialization of rotation against HD receive rotation** via `runExclusive` / `_serverRotationChain` — prevents tapscript tearing.
4. **Address-match loop for VHTLC recovery** — correct approach for resolving swaps that straddle a rotation.
5. **Persistence-before-commit pattern** in `_doRotateServerSigner` — mirrors `WalletReceiveRotator.rotate()` ordering; wallet never displays an unwatched address.

---

### Findings

#### 1. 🟡 Dropped `claimScript`/`refundScript` null assertions after VHTLC resolution

**Files:** `packages/boltz-swap/src/arkade-swaps.ts` (multiple call sites)

The old code had explicit guards:
```ts
if (!vhtlcScript.claimScript)
    throw new Error(`Swap ${id}: failed to create VHTLC script for reverse swap`);
```

The new `resolveVHTLCForLockup` relies on address-match as a proxy for script completeness. This is *logically correct* (address is derived from the taproot tree including claim/refund leaves), but the null guards were cheap defense-in-depth. If `createVHTLCScript` ever changes to produce a valid address with a null leaf (e.g., optional claim paths in a future VHTLC version), downstream code (`claimVHTLCwithOffchainTx`, refund signing) would fail with an opaque error instead of the clear "failed to create VHTLC script" message.

**Suggestion:** Add a post-resolution assertion in `resolveVHTLCForLockup` before returning:
```ts
if (!vhtlcScript.claimScript && !vhtlcScript.refundScript)
    throw new Error(`Swap ${args.swapId}: VHTLC address matched but script leaves are empty`);
```

This preserves the defense-in-depth at the single resolution point rather than at every call site.

#### 2. 🟡 `wallet.ts:3000` — address derivation reads `this.arkServerPublicKey` outside the snapshot

**File:** `packages/ts-sdk/src/wallet/wallet.ts:2999-3000` (pre-existing, but newly relevant)

```ts
const offchainTapscript = this.offchainTapscript;
const outputAddress = offchainTapscript.address(this.network.hrp, this.arkServerPublicKey);
```

With `arkServerPublicKey` now a mutable getter (backed by `_arkServerPublicKey`), this relies on JS single-threaded execution to guarantee the tapscript and server key are from the same rotation epoch. This is **currently safe** because `_doRotateServerSigner` sets all three fields synchronously, and no yield point exists between lines 2999-3000.

However, this is fragile: any future refactor that introduces an `await` between the two reads, or that makes `_doRotateServerSigner` async between the field swaps, would create a silent inconsistency (old tapscript + new server key → wrong change address → **lost funds**).

**Suggestion:** Snapshot the server key alongside the tapscript:
```ts
const offchainTapscript = this.offchainTapscript;
const serverPubKey = this.arkServerPublicKey; // pin to same epoch
const outputAddress = offchainTapscript.address(this.network.hrp, serverPubKey);
```

One-line addition that makes the safety invariant explicit. Same applies at `wallet.ts:2017-2020` (the `settle` path).

#### 3. 🟢 `candidateServerPubkeys` — clean and correct

**File:** `packages/boltz-swap/src/utils/vhtlc.ts:289-301`

Deduplication via `Set`, current-first ordering, empty-pubkey skip. No issues.

#### 4. 🟢 `signerRotation.ts` classification — correct edge cases

- `secondsUntilCutoff <= 0` at the exact cutoff instant → `expired`. Correct: window is **closed** at the boundary.
- `Number(cutoffDate)`: safe; Unix timestamps in seconds are well within `Number.MAX_SAFE_INTEGER`.
- `isCooperativelyMigratable`: correctly gates on `migratable | dueNow` only.

#### 5. 🟢 `DeprecatedSigner.cutoffDate` optional change — backward-compatible

**File:** `packages/ts-sdk/src/providers/ark.ts:503-512`

Breaking type change (`bigint` → `bigint | undefined`), but cross-repo search confirms no downstream consumer accesses `.cutoffDate` without null checking. The `RestArkProvider.getInfo()` parsing (`signer.cutoffDate != null`) correctly distinguishes missing-field from explicit-zero.

#### 6. 🟢 Migration fee calculation — mirrors `runPeriodicSettle`

**File:** `packages/ts-sdk/src/wallet/vtxo-manager.ts:1380-1420`

Fee estimation, dust-skipping, and output fee subtraction all follow the established pattern. The `below-dust` early exit prevents submitting uneconomical intents.

#### 7. 🟢 Migration error handling — correct non-throwing pattern

`migrateCore` catches settle errors and reports them in `DeprecatedSignerMigrationReport.error`. The `runMigrationPass` wrapper adds exponential backoff (30s base, 5min cap). Prevents poll-loop poisoning.

#### 8. 🟢 Service worker serialization — correct bigint transport

Wire types correctly stringify bigints. `MIGRATE_DEPRECATED_SIGNER_VTXOS` correctly added to the streaming set (line 830) so the bus deadline doesn't race the migration settle.

#### 9. 🟢 `WalletReceiveRotator.runExclusive` — correct serialization primitive

Returns the result, chains errors correctly. Matches the existing pattern.

#### 10. 🟢 Test coverage — comprehensive for unit tests

Good coverage of classification, ordering, exclusion, rotation, short-circuit, error reporting, opt-out, and service worker serialization. The `resolveVHTLCForLockup` test uses real keys and address comparison — not mocked.

---

### Cross-Repo Impact

**No downstream breakage found.** Searched all repos under the org for consumers of `DeprecatedSigner`, `IVtxoManager`, `arkServerPublicKey`, and `SettlementConfig`. No external `IVtxoManager` implementations exist. The field-to-getter change is backward-compatible for read-only access.

---

### Summary

| # | Severity | Finding | Location |
|---|----------|---------|----------|
| 1 | 🟡 Medium | Dropped claim/refund null assertions after VHTLC resolution | `arkade-swaps.ts` (multiple) |
| 2 | 🟡 Medium | Server-key snapshot not pinned alongside tapscript snapshot | `wallet.ts:3000`, `wallet.ts:2018` |
| 3–10 | 🟢 Clean | Architecture, classification, fees, errors, tests | — |

**Action items before merge:**
1. Add post-resolution script-leaf assertion in `resolveVHTLCForLockup` (finding 1)
2. Pin `arkServerPublicKey` snapshot alongside `offchainTapscript` snapshot in `_sendImpl` and `settle` (finding 2)
3. **Human sign-off required** — protocol-critical changes to VTXO lifecycle, signing, and address derivation

pietro909:
@coderabbitai full review

packages/ts-sdk/src/wallet/vtxo-manager.ts
```diff
@@ -462,6 +491,210 @@ export interface RenewVtxosOptions {
...
+        cutoffDate: c.classification.cutoffDate,
+    };
+}
+
+/**
+ * Ordering key for migration urgency: `dueNow` (no advertised cutoff, migrate
+ * immediately) sorts most urgent, then the soonest advertised cutoff. Smaller
+ * is more urgent.
+ */
+function urgencyKey(classification: SignerClassification): number {
```
pietro909: this is redundant: candidates should fall into the same bucket

packages/ts-sdk/src/wallet/vtxo-manager.ts
```diff
@@ -1116,6 +1386,437 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
...
+                classified,
+            })),
+        ];
+        candidates.sort((a, b) => {
+            const ck =
+                urgencyKey(candidateClassification(a)) - urgencyKey(candidateClassification(b));
+            if (ck !== 0) return ck;
+            return candidateValue(b) - candidateValue(a);
+        });
+        const capped = candidates.slice(0, MAX_VTXOS_PER_SETTLEMENT);
```
pietro909: This is done already at `packages/ts-sdk/src/wallet/wallet.ts:2442`

packages/ts-sdk/src/wallet/vtxo-manager.ts
```diff
@@ -1116,6 +1386,437 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
...
+        const deferred = candidates.length - capped.length;
+
+        // Price each input/output so the migration intent carries the operator's
+        // intent fees (mirrors plain settle). VTXOs use the offchain input fee;
+        // boarding UTXOs use the onchain input fee with the same `fee >= value`
+        // skip rule plain settle applies. Skip inputs whose fee exceeds value.
+        const dustAmount = getDustAmount(this.wallet);
+        const estimator = new Estimator(info.fees.intentFee);
```
pietro909: A forced migration shouldn't pay fees?

packages/ts-sdk/src/wallet/vtxo-manager.ts
```diff
@@ -1116,6 +1386,437 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
...
+                const inputFee = estimator.evalOffchainInput({
+                    amount: BigInt(v.value),
+                    type: "vtxo",
+                    weight: 0,
+                    birth: v.createdAt,
+                    expiry: v.virtualStatus.batchExpiry
+                        ? new Date(v.virtualStatus.batchExpiry)
+                        : undefined,
+                });
+                if (inputFee.satoshis >= v.value) continue;
```
pietro909: We shouldn't leave VTXOs back for a forced migration

packages/ts-sdk/src/wallet/vtxo-manager.ts
```diff
@@ -1116,6 +1386,437 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
...
+                inputs.push(v);
+                migratedRefs.push(classifiedToRef(cand.classified));
+                totalAmount += BigInt(v.value) - BigInt(inputFee.satoshis);
+            } else {
+                const coin = cand.classified.coin;
+                const inputFee = estimator.evalOnchainInput({ amount: BigInt(coin.value) });
+                if (inputFee.value >= coin.value) continue;
```
pietro909: same here

packages/ts-sdk/src/wallet/vtxo-manager.ts
```diff
@@ -1116,6 +1386,437 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
...
+        }
+
+        const arkAddress = await this.wallet.getAddress();
+        const outputFee = estimator.evalOffchainOutput({
+            amount: totalAmount,
+            script: hex.encode(ArkAddress.decode(arkAddress).pkScript),
+        });
+        totalAmount -= BigInt(outputFee.satoshis);
+
+        if (totalAmount < dustAmount) {
```
pietro909, line 1790: Don't check dustAmount: the migration is mandatory and must be allowed for every VTXO

packages/ts-sdk/src/wallet/signerRotation.ts
```diff
@@ -0,0 +1,142 @@
...
+    if (bytes.length === 33) return hex.encode(bytes.slice(1));
+    if (bytes.length === 32) return hex.encode(bytes);
+    throw new Error(`invalid signer pubkey length: expected 32 or 33 bytes, got ${bytes.length}`);
+}
+
+/**
+ * Build the {@link SignerAxis} from a server-info snapshot. Deprecated signers
+ * with an empty pubkey are skipped.
+ */
+export function signerAxisFromInfo(info: ArkInfo): SignerAxis {
```
pietro909: SignerAxis is very confusing name

packages/ts-sdk/src/wallet/vtxo-manager.ts
```diff
@@ -1192,6 +1475,437 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
...
+
+        // Cap to a single settlement: highest-value first (clears dust soonest,
+        // fewest passes to drain the bulk), bounded by BOTH the input count
+        // (MAX_VTXOS_PER_SETTLEMENT) AND a gross total within vtxoMaxAmount. The
+        // explicit-inputs settle path applies neither cap itself, so migration
+        // must — and since migration is fee-exempt the gross total IS the
+        // aggregated output amount, keeping it under the server ceiling. Overflow
+        // defers to the next pass.
+        const capped = capSettlementBatch(byValueDescending(sizedCandidates), vtxoMaxAmount);
+        const deferred = sizedCandidates.length - capped.length;
```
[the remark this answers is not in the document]
pietro909: we agreed not to do that for now. Putting on hold for now.

packages/ts-sdk/src/wallet/wallet.ts
```diff
@@ -905,25 +962,65 @@ export class ReadonlyWallet implements IReadonlyWallet {
...
+        const groups: BoardingUtxoGroup[] = [];
         for (const tapscript of tapscripts) {
             const address = tapscript.onchainAddress(this.network);
             const coins = await this.onchainProvider.getCoins(address);
             const utxos = coins.map((utxo) => extendCoinWithTapscript(tapscript, utxo));
             // Save boarding inputs using unified repository, keyed by the
             // address the UTXOs actually sit on.
             await this.walletRepository.saveUtxos(address, utxos);
-            all.push(...utxos);
+            groups.push({
```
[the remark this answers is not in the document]
pietro909: There is no ContractManager-driven boarding-UTXO sync to delegate to

ghost (changes requested):
## Arkana Code Review — PR #554 (Updated)
**Reviewer:** Arkana (automated, aggressive review)
**Scope:** Full review — all commits since force-push (entire PR rebased since last review on June 10)

---

### Verdict: REQUEST CHANGES

This is **protocol-critical code** (VTXO migration, signer key handling, transaction signing, forfeit/sweep paths). Requires human sign-off before merge regardless of code quality.

---

### 🟡 IMPORTANT Findings

**1. `vtxo-manager.ts` — Fee-exempt assumption on boarding migration settle**
The migration path (around line 1691-1715) passes gross `totalAmount` as the output, with a comment claiming "fee-exempt: every selected input moves at its full value." If arkd does NOT exempt migration settles from fees, the server will reject the intent because `sum(inputs) != sum(outputs) + fees`. This is a critical **server-side dependency** — verify arkd PR #1097 actually implements fee exemption for migration intents. If it doesn't, every boarding migration will silently fail (caught by try/catch, reported as `error` in the leg report, but the user gets no migration).

**2. `vtxo-manager.ts` — VTXOs without `batchExpiry` silently skipped from migration (line ~1887)**
VTXOs with no `batchExpiry` (unrolled/settled) are skipped with `continue`, but are still counted in per-signer report (`vtxoCount`, `totalValue`). This means the report shows holdings that can never be migrated cooperatively. Consumers will see a non-zero `totalValue` with `no-deprecated-vtxos` skip reason. Consider either: (a) excluding them from the report counts, or (b) adding a `nonMigratable` count to `DeprecatedSignerReport`.

**3. `ark.ts:342` — Loose `DIGEST_MISMATCH` detection**
`body.includes("DIGEST_MISMATCH")` is a substring match on the raw response body. If arkd ever returns an error containing this string in a different context, the SDK would false-positive: clear its digest, re-fetch info, emit `onServerInfoChanged`, and throw `DigestMismatchError` when the actual error was something different. Consider parsing JSON and checking a structured error code field, especially since this sits on **every mutating request** to arkd.

**4. `wallet.ts:1866` — Metadata loss on missing contract row during signer rotation**
In `_doRotateServerSigner`, the new offchain contract's metadata is sourced from `currentOffchainRow?.metadata`. If `currentOffchainRow` is `undefined` (e.g., corrupted contract repo, fresh wallet edge case), the new contract row loses its signing descriptor. On HD-rotated wallets, this means signing metadata is silently dropped, potentially causing signing failures post-rotation. Add a defensive guard or throw explicitly when the current offchain contract is missing.

**5. `wallet.ts:1809-1909` — Monotonic contract growth on repeated rotations**
Old-signer contracts are intentionally left `active` for migration discovery, but there's no cleanup path for contracts whose VTXOs have all been migrated or expired. Over many rotations, `getContracts` queries will degrade. Not a fund-safety issue, but an operational concern — consider documenting or implementing eventual cleanup.

**6. `ark.ts:337-340` — Silently swallowed body-read failure in digest detection**
If `response.clone().text()` throws (network interruption mid-body), `authedFetch` returns the original non-OK response without checking for `DIGEST_MISMATCH`. The client's digest stays stale until the next `getInfo()` call. Low probability but could delay rotation detection.

---

### 🟢 MINOR Findings

- **`index.ts`** — `DigestMismatchError` is not exported publicly. SDK consumers wanting to `instanceof` check for digest mismatches must import from the internal provider path. Consider exporting it.
- **`arkade-swaps.ts:2094`** — `senderXOnlyPublicKey` is initialized from `claimDetails.serverPublicKey` (the Boltz key), but the naming suggests the arkd server key. Add a clarifying comment.
- **`wallet.ts:1591`** — `console.warn` for server-signer rotation failure. For a protocol-critical event, structured error reporting (executive digest queue or similar) may be more appropriate than a `console.warn`.

---

### ✅ What Looks Good

- **VHTLC reconstruction** (`resolveVHTLCForLockup`): Correct. Deterministic probing, address-based matching, matched key threaded downstream to signing. All four call sites handle it correctly. No exploitable attack vector in multi-key probing.
- **Invoice payment hash validation** (`arkade-swaps.ts:473-476`): Correct. Hex-encoded SHA256 comparison catches malicious Boltz servers.
- **Signer classification** (`signerRotation.ts`): Clean implementation. x-only normalization handles both 33-byte and 32-byte keys. Classification logic is correct.
- **Service worker serialization**: Lossless bigint round-trip via `.toString()`/`BigInt()`. Thoroughly tested.
- **`walletReceiveRotator.ts`**: Well-structured. Build-locally-then-persist-then-swap pattern. Correct `runExclusive` serialization. Good `dispose()` drain.
- **Error handling on no-key-match** (`arkade-swaps.ts:2688-2692`): Loud failure with diagnostic message including candidate count.
- **Test coverage**: Comprehensive unit and e2e tests across signer classification, VHTLC reconstruction, migration reporting, service worker messaging.

---

### 📦 Cross-Repo Impact

- **`IVtxoManager` breaking change**: 2 new required methods (`migrateDeprecatedSignerVtxos`, `getDeprecatedSignerStatus`). Any external `IVtxoManager` implementation will break. Warrants a minor semver bump.
- **`ArkInfo.deprecatedSigners`**: Additive field, handled gracefully by `RestArkProvider` (defaults to `[]`). No downstream breakage.
- **coinflip**: Pinned to vendored `@arkade-os/sdk` v0.4.32 tarball — unaffected until updated.
- **arkd #1097** (`feat/deprecated-keys`): Server-side counterpart, currently OPEN. Both PRs implement the same signer rotation feature and must be coordinated for release.

---

### 🚨 Protocol-Critical Flag

This PR touches VTXOs, transaction signing, signer key handling, and cooperative settlement paths. **Requires human review and explicit sign-off before merge**, even though the implementation quality is high. The fee-exempt assumption (finding #1) is the highest-risk item — if arkd doesn't exempt migration settles, users silently lose the ability to migrate boarding UTXOs.

Kukks:
For the review record — two of Arkana's IMPORTANT findings look **already addressed** in the current code (the PR rebased since the June 13 review):

- **#3 (loose `DIGEST_MISMATCH` substring match):** `authedFetch` classifies via the *structured* grpc-gateway error — `maybeArkError(new Error(body))?.name !== "DIGEST_MISMATCH"` (`ark.ts:348-353`), not a raw `body.includes(...)`. A comment there already calls out the spurious-refresh risk of substring matching.
- **#6 (swallowed body-read failure):** the body-read `catch` surfaces it via `console.warn(...)` (`ark.ts:337-347`) rather than swallowing silently.

The other IMPORTANT items (#1 fee-exempt server dep, #2 `batchExpiry` report counts, #4 metadata guard, #5 contract growth) still look like open design/dependency calls.

ghost (changes requested):
## Arkana Code Review — PR #554 (Follow-up)
**Reviewer:** Arkana (automated, aggressive review)
**Scope:** Re-check after Kukks' feedback on June 13 review — no new commits since last review.

---

### Acknowledged: Findings #3 and #6 are addressed ✅

Kukks correctly noted:
- **#3 (DIGEST_MISMATCH detection):** Confirmed — `authedFetch` uses structured `maybeArkError(new Error(body))?.name !== "DIGEST_MISMATCH"` (ark.ts:353), not a raw substring match. Comment at lines 348-352 explicitly calls out the spurious-refresh risk of substring matching. This is solid.
- **#6 (body-read failure):** Confirmed — the `catch` block at ark.ts:345 surfaces via `console.warn(...)` with the caught exception, then returns the original response. Not silently swallowed.
- **Minor (DigestMismatchError export):** Now exported from `index.ts:137,352`. Resolved.

---

### Verdict: REQUEST CHANGES (3 open findings remain)

Still **protocol-critical** — VTXO migration, signer key handling, cooperative settlement. Requires human sign-off.

---

### 🟡 Still Open

**#1. `vtxo-manager.ts:1725-1734` — Fee-exempt assumption on boarding migration settle**
The boarding migration leg still passes gross `totalAmount` as the single output amount with no fee subtraction:
```ts
const totalAmount = capped.reduce((sum, c) => sum + BigInt(c.coin.value), 0n);
return this.wallet.settle({
    inputs: capped.map((c) => c.coin),
    outputs: [{ address: arkAddress, amount: totalAmount }],
}, ...);
```
The VTXO leg (line 1711) goes through `sendSelectedVtxosToSelf` which handles fees internally — only the boarding leg has this pattern. If arkd doesn't exempt migration settles from fees, every boarding migration silently fails. This is a **server-side dependency on arkd#822/arkd#1097** — please confirm the fee-exemption guarantee exists on the server side.

**#4. `wallet.ts:1904-1961` — No guard on missing contract row during signer rotation**
`_doRotateServerSigner` queries for the current offchain and boarding contract rows. If either query returns empty (`currentOffchainRow` is `undefined`), metadata is passed as `currentOffchainRow?.metadata` → `undefined`, silently dropping signing metadata. The comment at lines 1900-1903 acknowledges that index-0/static wallets carry no metadata, but there's no distinction between "row exists with no metadata" (fine) and "row doesn't exist at all" (dangerous on descriptor-backed wallets — HD signing metadata lost, wallet can't sign for rotated address). A defensive throw or warning when the expected row is missing would prevent silent corruption.

**#2. `vtxo-manager.ts:1882-1923` — Report inflates migratable counts with non-migratable VTXOs**
VTXOs lacking `batchExpiry` are skipped from cooperative migration (line 1923: `continue`), but are already counted in the per-signer `vtxoCount`/`totalValue` (lines 1882-1911). The comment at lines 1916-1922 says this is intentional ("reported for visibility"). However, consumers seeing non-zero `totalValue` with a `no-deprecated-vtxos` or `skipped` result will believe funds can be migrated when they can't. Consider splitting into `migratable` vs `recoveryOnly` counts in the report — or at minimum documenting this clearly in the `DeprecatedSignerReport` type.

---

### 🟢 Already Noted (lower priority, no change)

**#5. Contract growth on repeated rotations** — Old-signer contracts stay `active` indefinitely. Not fund-safety but operational — consider eventual cleanup or documenting the expected growth.

---

### Summary

3 of the original 6 IMPORTANT findings are resolved. The remaining 3 are design/safety calls that need human judgment:
1. Fee-exempt server dependency (fund safety)
2. Missing contract row guard (silent metadata loss)  
3. Report accuracy (UX correctness)

🚨 Protocol-critical: human sign-off required before merge.

ghost (changes requested):
## Arkana Code Review — PR #554 (Follow-up)
**Reviewer:** Arkana (automated, aggressive review)
**Scope:** Re-check after Kukks' feedback on June 13 review — no new commits since last review.

---

### Acknowledged: Findings #3 and #6 are addressed ✅

Kukks correctly noted:
- **#3 (DIGEST_MISMATCH detection):** Confirmed — `authedFetch` uses structured `maybeArkError(new Error(body))?.name !== "DIGEST_MISMATCH"` (ark.ts:353), not a raw substring match. Comment at lines 348-352 explicitly calls out the spurious-refresh risk of substring matching. This is solid.
- **#6 (body-read failure):** Confirmed — the `catch` block at ark.ts:345 surfaces via `console.warn(...)` with the caught exception, then returns the original response. Not silently swallowed.
- **Minor (DigestMismatchError export):** Now exported from `index.ts:137,352`. Resolved.

---

### Verdict: REQUEST CHANGES (3 open findings remain)

Still **protocol-critical** — VTXO migration, signer key handling, cooperative settlement. Requires human sign-off.

---

### 🟡 Still Open

**#1. `vtxo-manager.ts:1725-1734` — Fee-exempt assumption on boarding migration settle**
The boarding migration leg still passes gross `totalAmount` as the single output amount with no fee subtraction:
```ts
const totalAmount = capped.reduce((sum, c) => sum + BigInt(c.coin.value), 0n);
return this.wallet.settle({
    inputs: capped.map((c) => c.coin),
    outputs: [{ address: arkAddress, amount: totalAmount }],
}, ...);
```
The VTXO leg (line 1711) goes through `sendSelectedVtxosToSelf` which handles fees internally — only the boarding leg has this pattern. If arkd doesn't exempt migration settles from fees, every boarding migration silently fails. This is a **server-side dependency on arkd#822/arkd#1097** — please confirm the fee-exemption guarantee exists on the server side.

**#4. `wallet.ts:1904-1961` — No guard on missing contract row during signer rotation**
`_doRotateServerSigner` queries for the current offchain and boarding contract rows. If either query returns empty (`currentOffchainRow` is `undefined`), metadata is passed as `currentOffchainRow?.metadata` → `undefined`, silently dropping signing metadata. The comment at lines 1900-1903 acknowledges that index-0/static wallets carry no metadata, but there's no distinction between "row exists with no metadata" (fine) and "row doesn't exist at all" (dangerous on descriptor-backed wallets — HD signing metadata lost, wallet can't sign for rotated address). A defensive throw or warning when the expected row is missing would prevent silent corruption.

**#2. `vtxo-manager.ts:1882-1923` — Report inflates migratable counts with non-migratable VTXOs**
VTXOs lacking `batchExpiry` are skipped from cooperative migration (line 1923: `continue`), but are already counted in the per-signer `vtxoCount`/`totalValue` (lines 1882-1911). The comment at lines 1916-1922 says this is intentional ("reported for visibility"). However, consumers seeing non-zero `totalValue` with a `no-deprecated-vtxos` or `skipped` result will believe funds can be migrated when they can't. Consider splitting into `migratable` vs `recoveryOnly` counts in the report — or at minimum documenting this clearly in the `DeprecatedSignerReport` type.

---

### 🟢 Already Noted (lower priority, no change)

**#5. Contract growth on repeated rotations** — Old-signer contracts stay `active` indefinitely. Not fund-safety but operational — consider eventual cleanup or documenting the expected growth.

---

### Summary

3 of the original 6 IMPORTANT findings are resolved. The remaining 3 are design/safety calls that need human judgment:
1. Fee-exempt server dependency (fund safety)
2. Missing contract row guard (silent metadata loss)
3. Report accuracy (UX correctness)

🚨 Protocol-critical: human sign-off required before merge.

PR #556, opened 2026-06-11
feat(boltz-swap): optimistic Lightning sends via waitForSwapFunded

### LINKED ISSUE #555: feat: add optimisticResolveAt option to waitForSwapSettlement for submarine swaps
## Problem

`sendLightningPayment` in `arkade-swaps.ts` calls `waitForSwapSettlement`, which currently only resolves when the status reaches `transaction.claimed` — i.e. the recipient has swept their HTLC. This means callers are blocked until the *counterparty* acts, which can take seconds or never happen if the recipient is slow/offline.

## Proposed Solution

Add an `optimisticResolveAt` option to `waitForSwapSettlement` (and surface it through `sendLightningPayment`):

```ts
interface SwapSettlementOptions {
  /** Resolve optimistically when this status is reached, rather than waiting for transaction.claimed */
  optimisticResolveAt?: SwapStatus; // e.g. 'invoice.set' | 'transaction.mempool' | etc.
}
```

When `optimisticResolveAt` is provided, the function resolves as soon as that status is observed, returning the partial swap state. The caller is responsible for polling/updating in the background if needed.

## Why

- Other Lightning wallets show "sent" as soon as the payment is in-flight and the route is accepted — waiting for `transaction.claimed` is overly conservative.
- The swap is already persisted to IndexedDB at `invoice.set`, so an app crash won't lose state.
- Unblocks the wallet from showing a success/pending state immediately after sending, improving perceived UX significantly.

## Acceptance Criteria

- [ ] `waitForSwapSettlement` accepts an optional `optimisticResolveAt` status
- [ ] When provided, the promise resolves at that status rather than `transaction.claimed`
- [ ] Default behavior (no option) is unchanged — resolves at `transaction.claimed`
- [ ] Unit tests cover both paths
- [ ] Docs / JSDoc updated

## Downstream

Wallet follow-up is blocked on this: arkade-os/wallet (optimistic Lightning send UX)

tiero:
Closes #555. Unblocks arkade-os/wallet#667 (optimistic Lightning send UX).

## Problem

`sendLightningPayment` resolved only at the terminal `transaction.claimed` status — i.e. after Boltz swept the HTLC — blocking the caller on the counterparty, while most Lightning wallets show "sent" as soon as the payment is in flight.

## API

(The design evolved from the issue's `optimisticResolveAt` proposal during review — a named milestone instead of exposing raw Boltz statuses.)

```ts
// Resolve as soon as the lockup tx is observed instead of waiting for settlement
const result = await swaps.sendLightningPayment({
  invoice: 'lnbc...',
  waitFor: 'funded', // default: 'settled'
});
// result.preimage is undefined until the swap settles

// Same milestone for a swap you already hold
await swaps.waitForSwapFunded(pendingSwap);
```

- **`waitFor?: "settled" | "funded"`** on `SendLightningPaymentRequest`. Default `"settled"` — behavior and types of existing calls are unchanged (overloads keep the required `preimage` on the default path; `waitFor: "funded"` returns `OptimisticSendLightningPaymentResponse` with optional `preimage`).
- **`waitForSwapFunded(pendingSwap)`** resolves once `transaction.mempool` — or any *later* status in the submarine lifecycle — is observed. "At or beyond" matters because subscriptions only report the current status, so intermediate statuses (e.g. `transaction.confirmed` on 0-conf) can be skipped.
- Failure statuses observed before funding still reject (and auto-refund in `sendLightningPayment`).

## Semantics after optimistic resolution

Monitoring continues in the background until the swap reaches a terminal status, persisting every update to the repository (the `refundable` flag on a late failure, the preimage on claim), but the already-settled promise no longer rejects — acting on a late failure is the SwapManager's job (auto-refund) or the caller's via `restoreSwaps`/`recoverSubmarineFunds`.

If the app restarts before settlement, the in-process monitor is gone: `refreshSwapsStatus()` now backfills the missing preimage for submarine swaps it finds settled during startup reconciliation (still saving the status if the preimage fetch fails), so a wallet can flip its pending row to "completed" from the repository alone.

## Plumbing

- New `WAIT_FOR_SWAP_FUNDED`/`SWAP_FUNDED` service-worker message pair; `waitFor` rides the existing `SEND_LIGHTNING_PAYMENT` payload, so no envelope changes.
- Exposed through `ArkadeSwaps`, `IArkadeSwaps`, the Expo wrapper, and the service-worker runtime.
- `hasSubmarineStatusReached()` + `SubmarineProgressionStatus` exported as helpers.

## Tests

Unit tests cover: settled-only default, resolve at/beyond the funded milestone, no resolve on earlier statuses, rejection on failure before funding, persistence of late failures and late settlement after optimistic resolution, monitor errors after resolution, and the `refreshSwapsStatus` preimage backfill (success, fetch failure, not-yet-settled).

https://claude.ai/code/session_01HvHbtKFbABepS2JHqTew9D

### DIFF
--- a/packages/boltz-swap/src/boltz-swap-provider.ts
+++ b/packages/boltz-swap/src/boltz-swap-provider.ts
@@ -77,6 +77,36 @@ export const isSubmarineRefundableStatus = (status: BoltzSwapStatus): boolean =>
     return ["invoice.failedToPay", "transaction.lockupFailed", "swap.expired"].includes(status);
 };
 
+/**
+ * Canonical progression of a successful submarine swap, in order. Failure
+ * statuses are intentionally absent. Used to decide whether an observed
+ * status is at or beyond another status in the lifecycle.
+ */
+const SUBMARINE_STATUS_PROGRESSION: BoltzSwapStatus[] = [
+    "swap.created",
+    "invoice.set",
+    "transaction.mempool",
+    "transaction.confirmed",
+    "invoice.pending",
+    "invoice.paid",
+    "transaction.claim.pending",
+    "transaction.claimed",
+];
+
+/**
+ * Returns true if `status` is at or beyond `target` in the successful
+ * submarine swap progression. Returns false when either status is not part
+ * of the progression (e.g. failure statuses like "swap.expired").
+ */
+export const hasSubmarineStatusReached = (
+    status: BoltzSwapStatus,
+    target: BoltzSwapStatus,
+): boolean => {
+    const statusIndex = SUBMARINE_STATUS_PROGRESSION.indexOf(status);
+    const targetIndex = SUBMARINE_STATUS_PROGRESSION.indexOf(target);
+    return statusIndex >= 0 && targetIndex >= 0 && statusIndex >= targetIndex;
+};
+
 /** Returns true if the submarine swap completed successfully. */
 export const isSubmarineSuccessStatus = (status: BoltzSwapStatus): boolean => {
     return status === "transaction.claimed";
--- a/packages/boltz-swap/src/boltz-swap-provider.ts
+++ b/packages/boltz-swap/src/boltz-swap-provider.ts
@@ -79,10 +79,14 @@ export const isSubmarineRefundableStatus = (status: BoltzSwapStatus): boolean =>
 
 /**
  * Canonical progression of a successful submarine swap, in order. Failure
- * statuses are intentionally absent. Used to decide whether an observed
- * status is at or beyond another status in the lifecycle.
+ * statuses are intentionally absent. This defines relative ordering, NOT a
+ * guaranteed sequence: statuses can be skipped (e.g. "transaction.mempool"
+ * jumps straight to "invoice.pending" when Boltz accepts 0-conf, and a
+ * subscription only ever reports the current status). Consumers must treat
+ * any later status as implying the earlier ones — which is exactly what
+ * hasSubmarineStatusReached does.
  */
-const SUBMARINE_STATUS_PROGRESSION: BoltzSwapStatus[] = [
+const SUBMARINE_STATUS_PROGRESSION = [
     "swap.created",
     "invoice.set",
     "transaction.mempool",
@@ -91,19 +95,26 @@ const SUBMARINE_STATUS_PROGRESSION: BoltzSwapStatus[] = [
     "invoice.paid",
     "transaction.claim.pending",
     "transaction.claimed",
-];
+] as const satisfies readonly BoltzSwapStatus[];
+
+/**
+ * A status in the successful submarine swap progression — the only valid
+ * targets for optimistic settlement resolution.
+ */
+export type SubmarineProgressionStatus = (typeof SUBMARINE_STATUS_PROGRESSION)[number];
 
 /**
  * Returns true if `status` is at or beyond `target` in the successful
- * submarine swap progression. Returns false when either status is not part
- * of the progression (e.g. failure statuses like "swap.expired").
+ * submarine swap progression. Returns false when the observed status is not
+ * part of the progression (e.g. failure statuses like "swap.expired").
  */
 export const hasSubmarineStatusReached = (
     status: BoltzSwapStatus,
-    target: BoltzSwapStatus,
+    target: SubmarineProgressionStatus,
 ): boolean => {
-    const statusIndex = SUBMARINE_STATUS_PROGRESSION.indexOf(status);
-    const targetIndex = SUBMARINE_STATUS_PROGRESSION.indexOf(target);
+    const progression: readonly BoltzSwapStatus[] = SUBMARINE_STATUS_PROGRESSION;
+    const statusIndex = progression.indexOf(status);
+    const targetIndex = progression.indexOf(target);
     return statusIndex >= 0 && targetIndex >= 0 && statusIndex >= targetIndex;
 };
 
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -72,8 +72,10 @@ const result = await swaps.sendLightningPayment(
 );
 // Resolves as soon as the status is observed (or any later status in the
 // lifecycle — intermediate statuses can be skipped). result.preimage is
-// undefined when the swap hasn't settled yet; keep the SwapManager enabled
-// so settlement tracking and auto-refunds continue in the background.
+// undefined when the swap hasn't settled yet. Monitoring continues in the
+// background and keeps the stored swap up to date until a terminal status,
+// but a late failure no longer rejects — keep the SwapManager enabled so
+// auto-refunds are handled for you.
 ```
 
 The same option is accepted by `waitForSwapSettlement(pendingSwap, options)`.
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -81,6 +81,12 @@ const result = await swaps.sendLightningPayment({
 The same milestone is exposed for a swap you already hold as
 `waitForSwapFunded(pendingSwap)`, alongside `waitForSwapSettlement(pendingSwap)`.
 
+If the app restarts before the swap settles, the in-process monitoring is gone — call
+`refreshSwapsStatus()` on startup to reconcile: it polls every pending swap and persists
+the latest status, including the preimage for swaps that settled while the app was
+closed. Read the swap back from the repository (e.g. `getSwapHistory()`) to flip a
+pending UI row to "completed".
+
 ### ARK to BTC
 
 ```typescript
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -61,6 +61,23 @@ console.log('Paid:', result.txid);
 // SwapManager auto-refunds if payment fails
 ```
 
+By default the call resolves only once the swap fully settles (`transaction.claimed`),
+i.e. after Boltz has swept the HTLC. To show an optimistic "sent" state as soon as the
+payment is in flight — like most Lightning wallets — pass `optimisticResolveAt`:
+
+```typescript
+const result = await swaps.sendLightningPayment(
+  { invoice: 'lnbc500u1pj...' },
+  { optimisticResolveAt: 'invoice.pending' },
+);
+// Resolves as soon as the status is observed (or any later status in the
+// lifecycle — intermediate statuses can be skipped). result.preimage is
+// undefined when the swap hasn't settled yet; keep the SwapManager enabled
+// so settlement tracking and auto-refunds continue in the background.
+```
+
+The same option is accepted by `waitForSwapSettlement(pendingSwap, options)`.
+
 ### ARK to BTC
 
 ```typescript
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -71,8 +71,9 @@ const result = await swaps.sendLightningPayment({
   waitFor: 'funded',
 });
 // Resolves as soon as the lockup transaction is observed: the funds are
-// committed and the swap is refundable from here. result.preimage is
-// undefined since the swap hasn't settled yet. Monitoring continues in the
+// committed and the swap is refundable from here. result.preimage is not
+// reported on this path — the proof of payment is persisted to the stored
+// swap once it settles. Monitoring continues in the
 // background and keeps the stored swap up to date until a terminal status,
 // but a late failure no longer rejects — keep the SwapManager enabled so
 // auto-refunds are handled for you. With the SwapManager disabled, a late
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -63,22 +63,23 @@ console.log('Paid:', result.txid);
 
 By default the call resolves only once the swap fully settles (`transaction.claimed`),
 i.e. after Boltz has swept the HTLC. To show an optimistic "sent" state as soon as the
-payment is in flight — like most Lightning wallets — pass `optimisticResolveAt`:
+payment is in flight — like most Lightning wallets — pass `waitFor: 'funded'`:
 
 ```typescript
-const result = await swaps.sendLightningPayment(
-  { invoice: 'lnbc500u1pj...' },
-  { optimisticResolveAt: 'invoice.pending' },
-);
-// Resolves as soon as the status is observed (or any later status in the
-// lifecycle — intermediate statuses can be skipped). result.preimage is
-// undefined when the swap hasn't settled yet. Monitoring continues in the
+const result = await swaps.sendLightningPayment({
+  invoice: 'lnbc500u1pj...',
+  waitFor: 'funded',
+});
+// Resolves as soon as the lockup transaction is observed: the funds are
+// committed and the swap is refundable from here. result.preimage is
+// undefined since the swap hasn't settled yet. Monitoring continues in the
 // background and keeps the stored swap up to date until a terminal status,
 // but a late failure no longer rejects — keep the SwapManager enabled so
 // auto-refunds are handled for you.
 ```
 
-The same option is accepted by `waitForSwapSettlement(pendingSwap, options)`.
+The same milestone is exposed for a swap you already hold as
+`waitForSwapFunded(pendingSwap)`, alongside `waitForSwapSettlement(pendingSwap)`.
 
 ### ARK to BTC
 
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -75,7 +75,9 @@ const result = await swaps.sendLightningPayment({
 // undefined since the swap hasn't settled yet. Monitoring continues in the
 // background and keeps the stored swap up to date until a terminal status,
 // but a late failure no longer rejects — keep the SwapManager enabled so
-// auto-refunds are handled for you.
+// auto-refunds are handled for you. With the SwapManager disabled, a late
+// failure is only persisted as refundable and the funds stay locked until
+// you recover them via restoreSwaps()/recoverSubmarineFunds().
 ```
 
 The same milestone is exposed for a swap you already hold as
--- a/packages/boltz-swap/src/arkade-swaps.ts
+++ b/packages/boltz-swap/src/arkade-swaps.ts
@@ -2841,13 +2841,33 @@ export class ArkadeSwaps {
             if (isSubmarineFinalStatus(swap.status)) continue;
             promises.push(
                 this.getSwapStatus(swap.id)
-                    .then(({ status }) =>
-                        updateSubmarineSwapStatus(
+                    .then(async ({ status }) => {
+                        // A swap that settled while no monitor was attached
+                        // (e.g. a "funded" optimistic send followed by an app
+                        // restart) has no stored preimage yet — fetch it so
+                        // the repository carries the proof of payment. Never
+                        // fail the status update over a missing preimage.
+                        let additionalFields: Partial<BoltzSubmarineSwap> | undefined;
+                        if (isSubmarineSuccessStatus(status) && !swap.preimage) {
+                            try {
+                                const { preimage } = await this.swapProvider.getSwapPreimage(
+                                    swap.id,
+                                );
+                                additionalFields = { preimage };
+                            } catch (error) {
+                                logger.warn(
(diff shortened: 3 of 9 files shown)


ghost (changes requested):
## Arkana Code Review — arkade-os/ts-sdk#556

**Reviewer:** Arkana (aggressive review)
**Verdict:** Request changes

---

### ⚠️ PROTOCOL-CRITICAL: This PR touches submarine swap settlement (locked funds flow). Requires human sign-off before merge.

---

### 🔴 P0 — Status persistence stops after optimistic resolution

**`packages/boltz-swap/src/arkade-swaps.ts:1409`** — The `isResolved` guard at the top of `onStatusUpdate` causes ALL subsequent status updates to be silently dropped:

```typescript
const onStatusUpdate = async (status: BoltzSwapStatus) => {
    if (isResolved) return; // <-- everything after optimistic resolve is a no-op
```

After optimistic resolution at e.g. `invoice.pending`:
1. `isResolved = true`, promise resolves, caller gets the result
2. `monitorSwap` continues running (WebSocket stays open)
3. If `invoice.failedToPay` arrives → `onStatusUpdate` returns immediately → **`saveStatus()` is never called**
4. The swap's persisted status remains `invoice.pending` forever

**Impact:** The SwapManager has its own independent WebSocket monitor that should catch failures, so this is *survivable* — but only if the SwapManager is actually monitoring this swap. If it isn't (e.g., app restarts between optimistic resolve and failure, SwapManager not started, or `start()` called before the failure swap is passed in), the stored status is stale and the swap won't be refunded.

**Fix:** After setting `isResolved = true`, continue to persist status updates without resolving/rejecting:

```typescript
default:
    await saveStatus();
    if (optimisticResolveAt && hasSubmarineStatusReached(status, optimisticResolveAt)) {
        isResolved = true;
        resolve({ preimage: undefined });
        // Don't return — let the monitor keep persisting subsequent statuses
    }
    break;
```

And at the top, change the guard so it still saves:
```typescript
if (isResolved) {
    // Still persist status so SwapManager / restoreSwaps can see failures
    await saveStatus();
    return;
}
```

---

### 🟡 P1 — `optimisticResolveAt` accepts non-submarine statuses, silently hangs

**`packages/boltz-swap/src/types.ts:118-131`** — `SwapSettlementOptions.optimisticResolveAt` is typed as `BoltzSwapStatus`, which includes reverse-swap and chain-swap statuses (`invoice.settled`, `transaction.server.mempool`, etc.). If a caller passes one of these, `hasSubmarineStatusReached` returns `false` for every update → the promise hangs until `transaction.claimed` (or failure), with no indication the option was ignored.

**Fix:** Narrow the type to the statuses that actually appear in `SUBMARINE_STATUS_PROGRESSION`:

```typescript
export type SubmarineSwapProgressionStatus =
    | "swap.created" | "invoice.set" | "transaction.mempool"
    | "transaction.confirmed" | "invoice.pending" | "invoice.paid"
    | "transaction.claim.pending" | "transaction.claimed";

export interface SwapSettlementOptions {
    optimisticResolveAt?: SubmarineSwapProgressionStatus;
}
```

Or, at minimum, validate at runtime and throw early if the status isn't in the progression.

---

### 🟡 P1 — Documentation says "SDK stops monitoring" — it doesn't

PR description and JSDoc say: *"After an optimistic resolution the SDK stops monitoring the swap."*

In reality, `monitorSwap` continues running with an open WebSocket until the swap reaches a terminal status. The callbacks become no-ops but the connection stays alive. This is misleading to callers who may assume they need to set up their own monitoring immediately; in fact the monitor is still consuming the WS slot from Boltz.

**Fix:** Either actually unsubscribe (return a cleanup handle from `monitorSwap`) or correct the docs to say "the SDK stops *acting on* status updates but the WebSocket remains open until the swap reaches a terminal state."

---

### 🟢 P2 — Overload type edge case

**`packages/boltz-swap/src/arkade-swaps.ts:751-762`** — Passing `sendLightningPayment(args, {})` or `sendLightningPayment(args, { optimisticResolveAt: undefined })` resolves to the second overload returning `OptimisticSendLightningPaymentResponse` (optional `preimage`), but the implementation will wait for full settlement and always return the preimage. The caller's type is weaker than reality.

Not a runtime bug, but a downstream consumer doing `result.preimage.slice(...)` would get a type error even though `preimage` is always present. Consider adding a runtime check or adjusting the overloads so the non-optimistic return type is preserved when `optimisticResolveAt` is absent/undefined.

---

### 🟢 P2 — Missing test: monitor error after optimistic resolution

**`packages/boltz-swap/test/arkade-swaps.test.ts`** — Tests cover: optimistic resolve at exact status, beyond status, before status, and failure before optimistic point. Missing: what happens when `monitorSwap` throws (WebSocket error) *after* optimistic resolution. The `.catch()` handler at line 1465 checks `!isResolved`, which is correct, but a test would lock in the behavior.

---

### ✅ What looks good

- `SUBMARINE_STATUS_PROGRESSION` order matches the Boltz submarine lifecycle correctly (lockup before invoice progress)
- `hasSubmarineStatusReached` correctly returns `false` for failure statuses (not in progression array), so they always fall through to the explicit `case` handlers
- Failure statuses (`swap.expired`, `invoice.failedToPay`, `transaction.lockupFailed`) are handled in explicit `case` branches BEFORE the `default` where optimistic resolution lives — ordering is correct
- Backward compatibility: all new params are optional, service worker message protocol is backward-compatible, existing consumers (coinflip) won't break
- Test coverage is solid for the happy paths and the main edge cases
- TypeScript overloads preserve type safety for the non-optimistic call path

---

**Bottom line:** The design is sound and the feature is well-motivated. The P0 (status persistence gap) needs fixing before merge — it creates a narrow but real window for fund loss if the SwapManager isn't covering the swap. The P1 type narrowing prevents a class of silent misconfiguration bugs.

🏷️ **Protocol-critical — requires human reviewer sign-off.**

packages/boltz-swap/src/boltz-swap-provider.ts
```diff
@@ -77,6 +77,36 @@ export const isSubmarineRefundableStatus = (status: BoltzSwapStatus): boolean =>
...
+/**
+ * Canonical progression of a successful submarine swap, in order. Failure
+ * statuses are intentionally absent. Used to decide whether an observed
+ * status is at or beyond another status in the lifecycle.
+ */
+const SUBMARINE_STATUS_PROGRESSION: BoltzSwapStatus[] = [
+    "swap.created",
+    "invoice.set",
+    "transaction.mempool",
+    "transaction.confirmed",
```
tiero, line 93: it may skip from mempool with never have confirmed directly

ghost (approved):
## Arkana Follow-Up Review — arkade-os/ts-sdk#556 (commit ae26c1b7)

**Previous review:** Requested changes (4 findings)
**This review:** Focused on the new commit addressing those findings.

---

### ✅ P0 RESOLVED — Status persistence after optimistic resolution

The `isFinal` / `isSettled` split is the right fix. Clean separation of concerns:
- `isFinal` gates status processing (only set on terminal statuses)
- `isSettled` tracks promise settlement (set on optimistic resolve OR terminal)
- Optimistic resolution: `isSettled=true, isFinal=false` → monitor keeps running, `saveStatus()` keeps firing, `reject()`/`resolve()` are harmless no-ops on the already-settled promise

The callback wrapper (`.catch(error => logger.error(...))`) properly prevents unhandled rejections from late persistence failures. Good.

### ✅ P1 RESOLVED — Type narrowing for `optimisticResolveAt`

`SubmarineProgressionStatus` derived from `as const satisfies` — callers can now only pass statuses that actually exist in the submarine lifecycle. No more silent hangs from passing reverse-swap statuses.

### ✅ P1 RESOLVED — Documentation accuracy

README and JSDoc now correctly state monitoring continues in the background and the promise no longer rejects after optimistic resolution. Accurate.

### ✅ P2 RESOLVED — Missing test for monitor error after optimistic resolution

New test `"should not be affected by monitor errors after optimistic resolution"` covers this. Two other new tests (`"should keep persisting a failure..."` and `"should persist the preimage..."`) directly validate the P0 fix. Solid coverage.

### 🟢 P2 STILL OPEN (minor, non-blocking) — Overload type edge case

`sendLightningPayment(args, {})` still resolves to `OptimisticSendLightningPaymentResponse` (optional `preimage`) even though the implementation will wait for full settlement and always return the preimage. Not a runtime issue — just a type widening that could cause downstream consumers to add unnecessary null checks. Low priority.

---

**Verdict:** All material findings addressed. The fix is correct and well-tested. Approving.

⚠️ **Protocol-critical — still requires human reviewer sign-off before merge.** This touches submarine swap settlement flow (locked funds).

packages/boltz-swap/src/boltz-swap-provider.ts
tiero: Correct — and that's exactly why the check is "at or beyond" rather than exact-match: the array defines relative *ordering*, not a guaranteed sequence. If `transaction.confirmed` is skipped (0-conf) and the next observed status is `invoice.pending`, `hasSubmarineStatusReached("invoice.pending", target)` still returns true for any earlier target, so the wait resolves. Same applies to any other skipped status, since the WS subscription only ever reports the current one. Clarified this in the doc comment in ae26c1b.

---

ghost (approved):
## Arkana Follow-Up Review — arkade-os/ts-sdk#556 (commit 4d49a3af)

**Previous review:** Approved at ae26c1b7 (core logic fixes)
**This review:** Focused on the API refactor in the new commit.

---

### Summary of changes since last review

Commit `4d49a3af` is a pure API surface refactor per maintainer feedback — no changes to the core `isFinal`/`isSettled` logic I previously approved. Specifically:

1. **`SwapSettlementOptions` removed** — the `optimisticResolveAt` param is gone from the public API
2. **`waitFor?: "settled" | "funded"` added** directly to `SendLightningPaymentRequest` — simpler, more opinionated
3. **`waitForSwapSettlement` reverts to single-signature** (no options param) — always resolves at `transaction.claimed` with preimage
4. **New `waitForSwapFunded(pendingSwap)`** — hardcodes `"transaction.mempool"` as the resolve target via the shared `waitForSubmarineSwap` helper
5. **New service worker message pair** `WAIT_FOR_SWAP_FUNDED` / `SWAP_FUNDED` — properly added to request/response unions and `LONG_RUNNING_ARKADE_SWAPS_REQUEST_TYPES`

### ✅ Verified

- **Cross-repo impact:** No consumers of `SwapSettlementOptions` or `optimisticResolveAt` outside `ts-sdk`. Clean removal.
- **Service worker plumbing:** `waitFor` rides on the request payload → message handler calls `sendLightningPayment(message.payload)` → `ArkadeSwaps` checks `args.waitFor === "funded"`. Correct.
- **`ResponseWaitForSwapSettlement` payload** tightened from `{ preimage?: string }` to `{ preimage: string }` — correct since this path no longer has optimistic resolution.
- **`ExpoArkadeSwaps`** properly delegates both `waitForSwapSettlement` and `waitForSwapFunded`.
- **Tests** adapted to the new API surface; same coverage as before.
- **No conflicts** with other open PRs (#557, #554).

### 🟢 P2 STILL OPEN (minor, non-blocking) — Overload type edge case

Same note from last review: `sendLightningPayment({ invoice: "...", waitFor: undefined })` matches the second overload → `OptimisticSendLightningPaymentResponse` (optional `preimage`), but the implementation waits for full settlement. Not a runtime issue.

---

**Verdict:** Clean API simplification. Core logic unchanged from my previous approval. LGTM.

⚠️ **Protocol-critical — still requires human reviewer sign-off before merge.**

tiero:
Re Arkana's remaining P2 ("overload type edge case"): this was true for the old two-argument shape but no longer applies after 4d49a3a. The overloads now discriminate on the `waitFor` field — `{ invoice }`, `{ invoice, waitFor: undefined }`, and `{ invoice, waitFor: "settled" }` all match the strict overload (`waitFor?: "settled"` accepts explicit `undefined`; `exactOptionalPropertyTypes` is not enabled), so they keep the required `preimage: string`. Only `waitFor: "funded"` gets `OptimisticSendLightningPaymentResponse` with the optional preimage. Verified with `tsc --strict` against the project config. No open findings remain.

---

ghost (changes requested):
## Arkana Code Review — arkade-os/ts-sdk#556

**Reviewer**: Arkana (automated, aggressive)
**Verdict**: Request changes (one medium issue) + request human review (swap lifecycle / fund handling)

---

### Architecture & Design

The overall design is solid. The `isFinal` / `isSettled` flag split is the right abstraction — it cleanly separates "stop processing updates" from "promise is done," enabling background persistence after optimistic resolution. The `hasSubmarineStatusReached` "at or beyond" logic correctly handles Boltz's status-skipping behavior. The callback `.catch()` wrapper around `onStatusUpdate` fixes a real unhandled-rejection vector since `monitorSwap` fire-and-forgets the callback. Test coverage is thorough — 8 new test cases covering the happy path, beyond-target resolution, pre-funding rejection, late failure persistence, late settlement persistence, and post-resolution monitor errors.

---

### 🔴 Medium: Auto-refund bypassed on late failures with `waitFor: "funded"`

**`arkade-swaps.ts:783-790`**

When `waitFor === "funded"`, `sendLightningPayment` returns at line 789 after `waitForSwapFunded` resolves. The `catch` block (lines 797-808) that triggers `this.refundVHTLC(pendingSwap)` is never reached for failures that arrive *after* funding.

The background monitor correctly persists `refundable: true` to the repository, but the **active refund attempt** in the catch block is skipped entirely. This means:
- If SwapManager is enabled → auto-refund handled ✅
- If SwapManager is NOT enabled → funds sit in refundable state until the caller manually calls `restoreSwaps` / `recoverSubmarineFunds` ⚠️

The README and JSDoc say "keep the SwapManager enabled so auto-refunds are handled for you," which is the right guidance. But `sendLightningPayment` doesn't enforce or verify that SwapManager is active. A caller using the standalone `ArkadeSwaps` without SwapManager and `waitFor: "funded"` silently loses the safety net.

**Suggestion**: At minimum, emit a `logger.warn()` at the `waitFor: "funded"` entry point when `this.swapManager` is not set, so the caller gets a signal. Alternatively, document this constraint with a `@remarks` tag in the JSDoc. This is real money at stake if someone misuses the API.

---

### 🟡 Low: `isFinal` not set on monitor error after optimistic resolution

**`arkade-swaps.ts:1562-1568`**

```typescript
.catch((error) => {
    if (!isSettled) {
        isFinal = true;
        isSettled = true;
        reject(error);
    }
});
```

When the monitor throws after optimistic resolution (`isSettled === true`), this handler skips setting `isFinal = true`. In practice this is harmless — `monitorSwap` won't deliver more callbacks after it rejects — but it's an inconsistency with the rest of the code where `isFinal` and `isSettled` are always set together on terminal events. For defensive correctness:

```typescript
.catch((error) => {
    if (!isSettled) {
        isFinal = true;
        isSettled = true;
        reject(error);
    } else {
        isFinal = true;
        logger.warn(`Swap ${pendingSwap.id}: monitor failed after settlement: ${error}`);
    }
});
```

---

### 🟡 Low: TypeScript overload gotcha with typed request variables

**`arkade-swaps.ts:762-767`, `IArkadeSwaps` interface, Expo wrapper, service worker**

The overloads are correct for literal arguments:
- `sendLightningPayment({ invoice: "..." })` → matches overload 1 → `SendLightningPaymentResponse` (preimage required) ✅
- `sendLightningPayment({ invoice: "...", waitFor: "funded" })` → matches overload 2 → `OptimisticSendLightningPaymentResponse` ✅

But if a caller stores the request in a variable typed as `SendLightningPaymentRequest`:
```typescript
const req: SendLightningPaymentRequest = { invoice: "..." };
const res = await swaps.sendLightningPayment(req);
// res.preimage is string | undefined — overload 2 matched
```

This is a known TypeScript overload pattern limitation, not a bug. Just flagging it since downstream consumers (e.g., coinflip) that type their request objects will see the broader return type even on the default path. Worth a note in the README or a code comment near the overloads.

---

### ✅ What looks good

1. **`waitForSubmarineSwap` shared machinery** (`arkade-swaps.ts:1458-1570`): Clean refactor. The dual-flag approach is well-commented and the "at or beyond" semantics are correct.

2. **`hasSubmarineStatusReached` + `SUBMARINE_STATUS_PROGRESSION`** (`boltz-swap-provider.ts:85-123`): Correct ordering. Returns `false` for failure statuses (not in the progression array). The `as const satisfies` pattern gives good type safety.

3. **Callback error wrapping** (`arkade-swaps.ts:1551-1560`): Fixes a real bug — `monitorSwap` fire-and-forgets the callback, so an async `saveStatus` failure would be an unhandled rejection without this wrapper.

4. **`refreshSwapsStatus` preimage recovery** (`arkade-swaps.ts:2841-2873`): Correctly handles the app-restart-after-optimistic-send scenario. The try/catch around `getSwapPreimage` ensures a missing preimage doesn't block status persistence.

5. **Service worker plumbing**: `WAIT_FOR_SWAP_FUNDED` / `SWAP_FUNDED` message pair follows existing patterns. Added to `LONG_RUNNING_ARKADE_SWAPS_REQUEST_TYPES`. Consistent across message handler, runtime, and Expo wrapper.

6. **Test coverage** (`arkade-swaps.test.ts`): 270 new lines covering: default settled path unchanged, funded resolution, beyond-target resolution, pre-funding rejection, late failure persistence, late settlement + preimage persistence, monitor error after resolution, and `refreshSwapsStatus` preimage fetch (including failure case). Good edge case coverage.

7. **Types** (`types.ts`): `OptimisticSendLightningPaymentResponse` correctly extends the response shape with optional `preimage`. Clean separation from `SendLightningPaymentResponse`.

---

### ⚠️ Protocol-level flag

This PR changes when a submarine swap is considered "sent" from the caller's perspective. While it doesn't touch VTXO construction, transaction signing, or forfeit paths directly, it changes **fund lifecycle semantics** — the caller's funds are committed to the swap but the promise resolves before settlement. A late Boltz failure means real money is at risk until auto-refund runs.

**Requesting human review** per protocol-critical code policy. The design is sound, but the auto-refund bypass (🔴 above) should be addressed before merge.

---

*🤖 Reviewed by Arkana — arkade-os/ts-sdk#556, commit HEAD*

tiero:
Addressed Arkana's findings in 90d56e3:

- **Medium (auto-refund bypass):** `sendLightningPayment` with `waitFor: "funded"` now logs a warning when `this.swapManager` is null (it's enabled by default, so this only fires for callers who opted out), and the JSDoc `@remarks` + README spell out that a post-resolution failure is only persisted as refundable and must be recovered via `restoreSwaps`/`recoverSubmarineFunds`. I deliberately did **not** add a background `refundVHTLC` attempt on late failures: it would race with the SwapManager's own auto-refund (two concurrent refund attempts on the same VHTLC) and contradict the agreed semantics that post-resolution recovery belongs to the SwapManager or the caller.
- **Low (`isFinal` on monitor error):** the monitor-error handler now sets `isFinal = true` and logs a warning when the monitor fails after an optimistic resolution, as suggested.
- **Low (overload widening):** documented next to the overloads — a request stored in a variable typed `SendLightningPaymentRequest` widens the result to the optimistic response type even on the default path.

Unit suite is at 392 passing, including a new test for the SwapManager-disabled warning.

---

tiero:
Both CodeRabbit findings addressed in d9c7775:

- **Major (promise left pending on terminal-path errors):** Valid — a thrown `getSwapPreimage` or repository write in a terminal branch left the wait hanging forever. Fixed, but **not** by moving the `isFinal`/`isSettled` assignment after the awaits as proposed: `monitorSwap` doesn't await the callback, so deferring the flags would let a status arriving mid-await be double-processed (double resolve/save). Instead, the flags-first ordering is kept and the awaited operations can no longer hang the promise — `saveStatus` is now non-throwing (logged; the repository self-heals on the next update or `refreshSwapsStatus`), and the claim branch wraps the preimage fetch in try/catch and rejects on failure. Two new tests cover the no-hang guarantees (preimage fetch failure rejects; persistence failure still rejects with the domain error).
- **Minor (README preimage wording):** The suggested wording ("may already be present") doesn't match the code — `sendLightningPayment`'s funded path never includes a preimage in the response, even when the swap settles before the promise resolves. Reworded instead to say the preimage isn't reported on this path and the proof of payment is persisted to the stored swap on settlement.

---

pietro909 (approved):

PR #557, opened 2026-06-11
add digest header to http request

bordalix:
## Summary by CodeRabbit

* **New Features**
  * Added a centralized fetch utility that includes build-version metadata on requests and verifies Fetch API availability.
* **Chores**
  * Switched SDK providers to use the centralized fetch utility for HTTP calls.
* **Tests**
  * Updated test suites to mock the new fetch utility instead of stubbing the global fetch.

### DIFF
--- a/packages/ts-sdk/src/utils/fetch.ts
+++ b/packages/ts-sdk/src/utils/fetch.ts
@@ -1,4 +1,4 @@
-const digest = "83e847f06e157c8da464824a556098452323158cc34f65d40a72ffd8ff1b90e3";
+const version = "0.9.7";
 
 export function fetch(input: RequestInfo, init?: RequestInit): Promise<Response> {
     if (typeof globalThis.fetch !== "function") {
@@ -7,7 +7,7 @@ export function fetch(input: RequestInfo, init?: RequestInit): Promise<Response>
     if (init) {
         init.headers = {
             ...init.headers,
-            "X-Server-Digest": digest,
+            "X-Build-Version": version,
         };
     }
     return globalThis.fetch(input, init);
--- a/packages/ts-sdk/src/utils/fetch.ts
+++ b/packages/ts-sdk/src/utils/fetch.ts
@@ -1,0 +1,14 @@
+const digest = "83e847f06e157c8da464824a556098452323158cc34f65d40a72ffd8ff1b90e3";
+
+export function fetch(input: RequestInfo, init?: RequestInit): Promise<Response> {
+    if (typeof globalThis.fetch !== "function") {
+        throw new Error("Fetch API is not available in this environment.");
+    }
+    if (init) {
+        init.headers = {
+            ...init.headers,
+            "X-Server-Digest": digest,
+        };
+    }
+    return globalThis.fetch(input, init);
+}
--- a/packages/ts-sdk/src/utils/fetch.ts
+++ b/packages/ts-sdk/src/utils/fetch.ts
@@ -4,11 +4,7 @@ export function fetch(input: RequestInfo, init?: RequestInit): Promise<Response>
     if (typeof globalThis.fetch !== "function") {
         throw new Error("Fetch API is not available in this environment.");
     }
-    if (init) {
-        init.headers = {
-            ...init.headers,
-            "X-Build-Version": version,
-        };
-    }
-    return globalThis.fetch(input, init);
+    const headers = new Headers(init?.headers);
+    headers.set("X-Build-Version", version);
+    return globalThis.fetch(input, { ...init, headers });
 }
--- a/packages/ts-sdk/src/utils/fetch.ts
+++ b/packages/ts-sdk/src/utils/fetch.ts
@@ -1,10 +1,10 @@
-const version = "0.9.7";
+export const buildVersion = "0.9.7";
 
 export function fetch(input: RequestInfo, init?: RequestInit): Promise<Response> {
     if (typeof globalThis.fetch !== "function") {
         throw new Error("Fetch API is not available in this environment.");
     }
     const headers = new Headers(init?.headers);
-    headers.set("X-Build-Version", version);
+    headers.set("X-Build-Version", buildVersion);
     return globalThis.fetch(input, { ...init, headers });
 }
--- a/packages/ts-sdk/src/providers/ark.ts
+++ b/packages/ts-sdk/src/providers/ark.ts
@@ -7,6 +7,7 @@ import { maybeArkError } from "./errors";
 import type { IntentFeeConfig } from "../arkfee";
 import { Intent } from "../intent";
 import { DEFAULT_ARKADE_SERVER_URL } from "../networks";
+import { fetch } from "../utils/fetch";
 
 /** Output requested during settlement or transaction submission. */
 export type Output = {
--- a/packages/ts-sdk/src/providers/delegate.ts
+++ b/packages/ts-sdk/src/providers/delegate.ts
@@ -1,5 +1,6 @@
 import { Intent } from "../intent";
 import { SignedIntent } from "./ark";
+import { fetch } from "../utils/fetch";
 
 /**
  * Delegate identity and fee information returned by `getDelegateInfo`.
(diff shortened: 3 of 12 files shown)


ghost (changes requested):
## Arkana Code Review — `add digest header to http request`

### 🔴 Bug: Header not injected on GET requests (12 of 33 call sites)

**`packages/ts-sdk/src/utils/fetch.ts:7-11`**

The wrapper only adds `X-Build-Version` when `init` is truthy:

```ts
if (init) {
    init.headers = { ...init.headers, "X-Build-Version": version };
}
```

But 12 call sites across the codebase pass only a URL with no `init`:
- `ark.ts:261` — `getInfo()` (the single most important call — server identification)
- `indexer.ts:317,348,361,374,400,431,535,557,634,650` — all GET endpoints
- `delegate.ts:122` — `getDelegateInfo()`

These requests will **never** carry the version header. Fix:

```ts
export function fetch(input: RequestInfo, init?: RequestInit): Promise<Response> {
    if (typeof globalThis.fetch !== "function") {
        throw new Error("Fetch API is not available in this environment.");
    }
    const headers = new Headers(init?.headers);
    headers.set("X-Build-Version", version);
    return globalThis.fetch(input, { ...init, headers });
}
```

This handles all three `HeadersInit` shapes (`Headers`, `string[][]`, `Record<string,string>`), never mutates the caller's `init` object, and injects the header on every request.

---

### 🟡 `init.headers` spread drops `Headers` instances

**`packages/ts-sdk/src/utils/fetch.ts:8-10`**

`RequestInit.headers` is typed `HeadersInit = Headers | string[][] | Record<string, string>`. Spreading a `Headers` object produces `{}`, silently dropping every header the caller set. No current caller uses `new Headers()`, but this is a type-unsafe footgun. The `new Headers(init?.headers)` approach above resolves this.

---

### 🟡 Caller's `init` object is mutated

**`packages/ts-sdk/src/utils/fetch.ts:8`**

`init.headers = { ... }` mutates the caller's object. If any caller reuses an `init` config across calls (common pattern), headers accumulate or behave unpredictably. The fix above (`{ ...init, headers }`) avoids mutation.

---

### 🟡 Hardcoded version `"0.9.7"` ≠ package version `"0.4.34"`

**`packages/ts-sdk/src/utils/fetch.ts:1`**

The version string is hardcoded and already out of sync with `package.json` (`0.4.34`). This will silently drift on every release. Either import it from `package.json` or inject it at build time.

---

### 🟢 Everything else looks good

- Test refactoring from `global.fetch` stubs to `vi.mock("../src/utils/fetch")` is correct — `vi.hoisted` + module mock is the right Vitest pattern.
- Provider import changes are mechanical and correct.
- No protocol-critical code touched (no VTXO handling, signing, forfeit paths, etc.).

---

**Verdict:** The header injection gap on GET requests is a clear bug — the stated goal of the PR (add header to HTTP requests) is only achieved for ~64% of call sites. Please fix the wrapper to inject unconditionally.

ghost (changes requested):
## Arkana Code Review — `add digest header to http request`

### 🔴 Bug: Header not injected on GET requests (12 of 33 call sites)

**`packages/ts-sdk/src/utils/fetch.ts:7-11`**

The wrapper only adds `X-Build-Version` when `init` is truthy:

```ts
if (init) {
    init.headers = { ...init.headers, "X-Build-Version": version };
}
```

But 12 call sites across the codebase pass only a URL with no `init`:
- `ark.ts:261` — `getInfo()` (the single most important call — server identification)
- `indexer.ts:317,348,361,374,400,431,535,557,634,650` — all GET endpoints
- `delegate.ts:122` — `getDelegateInfo()`

These requests will **never** carry the version header. Fix:

```ts
export function fetch(input: RequestInfo, init?: RequestInit): Promise<Response> {
    if (typeof globalThis.fetch !== "function") {
        throw new Error("Fetch API is not available in this environment.");
    }
    const headers = new Headers(init?.headers);
    headers.set("X-Build-Version", version);
    return globalThis.fetch(input, { ...init, headers });
}
```

This handles all three `HeadersInit` shapes (`Headers`, `string[][]`, `Record<string,string>`), never mutates the caller's `init` object, and injects the header on every request.

---

### 🟡 `init.headers` spread drops `Headers` instances

**`packages/ts-sdk/src/utils/fetch.ts:8-10`**

`RequestInit.headers` is typed `HeadersInit = Headers | string[][] | Record<string, string>`. Spreading a `Headers` object produces `{}`, silently dropping every header the caller set. No current caller uses `new Headers()`, but this is a type-unsafe footgun. The `new Headers(init?.headers)` approach above resolves this.

---

### 🟡 Caller's `init` object is mutated

**`packages/ts-sdk/src/utils/fetch.ts:8`**

`init.headers = { ... }` mutates the caller's object. If any caller reuses an `init` config across calls (common pattern), headers accumulate or behave unpredictably. The fix above (`{ ...init, headers }`) avoids mutation.

---

### 🟡 Hardcoded version `"0.9.7"` ≠ package version `"0.4.34"`

**`packages/ts-sdk/src/utils/fetch.ts:1`**

The version string is hardcoded and already out of sync with `package.json` (`0.4.34`). This will silently drift on every release. Either import it from `package.json` or inject it at build time.

---

### 🟢 Everything else looks good

- Test refactoring from `global.fetch` stubs to `vi.mock("../src/utils/fetch")` is correct — `vi.hoisted` + module mock is the right Vitest pattern.
- Provider import changes are mechanical and correct.
- No protocol-critical code touched (no VTXO handling, signing, forfeit paths, etc.).

---

**Verdict:** The header injection gap on GET requests is a clear bug — the stated goal of the PR (add header to HTTP requests) is only achieved for ~64% of call sites. Please fix the wrapper to inject unconditionally.

ghost (approved):
## Arkana Code Review — Follow-up on `arkana fixes` commit

**Commit reviewed:** `8224a35` ("arkana fixes")

### ✅ All three previous findings are resolved

The fetch wrapper now correctly:
1. Injects `X-Build-Version` on **all** requests, including bare `fetch(url)` GETs — the `new Headers(init?.headers)` pattern handles `undefined` gracefully.
2. Properly handles all `HeadersInit` shapes (`Headers`, `string[][]`, `Record<string,string>`) via the `new Headers()` constructor.
3. Avoids mutating the caller's `init` object via `{ ...init, headers }`.

Clean, minimal fix. Exactly what was needed.

### 🟡 Still open: hardcoded version `"0.9.7"` ≠ `package.json` `"0.4.34"`

**`packages/ts-sdk/src/utils/fetch.ts:1`** — This was flagged in my first review and is still present. Not a blocker for merge, but it will send incorrect version telemetry from day one. Consider importing from `package.json` or injecting at build time in a follow-up.

---

**Verdict:** Approving. The core bug is fixed correctly. The version mismatch is a non-blocking nit that can be addressed separately.

Kukks (changes requested):
I think we need to also apply to the other providers like expo?

packages/ts-sdk/src/utils/fetch.ts
```diff
@@ -0,0 +1,10 @@
+const version = "0.9.7";
```
pietro909: won't force us to release SDK (and wallet) every time we release arkd?

bordalix: @pietro909 yes... it will be addressed in the future.

bordalix:
> I think we need to also apply to the other providers like expo?

@Kukks @pietro909 fixed by 159c7f7d

ghost (approved):
## Arkana Code Review — Follow-up on `add header to expo fetch` commit

**Commit reviewed:** `159c7f7d` ("add header to expo fetch")

### ✅ Addresses Kukks' feedback — expo path now covered

**`expoUtils.ts:11-15`** — The expo fetch wrapper correctly mirrors the pattern from `utils/fetch.ts`:
- `new Headers(init?.headers)` handles undefined/existing headers safely
- `headers.set("X-Build-Version", buildVersion)` injects the header
- Spread `{ ...init, headers }` avoids mutating the caller's `init`

**`fetch.ts:1`** — `buildVersion` exported (renamed from `version`) so expoUtils can import it. Single source of truth — no duplication.

**Fallback path (`expoUtils.ts:30`)** returns the wrapped `fetch` from utils, which already injects the header. Both expo and non-expo paths are covered.

No protocol-critical changes. LGTM.

pietro909 (approved):

PR #558, opened 2026-06-11
Add optimistic Lightning payment settlement with waitFor option

tiero:
## Summary
Adds support for optimistic Lightning payment settlement in submarine swaps, allowing callers to resolve as soon as the payment is in flight rather than waiting for full settlement. This enables UIs to show an optimistic "sent" state immediately while monitoring continues in the background.

## Key Changes

- **New `waitFor` parameter** on `sendLightningPayment()`: accepts `"settled"` (default, waits for `transaction.claimed`) or `"funded"` (resolves at `transaction.mempool` or later)
- **New `waitForSwapFunded()` method**: public API to wait until a submarine swap's lockup transaction is observed, without requiring the preimage
- **Refactored settlement monitoring**: unified `waitForSubmarineSwap()` internal method handles both optimistic and full settlement paths
- **Background persistence**: after optimistic resolution, monitoring continues persisting status updates (preimage on claim, refundable flag on failure) so the stored swap stays current for SwapManager
- **New `refreshSwapsStatus()` logic**: fetches and persists preimages for swaps that settled while the app was closed (e.g., after a "funded" optimistic send and restart)
- **New `SubmarineProgressionStatus` type**: defines the canonical successful swap progression for validating optimistic resolution targets
- **Updated response types**: `OptimisticSendLightningPaymentResponse` (with optional preimage) for `waitFor: "funded"` calls
- **Service worker and Expo wrapper updates**: propagate new method and overloaded signatures through abstraction layers

## Notable Implementation Details

- Optimistic resolution deliberately does NOT set the "final" flag, allowing background monitoring to persist subsequent statuses
- Monitor errors after optimistic resolution are logged but don't reject the already-settled promise
- Persistence failures in the status update callback are caught to prevent unhandled rejections
- The progression-based status checking (`hasSubmarineStatusReached`) handles skipped statuses since subscriptions only report the current one
- Documentation updated with examples and guidance on using SwapManager for automatic refund handling

https://claude.ai/code/session_01GC1cY1ZkPiEncsMxyWuisN

### DIFF
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -72,8 +72,10 @@ const result = await swaps.sendLightningPayment(
 );
 // Resolves as soon as the status is observed (or any later status in the
 // lifecycle — intermediate statuses can be skipped). result.preimage is
-// undefined when the swap hasn't settled yet; keep the SwapManager enabled
-// so settlement tracking and auto-refunds continue in the background.
+// undefined when the swap hasn't settled yet. Monitoring continues in the
+// background and keeps the stored swap up to date until a terminal status,
+// but a late failure no longer rejects — keep the SwapManager enabled so
+// auto-refunds are handled for you.
 ```
 
 The same option is accepted by `waitForSwapSettlement(pendingSwap, options)`.
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -81,6 +81,12 @@ const result = await swaps.sendLightningPayment({
 The same milestone is exposed for a swap you already hold as
 `waitForSwapFunded(pendingSwap)`, alongside `waitForSwapSettlement(pendingSwap)`.
 
+If the app restarts before the swap settles, the in-process monitoring is gone — call
+`refreshSwapsStatus()` on startup to reconcile: it polls every pending swap and persists
+the latest status, including the preimage for swaps that settled while the app was
+closed. Read the swap back from the repository (e.g. `getSwapHistory()`) to flip a
+pending UI row to "completed".
+
 ### ARK to BTC
 
 ```typescript
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -61,6 +61,23 @@ console.log('Paid:', result.txid);
 // SwapManager auto-refunds if payment fails
 ```
 
+By default the call resolves only once the swap fully settles (`transaction.claimed`),
+i.e. after Boltz has swept the HTLC. To show an optimistic "sent" state as soon as the
+payment is in flight — like most Lightning wallets — pass `optimisticResolveAt`:
+
+```typescript
+const result = await swaps.sendLightningPayment(
+  { invoice: 'lnbc500u1pj...' },
+  { optimisticResolveAt: 'invoice.pending' },
+);
+// Resolves as soon as the status is observed (or any later status in the
+// lifecycle — intermediate statuses can be skipped). result.preimage is
+// undefined when the swap hasn't settled yet; keep the SwapManager enabled
+// so settlement tracking and auto-refunds continue in the background.
+```
+
+The same option is accepted by `waitForSwapSettlement(pendingSwap, options)`.
+
 ### ARK to BTC
 
 ```typescript
--- a/packages/boltz-swap/README.md
+++ b/packages/boltz-swap/README.md
@@ -63,22 +63,23 @@ console.log('Paid:', result.txid);
 
 By default the call resolves only once the swap fully settles (`transaction.claimed`),
 i.e. after Boltz has swept the HTLC. To show an optimistic "sent" state as soon as the
-payment is in flight — like most Lightning wallets — pass `optimisticResolveAt`:
+payment is in flight — like most Lightning wallets — pass `waitFor: 'funded'`:
 
 ```typescript
-const result = await swaps.sendLightningPayment(
-  { invoice: 'lnbc500u1pj...' },
-  { optimisticResolveAt: 'invoice.pending' },
-);
-// Resolves as soon as the status is observed (or any later status in the
-// lifecycle — intermediate statuses can be skipped). result.preimage is
-// undefined when the swap hasn't settled yet. Monitoring continues in the
+const result = await swaps.sendLightningPayment({
+  invoice: 'lnbc500u1pj...',
+  waitFor: 'funded',
+});
+// Resolves as soon as the lockup transaction is observed: the funds are
+// committed and the swap is refundable from here. result.preimage is
+// undefined since the swap hasn't settled yet. Monitoring continues in the
 // background and keeps the stored swap up to date until a terminal status,
 // but a late failure no longer rejects — keep the SwapManager enabled so
 // auto-refunds are handled for you.
 ```
 
-The same option is accepted by `waitForSwapSettlement(pendingSwap, options)`.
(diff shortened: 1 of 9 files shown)


ghost (changes requested):
## 🔒 Protocol-Critical Review — Optimistic Lightning Settlement

**Verdict: Request changes + mandatory human sign-off.**

This PR changes when a submarine swap promise resolves — from "payment confirmed" to "funds committed but outcome unknown." That's a material shift in the trust/UX boundary. The implementation is well-structured, but I have findings that need addressing before merge.

---

### 🚨 FINDING 1 — No auto-refund after optimistic resolution (by design, but under-documented at call site)

**`arkade-swaps.ts:783-789`**

```typescript
if (args.waitFor === "funded") {
    await this.waitForSwapFunded(pendingSwap);
    return { amount: pendingSwap.response.expectedAmount, txid };
}
```

When `waitFor: "funded"`, the function returns inside the `try` block. The `catch` block (line 797) that auto-refunds via `refundVHTLC` is **never reached** for failures that arrive after the optimistic resolution. This is correct behavior — the background monitor persists `refundable: true` and SwapManager is expected to handle it — but the JSDoc on `sendLightningPayment` (line 753) says:

> "Auto-refunds on failures observed before the promise resolves."

This is technically true but buries the lede. A caller reading only the JSDoc could miss that **most failures happen AFTER funding** (e.g. `invoice.failedToPay` always follows `transaction.mempool` in the lifecycle). Suggest making the JSDoc explicit:

```
* ⚠️ With `waitFor: "funded"`, auto-refund only covers the narrow window
* before the lockup tx is observed. Most failures (routing failure, expiry)
* occur AFTER this point — ensure SwapManager is enabled for automatic
* refund handling, or monitor the swap status yourself.
```

### ⚠️ FINDING 2 — `waitForSwapFunded` @throws JSDoc is misleading

**`arkade-swaps.ts:1449-1451`**

```
* @throws {SwapExpiredError} If the swap expires before funding.
* @throws {InvoiceFailedToPayError} If Boltz fails to route the payment.
* @throws {TransactionLockupFailedError} If the lockup transaction fails.
```

`InvoiceFailedToPayError` almost always arrives **after** `transaction.mempool` in the Boltz lifecycle (lockup → route attempt → failure). Once the promise is optimistically resolved, this rejection is swallowed. The `@throws` tag implies the caller should catch it, but they can't — it's a no-op on the settled promise.

Suggest: either remove `InvoiceFailedToPayError` and `TransactionLockupFailedError` from `@throws` (since they're vanishingly unlikely before funding), or qualify them with "only if observed before the lockup transaction."

### ⚠️ FINDING 3 — Concurrent `onStatusUpdate` interleaving is safe but fragile

**`arkade-swaps.ts:1551-1559`**

```typescript
.monitorSwap(pendingSwap.id, (status) => {
    onStatusUpdate(status).catch((error) => ...);
})
```

Since `monitorSwap` doesn't `await` the callback, two status updates can interleave during `await saveStatus()`. Example: `transaction.mempool` enters default → yields at `saveStatus()` → `invoice.failedToPay` fires → sets `isFinal=true`, `isSettled=true` → `reject()` → mempool resumes → `resolve()` is a no-op.

This is **currently safe** because Promises can only settle once, and the `isFinal` guard prevents post-terminal processing. But it's subtle enough that a future maintainer could break it. Suggest adding an inline comment at the `isFinal`/`isSettled` declarations:

```
// CONCURRENCY NOTE: monitorSwap fires callbacks without awaiting them,
// so two onStatusUpdate calls can interleave across await points. Safety
// relies on: (1) Promise single-settlement semantics, (2) isFinal gating
// at the top of onStatusUpdate. Do not add early-returns or state mutations
// between the isFinal check and the switch without considering interleaving.
```

### 📋 FINDING 4 — Background monitor lifetime is unbounded

**`arkade-swaps.ts:1551-1568`**

After optimistic resolution, the `monitorSwap` WebSocket/polling continues until a terminal status. If the swap hangs in a non-terminal state (e.g. `invoice.pending` indefinitely due to a Boltz bug), the monitor runs forever — holding a WebSocket connection and never cleaning up.

For `waitFor: "settled"` this is acceptable (the caller is actively waiting). For `waitFor: "funded"` the caller has moved on and has no handle to cancel. Consider:
- A timeout on the background monitor (e.g. swap expiry time + buffer)
- Or at minimum, document this as a known limitation

### ✅ FINDING 5 — `hasSubmarineStatusReached` is correct

**`boltz-swap-provider.ts:112-118`**

The progression array and index comparison correctly handle:
- Skipped statuses (later status index ≥ target index → resolves)
- Failure statuses not in the progression (indexOf returns -1 → `false` → no false resolution)
- Terminal status `transaction.claimed` is in the switch's explicit `case`, never hits default → correct separation

### ✅ FINDING 6 — `refreshSwapsStatus` preimage reconciliation is sound

**`arkade-swaps.ts:2841-2873`**

The new code correctly:
- Only fetches preimage when `isSubmarineSuccessStatus(status) && !swap.preimage`
- Gracefully handles preimage fetch failure (warns, still persists status)
- Doesn't block status updates on preimage availability

### ✅ FINDING 7 — Cross-repo impact is minimal

Checked all repos under `/root/arkana/repos/`:
- **coinflip**: Calls `sendLightningPayment({ invoice })` without `waitFor` → hits first overload → `SendLightningPaymentResponse` with required `preimage`. **No breakage.**
- **coinflip e2e test**: Asserts `result.preimage.match(...)` — also calls without `waitFor`. **No breakage.**
- **rust-sdk / dotnet-sdk**: Parallel implementations, not directly affected. Should add equivalent feature if this lands.
- **IArkadeSwaps interface**: Updated with new overloads and `waitForSwapFunded`. Service worker wrappers correctly propagated. ✓

### ✅ FINDING 8 — Test coverage is strong

270 lines of new tests covering:
- Optimistic resolution at `transaction.mempool` ✓
- Resolution at a later-than-target status (skipped statuses) ✓
- No resolution on pre-target statuses ✓
- Rejection on pre-funding failures ✓
- Late failure persistence after resolution ✓
- Late settlement preimage persistence ✓
- Monitor errors after resolution (swallowed) ✓
- `refreshSwapsStatus` preimage fetch and failure ✓

**Missing test**: What happens if `saveStatus()` throws during the background monitor after optimistic resolution? The `.catch()` wrapper logs it, but the test doesn't verify the promise stays resolved (not re-rejected).

---

### Summary

| # | Severity | Finding | Action |
|---|----------|---------|--------|
| 1 | 🚨 Medium | Auto-refund scope under-documented for optimistic path | Strengthen JSDoc |
| 2 | ⚠️ Low | `@throws` on `waitForSwapFunded` misleading | Fix or qualify |
| 3 | ⚠️ Low | Concurrent callback interleaving is safe but fragile | Add concurrency comment |
| 4 | 📋 Info | Unbounded background monitor lifetime | Document or add timeout |
| 5-8 | ✅ | Correct | — |

**The code is well-written and the test coverage is thorough.** Findings 1-2 are documentation/JSDoc clarity issues, not logic bugs. Finding 3 is a maintainability concern. Finding 4 is a design consideration worth discussing.

**However: this is protocol-critical code.** It changes when users receive confirmation of fund commitment in submarine swaps, and the refund path for the optimistic case depends entirely on SwapManager being enabled. **Requesting human sign-off before merge**, per protocol-critical review policy.

🤖 Reviewed by Arkana (arkade-os/ts-sdk code review agent)

PR #560, opened 2026-06-12
fix: keep watching boarding addresses across operator signer rotation

Kukks:
## Problem

`getBoardingTapscripts()` is the single enumerator every boarding monitoring / fund-discovery path funnels through — `getBoardingAddresses`, `getBoardingUtxos`, `getBoardingTxs`, `notifyIncomingFunds` (the on-chain `watchAddresses` set), the `VtxoManager` boarding poll, balance, and sweep.

It filtered persisted `boarding` rows on the **current** signer key only:

```ts
if (c.params.serverPubKey !== serverPubKeyHex) continue;
```

When the operator rotates its signer key (`arkd` `ArkInfo.signerPubkey` changes; the old key moves into `ArkInfo.deprecatedSigners`), the wallet rebuilds its boarding tapscript with the new key. Every previously-persisted boarding row carries the **old** key, so all of them were skipped — a boarding address funded under the previous signer silently dropped out of the watch set, balance, history, and sweep, even though it still holds coins and still needs either a cooperative board before its `cutoffDate` or a unilateral CSV exit afterward.

In practice this shows up as two `active` boarding contracts in the repo where only one is monitored:

| address | created | csvTimelock |
|---|---|---|
| `ark1qrxn43a…` | 1h ago | 4194338 |
| `ark1qra883h…` | 2h ago | 4194338 |

## Fix

Keep a boarding row when its `serverPubKey` is the operator's **current** signer **or one of its deprecated signers** (`current ∪ deprecated`) — mirroring how VTXO discovery (`default`/`delegate` handlers, and `NArk`'s `IndexerVtxoDiscoveryProvider`) already scopes signers. A genuinely different operator's key is still excluded, preserving the guard against a repo recovered against another server.

The deprecated-signer set is parsed from `ArkInfo.deprecatedSigners` and cached at wallet construction from the info snapshot setup already fetches, so the hot boarding read path stays repo-only (offline-first) — no new network round-trip.

### Deliberately out of scope
- **`resolveBoardingBootTapscript`** (the display/receive address) stays pinned to the current signer: new deposits must not be advertised to a deprecated-key address.
- **Boarding discovery** (`BoardingContractHandler.discoverAt`) is unchanged — current-signer-only on a fresh-repo restore, consistent with `NArk`'s `BoardingUtxoDiscoveryProvider`. A continuous repo already has the row persisted, which this PR now keeps monitoring.

## Test plan
- New unit test in `walletBoardingRotation.test.ts`: a boarding row under a now-deprecated signer is watched (`getBoardingAddresses` returns it). Red→green verified.
- Existing "different ASP server" test still passes (foreign key still excluded).
- Full ts-sdk unit suite: **1416 passed, 1 skipped, 0 failed**.
- `tsc --noEmit` clean; prettier clean.

## Summary by CodeRabbit

* **Improvements**
  * Extended boarding address monitoring to include UTXOs funded under deprecated operator signer keys, enabling proper wallet recovery during operator key rotation.

* **Tests**
  * Added test coverage for deprecated signer key rotation scenarios.

### DIFF
--- a/packages/ts-sdk/src/wallet/wallet.ts
+++ b/packages/ts-sdk/src/wallet/wallet.ts
@@ -306,6 +306,18 @@ export class ReadonlyWallet implements IReadonlyWallet {
      */
     protected _boardingTapscript: DefaultVtxo.Script;
 
+    /**
+     * x-only hex of the operator's deprecated signer keys (from
+     * {@link ArkInfo.deprecatedSigners}). {@link getBoardingTapscripts} keeps a
+     * persisted boarding row whose `serverPubKey` is the current signer OR a
+     * member of this set, so funds at a boarding address derived under a
+     * now-rotated operator signer stay watched — while a row from a genuinely
+     * different operator is still excluded. Captured once at setup from the
+     * server-info snapshot so the hot boarding read path stays repo-only
+     * (offline-first).
+     */
+    protected readonly _deprecatedSignerHexes: ReadonlySet<string>;
+
     protected constructor(
         readonly identity: ReadonlyIdentity,
         readonly network: Network,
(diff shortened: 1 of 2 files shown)


Kukks:
Superseded by #554 (arkd signer-rotation support), which reworks getBoardingTapscripts with a parameterized signer set. Note: #554 currently widens only the migration path (getBoardingUtxosForSigners); the on-chain watch path (getBoardingAddresses/notifyIncomingFunds/balance) still defaults to current-signer-only, so the original symptom here — not watching multiple active boarding addresses across a signer rotation — must still be carried into #554 (cached current∪deprecated signer set + read/watch-path widening). Folding this work into #554 rather than landing separately. Branch retained.

PR #562, opened 2026-06-13
fix(electrum): de-flake getTxStatus by classifying RPCError-wrapped not-in-block errors

Kukks:
## Problem

`electrum.test.ts > getTxStatus > tracks a freshly-funded address …` flakes intermittently in CI with:

```
Error: No confirmed transaction matching the requested hash was found
```

It's the only e2e test that calls `blockchain.transaction.get_merkle`, which needs the funding tx *mined and merkle-indexed*. Right after a fresh block, electrs exposes the height via `listunspent` before `get_merkle` can answer, so `get_merkle` rejects — the index-lag race.

`getTxStatus` → `fetchTxMerkle` is *designed* to swallow that and return `{ confirmed: false }` (via `isTxNotInBlockError` / `isMissingHeightError`). **The classifiers only inspected `err.message`.** But `ws-electrumx-client` wraps a JSON-RPC error whose payload carries a numeric `code` as an `RPCError`: the human-readable server text moves to a non-standard **`.str`** field while `.message` becomes a generic `"<code-name> (code: N)"`. When the rejection arrives in that shape, the matcher misses it, the error escapes `getTxStatus`, and the test fails. The `"code"`-or-not variance under CI load is the intermittency.

This has recurred across multiple PRs' CI runs.

## Fix

Extract text from **both `.message` and `.str`** (plus the raw string form) via a shared `errorText(err)` helper, and use it in `isTxNotInBlockError` and `isMissingHeightError`. The not-in-block / missing-height races are now recognized regardless of how the transport wrapped the error, so `getTxStatus` reliably returns `{ confirmed: false }` — for every caller, not just the test. Genuine errors (auth/network/malformed) still propagate.

## Test plan

- New unit test reproducing the `RPCError`-wrapped variant (`.message` generic, text in `.str`) — red before the fix, green after.
- Existing `getTxStatus` swallow/propagate cases unchanged and passing.
- Full ts-sdk unit suite: 1416 passed / 0 failed; `tsc --noEmit` clean; prettier clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error detection in transaction status checks to handle server responses more reliably across different error formats.

* **Tests**
  * Added test coverage for transaction status checking with edge case error scenarios.

### DIFF
--- a/packages/ts-sdk/src/providers/electrum.ts
+++ b/packages/ts-sdk/src/providers/electrum.ts
@@ -1006,6 +1006,24 @@ function isHeaderSubscribeResult(v: unknown): v is HeaderSubscribeResult {
 }
 
 /**
+ * Best-effort text extraction from a thrown value. `ws-electrumx-client` wraps a
+ * JSON-RPC error whose payload carries a numeric `code` as an `RPCError`: the
+ * human-readable server text moves to a non-standard `.str` field while
+ * `.message` becomes a generic "<code-name> (code: N)". A classifier that reads
+ * only `.message` therefore misses the real reason, so pull from both (plus the
+ * raw string form) — the matchers below must see the server wording regardless
+ * of how the transport wrapped the error.
+ */
+function errorText(err: unknown): string {
+    if (typeof err === "string") return err;
+    if (err && typeof err === "object") {
+        const e = err as { message?: unknown; str?: unknown };
+        return [e.message, e.str].filter((v): v is string => typeof v === "string").join(" ");
+    }
+    return "";
+}
+
+/**
  * Recognise the "block header not yet indexable" failure shape returned by
  * electrum servers (electrs in particular) when `block.header(N)` runs
  * against a height that's already in `listunspent`/`get_merkle` but hasn't
@@ -1014,8 +1032,7 @@ function isHeaderSubscribeResult(v: unknown): v is HeaderSubscribeResult {
  * failures (auth/network) propagate.
  */
 function isMissingHeightError(err: unknown): boolean {
-    const msg = err instanceof Error ? err.message : typeof err === "string" ? err : "";
-    return msg.toLowerCase().includes("missingheight");
+    return errorText(err).toLowerCase().includes("missingheight");
 }
 
 /**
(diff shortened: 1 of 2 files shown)


ghost (approved):
## Review: APPROVE ✅

Verified against `ws-electrumx-client`'s `RPCError` source (`build/main/lib/rpc-error.js`). The bug is real and the fix is correct.

### Root cause confirmed

`RPCError` extends `Error` and does:
```js
super(`${message} (code: ${code})`);  // .message = generic mapped text
this.str = str;                        // .str = raw server error string
```

The old classifiers only read `.message` (via `err instanceof Error ? err.message : ...`), so the actual server text like `"No confirmed transaction matching the requested hash was found"` was invisible to the matchers when the error arrived as an `RPCError`.

### What I checked

1. **`errorText()` helper** — clean extraction from `.message`, `.str`, and raw strings. The `.join(" ")` + `.includes()` approach is safe; no risk of false positives from concatenation since the match strings are specific server wording.
2. **Callers of the classifiers** — `fetchTxMerkle` (line 301) and `getTxStatus` (line 878). Both use `isTxNotInBlockError` / `isMissingHeightError` as catch guards. The fix makes these guards match the same conditions they always intended to match — just now they also work when the transport wraps the error.
3. **Downstream impact** — `getTxStatus` is called from `wallet/unroll.ts` (lines 130, 263, 350) for VTXO confirmation tracking. A spurious throw here during index-lag races would break unroll polling, so this fix has real protocol-level value beyond test de-flaking.
4. **No public API change** — `errorText` and the classifiers are module-private. `getTxStatus` return type and semantics are unchanged. No cross-repo impact.
5. **Test coverage** — new test correctly simulates the `RPCError` shape (`Object.assign(new Error("server error (code: 2)"), { str: "..." })`), which matches the actual constructor behavior. Existing propagation test (`"Authentication required"`) ensures genuine errors still throw.
6. **No protocol-critical surface** — this is error classification in the provider layer, not transaction construction, signing, or forfeit logic.

Clean fix, well-scoped, good test. Ship it.

PR #565, opened 2026-06-15
Avoid Codex to run in circles when hitting the sandbox

pietro909:
## Summary by CodeRabbit

* **Documentation**
  * Added local environment troubleshooting guidance for intermittently failing sandbox wrapper commands, with recommendations to cease retry attempts when the error occurs and instead use safer, alternative approaches for configuration editing and scanning operations.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -69,6 +69,10 @@ pnpm run release:cleanup                      # Auto-detect dirty release artifa
 
 Tags are `@arkade-os/sdk/<version>` and `@arkade-os/boltz-swap/<version>` (no `v<version>`). Releasing SDK implies a dependent boltz-swap release because boltz-swap depends on SDK via `workspace:*`; override with `--boltz-bump <bump-or-version>`. The script runs tests, builds, commits, tags, publishes to npm (requires local npm credentials), and pushes commit + tags to `origin`.
 
+## Local Environment Notes
+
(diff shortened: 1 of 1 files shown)


ghost (changes requested):
## Review: AGENTS.md — Codex sandbox workaround

**Verdict: Request changes** (non-blocking, minor)

This is a 4-line documentation-only change to `AGENTS.md`. No code, no protocol impact, no cross-repo breakage risk.

### Issues

**1. Too environment-specific for a shared guidance file** (`AGENTS.md:72-75`)

`AGENTS.md` is the authoritative guidance for *all* AI assistants working on this repo (Claude, Codex, Copilot, Cursor, etc.). This block is a Codex-specific workaround for `bwrap` sandbox failures in OpenAI's execution environment. Other agents don't hit this issue and the advice ("use a precise Node.js exact-replacement script") is actively unhelpful for them — it steers them away from standard tools that work fine in their environments.

**Suggestion**: Move this to a `.codex/` config, `codex.md`, or a `.agents.md` scratch file (already gitignored per the repo's own convention at the bottom of `AGENTS.md`). If it must live in `AGENTS.md`, scope it explicitly:

```markdown
## Codex-Specific Environment Notes

> The following applies only to OpenAI Codex sessions.

- In the Codex sandbox, `bwrap` can fail with ...
```

**2. The workaround advice is vague**

"use a precise Node.js exact-replacement script" — what script? If this is meant to be actionable guidance, it should either link to or inline the script pattern. As written, a Codex agent encountering this for the first time still has to figure out what to do.

### Summary

The intent is good — preventing agents from retrying broken sandbox paths is the right instinct. But the fix should be scoped to the agent that has the problem, not injected into the shared guidance file that all agents read. Either move it to a Codex-specific config or clearly scope it with a header.

No protocol-critical code touched. No security concerns.

pietro909:
will move it to my own local folders since it's too specific

PR #567, opened 2026-06-15
fix(wallet): fail fast on unsignable boarding input during settle

Kukks:
## Summary

Onboarding certain boarding UTXOs fails — after every client signature and approval *succeeds* — with the opaque operator error:

> `failed to sign commitment tx: script <x> is not a wallet script`

Root cause is a **discovery↔signing asymmetry**, not batching:

- `Wallet.getBoardingTapscripts` (fund discovery) recognizes **three** boarding-script sources: the synthetic index‑0 baseline, the current display tapscript, and every persisted `boarding` contract row. So `getBoardingUtxos` will *select* a UTXO at any of them.
- `InputSignerRouter` (signing) resolves a boarding input only via a matching contract row **or** the single static `boardingPkScript` it was constructed with. Anything else is **silently skipped** — no signature, no error (intended for connector/cosigner inputs).

A selected-but-unresolved boarding input (e.g. a rotated boarding address whose contract row is missing/unresolvable in the repo) is therefore skipped → left unsigned. On the operator: `VerifyBoardingTapscriptSigs` drops unsigned inputs (`WithSkipUnsignedInputs`), so the leaf never merges into `round.CommitmentTx`; final `SignTransaction` then hits the key‑path branch the operator doesn't own → `"is not a wallet script"`. The trigger is **which boarding address** (its script's resolvability), **not** the number of inputs batched.

## What this PR does

This is a **defensive guard / diagnostic**, not the full routing fix. `handleSettlementFinalizationEvent` now verifies each boarding input received a tapscript spend signature after routing and, if not, throws **before submission** — naming the exact outpoint, the unresolved boarding address, and the boarding addresses the wallet *does* recognize. This converts a server-side mystery (and a confusing "all signatures succeeded, yet it failed") into an actionable client-side error.

It does **not** make an unresolvable boarding address signable. The full routing fix (make the signer resolve the same boarding-script set discovery spends, and/or restore the missing contract row / re-derive its descriptor) is tracked separately — it needs a repo-state classification of the offending address to choose the correct path safely.

## Test plan

- [x] New `test/wallet/unsignableBoardingInput.test.ts` pins the diagnostic message (outpoint, unresolved address, recognized set) and that error-building never throws on an undecodable script.
- [x] The silent-skip premise is already covered by `test/wallet/inputSignerRouter.test.ts` ("skips unknown script with no contract and no boarding match").
- [x] ts-sdk full unit suite: 1505 passed, 1 skipped, 0 failed.
- [x] boltz-swap unit suite: 389 passed (pre-commit hook).
- [x] `tsc --noEmit` clean; Prettier clean.

## Notes for reviewers

- The leaf-less rejection path does **not** trigger the operator's boarding-input ban (that's the leaf-present-but-unsigned branch), so affected UTXOs are not convicted.
- No change to signing behavior on the success path: the guard only fires for an input that received no signature — which would have failed at the operator anyway.

### DIFF
--- a/packages/ts-sdk/src/wallet/wallet.ts
+++ b/packages/ts-sdk/src/wallet/wallet.ts
@@ -3036,18 +3036,9 @@ export class Wallet extends ReadonlyWallet implements IWallet {
                     settlementPsbt = await this._signerRouter.sign(settlementPsbt, [
                         { index: i, lookupScript: script },
                     ]);
-                    // Defense-in-depth: the signer router silently skips any
-                    // input whose script it cannot resolve (no matching contract
-                    // row and not the recognized boarding script). For a
-                    // boarding input the wallet itself selected, a skip means we
-                    // would submit an unsigned, leaf-less-looking input that arkd
-                    // rejects far downstream with the opaque "script ... is not a
-                    // wallet script" — after every signature already "succeeded".
-                    // Fail here instead, naming the exact outpoint and the
-                    // boarding addresses the wallet actually recognizes, so an
-                    // unsignable boarding address (e.g. a rotated boarding script
-                    // whose contract row is missing from the repo) is immediately
-                    // diagnosable rather than a server-side mystery.
+                    // The signer router silently skips inputs it can't resolve,
+                    // which would leave this boarding input unsigned and cause
+                    // arkd to reject it as "not a wallet script". Fail early.
                     if (!settlementPsbt.getInput(i).tapScriptSig?.length) {
                         throw new Error(await this.unsignableBoardingInputError(input, script));
                     }
@@ -3130,17 +3121,8 @@ export class Wallet extends ReadonlyWallet implements IWallet {
         }
     }
 
-    /**
-     * Build the diagnostic thrown when a selected boarding input could not be
-     * signed (the signer router resolved no contract row and no recognized
-     * boarding script for it). Without this, the unsigned input flows to arkd
-     * and surfaces as the opaque "script ... is not a wallet script" two RPCs
-     * later. The message names the exact outpoint, the unresolved boarding
-     * address, and the boarding addresses the wallet *does* recognize — enough
-     * to classify the offender (rotated vs baseline, missing contract row)
-     * without introspecting wallet internals. Best-effort: every enrichment is
-     * guarded so building the error can never itself throw and mask the cause.
-     */
+    // Best-effort: every enrichment is guarded so a secondary failure can't
+    // mask the original signing error.
     private async unsignableBoardingInputError(
         input: { txid: string; vout: number },
         script: Bytes,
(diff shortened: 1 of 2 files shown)


ghost (approved):
## Code Review — fix(wallet): fail fast on unsignable boarding input during settle

**Verdict: Approve ✅ — but this is protocol-critical (settlement signing path), so requires human sign-off before merge.**

### Summary

Single-commit, additions-only PR. Adds a fail-fast guard after `_signerRouter.sign()` for boarding inputs in `handleSettlementFinalizationEvent`. When the signer router silently skips a boarding input (no matching contract row, no recognized boarding script), the wallet now throws with a diagnostic error **before** submitting the unsigned input to arkd — instead of letting it fail downstream with the opaque `"script ... is not a wallet script"`.

### What I verified

1. **Guard correctness** (`wallet.ts:3019`): `!settlementPsbt.getInput(i).tapScriptSig?.length` is the right check. Boarding inputs are signed via tapscript path (forfeit leaf is set on line 3008 via `tapLeafScript`), so a successful sign always produces `tapScriptSig` entries. Both `undefined` and empty-array are caught by `?.length` being falsy. The check runs on the *returned* transaction from `sign()`, not the pre-sign copy. Correct.

2. **No false positives on success path**: The guard sits between `sign()` and `hasBoardingUtxos = true; break;`. On a successful sign, `tapScriptSig` will be populated and the guard is a no-op. Zero behavioral change on the happy path.

3. **Error builder resilience** (`unsignableBoardingInputError`): Every enrichment (`Address.encode`, `getBoardingAddress`, `getBoardingAddresses`) is individually try/caught. The method cannot throw, which means the original signing failure is never masked. The `<undecodable>` and `<unavailable>` placeholders degrade gracefully.

4. **Signer router analysis** (`inputSignerRouter.ts:105-109`): Confirmed the silent-skip behavior. When `!contract && scriptHex !== boardingScriptHex`, the input is skipped with `continue` — no error, no signature. This is the exact gap the guard catches.

5. **Test coverage**: Two tests cover (a) diagnostic message content with correct outpoint, script hex, and recognized address set, and (b) non-throwing guarantee on undecodable scripts (garbage bytes). Both use the actual `Wallet.create` factory, not mocked internals. The `(wallet as any)` cast for private method access is acceptable for unit testing.

6. **Cross-repo impact**: None. Pure internal defensive code. No public API, type, or interface changes. No consumers affected.

7. **Import changes**: The `gh pr diff` showed import diffs beyond this commit's scope (likely base branch divergence). The PR metadata confirms 56 additions / 0 deletions in `wallet.ts` — clean additive change.

### No issues found

- No race conditions (guard is synchronous relative to the signing flow)
- No performance concern (one `getInput()` call per boarding input, already O(n) loop)
- `await` in `throw new Error(await this.unsignableBoardingInputError(...))` is correct — the builder needs async for `getBoardingAddress()`/`getBoardingAddresses()`
- Error message includes all necessary diagnostic data without leaking private keys or sensitive material

### ⚠️ Protocol-critical flag

This PR modifies the settlement signing hot path (`handleSettlementFinalizationEvent`). While the change is purely defensive and cannot alter success-path behavior, any code in the boarding→commitment→settlement pipeline warrants human review. **A human must approve before merge.**

Clean, well-scoped defensive fix. Ship it after human sign-off.

🤖 Reviewed by Arkana (model: opus)

pietro909 (approved):

PR #572, opened 2026-06-18
fix: Improve MissingSigningDescriptorError message

pietro909:
- simplify verbose docs
- improve error message to include _foreign contract_ possibility, not just rotation

## Summary by CodeRabbit

* **Documentation**
  * Enhanced error messages in wallet signing operations to clarify failure causes and provide expanded recovery guidance options.

### DIFF
--- a/packages/ts-sdk/src/wallet/inputSignerRouter.ts
+++ b/packages/ts-sdk/src/wallet/inputSignerRouter.ts
@@ -71,9 +71,8 @@ export class InputSignerRouter {
      * here) so the routing rules never live in two places.
      *
      * Throws {@link MissingSigningDescriptorError} for a non-baseline
-     * default/delegate contract whose `metadata.signingDescriptor` is
-     * missing — the same condition that would later abort signing. Failing
-     * here moves the failure earlier, before any PSBT is mutated.
+     * default/delegate/boarding contract whose `metadata.signingDescriptor` is
+     * missing.
      */
     async classify(jobs: InputSigningJob[]): Promise<InputRoutingPlan> {
         const identityIndexes: number[] = [];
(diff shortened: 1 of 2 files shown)


ghost (approved):
## Review: LGTM ✅

**Scope**: Documentation and error-message-only changes across 2 files. Zero behavioral changes.

### What changed

1. **`signingErrors.ts`** — JSDoc and error message now correctly explain both causes of `MissingSigningDescriptorError`:
   - Wallet rotation on an old build (existing cause)
   - Contract belongs to a different identity / storage reuse (new cause)
   - Recovery advice adds "or delete the contract" — valid since `deleteContract` is a supported operation on `ContractRepository`

2. **`inputSignerRouter.ts`** — Two comment simplifications:
   - JSDoc on `classify()`: adds "boarding" to the contract type list (matches `DESCRIPTOR_CAPABLE_CONTRACT_TYPES` set on line 30), trims redundant prose
   - Inline comment at `ownerPubKeyHex`: removes the `hex.encode`-is-already-lowercase explanation, keeps the important bit (persisted data may be uppercase)

### Checked

- **No code changes** — only JSDoc and inline comments touched
- **Error class API unchanged** — `MissingSigningDescriptorError` constructor signature (`contractScript`, `contractType`) is identical; no downstream breakage
- **Exported surface** — `wallet.ts` re-exports the class; consumers catch by type, not by message string — safe
- **Tests** — existing tests in `inputSignerRouter.test.ts` assert on the error *class*, not the message text, so they remain valid
- **Cross-repo** — no other repos import this error type (only `ts-sdk` internal + `ts-sdk-review-427` clone)

No protocol-critical logic touched. Clean approve.

PR #576, opened 2026-06-21
feat(boltz-swap): support descriptionHash on reverse swaps

tajava2006:
## Summary

Let callers commit a SHA256 description hash into a reverse-swap BOLT11 invoice instead of a plaintext description, by adding an optional `descriptionHash` to `createLightningInvoice` / `createReverseSwap`.

Motivation: a wallet built on this SDK that wants to support NIP-57 zaps must produce an invoice whose `description_hash` equals `SHA256(zap request)`, so the kind-9735 zap receipt verifies. The Boltz backend already accepts `descriptionHash` on `POST /v2/swap/reverse` (hex, optional) and forwards it to the node's hold invoice — the SDK was the only layer dropping it, so today every reverse-swap invoice can only carry a plaintext description. This adds the missing passthrough; nothing else is needed on the Boltz/node side.

## What changed

- `CreateLightningInvoiceRequest` and the low-level `CreateReverseSwapRequest` gain an optional `descriptionHash?: string` (hex, 32 bytes).
- `BoltzSwapProvider.createReverseSwap` sends it in the request body.
- A BOLT11 invoice carries a description or a description hash, never both, so when `descriptionHash` is set the plaintext `description` (including the default `"Send to Arkade address"`) is omitted.
- Minimal validation: `descriptionHash`, when present, must be 64 hex chars.

## Not changed (scope / safety)

- No change to the preimage, payment hash, VHTLC, claim, or settlement paths — this only affects the invoice's description field.
- Behavior is identical for existing callers: when `descriptionHash` is unset, the request is byte-for-byte what it was before (default description preserved).

## Test plan

- New `BoltzSwapProvider` unit tests: (1) `descriptionHash` is sent and the plaintext description is dropped when both are supplied; (2) a non-32-byte `descriptionHash` throws.
- `boltz-swap` unit suite green (405 passed); `tsc --noEmit` clean for both packages; `tsup` build clean (JS + DTS); prettier clean.

## Note for reviewers

- I could not run the regtest integration suite locally (Docker), so the end-to-end `descriptionHash` to invoice `h` field round-trip against a live LND/CLN node is unverified here. The backend wiring is confirmed by reading `boltz-backend` (`/v2/swap/reverse` accepts `descriptionHash`; LND `addinvoice` and the CLN hold plugin embed a precomputed hash, preferring it over memo).
- When both `description` and `descriptionHash` are passed, `descriptionHash` wins and `description` is silently dropped (mirrors LND/CLN-hold behavior) rather than throwing. Flag if you'd prefer a hard error instead.

## Summary by CodeRabbit

* **New Features**
  * Reverse swaps now support an optional SHA256 description hash. When provided, it’s sent in place of the plaintext description (including when the hash is an empty string).
  * Lightning invoice creation can also commit an optional description hash, and decoded invoices now expose `descriptionHash`.
  * Input validation ensures the description hash is a valid 32-byte / 64-hex-character SHA256.
* **Tests**
  * Added unit and end-to-end coverage for description vs. descriptionHash prioritization and for validation error cases.

### DIFF
--- a/packages/boltz-swap/src/arkade-swaps.ts
+++ b/packages/boltz-swap/src/arkade-swaps.ts
@@ -479,11 +479,14 @@ export class ArkadeSwaps {
         // build request object for reverse swap. A BOLT11 invoice carries
         // either a description or a description hash, never both, so prefer
         // descriptionHash when present and drop the plaintext description.
+        // Gate on `!== undefined` (not truthiness) so a present-but-invalid
+        // hash like "" is forwarded and rejected by the provider's hex check,
+        // rather than silently falling back to the description.
         const swapRequest: CreateReverseSwapRequest = {
             invoiceAmount: args.amount,
             claimPublicKey,
             preimageHash,
-            ...(args.descriptionHash
+            ...(args.descriptionHash !== undefined
                 ? { descriptionHash: args.descriptionHash }
                 : args.description?.trim()
                   ? { description: args.description.trim() }
--- a/packages/boltz-swap/src/arkade-swaps.ts
+++ b/packages/boltz-swap/src/arkade-swaps.ts
@@ -476,12 +476,18 @@ export class ArkadeSwaps {
         const preimageHash = hex.encode(sha256(preimage));
         if (!preimageHash) throw new SwapError({ message: "Failed to get preimage hash" });
 
-        // build request object for reverse swap
+        // build request object for reverse swap. A BOLT11 invoice carries
+        // either a description or a description hash, never both, so prefer
+        // descriptionHash when present and drop the plaintext description.
         const swapRequest: CreateReverseSwapRequest = {
             invoiceAmount: args.amount,
             claimPublicKey,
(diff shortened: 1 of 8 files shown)


bordalix (changes requested):
Hey @tajava2006, thanks for your contribution.
The PR looks solid, can we have some e2e tests?
For example:
- create reverse swap with description, expects decoded invoice to have description
- create reverse swap with both, expects decoded invoice to have only descriptionHash
- create reverse swap with descriptionHash, expects decoded invoice to have descriptionHash

tajava2006:
Added the three e2e tests (description / descriptionHash / both), decoding the invoice from the regtest stack — all green. To assert on it I also surfaced `descriptionHash` from `decodeInvoice`; that lookup needs a small local cast because light-bolt11-decoder's `Section` type omits `description_hash` — left it localized rather than patching that separate package (out of scope here).

Kukks (approved):

PR #578, opened 2026-06-23
feat(intent): set BIP-322 PSBT_GLOBAL_GENERIC_SIGNED_MESSAGE (0x09)

afsheenb:
Currently intent-proof PSBTs never set BIP-322's global 0x09 field (key=PSBT_GLOBAL_GENERIC_SIGNED_MESSAGE, value=the UTF-8 message being signed). Without this field a co-signer cannot recompute the to_spend commitment from PSBT-internal data, so it cannot reliably distinguish a genuine ownership proof from an ordinary fund-moving spend before contributing a partial signature.

In this PR we thread the (already canonical) message into craftToSignTx() and set the 0x09 global field to its UTF-8 encoding, which is the same value the proof already commits to via the to_spend tagged hash.

@scure/btc-signer exposes no public setter for PSBT global fields, so the unknown global entry is written directly; 0x09 is unused in the global keymap and round-trips as an unknown field that toPSBT()/ fromPSBT() preserves.

## Summary by CodeRabbit

* **New Features**
  * Enhanced signing process to store message data within PSBT format for verifier and co-signer access

* **Tests**
  * Added comprehensive test coverage for message preservation in PSBT signing and round-trip conversion validation

### DIFF
--- a/packages/ts-sdk/src/intent/index.ts
+++ b/packages/ts-sdk/src/intent/index.ts
@@ -67,7 +67,7 @@ export namespace Intent {
         const toSpend = craftToSpendTx(message, inputs[0].witnessUtxo.script);
 
         // Create the transaction to sign.
-        return craftToSignTx(toSpend, inputs, outputs);
+        return craftToSignTx(toSpend, inputs, outputs, message);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


afsheenb:
@coderabbitai both recommended fixes applied - docstrings updated and the find condition
in globalSignedMessage has been tightened up as suggested. All good?

Issue #522, opened 2026-05-26
Provide a first-class API to reset/clear a wallet's persisted local data (stale balance survives a chain reset)


Kukks:
## Summary

There's no first-class way to **reset/clear a wallet's persisted local data**. After the backing chain is reset (e.g. a regtest wipe), or when otherwise invalidating local state, the wallet keeps reporting a **stale balance** from VTXOs that no longer exist, and the SDK offers no supported way to clear them.

## Why `clearSyncCursor()` + reconnect isn't enough

- `clearSyncCursor()` only resets the sync high-water mark; it does **not** delete stored VTXOs.
- The delta sync **upserts** what the indexer returns but never **deletes** rows the indexer stops reporting. So after a chain reset, the old VTXOs linger in the repo and `getBalance()` / `getVtxos()` keep summing them ("ghost" rows).

Net effect: a wallet pointed at a freshly-reset chain shows the **old** balance indefinitely, with no public API to fix it.

## Why clearing it externally is awkward today

- The IndexedDB repositories implement `clear()` (`IndexedDBWalletRepository`/`IndexedDBContractRepository`), but it isn't surfaced on the `Wallet` — a consumer has to reach into `wallet.walletRepository.clear()` / `wallet.contractRepository.clear()` directly, which feels like touching internals.
- Deleting the DB out-of-band via `indexedDB.deleteDatabase(name)` is **blocked**: `openDatabase` (repositories/indexedDB/manager.ts) doesn't register `onversionchange`/`onclose` to close the live connection, so the delete request blocks indefinitely and a blocked request is abandoned on navigation without finalizing.
- `wallet.dispose()` only disposes the ContractManager (stops the watcher); it does **not** close the repository connections, so even after `dispose()` the IndexedDB handle stays open.

## Requests (any subset)

1. **Expose a reset API on `Wallet`**, e.g. `wallet.clearLocalData()` / `resetLocalState()`, that clears the wallet + contract repositories (the repos already implement `clear()`). This is the smallest, most direct fix.
2. **Reconcile deletions during sync** — drop local VTXOs the indexer no longer reports for owned scripts, so a chain reset/reorg is reflected without manual intervention. (Overlaps with #480 "ghost unspent rows" and #477's `isSpent` upsert.)
[...]

PR #579, opened 2026-06-23
feat(wallet): reset persisted local data + unblock IndexedDB deletion

Dunsin-cyber:
part of #522

ffter a chain reset, network switch, or logout, the wallet shows a stale balance from coins that no longer exist, with no supported way to clear it.

**What this does**
- Makes `clear()` a full reset on every wallet type + the shared interface: wipes cached coins, history, sync cursor, and contracts in one call.
- Stops the background sync first so nothing re-saves stale data mid-wipe.
- Unblocks `indexedDB.deleteDatabase()` (it used to hang because the wallet never released the DB connection).
- Folds the full reset into the existing service-worker `clear()` (it previously deleted only the current address's VTXOs).

**Scope**
- Covers requests 1 and 3 of #522.
- Follow-up (separate PR): sync auto-drops coins the server stops reporting, so the stale balance self-heals.

## Summary by CodeRabbit

* **New Features**
  * Added/extended a public `clear()` wallet action to fully wipe locally persisted wallet data (including cached VTXOs and stored contracts) across read-only, service worker, and Expo wallet flows.
* **Bug Fixes**
  * Improved IndexedDB behavior on `versionchange` to avoid stale open connections blocking database deletion.
  * Expo background shim `clear` now consistently follows the expected “not implemented” behavior.
* **Tests**
  * Added coverage for IndexedDB deletion unblocking and for `clear()` removing cached VTXOs and stored contracts.
* **Chores**
  * Updated the `regtest` subproject pointer.

### DIFF
--- a/packages/ts-sdk/src/wallet/expo/wallet.ts
+++ b/packages/ts-sdk/src/wallet/expo/wallet.ts
@@ -260,13 +260,13 @@ export class ExpoWallet implements IWallet {
      * Stop foreground polling and wipe all locally persisted wallet data.
      * Does not unregister the OS background task or persisted queue config.
      */
-    async clearLocalData(): Promise<void> {
+    async clear(): Promise<void> {
         if (this.foregroundIntervalId) {
             clearInterval(this.foregroundIntervalId);
             this.foregroundIntervalId = undefined;
         }
 
-        await this.wallet.clearLocalData();
+        await this.wallet.clear();
     }
 
     // ── IWallet delegation ───────────────────────────────────────────
--- a/packages/ts-sdk/src/wallet/expo/wallet.ts
+++ b/packages/ts-sdk/src/wallet/expo/wallet.ts
@@ -118,6 +118,8 @@ export class ExpoWallet implements IWallet {
     readonly indexerProvider: Wallet["indexerProvider"];
 
     private foregroundIntervalId?: ReturnType<typeof setInterval>;
+    private foregroundPollChain: Promise<void> = Promise.resolve();
+    private cleared = false;
 
     private constructor(
         private readonly wallet: Wallet,
@@ -207,11 +209,16 @@ export class ExpoWallet implements IWallet {
 
     private startForegroundPolling(intervalMs: number): void {
         this.foregroundIntervalId = setInterval(() => {
-            this.runForegroundPoll().catch(console.error);
+            // Serialize polls so clear() can await the in-flight one before wiping.
+            this.foregroundPollChain = this.foregroundPollChain
+                .then(() => this.runForegroundPoll())
+                .catch(console.error);
         }, intervalMs);
     }
 
     private async runForegroundPoll(): Promise<void> {
+        if (this.cleared) return;
+
         await runTasks(this.taskQueue, this.processors, this.deps);
 
         // Consume results immediately (no background handoff needed)
@@ -219,6 +226,8 @@ export class ExpoWallet implements IWallet {
         if (results.length > 0) {
             await this.taskQueue.acknowledgeResults(results.map((r) => r.id));
         }
+
+        if (this.cleared) return;
 
         // Re-seed for the next tick
         await this.seedContractPollTask();
@@ -235,6 +244,13 @@ export class ExpoWallet implements IWallet {
             createdAt: Date.now(),
         };
         await this.taskQueue.addTask(task);
+    }
+
+    private async removeContractPollTasks(): Promise<void> {
+        const tasks = await this.taskQueue.getTasks(CONTRACT_POLL_TASK_TYPE);
+        for (const task of tasks) {
+            await this.taskQueue.removeTask(task.id);
+        }
     }
 
     // ── Lifecycle ────────────────────────────────────────────────────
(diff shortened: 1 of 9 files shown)


Kukks (approved):

packages/ts-sdk/src/wallet/index.ts
```diff
@@ -894,4 +894,10 @@ export interface IReadonlyWallet {
 
     /** Readonly asset manager bound to this wallet instance. */
     assetManager: IReadonlyAssetManager;
+
+    /**
+     * Wipe all locally persisted wallet data (VTXOs, UTXOs, history, sync
+     * cursor, contracts). Create a fresh wallet instance afterward.
+     */
+    clearLocalData(): Promise<void>;
```
pietro909: The docs say "Create a fresh wallet instance afterward." but this method only wipes the storage, doesn't create anything.

Am I wrong @Dunsin-cyber ?

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
```diff
@@ -934,6 +934,20 @@ export class ServiceWorkerReadonlyWallet implements IReadonlyWallet {
...
+     * (VTXOs, UTXOs, history, sync cursor, contracts). Re-init the wallet after.
+     */
+    async clearLocalData(): Promise<void> {
+        const message: RequestClear = {
+            id: getRandomId(),
+            tag: this.messageTag,
+            type: "CLEAR",
+        };
+        await this.sendMessage(message);
+        await Promise.all([this.walletRepository.clear(), this.contractRepository.clear()]);
```
pietro909: This runs page-side: its job is only to  send the message to the worker, where the logic is. 

The fact that repositories instances are available and share the same underlying IndexedDB is incidental and shouldn't be relied on.

pietro909: This runs page-side: its job is only to  send the message to the worker, where the logic is. 

The fact that repositories instances are available and share the same underlying IndexedDB is incidental and shouldn't be relied on.

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
```diff
@@ -966,7 +980,7 @@ export class ServiceWorkerReadonlyWallet implements IReadonlyWallet {
         }
     }
 
-    /** Clear cached wallet state from both the page and service worker storage. */
+    /** @deprecated Use {@link clearLocalData} for a full reset; clear() leaves contract rows. */
     async clear() {
```
pietro909: why not reuse `clear()` rather than introduce a new method?

pietro909 (changes requested):
Overall good direction, the two main issues I have are:

1. logic lives in src/wallet/wallet.ts only, not in the service worker wrapper 
2. having `clear` and `clearLocalData` is code smell because they are supposed to do exactly the same thing

packages/ts-sdk/src/wallet/index.ts
Dunsin-cyber: yep, you're right, that doc line is misleading.

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
Dunsin-cyber: i introduced a separate method to avoid breaking existing implementations.  

the new modification wipes all locally persisted data so i kept it isolated from `clear()` out of caution. happy to fold it into `clear()` directly if the implications are acceptable.

pietro909: Thank you, I appreciate your note. I believe it's ok to use `clear()` on Wallet since this looks more of a fix for that functionality than a new one.

Dunsin-cyber: okay, got it

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
```diff
@@ -980,22 +966,18 @@ export class ServiceWorkerReadonlyWallet implements IReadonlyWallet {
...
-        // Clear page-side storage to maintain parity with SW
-        try {
-            const address = await this.getAddress();
-            await this.walletRepository.deleteVtxos(address);
-        } catch (_) {
-            console.warn("Failed to clear vtxos from wallet repository");
-        }
-
         await this.sendMessage(message);
+        await Promise.all([this.walletRepository.clear(), this.contractRepository.clear()]);
```
pietro909: this part must occur in wallet/wallet.ts not here. This is the page-facing part.

Dunsin-cyber: yes. got it. i followed the existing clear() here, which already wiped the data straight from the page side, so i carried the same pattern over instead of questioning it.

packages/ts-sdk/src/wallet/expo/wallet.ts
```diff
@@ -256,6 +272,23 @@ export class ExpoWallet implements IWallet {
...
+        this.cleared = true;
+
+        if (this.foregroundIntervalId) {
+            clearInterval(this.foregroundIntervalId);
+            this.foregroundIntervalId = undefined;
+        }
+
+        await this.foregroundPollChain;
+        await this.wallet.clear();
+        await this.removeContractPollTasks();
```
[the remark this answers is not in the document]
Dunsin-cyber: the `cleared` flag set at the top of `clear()` already stops every foreground poll at its entry guard, so no queued task can repopulate data during the wipe and reordering the drain has no effect.

pietro909 (approved):

PR #583, opened 2026-06-24
feat(boltz-swap): swap activity resolver + claimTxid plumbing for all swap types

Kukks:
## Swap activity resolver (Phase 2 of the activity-history feature)

**Stacked on #582** (the activity-history framework) — base branch is `feat/activity-history`.

### What
A `swapActivityResolver(swapRepo)` for the SDK's `wallet.getActivityHistory()` (#582) that labels a wallet transaction as the Boltz swap it settled:

```ts
wallet.activity.use(swapActivityResolver(swapRepository));
// the claim tx of a swap now renders as one "Lightning swap" / "Chain swap" row
```

`prepare()` indexes the swap repository by each swap's `claimTxid`; `resolve()` is a pure, synchronous lookup against `ArkTransaction.key`.

### How (the correlation)
A wallet tx correlates to a swap by the swap's on-chain claim/settlement txid, which the swap record didn't previously persist. This PR adds `claimTxid?: string` to the swap types and persists it at **every** swap-completion path. All persists are **purely additive** — the claim/broadcast logic is untouched, so the blast radius is correctness, not funds:

| Swap type | Correlating txid | Persist point |
|-----------|-----------------|---------------|
| Chain (claimArk / claimBtc) | claim txid | claim execution + both `resolveChainClaimTxid` monitor paths |
| Reverse (LN → Ark) | `getReverseSwapTxId` at settlement | `invoice.settled` — mutated so it survives subsequent saves |
| Submarine (Ark → LN) | wallet's lockup send txid | right after `wallet.send` — mutated so it survives the settlement save |

Reverse/submarine mutate the in-memory swap so the later settlement save carries `claimTxid`; chain persists at its terminal point.

### Verification
- boltz-swap `tsc` clean; **monorepo build clean** (the framework's interface change + this PR typecheck across all packages).
- boltz-swap unit suite: **all green** — the additive persists did not regress the swap flows.
- Resolver unit tests: chain `Chain swap` label + reverse/submarine `Lightning swap` label.


pietro909:
I'm not convinced this is a good addition to the SDK.

This is what lands in for a consumer once the PR is merged (2 out of 4 despite the large effort):

Direction  | Consumer-visible after 0+1
--------------|-------------------------------------
   Submarine (LN send) |  ✓ labeled                       
 Chain BTC→ARK       | ✓ labeled     
Reverse (LN Receive) | ✗ removed or unverified  
Chain ARK→BTC | ✗ removed or unverified  


While working on Activity History for trixie (see https://github.com/pietro909/trixie-wallet/blob/master/app/services/arkade/activity-history.ts) I already hit this problem.
I eventually rejected the _captured-txid_ approach entirely and rolled my own correlation logic because I needed all directions and restore-survivability, and `claimTxid` gives neither. 

A different approach using *VHTLC-script + indexer* join (NArk's actual join key) covers all four, survives restore, and removes the plumbing rather than adding to it, since it correlates at read time from data the swap already stores. Mechanism verified against the indexer's  `spentBy/settledBy`; still needs an e2e to confirm before implementing.

Happy to discuss this further @Kukks !

Kukks:
Go for it.

PR #584, opened 2026-06-24
feat(wallet): collab-exit + asset-mint activity built-ins + tx tag

Kukks:
## Collab-exit + asset-mint activity built-ins (+ tx `tag`)

**Stacked on #582** (the activity-history framework) — base `feat/activity-history`. Adds the last two built-in resolvers to exercise the framework with, plus the small primitive collab-exit keys off.

### `tag` on ArkTransaction (additive)
The history builder already computes a category for every tx (`ExtendedArkTransaction`: `"offchain" | "boarding" | "exit" | "batch"`) — it just wasn't surfaced on the public type. This exposes it as an optional `ArkTransaction.tag`. **Purely additive**: the value is already set at runtime and `getTransactionHistory()` returns the builder output directly (`wallet.ts:868`), so behavior is unchanged — only the type gains a field.

### Built-ins
- **`collabExitResolver`** — labels collaborative exits (VTXOs forfeited to chain in a batch) off `tag === "exit"`, grouped by commitment txid. A robust marker, not a heuristic.
- **`assetMintResolver`** — labels the genesis tx of a minted asset. An asset id encodes its genesis txid (`AssetId`), so a tx is a mint exactly when it carries the asset and its `arkTxid` is that genesis. Reissues/transfers carry the asset but key off a different genesis, so they stay plain.

Both are pre-registered in `createDefaultActivityRegistry()` (now `boarding`, `collab-exit`, `asset-mint`) and exported from the package root.

### Verification
- New unit tests for both resolvers (incl. a real `AssetId`-derived genesis match) + the updated factory order.
- ts-sdk unit suite: **1519 pass** / 1 skip (no regression).
- **Monorepo build clean** — the additive `tag` breaks no consumer.

## Summary by CodeRabbit

* **New Features**
  * Added activity tracking for collaborative exits.
  * Added activity tracking for minted and received assets, including asset identifiers and amounts.
  * Added transaction tags to transaction data, supporting built-in and custom tags.
  * Exposed new activity resolvers and transaction tag types through the SDK.

* **Tests**
  * Added coverage for collaborative exits, asset minting, asset receipt, and default activity registration.


pietro909 (approved):
@Kukks good to go from my side.
I've addressed minor findings:
- type duplication
- "asset mint" attribution on receive
-  a hand-built exit-tagged tx with an empty key would land as `exit:` group

PR #585, opened 2026-06-25
fix(wallet): reconcile VTXOs the indexer stops reporting (stale balance after chain reset)

Dunsin-cyber:
### Remaining part of #522.

After a chain reset or reorg, the indexer stops reporting a wallet's old coins, but the sync only ever adds and updates, never removes. So the wallet keeps showing a stale balance from coins that no longer exist.

## Status

Implemented and tested.

### What it does

 - On sync, checks the wallet's own coins against the indexer by outpoint and deletes the ones it no longer reports, so the balance self heals.
 - Deletes a coin only after it is absent 3 checks in a row, throttled to once every 15s, so a transient blip never drops real money.
 - Adds deleteVtxosByOutpoint to all four backends so the delete is precise (in line with #480).

### Related (already fixed, separate PR #579)

Request 1: clearLocalData() to wipe local data.
Request 3: unblock indexedDB.deleteDatabase().

## Summary by CodeRabbit

* **New Features**
  * Automatic reconciliation now prunes “vanished” stored VTXOs after repeated consecutive absence from the indexer, including during refresh and reconnect/boot.
  * Added outpoint-based VTXO deletion support across all wallet storage backends, including a helper to delete VTXOs for a single contract.
* **Bug Fixes**
  * Safer pruning: VTXOs are only removed after reaching a consecutive-miss threshold, reducing the risk of deleting due to temporary gaps.
* **Tests**
  * Added/updated unit and end-to-end coverage for outpoint deletion and chain-reset reconciliation behavior.


Dunsin-cyber:
ready for review @pietro909 @Kukks

packages/ts-sdk/src/contracts/contractManager.ts
```diff
@@ -1292,9 +1306,90 @@ export class ContractManager implements IContractManager {
...
+                }
+            }
+        } catch (error) {
+            console.error("reconcileVanishedVtxos: vanished-vtxo reconciliation failed:", error);
+        }
+    }
+
+    /** Outpoints the indexer still knows, across all pages (absent = truly gone). */
+    private async fetchPresentOutpoints(vtxos: ExtendedVirtualCoin[]): Promise<Set<string>> {
+        const outpoints = vtxos.map((v) => ({ txid: v.txid, vout: v.vout }));
```
pietro909, line 1448: For a large number of outpoints (~104) this results in a long URL which crosses the common 8 KB server/proxy header limit.
Check how src/wallet/serviceWorker/wallet-message-handler.ts:1853 batches the identical call at BATCH_SIZE = 100.

packages/ts-sdk/test/sqlite-wallet-repository.test.ts
```diff
@@ -8,6 +8,316 @@ import type { TapLeafScript } from "../src/script/base";
...
+// A lightweight in-memory SQL engine that supports the subset of SQL
+// used by SQLiteWalletRepository: CREATE TABLE, INSERT OR REPLACE,
+// SELECT, DELETE, and ORDER BY.
+
+interface TableDef {
+    primaryKey: string[];
+    rows: Map<string, Record<string, unknown>>;
+}
+
+function createMockSQLExecutor(): SQLExecutor {
```
pietro909: This shadows the imported `createMockSQLExecutor` at line 9.
Why we need it?

packages/ts-sdk/src/contracts/contractManager.ts
```diff
@@ -1292,9 +1306,90 @@ export class ContractManager implements IContractManager {
...
     }
 
+    /**
+     * Delete unspent VTXOs the indexer no longer knows, checked by outpoint.
+     * Requires {@link RECONCILE_ABSENCE_THRESHOLD} consecutive absences so a
+     * transient empty response never drops a real coin; a failed lookup never deletes.
+     */
+    private async reconcileVanishedVtxos(contracts: Contract[]): Promise<void> {
+        const now = Date.now();
+        if (now - this.lastReconcileAt < RECONCILE_MIN_INTERVAL_MS) return;
```
pietro909: `lastReconcileAt` is shared by all calls to this function: if ContractManager and VtxoManager race for different contracts within the same 15s window we discard legit calls.

packages/ts-sdk/src/contracts/contractManager.ts
```diff
@@ -1292,9 +1306,90 @@ export class ContractManager implements IContractManager {
...
+                    const misses = (this.vanishMissCounts.get(key) ?? 0) + 1;
+                    if (misses >= RECONCILE_ABSENCE_THRESHOLD) {
+                        this.vanishMissCounts.delete(key);
+                        vanished.push(v);
+                    } else {
+                        this.vanishMissCounts.set(key, misses);
+                    }
+                }
+                if (vanished.length > 0) {
+                    await deleteVtxosForContract(this.config.walletRepository, contract, vanished);
```
pietro909, line 1422: shall we notify listeners of `onVtxoSpent`? while not a "spend" action, this in practice leads to a similar result (a potential change in balance, coins shown on a UI, ... )

packages/ts-sdk/src/contracts/vtxoOwnership.ts
```diff
@@ -97,3 +97,21 @@ export async function saveVtxosForContract(
...
+export async function deleteVtxosForContract(
+    repo: WalletRepository,
+    contract: Pick<Contract, "script" | "address">,
+    vtxos: Array<Pick<ExtendedVirtualCoin, "txid" | "vout">>,
+): Promise<void> {
+    if (vtxos.length === 0) return;
+    if (repo.deleteVtxosByOutpoint) {
+        return repo.deleteVtxosByOutpoint(vtxos.map((v) => ({ txid: v.txid, vout: v.vout })));
+    }
+    // Fallback: rewrite the address bucket without the removed coins.
```
pietro909, line 117: the fallback isn't atomic: if `repo.saveVtxos` fails we end up with an empty address

pietro909 (changes requested):
There are many moving parts in this area: please check the comments.

Dunsin-cyber:
> There are many moving parts in this area: please check the comments.

got it. would make PR addressing them soon.

packages/ts-sdk/test/sqlite-wallet-repository.test.ts
Dunsin-cyber: ahh. i see, happened from merging master in

packages/ts-sdk/src/contracts/vtxoOwnership.ts
Dunsin-cyber: yes, this fallback only runs for custom backends missing the atomic `deleteVtxosByOutpoint` (all shipped backends have it), the coarse primitives can't be made atomic without a transaction API, and the repo is a cache, so the worst case is a temporarily empty bucket the next sync refills, not lost funds.

packages/ts-sdk/src/contracts/contractManager.ts
Dunsin-cyber: yes this is a good idea, i've added a new event, called `vtxo_vanished` to handle this

packages/ts-sdk/src/contracts/contractManager.ts
```diff
@@ -1338,9 +1361,103 @@ export class ContractManager implements IContractManager {
...
+    }
+
+    /** Outpoints the indexer still knows, across all pages (absent = truly gone). */
+    private async fetchPresentOutpoints(vtxos: ExtendedVirtualCoin[]): Promise<Set<string>> {
+        const outpoints = vtxos.map((v) => ({ txid: v.txid, vout: v.vout }));
+        const present = new Set<string>();
+        const BATCH_SIZE = 100;
+
+        for (let i = 0; i < outpoints.length; i += BATCH_SIZE) {
+            const { vtxos: found } = await getNormalizedVtxos(this.config.indexerProvider, {
```
pietro909, line 1454: This ignores pagination. If we use default page size we get it 5 times larger than the batch size and it should cover for all the requested VTXOs:

```
const { vtxos: found } = await getNormalizedVtxos(this.config.indexerProvider, {
    outpoints: outpoints.slice(i, i + BATCH_SIZE),
    pageSize: DEFAULT_PAGE_SIZE,   // <---this one
});
```

packages/ts-sdk/src/contracts/contractManager.ts
```diff
@@ -1338,9 +1361,103 @@ export class ContractManager implements IContractManager {
...
+                        }
+                    }
+                }
+            }
+        } catch (error) {
+            console.error("reconcileVanishedVtxos: vanished-vtxo reconciliation failed:", error);
+        }
+    }
+
+    /** Outpoints the indexer still knows, across all pages (absent = truly gone). */
```
pietro909: "across al pages" is probably not true anymore since there is no paging loop?

packages/ts-sdk/src/contracts/contractManager.ts
```diff
@@ -395,6 +403,14 @@ export interface ContractManagerConfig {
      */
     onVtxosSpent?: (vtxos: Outpoint[]) => Promise<void>;
 
+    /**
+     * Optional exit-data prune hook for coins that vanished without a spend
+     * (chain reset/reorg). Reconcile is their only cleanup point, so a
+     * configured virtualTxRepository can drop their now-unusable branch here.
+     * Absent ⇒ no-op.
+     */
+    onVtxosVanished?: (vtxos: Outpoint[]) => Promise<void>;
```
pietro909, line 412: The new callback is never wired up.
This should at least occur in `src/wallet/wallet:ts` where we set up `onVtxosSpent` and call `pruneExitBranches`.

EDIT:
I'm also confused about the role of this event. In `wallet.notifyIncomingFunds()` we relay the events as:

```
      eventCallback({
          type: "vtxo",
          newVtxos: event.type === "vtxo_received" ? annotated : [],
          spentVtxos: event.type === "vtxo_spent" ? annotated : [],
      });
```

The `vtxo_vanished` is entirely ignored. Is this something we want? What's your idea for that @Dunsin-cyber ?

pietro909 (changes requested):
I'm putting this one on hold for some time @Dunsin-cyber because there are some considerations I need to study.

This PR is deleting VTXO data which under certain circumstances can be dangerous. Is there a way to have the offending VTXOs (_vanished_) be excluded from the logic but still backed up?

packages/ts-sdk/src/contracts/contractManager.ts
Dunsin-cyber: that was an oversight on my end. the `vtxo_vanished` should be wired up too in the `wallet.notifyIncomingFunds()`.

Dunsin-cyber:
> I'm putting this one on hold for some time @Dunsin-cyber because there are some considerations I need to study.
> 

alright, got it. i'll do more vetting on my end too.

PR #586, opened 2026-06-26
feat(payment): modular payment router (ark/onchain/lightning/chain-swap rails)

Kukks:
## Summary

Adds a registry-based **modular payment router** to `@arkade-os/sdk` (with swap rails in `@arkade-os/boltz-swap`) so consumers (coinflip, arkade-os/wallet, future apps) stop re-implementing the same three things: classifying a payment target, choosing a rail, and parsing/building the unified BIP21 URI. It mirrors the existing `ActivityRegistry`: pluggable **id-keyed** entries, defaults registered via a **factory** (not hard-coded into a class), apps extend/override/remove by id.

The router operates on the **raw request string** — no assumed structure (most targets are bare `lnbc…`/`tark1…`/`bc1…`, not URIs). `options(raw)` returns every matching + available rail ranked by preferences; `route()` takes the top; execution returns an observable **`PaymentHandle`**, never a forced result (swaps settle out-of-band).

## Rails

| id | package | matches | executes via |
|----|---------|---------|--------------|
| `ark` | ts-sdk | bare ark address / `ark=` BIP21 param | `Wallet.sendBitcoin` (off-chain) |
| `onchain` | ts-sdk | BTC address / on-chain BIP21 part | collaborative exit (`Wallet.settle`) |
| `lightning` | boltz-swap | bolt11 / `lightning=` BIP21 param | `ArkadeSwaps.sendLightningPayment` (submarine) |
| `onchain-swap` | boltz-swap | BTC address / on-chain BIP21 part | Ark→BTC chain swap (`arkToBtc` → fund lockup → `waitAndClaimBtc`) |

Two factories:
- `createDefaultPaymentRouter(wallet)` (ts-sdk) — `ark` + `onchain`, priority `["ark", "lightning", "onchain"]`.
- `createDefaultPaymentRouter(wallet, swaps)` (boltz-swap) — all four, priority `["ark", "lightning", "onchain-swap", "onchain"]`.

For a plain BTC address the boltz factory **prefers the chain swap** (`onchain-swap`) over the collaborative exit (`onchain`), but both stay registered — `options()` surfaces both and an app/user can override `priority`/`disabled` to force the exit. The default expresses a *preference*, not a *restriction*.

This matches the reference implementation: NArk's `Bip21PaymentInfo.PreferredMethod` (`NArk.Abstractions/Payments/Bip21PaymentInfo.cs`) resolves a bare on-chain address to `ChainSwap`, not `CollaborativeExit`, and its `ArkPaymentMethod` enum (`NArk.Abstractions/Payments/ArkPayment.cs`) is the same four-way split as the rails here — `ArkSend`, `CollaborativeExit`, `SubmarineSwap`, `ChainSwap`. The rails keep string ids rather than an enum so apps can register their own.

## Quote semantics

`RouteQuote`'s three amounts are **receiver-exact** and mean the same thing on every rail, so `options()` is rankable on cost and swapping rails never changes what the recipient gets:

- `amount` — sats delivered **to the recipient**
- `fee` — sats charged **on top**
- `total` — `amount + fee`; what leaves the wallet

`Ramps.offboard` *deducts* its fee, so the `onchain` rail grosses the amount up to honour this; the swap rails add on top and pass it through. The gross-up is a **fixpoint, not a subtraction** — `outputToArgs` feeds `amount` into the CEL fee program, so at 1% a naive `amount + fee(amount)` leaves the recipient a sat short. Iterating upward from `amount` converges on the least fixpoint, and every intermediate value under-estimates, so a pathological schedule errs toward the recipient.

`fee` is a pre-send display/ranking figure, not a guarantee, wherever the true cost is fixed later:

| rail | source | exact? |
|---|---|---|
| `ark` | none — off-chain send | exact, 0 |
| `onchain` | `getInfo().fees.intentFee` via `Estimator` | exact for the output fee |
| `onchain-swap` | `getFees("ARK","BTC")` | estimate; Boltz fixes it at swap creation |
| `lightning` | `getFees().submarine` | estimate; Boltz fixes it at swap creation |

The exit's per-input intent fees are excluded — they depend on the VTXO selection `offboard` makes at settlement, so a guess would make `fee` less honest, not more.

## Design notes

- **`onchain` rail = collaborative exit, not `sendBitcoin`.** `Wallet.sendBitcoin`/`Wallet.send` are Ark-only — they validate the destination with `ArkAddress.decode` and throw on a BTC address. Sending BTC on-chain from an Ark wallet requires a collaborative exit: select VTXOs and `Wallet.settle({ inputs, outputs })`. The rail encapsulates that (the dedup target of the wallet app's hand-rolled `lib/asp.ts: collaborativeExit`) and selects soonest-expiring VTXOs. With swaps available it ranks below the chain swap by default but stays selectable.
- **Target extraction is shared, not per-rail.** `payment/targets.ts` in core owns `arkTarget`/`btcTarget`/`invoiceTarget`, so the boltz rails classify identically to the built-in ones. Predicates are format-only by design — core carries no bolt11 dependency, so `isLightningInvoice` cannot decode; rails re-validate before spending.
- **Reuses the existing `BIP21` codec and `ArkAddress`** rather than introducing a new parser/encoder. `BIP21` is now re-exported from the SDK root (it previously wasn't).
- **Pricing the exit needs its destination script**, so `Ramps.offboard`'s network-probing decode is extracted to `offboardDestinationScript` and shared — the quote is evaluated against the exact script settlement uses, which matters for a script-size-dependent fee program. It stays package-internal (the root barrel re-exports `Ramps` by name, not `export *`). Consequence: quoting now decodes the address, so an invalid one fails at `quote()` rather than inside `offboard()`.
- **Dependency direction preserved:** ts-sdk core holds the router, interfaces, predicates, and Wallet-only rails; boltz-swap holds the swap rails. Core never imports boltz-swap — `RouterContext.swaps` is typed `unknown` in core and cast to `ArkadeSwaps` only inside the boltz rails (which are also `available()`-gated on `ctx.swaps`).

## Test plan

- `pnpm -r build` — exit 0 (both packages, incl. dts)
- `pnpm lint` — clean (prettier, both packages)
- `pnpm test:unit` — ts-sdk **2193 passed** (+1 pre-existing skip), boltz-swap **525 passed**, 0 failures
  - Baseline before this round: ts-sdk 2191 + 1 skip, boltz-swap 522. The deltas are exactly the tests added here (+2 / +3); no regressions.
  - One pre-existing flake appears in full-suite runs (`refundArk — multi-VTXO iteration … 2s delay between calls`, `Aborting after running 10000 timers`). Confirmed unrelated: it installs its own `vi.useFakeTimers()`, `vi.restoreAllMocks()` runs between tests, and a clean tree shows the same behaviour. Not fixed here.
- New unit tests written strict-TDD (red → green per task): router ranking / `disabled` / `available` filtering / `tieBreak`; each rail's `match` / `available` / `quote` / `send` against mocked `Wallet` / `ArkadeSwaps`; both factory fan-outs (`bitcoin:<btc>?ark=<ark>&lightning=<bolt11>` → `["ark", "lightning", "onchain-swap", "onchain"]`); and the "chain swap preferred, collaborative exit still selectable" guarantee.

## Out of scope (follow-ups)

- **Phase 4 — consumer migration** (coinflip `detectSend` + wallet `Send/Form+Details` rail ladder → the router) lands as separate PRs after this SDK release.
- **Asset-aware routing.** `BIP21Params` has no `assetid` field and no rail inspects one, so an asset URI currently matches the BTC rails. NArk treats `AssetId` as forcing Ark-only delivery; aligning needs a follow-up.
- Receive-side `encodeUnifiedAddress` (a thin `BIP21.create` wrapper) is still deferred. Per-rail fee resolution is **no longer deferred** — see Quote semantics above.
- **`route()` next-option fallback** — tracked in #658. A chain-swap failure can be mid-flight (lockup funded, claim pending), so falling back to `onchain` there would select VTXOs the swap still needs; it needs a conservative per-rail "safe to retry" signal. The `available()` limits gate already closes the concrete sub-minimum case before `send()` is reached.

## Summary by CodeRabbit

* **New Features**
  * Added default payment router creation and expanded routing to select Ark, Lightning, on-chain, and on-chain-swap payment rails.
  * Introduced Lightning and on-chain-swap rails with quote/send flows, including BIP21 target parsing support.
  * Expanded SDK exports (payment utilities, BIP21 helpers, and routing targets).

* **Bug Fixes**
  * Improved Lightning invoice normalization/decoding and tightened quote-time amount validation.
  * Enhanced BTC settlement failure details for refundable failures.

* **Tests**
  * Added/expanded coverage for routing/priority, rail matching & availability gating, amount parsing, caching behavior, and settlement/claim flows.


packages/ts-sdk/test/payment/factory-core.test.ts
```diff
@@ -0,0 +1,28 @@
...
+        expect(opts.map((o) => o.railId)).toEqual(["ark", "onchain"]);
+    });
+
+    it("routes a bare ark address to ark and a bare BTC address to onchain", async () => {
+        const router = createDefaultPaymentRouter({} as any);
+        expect((await router.route(arkAddr, 500)).railId).toBe("ark");
+        expect((await router.route(btcAddr, 500)).railId).toBe("onchain");
+    });
+
+    it("is re-exported from the SDK root with PaymentRouter, BIP21 and predicates", () => {
```
pietro909: I would suggest we start removing explicit "export" tests: that's compiler's role and units will cover the runtime part.

packages/ts-sdk/src/payment/rails/onchain.ts
```diff
@@ -0,0 +1,91 @@
...
+    return {
+        id: "onchain",
+        match: (raw) => btcTarget(raw) !== undefined,
+        quote: async (raw, amount, ctx: RouterContext) => {
+            const address = btcTarget(raw)!;
+            const amt = amount ?? encodedAmountSats(raw) ?? 0;
+            return {
+                railId: "onchain",
+                amount: amt,
+                fee: 0,
```
pietro909: Where do we expect the `fee` to come from in the future?

packages/ts-sdk/src/payment/rails/onchain.ts
```diff
@@ -0,0 +1,91 @@
...
+ * URI, selects soonest-expiring VTXOs to cover the amount, and `Wallet.settle`s
+ * them to the address with change back to the wallet.
+ */
+export function onchainRail(): PaymentRail {
+    return {
+        id: "onchain",
+        match: (raw) => btcTarget(raw) !== undefined,
+        quote: async (raw, amount, ctx: RouterContext) => {
+            const address = btcTarget(raw)!;
+            const amt = amount ?? encodedAmountSats(raw) ?? 0;
```
pietro909: Does a zero satoshis payment make sense? Will probably fail later down the line.

packages/ts-sdk/src/payment/types.ts
```diff
@@ -0,0 +1,62 @@
+import type { Wallet } from "../index";
+
+export type PaymentStatus = "pending" | "sent" | "settled" | "failed";
```
pietro909, line 3: I cannot find code emitting `failed`. 
`makeHandle` swallows errors and no handler emilts `failed` at any time: if we don't use it it's better to remove it.

packages/ts-sdk/src/payment/rails/onchain.ts
```diff
@@ -0,0 +1,91 @@
...
+                    }),
+            };
+        },
+    };
+}
+
+/**
+ * Select soonest-expiring VTXOs covering `amount` and settle them to `address`,
+ * returning change to the wallet's offchain address. Mirrors the wallet app's
+ * hand-rolled collaborative exit (`lib/asp.ts`), the dedup target of this rail.
```
pietro909: Probably  `lib/asp.ts` refers to the wallet app, this block is more similar to `wallet/ramps.ts` instead.

Which makes me think why we aren't reusing `Ramps.offboard`? It's fees and dust aware, with different coin selection preference.

packages/ts-sdk/src/payment/handle.ts
```diff
@@ -0,0 +1,43 @@
...
+    id: string,
+    run: (emit: (u: Update) => void) => Promise<RouteResult>,
+): PaymentHandle {
+    let status: PaymentStatus = "pending";
+    let last: Update = { status };
+    const subs = new Set<(u: Update) => void>();
+    const emit = (u: Update) => {
+        status = u.status;
+        last = u;
+        for (const f of subs) f(u);
```
pietro909: we don't want a subscriber's error to propagate here: the failure would break the cycle and even the handler

packages/ts-sdk/src/payment/predicates.ts
```diff
@@ -0,0 +1,25 @@
...
+    /^lnurl/i.test(raw) || /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(raw);
+
+/** True for a bech32 (segwit) or base58 (legacy) Bitcoin address — format only,
+ *  not a checksum check (the rail/SDK validates fully before spending). */
+export const isBtcAddress = (raw: string): boolean =>
+    /^(bc1|tb1|bcrt1)[qpzry9x8gf2tvdw0s3jn54khce6mua7l]{6,90}$/i.test(raw) ||
+    /^[13][a-km-zA-HJ-NP-Z1-9]{25,34}$/.test(raw);
+
+/** True if the string decodes as an Arkade address. */
+export const isArkAddress = (raw: string): boolean => {
```
pietro909: we have ` isValidArkAddress` (see src/wallet/utils.ts:194)

packages/ts-sdk/src/payment/rails/onchain.ts
```diff
@@ -0,0 +1,91 @@
...
+    for (const v of sorted) {
+        if (selectedAmount >= amount) break;
+        selected.push(v);
+        selectedAmount += v.value;
+    }
+    if (selectedAmount < amount) {
+        throw new Error("Insufficient funds for collaborative exit");
+    }
+
+    const outputs = [{ address, amount: BigInt(amount) }];
```
pietro909: `BigInt` may throw for decimals, needs a guard or it will fail in-flight

packages/ts-sdk/src/payment/types.ts
```diff
@@ -0,0 +1,62 @@
...
+    swaps?: unknown;
+    prefs: RouterPreferences;
+}
+
+/** A payment rail — registered by id, mirrors the ActivityRegistry resolver shape. */
+export interface PaymentRail {
+    id: string;
+    match(raw: string, ctx: RouterContext): boolean;
+    available?(ctx: RouterContext): boolean | Promise<boolean>;
+    quote(raw: string, amount: number | undefined, ctx: RouterContext): Promise<RouteQuote>;
```
pietro909: The `amount` field is unused for some rails like a submarine swap, and having it in the second position before a mandatory `ctx` ends up in poor ergonomics: consumers may end up parsing the invoice to extract the amount that is then ignored.

Maybe we can use a object for the first argument.

pietro909 (changes requested):
I like the vision, but we need an iteration around:

1. code reuse
2. error management
3. payment status model

packages/ts-sdk/src/payment/handle.ts
```diff
@@ -0,0 +1,53 @@
...
+    };
+    const emit = (u: Update) => {
+        status = u.status;
+        last = u;
+        for (const f of subs) notify(f, u);
+    };
+    const done = run(emit);
+    // Swallow the unhandled-rejection warning for fire-and-forget callers that
+    // never await settled(); awaiters still receive the original rejection.
+    done.catch(() => {});
```
pietro909: 🔴 **High — a failed payment never reaches `subscribe()` consumers.**

Rails only `emit()` on success; on rejection this line swallows the error and the handle stays on `"pending"`/`"sent"` forever. `PaymentStatus` defines `"failed"` but nothing ever emits it, so a consumer using the documented progress stream never learns a payment failed — only `await settled()` rejects. Swaps fail routinely (invoice expiry, liquidity, sub-limit), so this is the common path, and the subscriber closure leaks too.

```suggestion
    // Fire-and-forget callers never await settled(); still surface failure on the
    // observable stream. settled() returns the original `done`, so its awaiters
    // keep receiving the underlying rejection.
    done.catch(() => emit({ status: "failed" }));
```

packages/boltz-swap/src/payment/index.ts
```diff
@@ -0,0 +1,33 @@
...
+ * (`onchain-swap`) is preferred over the collaborative exit (`onchain`), but
+ * both stay registered: the default is a *preference*, not a restriction —
+ * `options()` surfaces both and an app/user can override `priority` (or
+ * `disabled`) to force the collaborative exit.
+ */
+export function createDefaultPaymentRouter(wallet: Wallet, swaps: ArkadeSwaps): PaymentRouter {
+    return new PaymentRouter({
+        wallet,
+        swaps,
+        prefs: { priority: ["ark", "lightning", "onchain-swap", "onchain"] },
```
pietro909, line 38: 🔴 **High — default ranks a swap that can fail above a rail that won't, with no fallback.**

Priority puts `onchain-swap` ahead of `onchain` for any BTC address, but `onchain-swap` never checks Boltz min/max limits (`available()` only tests `swaps != null`; `quote()` only validates positive-int), and `router.route()` has no fallback past `options[0]`. A sub-minimum BTC send picks the chain swap and fails at `send()` — silently, per the `handle.ts` finding — even though `onchain` (collaborative exit) would have completed it. Compounding: `waitAndClaimBtc` doesn't auto-refund without a `SwapManager`, so the Ark lockup can be stranded.

**Fix:** gate `onchain-swap.quote()` on `getLimits("ARK","BTC")` **and** add a next-option fallback in `router.route()`.

packages/ts-sdk/src/payment/rails/onchain.ts
```diff
@@ -0,0 +1,56 @@
...
+        quote: async (raw, amount, ctx: RouterContext) => {
+            const address = btcTarget(raw)!;
+            // Reject missing/zero/fractional amounts up front: 0 sats would
+            // silently settle nothing, and BigInt(amt) throws on non-integers.
+            const amt = resolveSendAmount("onchain", raw, amount);
+            return {
+                railId: "onchain",
+                amount: amt,
+                // Provisional: Ramps.offboard deducts the real intent + network
+                // fees from the amount at settlement time.
```
pietro909: 🟠 **Medium — quotes report `fee: 0` / `total: amount` while the real debit differs.**

Every rail reports `fee: 0`, but here `Ramps.offboard` *deducts* fees so the recipient receives `amount - fee`; the boltz rails instead *add* fees on top (`lightning` funds `expectedAmount`, `onchain-swap` sends `amt + claim fee`). The `fee`/`total` fields exist for pre-send display and cost-based ranking — `0` makes them misleading and ranking impossible. Fee data is queryable (`getFees`, `getInfo().fees`); populate it or explicitly mark the quote provisional.

packages/boltz-swap/src/payment/onchain-swap.ts
```diff
@@ -0,0 +1,54 @@
...
+                railId: "onchain-swap",
+                amount: amt,
+                fee: 0,
+                total: amt,
+                send: async () =>
+                    makeHandle("onchain-swap", async (emit) => {
+                        const swaps = ctx.swaps as ArkadeSwaps;
+                        const { arkAddress, amountToPay, pendingSwap } = await swaps.arkToBtc({
+                            btcAddress: address,
+                            receiverLockAmount: amt,
```
pietro909, line 95: 🟠 **Medium — `amount` means different things across the two on-chain rails.**

Here `receiverLockAmount: amt` → the recipient receives `amt`. The core `onchain` rail passes the same `amt` to `Ramps.offboard`, which does `amount -= fee` → recipient receives `amt - fee`. `router.options()` surfaces both as interchangeable with identical `RouteQuote.amount`, so an app offering both (or a user overriding `priority: ["onchain"]`) silently delivers a different amount. **Fix:** normalize `amount` to "delivered to recipient" across rails (gross up in the collaborative-exit rail), or expose the semantic per rail.

packages/ts-sdk/src/payment/predicates.ts
```diff
@@ -0,0 +1,16 @@
+/** True if the string decodes as an Arkade address (canonical SDK check). */
+export { isValidArkAddress } from "../wallet/utils";
+
+/** True for a BOLT11 invoice (with or without a `lightning:` prefix). */
+export const isLightningInvoice = (raw: string): boolean =>
+    /^ln(bc|tb|bcrt|bs)/i.test(raw.replace(/^lightning:/i, ""));
+
+/** True for an LNURL or a Lightning address (`user@host`). */
+export const isLnurl = (raw: string): boolean =>
```
pietro909, line 25: 🟡 **Low — `isLnurl` is exported but consumed by no rail.**

LNURL / Lightning-address payments silently don't route: `route("user@host")` throws "no rail for". Either wire a rail that handles LNURL / LN-address, or drop the export so the public API doesn't imply support.

packages/ts-sdk/src/payment/rails/ark.ts
```diff
@@ -0,0 +1,44 @@
...
+            const address = arkTarget(raw)!;
+            const amt = resolveSendAmount("ark", raw, amount);
+            return {
+                railId: "ark",
+                amount: amt,
+                fee: 0,
+                total: amt,
+                send: async () =>
+                    makeHandle("ark", async (emit) => {
+                        const txid = await ctx.wallet.sendBitcoin({ address, amount: amt });
```
pietro909: 🟡 **Low — reference rail uses deprecated `sendBitcoin`.**

`wallet.sendBitcoin` is deprecated in favor of `wallet.send` (which the boltz `onchain-swap` rail already uses). Align the reference rail to `send`.

packages/ts-sdk/src/payment/predicates.ts
pietro909: future work

pietro909 (changes requested):

packages/boltz-swap/src/payment/index.ts
pietro909: Half of this landed; leaving the thread open for the other half.

**Done — the limits gate.** `onchainSwapRail.available()` now calls `getLimits("ARK","BTC")` + `getFees("ARK","BTC")` and brackets the *source* (user-lock) amount, since Boltz bounds that rather than the receiver amount this rail knows. Out-of-range drops the rail during `options()`, so `onchain` becomes `options[0]` and the collaborative exit completes the send — the sub-minimum case never reaches `send()`. `764b11ed` also bails out at a fee rate ≥ 100%, which previously inverted the max check into passing.

The two compounding factors are gone as well: `makeHandle` emits a terminal `failed` carrying the error, and the stranded-lockup path is documented on the rail (`SwapManager` refunds automatically; without one, `failed` carries `isRefundable` + `pendingSwap` for `refundArk`).

**Not done — the `route()` next-option fallback.** Deliberate, for now: `send()` returns a handle rather than a result, and a chain-swap failure can be mid-flight (lockup funded, claim pending). Retrying on `onchain` at that point risks spending VTXOs the swap still needs. A send-time fallback needs a "is this failure safe to retry" signal per rail, which doesn't exist yet.

Happy to add it if you'd rather have it in this PR — otherwise it wants its own change.

pietro909:
Validated all 10 findings from https://github.com/arkade-os/ts-sdk/pull/586#pullrequestreview-4806701438 against `764b11ed`. Seven were real and are fixed, one was already stale, two had wrong premises but pointed at something real.

**Fixed**

- (2) `isBtcAddress` now rejects mixed-case bech32 per BIP173 (all-upper still accepted) and — separately — accepts legacy testnet/regtest `m`/`n`/`2`, which it previously rejected outright. That gap would have broken regtest routing for any consumer migrating onto the predicate.
- (3) The HRP set is corrected, but not as described: signet invoices already matched, because `/^ln(bc|tb|…)/` is a prefix test and `tb` matches `lntbs…`. The real bugs were the bogus `bs` alternative (making `lnbs…` a false positive), a dead `bcrt` alternative shadowed by `bc`, and a missing simnet `sb`. Now `/^ln(bcrt|bc|tbs|tb|sb)[0-9a-z]+$/i`, anchored so `lnbc!!!` no longer matches. Table-tested against real amount-in-HRP invoices (`lnbc2500u1…`).
- (4) Commented. Worth noting the docstring's contract has a narrower hole than described: a `wallet.send()` rejection *after* the lockup broadcast leaves funds locked with no `pendingSwap` on the handle. Reconcile via `swaps.getPendingChainSwaps()`.
- (5) `available()` returns `false` at a fee rate ≥ 100% instead of letting a negative divisor invert the max gate.
- (6) `options()` warns before dropping a rail whose `available()` threw, matching the existing `console.warn` idiom in `providers/ark.ts` / `wallet/vtxo.ts`.
- (7) Documented on the rail, pointing at `Ramps.offboard(address, feeInfo)` for the amountless full sweep.
- (8) `makeHandle` clears subscribers on a terminal update; post-terminal `subscribe()` still replays, then returns a no-op unsubscribe.

**Already stale** — (1) the third `isValidArkAddress` copy in `boltz-swap/src/utils/decoding.ts` is already a re-export from `@arkade-os/sdk`. The two copies inside ts-sdk (`wallet/utils.ts` and `utils/arkTransaction.ts`) are still there; collapsing them touches a pre-existing root export, so it's a follow-up rather than part of this PR.

**Premise off, but useful**

- (9) The consumer sweep checked four repos that don't exist in this workspace (`coinflip`, `checkout`, `btcpay-arkade`, `lnurl-server`) and missed four that do. Actual state: `trixie-wallet` (40 files on the SDK), `chimera-wallet-pwa` (20), `expo-wallet` (10) and `wdk-arkade` (12) alongside `wallet` (32). Three of them ship their own `isLightningInvoice`, not one — and `wallet`/`chimera` implement it as a full bolt11 decode, so the SDK predicate is strictly weaker. Nothing collides today, though: every consumer uses named imports, and none does `export *` from the SDK, so the suggested `export {} from …` remedy isn't needed. That widened blast radius is why (3) got a real fix rather than a doc note, and why `predicates.ts` now states up front that it classifies rather than validates.
- (10) `createChainSwap()` does not refetch at swap time — it reads the same 15-minute cache, by design, and uses it to derive `userLockAmount = receiverLockAmount + minerFees.user.claim`. So the staleness reaches swap creation, not just routing hints. Bounded, because the funded amount comes from Boltz's authoritative `lockupDetails.amount`. Filed as a follow-up. (There is no `CHANGELOG.md` or `.changeset/` in this repo to note it in.)

The cross-repo check also turned up two things the review didn't look for, both now in the PR body:

- NArk has the analogue this PR lacked a reference for. `ArkPaymentMethod` is the same four-way split as the rails, and `Bip21PaymentInfo.PreferredMethod` resolves a bare on-chain address to `ChainSwap` rather than `CollaborativeExit` — independent precedent for the contested factory default.
- `Bip21PaymentInfo.AssetId` forces Ark-only delivery in NArk. Here, `BIP21Params` has no `assetid` field and no rail inspects one, so an asset URI currently matches the BTC rails. `trixie-wallet` already implements that rule locally. Listed under out-of-scope; of everything here it's the one I'd argue is a correctness gap rather than polish.

Also fixed in `764b11ed`, not from this review: `btcTarget` was byte-identical in ts-sdk and boltz-swap. Target extraction now lives in `payment/targets.ts` in core (`arkTarget`/`btcTarget`/`invoiceTarget`) so plugin rails classify identically to the built-in ones.

`pnpm -r build` clean, ts-sdk 2191 passed (+1 pre-existing skip), boltz-swap 514 passed, lint clean.

packages/boltz-swap/src/arkade-swaps.ts
```diff
@@ -1672,10 +1672,14 @@ export class ArkadeSwaps {
...
                     case "transaction.failed":
                         await updateSwapStatus();
+                        // Refundable terminal failure — attach the pending swap so a
+                        // no-manager caller can drive `refundArk(pendingSwap)` off the
+                        // handle's `failed` event (the SwapManager path refunds itself).
                         reject(
                             new TransactionFailedError({
                                 message: data.failureReason,
                                 isRefundable: true,
+                                pendingSwap: swap,
```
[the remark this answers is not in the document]
pietro909: Fixed in `f810911a`, at the root rather than the call site.

The cause is spread order, not this call: every subclass did `super({ message: "<default>", ...options })`, so an explicit `message: undefined` overwrote the default and fell through to `SwapError`'s generic `"Error during swap."`. Patching the two `transaction.failed` sites would have left the same trap for every other caller — `QuoteRejectedError` had already tried to guard with `options.message ?? defaultMessage(options)` and its trailing `...options` silently undid it, which is reachable from `fromTransportError` whenever the transported payload carries no message.

So all 10 constructors now spread first and resolve `message` last, and `ErrorOptions.message` documents that `undefined` keeps the subclass default. The third site (`waitAndClaimArk`, line 773) was passing `?? "Transaction failed"` — a divergent string for the same condition — and now relies on the default too, so all three read `"The transaction has failed."`.

Tests: new `test/errors.test.ts` covers the explicit-`undefined` path across three subclasses plus the base default; the two existing assertions of `"Error during swap."` are updated.

packages/boltz-swap/src/boltz-swap-provider.ts
```diff
@@ -1631,6 +1646,34 @@ export class BoltzSwapProvider {
...
+        const cached = this.pairsCache.get(path);
+        if (cached && Date.now() - cached.at < BoltzSwapProvider.PAIRS_CACHE_TTL_MS) {
+            return cached.value as T;
+        }
+        const response = await this.request<unknown>(path, "GET");
+        if (!validate(response)) {
+            throw new SchemaError({ message: schemaErrorMessage });
+        }
+        this.pairsCache.set(path, { at: Date.now(), value: response });
+        return response;
```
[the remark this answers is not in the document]
pietro909: Fixed in `f810911a`. Confirmed both exposure paths: `getChainFees` returns `response[from][to].fees` — the cached sub-object itself — and `getFees` lifts `minerFees` out by reference into a fresh outer object. `getLimits`/`getChainLimits` were already safe, returning scalar `{min, max}`.

`pairsMetadata` now returns a detached copy on both the hit and the miss path (the miss path matters too — the first caller would otherwise hold the very object just cached). JSON round-trip rather than `structuredClone`: these are plain JSON payloads, and Hermes/older RN runtimes lack the global.

Regression test added, and worth flagging why it's shaped the way it is: my first version passed *without* the fix. `createFetchResponse` hands the fixture out by reference, so a leaked reference corrupts the fixture too and any comparison against it succeeds vacuously. The test now asserts literal values, mutates through both paths, and I verified it fails on `detach = v => v` and passes with the fix.

packages/boltz-swap/src/payment/onchain-swap.ts
```diff
@@ -0,0 +1,89 @@
...
+            // gross-up inverted (the fee is charged on the user-lock total, so
+            // divide by (1 - feeRate); adding feeRate * serverLock under-estimates
+            // near max) — is the upper bound. Gate min on the lower bound
+            // (conservative) and max on the upper bound; an amount in the
+            // ambiguous band self-heals to the `onchain` collaborative exit.
+            const serverLockAmount = amt + fees.minerFees.user.claim;
+            const feeRate = fees.percentage / 100;
+            // A rate at or above 100% makes the gross-up divisor zero or
+            // negative, which would flip the max check into passing.
+            if (!(feeRate < 1)) return false;
```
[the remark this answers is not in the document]
pietro909: Fixed in `f810911a` — `if (!(feeRate >= 0 && feeRate < 1)) return false`.

Correct catch on negative: a negative rate makes the divisor exceed 1, shrinking the gross-up estimate, so `userLockAmount <= max` can pass for an amount that is actually out of range. Non-finite was already covered — `NaN < 1` is false, so the old `!(feeRate < 1)` rejected it, and `Infinity` fails the same test — but the range form states the intent directly instead of relying on that. Test cases extended to `[100, 150, -5, NaN]`.

packages/ts-sdk/src/payment/rails/onchain.ts
Kukks: Fixed in `749f43e4` — populated rather than marked provisional, and the underlying divergence you flagged on the sibling thread is gone with it.

`RouteQuote` now defines the three fields once, receiver-exact, and every rail honours them:

- `amount` — sats delivered **to the recipient**
- `fee` — sats charged **on top**
- `total` — `amount + fee`, the wallet debit

Correct on the asymmetry: `Ramps.offboard` deducts, the swap rails add. Rather than expose that per rail, the exit rail grosses up so all four agree, and `options()` stays rankable.

One thing that didn't survive first contact: the gross-up is a **fixpoint, not a subtraction**. `outputToArgs` passes `amount` into the CEL program (`arkfee/estimator.ts:153`), so the fee can be amount-dependent. At 1% on 10 000 sats, `amount + fee(amount)` = 10 100, but `fee(10100)` = 101, so the recipient lands on 9 999 — a sat short. Iterating upward from `amount` converges on the least fixpoint (10 102, fee 102). The test uses exactly that case so the naive form fails it.

Fee sources, which also answers your question on the sibling thread:

| rail | source | exact? |
|---|---|---|
| `ark` | none — off-chain send | exact, 0 |
| `onchain` | `getInfo().fees.intentFee` via `Estimator` | exact for the output fee |
| `onchain-swap` | `getFees("ARK","BTC")` | estimate |
| `lightning` | `getFees().submarine` | estimate |

`fee` is documented as a pre-send display/ranking figure, not a guarantee: Boltz returns the authoritative amount at swap creation, and the exit's per-input intent fees depend on the VTXO selection made at settlement, so they're excluded.

Two side effects worth flagging:

- Pricing the exit needs the destination script, so `Ramps.offboard`'s network-probing decode is extracted to `offboardDestinationScript` and shared — the quote is evaluated against the exact script settlement uses, which matters for a script-size-dependent program. It stays package-internal; the root barrel re-exports `Ramps` by name, not `export *`.
- Quoting now decodes the address, so an invalid one fails at `quote()` instead of inside `offboard()`. That incidentally closes arkana's finding #2 about `isBtcAddress` being regex-only. It also exposed that the payment test fixtures used a BTC address with a **bad checksum** — it only ever passed because `offboard` was mocked. Replaced with a valid vector across all seven files.

packages/boltz-swap/src/payment/onchain-swap.ts
Kukks: Fixed in `749f43e4` — normalized to "delivered to recipient" rather than exposed per rail.

You had the asymmetry exactly right: `receiverLockAmount: amt` here means the recipient gets `amt`, while the core rail hands the same `amt` to `Ramps.offboard`, which does `amount -= outputFee` before building the output — so the recipient got `amt - fee`. Three of the four rails were already receiver-exact, so the exit was the odd one out and grossing it up was the smaller change.

Of the two options you offered, normalizing won over a per-rail semantic flag because a flag pushes reconciliation onto every consumer and gives up the cost-based ranking that `fee`/`total` exist for — which is most of the point of surfacing both on-chain rails in `options()` at all.

This rail also stopped quoting `fee: 0`: it now reports the user-lock gross-up, sharing the reconstruction with `available()` via `grossUpUserLock` so the gate and the quote can't drift apart. Same fixture as the existing gate test — 100 000 sats at 2% + 1000 server + 500 claim gives `fee: 3572`, `total: 103_572`, `amount: 100_000`.

`quote()` now also rejects a fee rate outside `[0, 100)` instead of silently emitting a wrong total, mirroring the `available()` bail-out from `764b11ed`.

packages/ts-sdk/src/payment/rails/onchain.ts
Kukks: Answered in code as of `749f43e4` — the rail no longer reports `fee: 0`.

For this rail specifically: `getInfo().fees.intentFee`, evaluated through `Estimator.evalOnchainOutput` against the decoded destination script. That's the same call `Ramps.offboard` makes at settlement, and the rail now fetches the `FeeInfo` once at quote time and reuses that object for the offboard, so the quoted number and the settled one can't drift.

Across the four rails:

| rail | source | exact? |
|---|---|---|
| `ark` | none — off-chain send is free | exact, 0 |
| `onchain` | `getInfo().fees.intentFee` via `Estimator` | exact for the output fee |
| `onchain-swap` | `getFees("ARK","BTC")` | estimate |
| `lightning` | `getFees().submarine` | estimate |

The two swap rails are estimates by construction — Boltz only fixes the real number when the swap is created, so the quote uses its advertised pricing and is superseded by `expectedAmount` / the lockup amount at send time. `RouteQuote` documents `fee` as a display-and-ranking figure rather than a guarantee.

One deliberate exclusion on this rail: `offboard` also shaves a per-input intent fee off each selected VTXO, so the wallet's true debit is `total` plus those. They depend on the VTXO selection made at settlement, which the rail doesn't control (offboard selects internally), so folding a guess into `fee` would make it less honest, not more. That's called out in the `RouteQuote` docstring.

packages/boltz-swap/src/payment/index.ts
Kukks: Tracked as #658 so this thread can close.

The remaining half — the `route()` next-option fallback — stays out of this PR, for the reason already in the thread: `send()` returns a handle, and a chain-swap failure can be mid-flight (lockup funded, claim pending), so retrying on `onchain` there would select VTXOs the swap still needs. A blind fallback converts one stuck payment into a double-spend attempt. It needs a per-rail "safe to retry" signal that is conservative by default (unknown error ⇒ not retryable), which is a design change rather than a patch. #658 has the sketch.

Nothing unsafe ships meanwhile: the `available()` limits gate closes the concrete case you raised before `send()` is ever reached, and `options()` still surfaces every viable rail so an app can implement its own policy today.

Separately, both `fee: 0` sibling threads are now fixed rather than deferred — `749f43e4` makes `RouteQuote` receiver-exact across all four rails with real fee sources, so `options()` is rankable on cost. Details on those threads.

Kukks:
Pushed `749f43e4`, `ea558833`, `2ac5a323`.

**pietro's four open threads** — three fixed in code, one tracked:

- **Quote semantics (both 🟠 Medium threads).** `RouteQuote` now defines `amount`/`fee`/`total` once, receiver-exact, and all four rails honour it: `amount` is what the recipient gets, `fee` is charged on top, `total` is the wallet debit. The collaborative exit grosses up (`Ramps.offboard` deducts); the swap rails pass through. Real fee sources replace `fee: 0` — `getInfo().fees.intentFee` via `Estimator` for the exit, `getFees(...)` for both swap rails.
- **"Where does `fee` come from?"** Answered on the thread with the per-rail source table; the two swap rails are documented as estimates superseded by Boltz at swap creation.
- **🔴 High — `route()` next-option fallback.** Still deferred, now tracked in #658 with the reasoning: a chain-swap failure can be mid-flight, so falling back would select VTXOs the swap still needs. The `available()` limits gate already closes the concrete sub-minimum case before `send()` is reached.

Worth flagging: the gross-up is a **fixpoint, not a subtraction**. `outputToArgs` feeds `amount` into the CEL fee program, so at 1% on 10 000 sats the naive `amount + fee(amount)` lands the recipient a sat short. The test uses exactly that case so the naive form fails it.

Two side effects fell out. Pricing the exit needs its destination script, so `Ramps.offboard`'s network-probing decode is now shared via `offboardDestinationScript` — meaning quoting decodes the address, and an invalid one fails at `quote()` instead of inside `offboard()` (this also closes arkana's finding #2 about `isBtcAddress` being regex-only). That in turn exposed that the payment test fixtures used a BTC address with a **bad checksum**, which only ever passed because `offboard` was mocked; replaced with a valid vector across all seven files.

**arkana's findings:** C (the flaky `retryAt` assertion) is fixed in `ea558833` by pinning the clock. A and B are pre-existing `master` code — this PR touches only two small spots in `arkade-swaps.ts` — so they're verified and filed as #659 rather than folded in here.

**Bonus, separately revertible (`2ac5a323`):** both lint guard scripts compared `node:path` `relative()` output against forward-slash literals, so on Windows `check-virtual-status` reported all seven allowlisted files and `check-provider-boundary` skipped every file and passed vacuously. No behaviour change on CI, where `sep` is already `/`. Revert with `git revert 2ac5a323` if you'd rather it landed on its own.

**Gate:** lint clean, `pnpm run build` exit 0, ts-sdk **2193 passed** + 1 skip, boltz-swap **525 passed** — against a recorded baseline of 2191 + 1 skip and 522, so the deltas are exactly the tests added here with no regressions.

PR #590, opened 2026-06-27
feat(swaps): deterministic boltz preimage

d4rp4t:
Replace randomBytes(32) in reverse and chain swap creation with a preimage derived from the wallet's signing key, matching the [dotnet-sdk scheme](https://github.com/arkade-os/dotnet-sdk/pull/116). Test vectors were created in https://github.com/arkade-os/dotnet-sdk/pull/151

## Summary by CodeRabbit

* **New Features**
  * Swap preimages are now derived consistently from wallet identity, making swap creation and restoration more reproducible.
  * Added support for deterministic signing capabilities in the SDK, with a new way to detect compatible identities.
  * Expanded test coverage with cross-environment vectors to verify consistent preimage behavior.

* **Bug Fixes**
  * Restored swaps now recover missing local preimage data when it matches the expected hash, improving swap recovery reliability.


packages/boltz-swap/tsconfig.test.json
```diff
@@ -0,0 +1,12 @@
...
+    "extends": "../../config/tsconfig.base.json",
+    "compilerOptions": {
+        "lib": ["es2022", "dom"],
+        "types": ["node"],
+        "noEmit": true,
+        "noUnusedLocals": true,
+        "noUnusedParameters": true
+    },
+    "include": ["src/**/*.ts", "test/**/*.ts", "../../config/types/**/*.d.ts"],
+    "exclude": ["node_modules", "dist"]
```
pietro909, line 11: why this change? The file doesn't seem used

d4rp4t: i couldnt import json test vectors

packages/boltz-swap/src/utils/preimage.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Falls back to random bytes for identities that cannot guarantee determinism.
+ */
+export async function derivePreimage(
+    identity: Identity | ReadonlyIdentity,
+    index: number,
+): Promise<Uint8Array> {
+    if (!isDeterministicSignCapable(identity)) {
+        return randomBytes(32);
+    }
```
Dunsin-cyber, line 35: if the identity can't sign deterministically this silently returns a random preimage, so the swap will never be recoverable after restore. the PWA wallet's identities (SingleKey/MnemonicIdentity) are all deterministic so nothing hits this today, but in case an SDK consumer plugs in a custom signer, a `logger.warn` here would let them find out at create time, not after data loss. 

e.g
`logger.warn("identity cannot sign deterministically; preimage is random and will not be recoverable after")`

Dunsin-cyber (changes requested):

pietro909:
Cross-ref: #662 (HD swap keys + multi-key restore) leaves the preimage residuals to this PR — its `DescriptorIdentity` implements `signSchnorrDeterministic` for the swap's own index, so once this merges `derivePreimage` can be pointed at the swap's descriptor identity instead of `wallet.identity`, and the restore re-derivation retried with the attributed owner key.

pietro909:
Superseded by https://github.com/arkade-os/ts-sdk/pull/665 thank you for spearheading this one @d4rp4t !

PR #598, opened 2026-07-02
Fix/invoice expiry timestamp docs

bilthon:
`CreateLightningInvoiceResponse.expiry` (and `DecodedInvoice.expiry`) were documented as an absolute Unix-seconds timestamp, but the value is actually a relative delta: seconds from invoice creation until it expires. This mismatch caused a real bug as code that trusted the docstring multiplied `expiry` by 1000, as if it were an absolute timestamp and ended up treating every invoice as expired already.

This fixes the docstring to describe the value correctly, and adds a `timestamp` field (the invoice's own creation time, Unix seconds) to both types so callers can compute an absolute deadline (`timestamp + expiry`) using the invoice's own embedded time rather than their local clock. `decodeInvoice()` and `createLightningInvoice()` are updated to populate it, with a test covering the new field.

Both changes are additive/non-breaking.

## Summary by CodeRabbit

* **New Features**
  * Lightning invoice details now include the invoice creation timestamp, and decoded invoices now expose it as well.
* **Bug Fixes**
  * Invoice decoding now returns a timestamp consistently, defaulting to `0` when the BOLT11 timestamp section is missing.
* **Documentation**
  * Updated expiry documentation to clarify it represents seconds remaining from invoice creation until expiration.
* **Tests**
  * Extended invoice decoding tests to verify timestamp extraction.

### DIFF
--- a/packages/boltz-swap/src/types.ts
+++ b/packages/boltz-swap/src/types.ts
@@ -81,7 +81,7 @@ export interface CreateLightningInvoiceRequest {
 export interface CreateLightningInvoiceResponse {
     /** The on-chain amount in satoshis (after Boltz fees). */
     amount: number;
-    /** Invoice expiry timestamp (Unix seconds). */
+    /** Invoice expiry: seconds from invoice creation (timestamp) until it expires. */
     expiry: number;
     /** The BOLT11-encoded Lightning invoice string. */
     invoice: string;
--- a/packages/boltz-swap/src/types.ts
+++ b/packages/boltz-swap/src/types.ts
@@ -83,7 +83,7 @@ export interface CreateLightningInvoiceResponse {
     amount: number;
     /** Invoice expiry: seconds from invoice creation (timestamp) until it expires. */
     expiry: number;
-    /** Invoice creation timestamp (Unix seconds). */
+    /** Invoice creation timestamp (Unix seconds). Always present in a valid BOLT11 invoice; `0` should not occur in practice. */
     timestamp: number;
     /** The BOLT11-encoded Lightning invoice string. */
     invoice: string;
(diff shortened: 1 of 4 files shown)


packages/boltz-swap/src/utils/decoding.ts
```diff
@@ -12,6 +12,7 @@ export const decodeInvoice = (invoice: string): DecodedInvoice => {
     const millisats = BigInt(decoded.sections.find((s) => s.name === "amount")?.value ?? "0");
     return {
         expiry: decoded.expiry ?? 3600,
+        timestamp: decoded.sections.find((s) => s.name === "timestamp")?.value ?? 0,
```
pietro909, line 17: A bit divided on the default zero value. I see this is the same pattern followed for `expiry` so it's not really some you are introducing... very Go-ish. 

In the future, we may want to validate that a number is actually a number and return an error if not or an optional type. Noting down.

packages/boltz-swap/src/types.ts
```diff
@@ -81,8 +81,10 @@ export interface CreateLightningInvoiceRequest {
 export interface CreateLightningInvoiceResponse {
     /** The on-chain amount in satoshis (after Boltz fees). */
     amount: number;
-    /** Invoice expiry timestamp (Unix seconds). */
+    /** Invoice expiry: seconds from invoice creation (timestamp) until it expires. */
     expiry: number;
+    /** Invoice creation timestamp (Unix seconds). */
+    timestamp: number;
```
pietro909, line 87: Let's document that `0` is a fallback value meaning "no timestamp was provided" for the consumers to decide how to handle it.

pietro909 (changes requested):
Thank you for your contribution @bilthon 

My request for the change is [this](https://github.com/arkade-os/ts-sdk/pull/598#discussion_r3518546990) to make it clear that `0` is the fallback value in the docs.

packages/boltz-swap/src/types.ts
bilthon: Adding a zero there was more of a way to get around the TS type checker than anything else. According to the [BOLT11](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md#data-part) spec, the timestamp field is mandatory. So we don't really expect a valid BOLT11 invoice to come without one.

I'm adding this as a clarification to the code regardless, as requested.

pietro909 (approved):

PR #601, opened 2026-07-07
fix(sdk): wrap transport-level fetch failures in a typed FetchError

### LINKED ISSUE #593: baseFetch should handle fetch errors
Function definition is [here](https://github.com/arkade-os/ts-sdk/blob/5f87cd8fe6c33206233cadb7a0538df2c7fc6680/packages/ts-sdk/src/utils/fetch.ts#L16)

Kukks:
## Problem
`baseFetch` (`packages/ts-sdk/src/utils/fetch.ts`) — and the Ark-server `fetch` wrapper that delegates to it — returned `globalThis.fetch(...)` directly, so a transport-level rejection (DNS failure, connection refused, TLS/CORS) propagated as the raw, contextless platform error: `TypeError: Failed to fetch` in browsers, the opaque `TypeError: fetch failed` in Node. Callers can't tell *which* request failed, nor cleanly discriminate a network failure from other errors.

Closes #593.

## Change
- Add `FetchError extends Error` (colocated in `utils/fetch.ts`, matching the repo's single-module error convention — `DigestMismatchError`, `DustChangeError`, etc.). It carries `readonly url` / `readonly method` and forwards the original error as `cause`.
- `baseFetch` now `.catch()`es the platform `fetch` rejection and re-throws a `FetchError` naming the request (`method URL`). The Ark-server `fetch` wrapper inherits this for free (it delegates).
- Re-export `FetchError` from `src/index.ts` (next to `DigestMismatchError`) so consumers can `instanceof`-discriminate it.
- The "Fetch API is not available" guard keeps its **synchronous** throw (programmer/environment error, not transport) — the `.catch` approach is deliberate so that semantic is preserved.
- No retry — retrying stays the caller's responsibility.
- Widened `input` from `RequestInfo` to `RequestInfo | URL` (matches the platform `fetch` signature; backward-compatible).

## Test plan
- 10 new unit tests in `test/fetch.test.ts` (written red-first): transport rejection → `FetchError` with correct `name`/message/`cause`; method/URL derivation from `string | URL | Request`; happy-path passthrough; the unavailable-fetch guard still throws synchronously and is **not** wrapped; the Ark-server `fetch` wrapper surfaces the wrapped error too.
- Full ts-sdk unit suite green (1555 passed); pre-commit hook (both packages' `test:unit` + lint) passes.

## Summary by CodeRabbit

* **New Features**
  * Added a public fetch error type that includes the request URL, HTTP method, and original failure details.

* **Bug Fixes**
  * Network request failures are now reported more consistently with clearer error messages.
  * Fetch helpers now support both string/URL inputs and preserve successful responses unchanged.

### DIFF
--- a/packages/ts-sdk/src/utils/fetch.ts
+++ b/packages/ts-sdk/src/utils/fetch.ts
@@ -8,16 +8,48 @@ export const buildVersion = "0.9.9";
 export const sdkVersion = `ts-sdk/${version}`;
 
 /**
+ * Thrown by {@link baseFetch} — and therefore by the Ark-server {@link fetch}
+ * wrapper, which delegates to it — when the underlying platform `fetch` rejects
+ * at the transport layer: DNS failure, connection refused, TLS or CORS error,
+ * etc. The raw platform rejection (`TypeError: Failed to fetch` in browsers, the
+ * opaque `TypeError: fetch failed` in Node) says nothing about which request
+ * failed; this wraps it with the request method and URL and preserves the
+ * original as {@link Error.cause}. It does NOT retry — retrying is the caller's
+ * responsibility.
+ */
+export class FetchError extends Error {
+    /** The request URL that failed, when derivable from the `fetch` input. */
+    readonly url?: string;
+    /** The HTTP method of the failed request (defaults to `"GET"`). */
+    readonly method?: string;
+
+    constructor(message: string, options: { url?: string; method?: string; cause?: unknown }) {
+        super(message, { cause: options.cause });
+        this.name = "FetchError";
+        this.url = options.url;
+        this.method = options.method;
+    }
+}
+
+/**
  * Guarded passthrough to the platform `fetch` with no Arkade-specific headers.
  * Use for any service that is NOT the Ark server (delegate, Esplora, …): those
  * origins reject unknown request headers such as `X-Build-Version` in the CORS
  * preflight.
+ *
+ * A transport-level rejection from the platform `fetch` is re-thrown as a
+ * {@link FetchError} that names the request and keeps the original error as
+ * `cause`, so callers can tell *which* request failed instead of receiving an
+ * opaque `TypeError: Failed to fetch`.
  */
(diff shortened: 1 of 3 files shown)


packages/ts-sdk/src/utils/fetch.ts
```diff
@@ -7,17 +7,49 @@ export const buildVersion = "0.9.9";
...
+/**
+ * Thrown by {@link baseFetch} — and therefore by the Ark-server {@link fetch}
+ * wrapper, which delegates to it — when the underlying platform `fetch` rejects
+ * at the transport layer: DNS failure, connection refused, TLS or CORS error,
+ * etc. The raw platform rejection (`TypeError: Failed to fetch` in browsers, the
+ * opaque `TypeError: fetch failed` in Node) says nothing about which request
+ * failed; this wraps it with the request method and URL and preserves the
+ * original as {@link Error.cause}. It does NOT retry — retrying is the caller's
+ * responsibility.
+ */
```
pietro909, line 14: too verbose: the note about retry logic is irrelevant and even the reasons for this error to exist boil down to "wrap platform and transport layer errors with context preserving the original error"

packages/ts-sdk/test/fetch.test.ts
```diff
@@ -0,0 +1,144 @@
...
+        expect(err).toBeInstanceOf(FetchError);
+        expect(err.name).toBe("FetchError");
+        expect(err.message).toContain("https://esplora.test/address/abc/utxo");
+        expect(err.message).toContain("GET");
+        expect(err.url).toBe("https://esplora.test/address/abc/utxo");
+        expect(err.method).toBe("GET");
+        expect(err.cause).toBe(cause);
+    });
+
+    it("reflects the request method from init in the wrapped error", async () => {
```
pietro909: this and the previous test can be merged without losing much. The only missing part would be the implicit "GET" fallback.

pietro909 (changes requested):
Please let's make an effort to limit obvious docs/comments and unit tests. They are cheap but still represent a burden when reading the code.

PR #606, opened 2026-07-08
feat(sdk): pre-signed unilateral exit packages

Kukks:
## Summary

Adds a `UnilateralExit` namespace to `@arkade-os/sdk` that pre-signs the transactions needed to move VTXOs onchain, plus a keyless `Executor` to drive them. Two funding modes:

- **Funded** (`mode: "funded"`, default) — a splitter tx (broadcast at `prepare()` time) pre-funds pre-signed CPFP fee children. The package executes **fully keyless**: the executor needs no keys at all.
- **Graph** (`mode: "graph"`) — transports only the tx graph + pre-signed sweeps; **no splitter, no pre-signed children**. The executor funds and signs the CPFP bumps at execution time from its own fee wallet, enabling a "send funds to this address and we proceed" flow.

Three entry points:

- **`UnilateralExit.estimate(opts)`** — quote (tx count, fees, funding required + shortfall, funding address, recovered amount, validity horizon, per-VTXO breakdown incl. skips). No funds, no signing.
- **`UnilateralExit.prepare(opts)`** — signs everything and returns a versioned, language-agnostic JSON `ExitPackage`. Funded mode broadcasts the splitter as a side effect.
- **`UnilateralExit.Executor`** — keyless, stateless, idempotent async-iterator over the package. Graph packages take an `ExitFeeWallet` (implemented by `OnchainWallet.bumpAnchor`).

### How it works

- **Fan-out splitter** (funded) creates one dedicated funding output per unroll step; each fee child spends `[P2A anchor + its own funding output]`, so one externally-bumped/conflicted step never invalidates the others.
- **One CSV sweep per VTXO** (not a batch) — a keyless executor can't re-sign, so an invalidated input never blocks the rest.
- **Shared ancestors deduped into a DAG**: shared txs become one step/child/funding-output; `forVtxos` records which VTXOs each step serves.
- **Contract-aware exit paths** via the handler registry (`getAllSpendingPaths`, `collaborative:false`): a VHTLC claimed by the receiver embeds the preimage in the pre-signed sweep witness. A universal sign-and-verify guard skips paths needing other signers, with a reason.

### Reuse over duplication

Shared primitives extracted from the existing online flow rather than copied: `finalizeVirtualTx` (from `Unroll.Session`), `buildAnchorChild` (pure core of `bumpP2A`), `buildBumpPackage` (shared by `bumpP2A` + the new `bumpAnchor`), and handler-driven `resolveUnilateralPath` (which `completeUnroll`/`prepareUnrollTransaction` migrate to — identical for default VTXOs; contract VTXOs become exitable there too). `bumpP2A` also gained a dust guard.

## Test plan

- **Unit** (`test/exit-*.test.ts`): codec round-trip + version/step validation; `finalizeVirtualTx`; `buildAnchorChild` fee/dust; sweep builder incl. condition-witness stack order; handler path resolution; DAG dedupe/topo-sort/cycle detection; estimate/prepare math incl. quote↔prepare consistency, the deposit-UTXO funding case, and graph-mode bump steps; executor state machine (skip-if-confirmed, CSV maturity blocks & time, dead-branch isolation, non-BIP68-final retry, graph bump via fee wallet). **Full ts-sdk unit suite: 1585 passing / 1 skipped.**
- **e2e** (`test/e2e/unilateralExit.test.ts`, regtest, verified from the feature worktree — **7/7** across `onchain` + `vhtlc` + `unilateralExit`): single default VTXO end-to-end; graph mode funding from an independent ephemeral wallet; two VTXOs sharing a commitment (DAG dedupe); VHTLC receiver claim (ground truth for condition-witness order). The refactored consumers (`onchain`, `vhtlc` incl. unilateral claim) pass unchanged.

## Notes

- Package files can embed condition witnesses (e.g. VHTLC preimages) — treat as confidential until broadcast.
- Companion static executor UI (`arkade-unilateral-exit`) is built to consume this API — supports both modes, deploys to GitHub Pages. Ships in its own repo.

## Summary by CodeRabbit

* **New Features**
  * Added unilateral exit “packages” end-to-end: estimation/quoting, pre-signed preparation, and a keyless executor that streams lifecycle events.
  * Introduced `"graph"` mode with deterministic bump ordering, deduped shared ancestry handling, and contract-aware unilateral path resolution.
  * Added SDK support for constructing anchor-child transactions and CSV-based sweep transactions (including conditional witness handling).
* **Documentation**
  * Documented the pre-signed unilateral exit package workflow in the TS SDK README.
* **Refactor**
  * Updated unroll and on-chain bump/anchoring flows to use the new shared exit/transaction building helpers.
* **Tests**
  * Added/expanded unit, codec, path, executor, and E2E coverage for the full unilateral exit flow.


packages/ts-sdk/README.md
```diff
@@ -797,6 +797,60 @@ await Unroll.completeUnroll(
 - The `completeUnroll` method can only be called after all virtual outputs are fully unrolled and the timelock has expired
 - You need sufficient onchain funds in the `OnchainWallet` to pay for P2A transaction fees
 
+### Unilateral Exit Packages (pre-signed)
+
+`Unroll.Session` requires the wallet (keys + indexer access) to stay online for the whole
+multi-day exit. `UnilateralExit` removes that requirement: it pre-signs **every** transaction
+needed to land the funds onchain and emits a versioned JSON package that anything with an
+Esplora-compatible endpoint can execute — no keys, no Arkade infrastructure.
+
```
pietro909, line 807: let's add a note about assets? like "BTC-only; do not pass asset-bearing VTXOs"

packages/ts-sdk/src/wallet/exit/estimate.ts
```diff
@@ -0,0 +1,313 @@
...
+        indexer: wallet.indexerProvider,
+        onchain: wallet.onchainProvider,
+    });
+
+    // Fetch pending virtual txs and key their PSBTs by unsigned txid (the
+    // id is witness-independent, so it is stable pre-finalization).
+    const pendingNodes = dag.filter((n) => !n.confirmed);
+    const psbts = new Map<string, string>();
+    if (pendingNodes.length > 0) {
+        const res = await wallet.indexerProvider.getVirtualTxs(pendingNodes.map((n) => n.txid));
```
pietro909: Shall we guard against pagination in `res.page`? The risk is to throw later with an `indexer did not return virtual tx ` which is misleading

packages/ts-sdk/src/wallet/exit/chain.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+export async function buildExitDag(params: {
+    vtxos: { txid: string; vout: number }[];
+    indexer: IndexerProvider;
+    onchain: OnchainProvider;
+}): Promise<DagNode[]> {
+    const byTxid = new Map<string, ChainTx & { forVtxos: Set<string> }>();
+
+    for (const vtxo of params.vtxos) {
+        const { chain } = await params.indexer.getVtxoChain({
```
pietro909: Same as ./estimate.ts: the call is paginated and, while an edge case, we should handle pagination.

packages/ts-sdk/README.md
Kukks: Good call — added a note in 1bc4c830: unilateral exit handles BTC value only, and asset-bearing VTXOs (non-empty `assets`) must not be passed.

packages/ts-sdk/src/wallet/exit/estimate.ts
Kukks: Fixed in 1bc4c830 — `getVirtualTxs` now walks all pages (a short page, or absent `page` metadata, marks end of history — matching the pattern in `contracts/handlers/helpers.ts`), so a chain deeper than one page no longer drops txs and throws the misleading `indexer did not return virtual tx`.

packages/ts-sdk/src/wallet/exit/chain.ts
Kukks: Fixed in 1bc4c830 — `getVtxoChain` in `buildExitDag` now paginates the same way.

pietro909 (approved):

PR #607, opened 2026-07-08
feat(arkade): typed program params, $param timelocks, condition+csv closure

tiero:
Aligns the `Program` artifact shape (`src/arkade/program.ts`) with the ArkadeScript compiler's direction, while keeping the SDK code minimal. These are **breaking changes to the artifact shape at version 0** — since PR #319 is unmerged, `SUPPORTED_PROGRAM_VERSION` stays `0` and no migration code is needed.

> **Stacking:** this branch is built on top of PR #319 (`arkade-script-final`). The base of this PR is a throwaway branch (`base/pr-319-arkade-script`) pinned at #319's head so the diff shows only the four commits below. Once #319 merges, retarget this PR onto `master`.

## Changes (one per commit)

1. **`$param`-referenceable timelocks** — `TapscriptSegment.csv` is now `{ type: "blocks" | "seconds"; value: bigint | string }` and `cltv` is `bigint | string`. A string must be a `"$param"` reference, resolved against the program's constructor args at compile time (same convention as `SignerRef`). A single shared helper `resolveTimelockValue` handles `bigint | string → bigint` (bigint passes through; `"$name"` binds via `bindValue` and must resolve to a bigint/number; any other string errors). Serialization emits literals as decimal strings and `"$param"` refs as-is; parsing keeps `"$…"` a string and `BigInt()`s everything else.

2. **The fifth arkd closure: condition + CSV** — `asm` + `csv` on the same tapscript segment is now valid and encodes via the existing `ConditionCSVMultisigTapscript`. `asm` + `cltv` and `csv` + `cltv` remain errors, with messages naming the invalid pair (arkd recognizes no condition+CLTV or dual-timelock closure).

3. **Typed constructor params, authoritative when present** — `Program.params` is now `readonly (string | { name: string; type: ArkadeArgType })[]`, reusing the existing `ArkadeArgType` / `InputRef` union (`inputName` was generalized to cover params too). When any entry is typed, the exported `validateProgram(program, args)` (called from `compileFunctions`) enforces that every declared param is bound, every `"$name"` reference anywhere in the program is declared (collected by one small walker), and typed values match their type (bytes-like → `Uint8Array`; `pubkey` 32 bytes, `sig` 64 bytes; `int` → bigint/number). Typed declarations also direct persisted-arg decoding in `deserializeArkadeContractParams` (bytes-like require `0x` hex; int decodes a number or decimal string) — the regression this prevents is a typed bytes value without a `0x` prefix silently becoming a bigint under the old heuristic. Bare-string params keep the legacy advisory behavior, so existing hand-written programs and all current tests are unaffected. Error messages name the parameter and expected type.

4. **Optional program `name`** — added `name?: string` to `Program`. Metadata only: round-tripped by `stringifyArtifact` / `parseArtifact`, never influencing compilation or the taproot tree.

Also updated the module-level doc comment in `program.ts` and the `Program`-related JSDoc/example in `contract.ts`.

## Tests

New unit tests in `packages/ts-sdk/test/arkade-program.test.ts` (vitest, no e2e):

- Full-feature round-trip: typed params, `"$exit"` csv value, literal cltv, condition+csv function, arkadeScript with witness, and a bigint token above `Number.MAX_SAFE_INTEGER` survive `parseArtifact(JSON.parse(stringifyArtifact(p)))` deep-equal (with the documented script-num canonicalization).
- `"$param"` csv/cltv leaves byte-identical to `CSVMultisigTapscript.encode` / `CLTVMultisigTapscript.encode` with the literal timelock.
- condition+csv leaf byte-identical to `ConditionCSVMultisigTapscript.encode`.
- Validation errors (asserted on message substrings): undeclared `"$param"` reference, declared param missing from args, typed pubkey bound to 33 bytes, typed int bound to bytes, `asm`+`cltv` rejected, `csv`+`cltv` rejected.
- Type-directed persistence: typed int from a decimal string deserializes to bigint; typed bytes without `0x` throws (with a comment marking the regression this prevents).
- Program `name` proven not to affect the compiled tree.

Back-compat for untyped `params` is covered by the pre-existing suite, which passes unchanged.

## Verification

- `pnpm run lint` (prettier) passes.
- Both packages build (`pnpm run build`).
- Full non-e2e vitest suite passes (ts-sdk + boltz-swap, run by the pre-commit hook on each commit).

No new dependencies, no `SUPPORTED_PROGRAM_VERSION` bump, no changes to the `boltz-swap` package.

https://claude.ai/code/session_019YqJ7z6yuyzh8qDHiqYEFW

---
_Generated by [Claude Code](https://claude.ai/code/session_019YqJ7z6yuyzh8qDHiqYEFW)_


louisinger:
included to louisinger/arkade-script-final branch (36f61c9462b2004f39ad6bab7e5178cb8791c929)

PR #608, opened 2026-07-09
feat(sdk): VTXO exit-data availability — local-first resolver + capture/prune

Kukks:
## Summary

Makes unilateral exit work without the Ark indexer: capture each VTXO's exit chain data locally on receive, and read it local-first at exit time. This activates `VirtualTxRepository` (the declared follow-up of #494) and adds a modular resolver, ported from NArk's design.

**Stacked on #606** (base `worktree-unilateral-exit-package`) — merge after #606 lands, or rebase onto `master` once it does.

### Resolver seam (local-first reads)
- `ExitDataSource` + `ExitChainResolver` (`src/wallet/exit/resolver.ts`): an ordered source chain — local repo → provider slot → indexer — with best-effort read-through persist.
- `IndexerExitDataSource` (paginated, behavior-preserving) and `RepositoryExitDataSource` (Full derives physical `spends` from stored PSBTs; Lite reports a structure miss so the resolver falls through).
- Threaded into `buildExitDag` and `computeExitLayout`. With no `virtualTxRepository` configured it is a pure no-op that falls through to the indexer.

### Capture + prune
- `captureExitBranch` / `pruneExitBranches` (`src/wallet/exit/capture.ts`): idempotent, dust-gated (`minExitWorthSats`), Lite/Full modes (**default Lite**).
- Wired into `ContractManager` via best-effort `onVtxosPersisted` / `onVtxosSpent` callbacks bound by the wallet — keeping `contracts/` independent of `wallet/exit/`. Capture fires after the sync write path and at `reconcilePendingFrontier`; prune fires on `vtxo_spent`.
- Prune is ref-counted (via the repo): ancestors shared with surviving VTXOs are kept, so change / self-spend "keep + extend" falls out of the normalized `VirtualTx` + `VtxoBranch` storage — no special-casing.

### Configurability & extensibility
- **Lite/Full + dust:** `storage.exitDataCapture.mode` — **default "lite"** (structure only, cheap; most VTXOs never exit); set `"full"` to store PSBTs for indexer-independent exit. `minExitWorthSats` (default 1000) skips dust.
- **Provider source slot:** `storage.exitDataCapture.sources?: ExitDataSource[]` — extra sources tried before the indexer for both capture and exit reads. `ExitDataSource` / `ExitChainResolver` / `createExitChainResolver` / `ExitCaptureMode` are exported so an integration can implement a provider-backed source.

## Test plan
- Unit: resolver composition + read-through + source ordering, indexer/repository sources, capture/prune helpers (Full, Lite, dust, idempotent re-receive, ref-counted prune).
- Full ts-sdk unit suite: **1742 passed / 1 skipped**; tsc clean; behavior-preserving with no repo configured.
- e2e (regtest): capture on receive (Full) → chain resolves from the repo with the indexer offline → prune on spend.

## Still deferred
- A concrete provider *implementation* (wire protocol + client) — the slot and interface are in place; the protocol needs a target service to define.

### DIFF
--- a/packages/ts-sdk/src/wallet/exit/capture.ts
+++ b/packages/ts-sdk/src/wallet/exit/capture.ts
@@ -16,6 +16,10 @@ export type ExitCaptureMode = "lite" | "full";
 /** BTC-value floor for capture; below it, exit fees would exceed the VTXO (NArk parity). */
 export const DEFAULT_MIN_EXIT_WORTH_SATS = 1000;
 
+/** Default capture mode: Lite (structure only). Full (store PSBTs) is opt-in for
+ *  wallets that want indexer-independent exit — most VTXOs never exit (NArk parity). */
+export const DEFAULT_EXIT_CAPTURE_MODE: ExitCaptureMode = "lite";
+
 /** ChainTxType (indexer string enum) → ChainedTxType (repo numeric enum).
  *  Duplicated from unroll.ts:27 to keep the exit path free of a cross-module dep. */
 function chainTxTypeToChainedExit(t: ChainTxType): ChainedTxType {
--- a/packages/ts-sdk/src/wallet/exit/capture.ts
+++ b/packages/ts-sdk/src/wallet/exit/capture.ts
@@ -1,0 +1,111 @@
+import { base64 } from "@scure/base";
+import { ChainTx, ChainTxType } from "../../providers/indexer";
+import {
+    ChainedTxType,
+    VirtualTx,
+    VirtualTxRepository,
+    VtxoBranch,
+} from "../../repositories/virtualTxRepository";
+import { Transaction } from "../../utils/transaction";
+import { Outpoint } from "../index";
+import { topoSortByDeps } from "./chain";
+import { ExitChainResolver } from "./resolver";
+
+export type ExitCaptureMode = "lite" | "full";
+
+/** BTC-value floor for capture; below it, exit fees would exceed the VTXO (NArk parity). */
+export const DEFAULT_MIN_EXIT_WORTH_SATS = 1000;
+
+/** ChainTxType (indexer string enum) → ChainedTxType (repo numeric enum).
+ *  Duplicated from unroll.ts:27 to keep the exit path free of a cross-module dep. */
+function chainTxTypeToChainedExit(t: ChainTxType): ChainedTxType {
+    switch (t) {
+        case ChainTxType.COMMITMENT:
+            return ChainedTxType.Commitment;
+        case ChainTxType.ARK:
+            return ChainedTxType.Ark;
+        case ChainTxType.TREE:
+            return ChainedTxType.Tree;
+        case ChainTxType.CHECKPOINT:
+            return ChainedTxType.Checkpoint;
+        default:
+            return ChainedTxType.Unspecified;
+    }
+}
+
+/** Order a chain ancestors-first so position 0 is the root/commitment. */
+function orderAncestryFirst(chain: ChainTx[]): ChainTx[] {
+    const ids = new Set(chain.map((c) => c.txid));
+    return topoSortByDeps(
+        chain,
+        (c) => c.txid,
+        (c) => c.spends.filter((s) => ids.has(s)),
+    );
+}
+
+/**
+ * Fetch and persist a received VTXO's unilateral-exit branch. Idempotent (skips
+ * if a branch is already stored) and dust-gated. Full mode stores PSBTs so the
+ * exit needs no indexer; Lite stores structure only. Throws on failure — callers
+ * wrap best-effort so capture never blocks receive.
+ */
+export async function captureExitBranch(params: {
+    resolver: ExitChainResolver;
+    repository: VirtualTxRepository;
+    vtxo: Outpoint;
+    value: number;
+    mode: ExitCaptureMode;
+    minExitWorthSats: number;
+}): Promise<void> {
+    const { resolver, repository, vtxo, value, mode, minExitWorthSats } = params;
+    if (value < minExitWorthSats) return;
+    if (await repository.hasBranch(vtxo)) return;
+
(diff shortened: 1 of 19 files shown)


packages/ts-sdk/src/wallet/exit/capture.ts
```diff
@@ -0,0 +1,115 @@
...
+} from "../../repositories/virtualTxRepository";
+import { Transaction } from "../../utils/transaction";
+import { Outpoint } from "../index";
+import { topoSortByDeps } from "./chain";
+import { ExitChainResolver } from "./resolver";
+
+export type ExitCaptureMode = "lite" | "full";
+
+/** BTC-value floor for capture; below it, exit fees would exceed the VTXO (NArk parity). */
+export const DEFAULT_MIN_EXIT_WORTH_SATS = 1000;
```
pietro909, line 17: How stable will this value be? Worth having it from configuration?

packages/ts-sdk/src/wallet/exit/capture.ts
```diff
@@ -0,0 +1,115 @@
...
+
+/** Default capture mode: Lite (structure only). Full (store PSBTs) is opt-in for
+ *  wallets that want indexer-independent exit — most VTXOs never exit (NArk parity). */
+export const DEFAULT_EXIT_CAPTURE_MODE: ExitCaptureMode = "lite";
+
+/** ChainTxType (indexer string enum) → ChainedTxType (repo numeric enum).
+ *  Duplicated from unroll.ts:27 to keep the exit path free of a cross-module dep. */
+function chainTxTypeToChainedExit(t: ChainTxType): ChainedTxType {
```
pietro909, line 25: I get why we don't want the cross-and-potentially-cyclical dependency between `unroll.ts` and `capture.ts` but this helper maps from _indexer_ to _repository_ domains. Probably the right place is somewhere in `../utils.ts` given that the first-level `wallet` depends on both.

packages/ts-sdk/src/wallet/exit/capture.ts
```diff
@@ -0,0 +1,115 @@
...
+    vtxo: Outpoint;
+    value: number;
+    mode: ExitCaptureMode;
+    minExitWorthSats: number;
+}): Promise<void> {
+    const { resolver, repository, vtxo, value, mode, minExitWorthSats } = params;
+    if (value < minExitWorthSats) return;
+    if (await repository.hasBranch(vtxo)) return;
+
+    const ordered = orderAncestryFirst(await resolver.getVtxoChain(vtxo));
```
pietro909, line 68: this will hit the indexer every time, right after we've just hit it because of `syncContracts` called when we receive/spend a VTXO.

Do we have a chance to cache some data somehow?

packages/ts-sdk/src/wallet/exit/capture.ts
```diff
@@ -0,0 +1,115 @@
...
+                psbtByTxid.set(Transaction.fromPSBT(base64.decode(psbt)).id, psbt);
+            }
+        }
+    }
+
+    const virtualTxs: VirtualTx[] = ordered.map((c) => {
+        const expires = Number(c.expiresAt);
+        return {
+            txid: c.txid,
+            psbt: psbtByTxid.get(c.txid) ?? null,
```
pietro909, line 86: This makes the `mode === "full"` a best effort without explicitly mentioning it. How reliable should it be ?

pietro909 (changes requested):
I'm skeptical about the `mode` (full vs lite) because it's best effort and an upgrade to "full" won't backfill the repository, leading to inconsistency.

packages/ts-sdk/src/wallet/exit/capture.ts
```diff
@@ -0,0 +1,115 @@
...
+    const { resolver, repository, vtxo, value, mode, minExitWorthSats } = params;
+    if (value < minExitWorthSats) return;
+    if (await repository.hasBranch(vtxo)) return;
+
+    const ordered = orderAncestryFirst(await resolver.getVtxoChain(vtxo));
+
+    const psbtByTxid = new Map<string, string>();
+    if (mode === "full") {
+        const nonCommitment = ordered
+            .filter((c) => chainTxTypeToChainedExit(c.type) !== ChainedTxType.Commitment)
```
pietro909, line 73: `ChainTxType.UNSPECIFIED` is treated as root-type in `chain.ts:13` and `unroll.ts:195` but here will be part of `nonCommitment` and will try to fetch virtua Txs for it. Which returns no txs.

My understanding of `UNSPECIFIED` is that is a _must never happen_ case, yet we are modeling it.

packages/ts-sdk/src/wallet/exit/capture.ts
Kukks: lets make it optional config and falls back to this

packages/ts-sdk/src/wallet/exit/capture.ts
Kukks: ACK

packages/ts-sdk/src/wallet/exit/capture.ts
Kukks: indeed it is heavy.  The only way we can optimzie is by being smart: Is this vtxo a change output of our own previous vtxos?  That would mean that ensuring storing the tx as we construct and finalize it locally is actuall all the data we needed. (no calls to the indexer at all in that case)

packages/ts-sdk/src/wallet/exit/capture.ts
Kukks: Baby steps.

Kukks:
>I'm skeptical about the mode (full vs lite) because it's best effort and an upgrade to "full" won't backfill the repository, leading to inconsistency.

We can maybe do an  interval based backfill strategy as a followup or a dedicated function to do this when you switch from lite to full

PR #609, opened 2026-07-10
Offline-first wallet: survive an offline/uncooperative operator

pietro909:
Makes the wallet degrade gracefully instead of crashing when the Arkade operator/indexer is unreachable.

- **Cached server-info fallback** — a previously synced wallet boots from a persisted `ArkInfo` snapshot when live `getInfo` is down; a no-cache boot fails with a typed retryable error. The snapshot is written only after construction validates the response, so a terminal response can't poison the cache.
- **Typed provider errors** — indexer/ark transport and 429/5xx failures normalize to a retryable `ProviderUnavailableError`; 4xx/schema/network-mismatch stay terminal.
- **Offline-first reads** — `ContractManager` boot/reads/hydration and `getBalance`/`getVtxos`/history serve repository state on retryable failure (no partial mutation, no cursor advance); cooperative send/settle/refresh still fail with the typed error.
- **Degradation diagnostics** — `ProviderConnectionState` composes boot (cache) and indexer-sync (repository) degradation, surfaced across in-process, service-worker (`GET_STATUS` / `GET_CONTRACT_SYNC_STATE`), and Expo foreground wallets.

Implements Step 2 of `plans/unilateral-exit-sdk-readiness-v2.md`. Experimental exit APIs stay experimental.

## Summary by CodeRabbit

* **New Features**
  * Added provider connection and contract sync diagnostics across wallet, Expo, and service-worker APIs.
  * Added offline-first server-info snapshot caching (live vs cache reporting) for temporary provider outages.
  * Added typed provider-unavailable errors plus a retryability classifier, and exposed additional provider status helpers.
* **Bug Fixes**
  * Retryable provider/indexer/operator sync failures now degrade health and continue serving repository-backed data instead of failing operations.
  * Improved Arkade/indexer HTTP unavailability detection and made transaction-history enrichment best-effort.
* **Tests**
  * Expanded coverage for snapshot caching, offline reads, degraded/online transitions, provider availability classification, and related e2e timing.

### DIFF
--- a/packages/ts-sdk/src/contracts/contractManager.ts
+++ b/packages/ts-sdk/src/contracts/contractManager.ts
@@ -1,5 +1,6 @@
 import { hex } from "@scure/base";
 import { IndexerProvider } from "../providers/indexer";
+import { isRetryableProviderError } from "../providers/availability";
 import { WalletRepository } from "../repositories/walletRepository";
 import {
     Contract,
@@ -153,6 +154,17 @@ export interface ScanContractsOptions {
     deps: DiscoveryDeps;
 }
 
+/**
+ * Freshness of the ContractManager's provider-backed sync. `degraded` means the
+ * most recent sync (boot, best-effort read, or contract hydration) hit a
+ * retryable indexer/operator failure and the manager is serving repository
+ * state; it returns to `online` on the next successful sync. This only
+ * describes sync freshness — never wallet data itself.
+ */
+export type ContractSyncState =
+    | { mode: "online"; lastSyncedAt?: number }
+    | { mode: "degraded"; reason: string; lastSyncedAt?: number };
+
 export interface IContractManager extends Disposable {
     /**
      * Create and register a new contract.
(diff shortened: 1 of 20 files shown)


Kukks:
Cross-link: I opened #610 as the exit-side follow-up to this.

#609 makes `create`/reads survive an unreachable operator but (intentionally) leaves the experimental exit path untouched — and `selectExitVtxos`' explicit-outpoint branch still calls `indexerProvider.getVtxos({ outpoints })` directly, so `prepare({ vtxos })` throws when the operator is down even though the VTXOs are cached locally. #610 (stacked on #608) falls back to the local repo there.

It currently ships a tiny `isOperatorUnreachable` shim so it builds standalone on #608; the plan is to swap it for your exported `isRetryableProviderError` once this lands (a one-liner) and drop the shim — so the only touchpoint between the two is the classifier, which converges onto yours.

Verified the integration end-to-end: merging #608 + #609 + #610 is a clean 2-file / 4-hunk union (your degrade-wrapper composes with #608's prune callback; snapshot-save alongside `exitDataCapture`), `tsc`/`build` clean, full unit suite green, and an offline `prepare({ vtxos })` e2e builds the exit package from Full-mode local data with the indexer pointed at a dead port. Happy to rebase #610 onto master once #609 is in.

Kukks (approved):

PR #611, opened 2026-07-10
test(sdk): regtest e2e for offline wallet create + cached reads

Kukks:
Adds an end-to-end regtest test for #609's offline-first wallet. #609's coverage is unit-level; this exercises the real degraded-boot path against the live stack.

The test:
1. Creates a wallet online against the operator (persists the `ArkInfo` snapshot + funds a VTXO into shared repos).
2. Re-creates the same wallet (same identity + repos) pointed at an unreachable operator (port 9 → deterministic retryable failure) — `Wallet.create` **succeeds** from the cached snapshot.
3. Asserts reads serve cached state (`getVtxos()` returns the same outpoints) and the connection state reports degraded: `getProviderConnectionState().mode === "degraded"` and `serverInfoSource === "cache"`.

Verified passing against this branch (`193f8e6a`) on the local regtest stack (~9s). Offered as a stacked test-only add for #609 — feel free to fold it in or close if you'd rather cover this differently.

Context: came out of a parallel offline-first effort; #609 supersedes that wallet-core work, and the exit-side follow-up is #610.

### DIFF
--- a/packages/ts-sdk/src/contracts/contractManager.ts
+++ b/packages/ts-sdk/src/contracts/contractManager.ts
@@ -1,5 +1,6 @@
 import { hex } from "@scure/base";
 import { IndexerProvider } from "../providers/indexer";
+import { isRetryableProviderError } from "../providers/availability";
 import { WalletRepository } from "../repositories/walletRepository";
 import {
     Contract,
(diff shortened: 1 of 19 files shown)


pietro909:
merged manually in https://github.com/arkade-os/ts-sdk/pull/609/commits/010d6fb241f1ad4885e25f23f704a3e81fb127d3

PR #623, opened 2026-07-17
feat(ArkadeCash): recover server-swept ArkCash via import

pietro909:
PR 4 of the ArkCash hybrid-L3 stack (`plans/pr-337-new-plan.md`), on top of #622.

`claimCash`'s swept branch flips from report to recover: each server-swept VTXO is imported as a signable `recoveryOnly` contract (arkcash key filed in the keyring) and settled back by the wallet's own recovery machinery in its own isolated, self-serializing intent — kept out of the wallet's balance / renewal / recovery / coin-selection on both the core and service-worker paths. The key is purged once recovery completes. Adds a `recovering` bucket to the claim result.

Stack: #621 → #622 → this.
Follow-up: Phase 4 — regtest e2e + doc closure.


pietro909:
Stacked on top of this: #624 (Phase 4 — e2e for server-swept recovery + v0.9.14 P2TR-only query fix).

packages/ts-sdk/src/wallet/wallet.ts
```diff
@@ -4378,8 +4455,43 @@ export class Wallet extends ReadonlyWallet implements IWallet {
...
+                // Kick recovery promptly rather than waiting for the poll loop.
+                // Non-blocking: recovery is a full settlement (a batch round)
+                // retried on the wallet's normal cadence, so claimCash must not
+                // await it. A kick failure is harmless — the poll loop retries.
+                void this.kickImportedRecovery();
+            } catch (error) {
+                // The import is local (keyring + contract writes) and rarely
+                // fails; if it does, report the funds rather than dropping them
+                // silently. They stay recoverable in principle by re-running.
+                console.error("Failed to import swept arkadeCash for recovery:", error);
```
pietro909, line 4501: Shall we `deleteKey` here? The key is useless without the contract and a re-run re-imports it anyway.
It could be good house-keeping.

packages/ts-sdk/src/wallet/wallet.ts
```diff
@@ -2916,6 +2975,17 @@ export class Wallet extends ReadonlyWallet implements IWallet {
         // baseline contracts.
         const boot = await WalletReceiveRotator.resolveBoot(config, setup);
 
+        // Wrap the resolved provider (or a static one over the wallet
+        // identity, for the common non-HD wallet) with a keyring so
+        // `claimCash` can import a foreign signable contract for recovery
+        // and the signer router can resolve it. This does NOT feed HD
+        // boarding/receive rotation — that still keys off `boot?.provider`
+        // below — so a static wallet keeps its single fixed address.
+        const keyring = await KeyringDescriptorProvider.create(
```
pietro909: This throws loudly by design, which is ok. 
But a failure here bricks the whole wallet: main funds not accessible to protect an auxiliary recovery key.

Perhaps better to catch and log it as an error.

PR #624, opened 2026-07-17
test(ArkadeCash): e2e for server-swept ArkCash recovery

pietro909:
Phase 4 of the ArkCash swept-VTXO follow-up (hybrid L3). Stacked on #623.

Stack: #621 → #622 → #623 → this.

## Changes
- **fix:** `claimCash` queries only the P2TR script. The v0.9.14 indexer rejects non-P2TR scripts (`invalid script, must be P2TR`), so the old two-script query (P2TR + OP_RETURN subdust) 400'd, breaking every claim. Subdust arkcash lives at the OP_RETURN script and is no longer reported; `createCash`'s dust floor prevents minting it.
- **test:** e2e for the swept → import → isolated-recovery → purge arc and the crash/re-run idempotency variant. Adds a `waitForCashVtxo` helper that also fixes pre-existing races in the sweep-or-report tests (claim fired before the cash VTXO was indexed).

## Not included
- Rotation variant e2e (deferred): heavy arkd-restart test; recovery reuses the already-e2e-covered `rotateForRecoverableInputs` machinery.

## Verification
- lint + 1870 unit tests green
- 7 ArkCash e2e green and stable against the v0.9.14 regtest stack


pietro909:
Follow-up stacked on top: #627 — full subdust ArkCash claim via top-up recovery (Phase 5).

PR #613, opened 2026-07-10
boltz-swap: non interactive claim

louisinger:
https://github.com/ArkLabsHQ/fulmine/pull/452
https://github.com/BoltzExchange/boltz-backend/pull/1422

## Summary by CodeRabbit

* **New Features**
  * Added optional non-interactive Lightning reverse swaps with automated covclaimd-based claim handling.
  * Added covclaimd-driven non-interactive support for chain swaps to ARK, including automated claim reveal and claim-script enforcement.
  * Introduced `covclaimdUrl` and `nonInteractive` options, propagated through service-worker initialization.

* **Bug Fixes**
  * Improved non-interactive VHTLC script/address generation to support delegated claims.

* **Tests**
  * Added unit and end-to-end coverage for covclaimd claim delegation and ECIES preimage handling.

* **Chores**
  * Updated regtest reference and regtest image overrides; added required cryptography dependency.

PR #625, opened 2026-07-17
feat(boltz-swap): walletless non-interactive reverse swap for arbitrary receivers

Kukks:
Builds on #613. That PR added non-interactive reverse swaps, but through `ArkadeSwaps.createReverseSwap`, which reads the receiver from `this.wallet` (`claimAddress = await this.wallet.getAddress()`, and the claim key from the wallet identity). That's right for a wallet claiming its own swap while offline, but it means a *service* — say an LNURL server minting invoices for users whose keys it doesn't hold — can't use it, because it has no wallet for the receiver.

The primitives underneath are already receiver-agnostic (`nonInteractiveClaim.claimAddress`, `enforcePayTo(receiverTapKey)`, and `CovclaimdProvider` all take arbitrary values), so this adds a small walletless helper that drives them directly:

```ts
const { invoice, preimage } = await createNonInteractiveReverseSwap({
  swapProvider, covclaimd, amount,
  claimPublicKey, // receiver's compressed pubkey
  claimAddress,   // receiver's Arkade address
});
```

It's `createReverseSwap` in `nonInteractive` mode minus the wallet: generate a preimage, request the swap with the caller-supplied claim identity, verify the invoice commits to the preimage hash, then hand covclaimd the ECIES-encrypted preimage plus the `enforcePayTo` covenant so it can only pay the receiver. Same anti-theft guarantee as #613 — covclaimd learns the preimage but can't redirect the funds. I also exported `CovclaimdProvider`, which a caller needs but the package index didn't re-export.

It's additive — no change to `ArkadeSwaps` or existing behaviour. I made it a standalone function since a server has no wallet to hang a method on, but if you'd rather it live on `ArkadeSwaps` (e.g. `createReverseSwapFor`) or be named differently, happy to move it. I couldn't run the regtest/e2e stack locally, so the on-chain path is unverified from my side.

For tests I added a unit test next to #613's own non-interactive one: it asserts Boltz is asked to pay the receiver's `claimAddress`/`claimPublicKey`, and that covclaimd's reveal packet decrypts back to the swap's preimage under the pay-to-receiver covenant. `typecheck` and the full `boltz-swap` unit suite (444 tests) pass.

Context: this unblocks server-side offline Lightning receive in the lnurl-server — a service-side companion to the wallet-side flow #613 enables.

### DIFF
--- a/packages/boltz-swap/src/index.ts
+++ b/packages/boltz-swap/src/index.ts
@@ -8,6 +8,12 @@ export const sdkVersion = `boltz-swap/${version}`;
 
 export { ArkadeSwaps } from "./arkade-swaps";
 export type { QuoteSwapOptions } from "./arkade-swaps";
+export { createNonInteractiveReverseSwap } from "./non-interactive-reverse-swap";
+export type {
+    CreateNonInteractiveReverseSwapArgs,
+    NonInteractiveReverseSwap,
+} from "./non-interactive-reverse-swap";
+export { CovclaimdProvider } from "./covclaimd-provider";
 export type { BoltzSwapStatus, SubmarineProgressionStatus } from "./boltz-swap-provider";
 export {
     BoltzSwapProvider,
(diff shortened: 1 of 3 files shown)


pietro909:
All changed files are under packages/boltz-swap which is discontinued. Closing @Kukks

PR #630, opened 2026-07-20
Fix restore under indexer rate limiting

pietro909:
Restoring a large HD wallet floods the indexer with small requests and trips the operator's rate limiter. Two problems follow from that, fixed here.

**A rate-limited probe was read as "index unused."** A `discoverAt` rejection incremented the gap counter, so the scan's window closed on failed requests rather than on absent funds — restore could return a wallet missing money while reporting success. A failed probe is now *indeterminate*: it does not advance the gap counter, and the scan stops there and reports where.

`ScanResult` gains `highestConfirmedUsedIndex` and `truncatedAt`; `lastIndexUsed` stays as a deprecated alias, so consumers are unaffected. Confirmed hits found above the truncation point are still persisted and still advance the HD watermark — otherwise a caller that swallows the restore error could be handed an already-funded index as a fresh receive address.

**Nothing throttled or backed off.** There was no concurrency limit anywhere and no retry on any indexer GET; `Retry-After` was never read. New origin-keyed gate holding a shared cooldown and an in-flight cap:

- Idempotent GETs (indexer, `getInfo`, delegate info) wait behind the cooldown and report into it.
- Intent/settlement POSTs report only — never delayed, never retried. Batch sessions have deadlines and these POSTs are not idempotent.
- Waiters resume with independent jitter, so a herd released by one expiring cooldown does not re-form into the burst that earned the 429.
- `indexerFetch` retries availability failures a bounded number of times. `Retry-After` is read off the response and fed to the gate, never attached to the typed error — per-instance `Error` properties do not survive the service-worker `postMessage` boundary.

### Notes for review

- Restore will now fail *more often but honestly* — it reports an unverified range instead of a plausible wrong answer. The retry absorbs transient 429s before they reach the scanner, which is what keeps that tolerable.
- `highestConfirmedUsedIndex` is a required field, so this is source-breaking for anyone *implementing* `IContractManager` (nothing in the repo does). It cannot be optional without being untrustworthy.
- The boarding-before-default first-wins tie-break is untouched; its test still pins it.

Follow-up, not in this PR: collapsing the per-index request fan-out using the indexer's existing batch support, which is what actually cuts request volume.

### Testing

New `test/rate-gate.test.ts` covers cooldown sharing, origin scoping, jittered resume, and the report-only path. `test/restore.test.ts` gains the indeterminate-probe regression and the address-reuse guard. Full unit suite green.

Not done: re-capturing a HAR against a live operator to diff request counts.

## Summary by CodeRabbit

* **New Features**
  * Enhanced contract scanning to report confirmed vs indeterminate results, including the highest confirmed index and an optional truncation point when verification can’t complete.
  * Added shared request throttling with 429-based cooldown tracking across provider requests.
  * Added attempt-limited retries for temporary indexer availability failures.
* **Bug Fixes**
  * Wallet restoration now advances from the highest confirmed index and prevents reusing a previously funded receive address even when scanning is truncated.
  * Restore scan errors now explicitly indicate the unverified range and that retrying is safe.

### DIFF
--- a/packages/ts-sdk/src/contracts/contractManager.ts
+++ b/packages/ts-sdk/src/contracts/contractManager.ts
@@ -308,15 +308,17 @@ export interface IContractManager extends Disposable {
      *
      * Error contract (safety-critical — see spec §4):
      * - A handler's `discoverAt` rejecting is **collected** into
-     *   `handlerErrors` and the loop **continues**; it never aborts the
-     *   scan or throws.
+     *   `handlerErrors` and makes its index *indeterminate*: it never
+     *   advances the gap counter, and the scan **stops verifying** there
+     *   rather than closing a window it never observed close. It still
+     *   never throws.
      * - A fatal operational error — `materialize()` throwing, or
      *   `createContract` rejecting — **propagates** out of `scanContracts`
      *   (it invalidates the gap-window signal, so a silent truncation
      *   would risk hiding user funds).
      *
      * @param opts See {@link ScanContractsOptions}.
-     * @returns `{ lastIndexUsed, handlerErrors }` — the caller surfaces
+     * @returns See {@link ScanResult}. The caller surfaces `truncatedAt` /
      *   `handlerErrors` *after* the inline VTXO pull.
      */
     scanContracts(opts: ScanContractsOptions): Promise<ScanResult>;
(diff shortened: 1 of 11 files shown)


pietro909:
@coderabbitai please review

PR #648, opened 2026-07-27
fix: service-worker correctness batch

pietro909:
Three independent fixes in `packages/boltz-swap/src/serviceWorker/`, one commit each.

- **Init race** — `MESSAGE_BUS_INITIALIZING` is a superset of `MESSAGE_BUS_NOT_INITIALIZED`, so a concurrent bus init was read as a dead bus and burned both retries reinitializing with no backoff. The runtime now backs off and retries without consuming an attempt, mirroring `ServiceWorkerWallet`. Both classifiers move to `worker/errors` and are exported, replacing duplicated string matching in two packages.
- **Event broadcast** — the handler's fan-out called `clients.matchAll()` bare, so pages loaded before the SW claimed them missed every swap event. Now uses the same options as the core bus.
- **`SM-START` arity** — the proxy's `start()` sends no payload while the interface demanded `start(pendingSwaps)`. No swaps are lost (the SW loads them from the repository), so the parameter is now optional on both `SwapManagerClient` and `SwapManager`.

Five tests added. Full unit suite green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service worker communication with pages that are not yet controlled.
  * Added bounded retry and backoff handling during message-bus initialization, preventing unnecessary reinitialization and indefinite waits.
  * Improved swap manager startup by automatically loading stored swaps when no list is provided.

* **New Features**
  * Added shared SDK helpers for identifying message-bus initialization errors.

* **Tests**
  * Added coverage for pending swap loading, service worker event delivery, startup handling, and initialization retry scenarios.

### DIFF
--- a/packages/boltz-swap/src/serviceWorker/arkade-swaps-message-handler.ts
+++ b/packages/boltz-swap/src/serviceWorker/arkade-swaps-message-handler.ts
@@ -774,7 +774,12 @@ export class ArkadeSwapsMessageHandler
     private async broadcastEvent(event: SwapManagerEventMessage): Promise<void> {
         const sw: any = self as any;
         if (!sw?.clients?.matchAll) return;
-        const clients = await sw.clients.matchAll();
+        // Same options as the core bus fan-out: pages loaded before the SW claimed
+        // them are uncontrolled and would otherwise miss every swap event.
+        const clients = await sw.clients.matchAll({
+            includeUncontrolled: true,
+            type: "window",
+        });
         for (const client of clients) {
             try {
                 (client as any).postMessage(event);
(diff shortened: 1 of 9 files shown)


pietro909:
Note for later: the init-wait backoff policy (`sleep`, `INIT_WAIT_BACKOFF_CAP_MS`, `MAX_INIT_WAITS`, `initWaitBackoffMs`) is now duplicated verbatim between `packages/ts-sdk/src/wallet/serviceWorker/wallet.ts` and `packages/boltz-swap/src/serviceWorker/arkade-swaps-runtime.ts`. The error classifiers were promoted to `worker/errors`; the backoff policy should follow so the wallet and the plugin can't silently diverge if it's ever tuned.

Deliberately not addressed here — it will be consolidated as part of the broader "Swap entity in core SDK" work.

PR #654, opened 2026-07-28
feat: port arkade-os/packages into the monorepo

Kukks:
Ports the five publishable packages from [arkade-os/packages](https://github.com/arkade-os/packages) into this monorepo, wires them into CI and the release orchestrator, and adds a unified example that exercises them together.

The source repo's four demo apps do not come across. It should be archived once a release from here proves the flow.

## What moved

| Package | npm status before | Notes |
| --- | --- | --- |
| `@arkade-os/wallet-providers` | unpublished | UniSat, OKX, Leather, Phantom `Identity` implementations |
| `@arkade-os/sats-connect` | unpublished | `SatsConnectIdentity` + `ArkadeWallet` |
| `@arkade-os/sats-connect-react` | unpublished | React bindings — first JSX in this repo |
| `@arkade-os/checkout` | unpublished | Next.js checkout widget + route handlers |
| `@arkade-os/snap` | **live at 0.1.2** | MetaMask Snap — the only one with real users |

Four of the five had never been published (verified against the npm registry), so there were no consumers and no old contract to break. That is what made this cheap.

## Dependency and release policy

**Only `boltz-swap` fans out from an SDK release.** It pins the SDK via `workspace:*`, which pnpm rewrites to an *exact* version on publish, so an SDK bump invalidates it. The five new packages use peer ranges or `workspace:^`, both of which publish as caret ranges and survive an SDK patch or minor untouched.

`release.mjs` grows from 2 targets to 7, and `all` now means every package rather than just two. The fan-out logic is deliberately **not** generalised — verified by dry run that `sdk patch` still selects only sdk + boltz-swap, while each web package selects only itself.

## Snap: 0.1.2 can never be republished

The ported source builds a different bundle than what is on npm:

```
published 0.1.2   d2S8THO6VgRbOlJSOJwNw4R8CYPOHd500J7GVDiHoSs=
this tree         aXZE8MGGsgfKCuCrnJHWOoJuoa8AUtQf5V3h3dhpyZc=
```

Expected — reformatting and dependency changes altered it. Republishing 0.1.2 with a different bundle would break the shasum for every already-installed MetaMask user, so **snap releases must always move forward**.

`scripts/snap-release-hook.mjs` handles this: it mirrors the version into `snap.manifest.json`, lets `mm-snap build` regenerate the shasum, then validates and **aborts before any commit, tag or publish** on mismatch. The shasum is a composite hash over all snap source files (sort by path, sha256 each, concatenate, sha256 again) covering bundle *and* icon — never a plain hash of `dist/bundle.js` — so the hook delegates entirely to `mm-snap` and never recomputes it.

The bundle is deterministic for a fixed dependency tree, but not stable across dependency changes. The committed shasum is therefore only a cache; the hook is the source of truth.

## Defects fixed in `checkout`

1. `next` and `react` were peer dependencies but not dev dependencies, so the package could not typecheck standalone. Nothing else in the workspace depends on `next`, making this a hard blocker for `pnpm -r typecheck`.
2. `cli/create.js` was a committed build artifact duplicating `cli/create.ts`. Only the `.ts` is ported, and `files` no longer publishes the raw `cli/` directory.
3. `storage.ts` lazily `require`s `@vercel/kv` behind a `KV_REST_API_URL` guard but declared it nowhere — any consumer setting that variable hit `MODULE_NOT_FOUND`. Now an optional peer dependency.
4. `tsup.config.ts` used `as const`, making `format` a readonly tuple that tsup's mutable `Format[]` rejects. The source repo never saw it because its tsconfig only included `src/`.

**Two pre-existing defects are deliberately carried unchanged**, since fixing either changes the public API:

- `src/index.ts` re-exports `POST`/`GET` from `./server/route`, so importing the package root pulls `next/server` into client bundles. Confirmed at runtime: the built `dist/index.js` cannot be imported in a bare Node process (`ERR_MODULE_NOT_FOUND: next/navigation`).
- `storage.ts` uses `require()` in a `"type": "module"` package.

## CI

The `check` job's hand-enumerated per-package steps collapse into `pnpm -r`, which does not scale past two packages. Two things worth flagging:

- **Build now precedes type-check.** Packages resolve the SDK's types from `packages/ts-sdk/dist`, so `pnpm -r typecheck` on an unbuilt tree fails with `TS2307`. Verified by moving `dist` aside.
- **`scripts/check-package-scripts.mjs` is new and load-bearing.** `pnpm -r <script>` silently *skips* packages that lack the script rather than failing, so without this guard a package that forgets `test:unit` looks green while running nothing. It also refuses to pass vacuously if it finds no packages.

The `integration` job is unchanged — the new packages have no regtest e2e.

## Unified example

`examples/integration-playground/` is a private Next.js app importing all the packages, under a new `examples/*` workspace root. Next.js is required rather than preferred: `checkout` exports handlers importing `next/server`, which cannot run under Vite.

**The point is that it builds.** Bundling the packages together exercises their real export maps, ESM/CJS dual-package resolution, and cross-package `.d.ts` correctness — failure modes per-package unit tests cannot reach. CI typechecks and builds it; the publish-shape step is scoped to `packages/*` so the private example is never packed.

Each panel's connect sequence is lifted from the corresponding provider's own `@example` block, so it doubles as documentation.

## Also included

- **`fix(ts-sdk)`: lint guards were broken on Windows.** Both `check-virtual-status.mjs` and `check-provider-boundary.mjs` compared a `node:path` `relative()` result against a forward-slash allowlist. On Windows that returns backslashes, so neither matched. The first flagged all 7 allowlisted files and failed `pnpm lint`; the second silently skipped every file and reported success having inspected **zero** — measured 0 files before the fix, 194 after. Green on Linux CI, which is why it went unnoticed. This is out of the port's scope but blocked committing anything on Windows.
- **React pinned to `19.2.8`** in workspace overrides. React requires `react` and `react-dom` to match exactly; with `node-linker=hoisted` plus `auto-install-peers`, Next's peer auto-install and the example's own dependency resolved to different patches and `next build` failed. All eight workspace members typecheck clean with the pin.

## BLOCKED: the snap will not build against current `master`

Merging `master` in (the branch was 528 commits behind) surfaces a failure that has nothing to do with the ported code, and `pnpm -r build` fails on it — reproducibly, on CI as well as locally:

```
✖ Error: Failed to evaluate snap bundle in SES.
  SES_UNCAUGHT_EXCEPTION: [TypeError: Cannot read properties of undefined (reading 'browser')
```

Traced to the line: `@noble/hashes/esm/utils.js` is **inlined** into `packages/ts-sdk/dist` by tsup and carries `import * as nc from 'crypto'` as its `randomBytes` Node fallback. `snap.config.ts` sets `polyfills: { crypto: true, stream: true }`, so webpack satisfies that with `crypto-browserify`, which drags in `readable-stream`, which reads a **bare** `process.browser` — undefined under SES lockdown.

Dropping those two polyfills makes the bundle evaluate. **Do not do that:** the inlined `randomBytes` has no `globalThis.crypto` branch, so with the polyfill gone it throws `crypto.getRandomValues must be defined` — a snap that builds, evaluates, passes CI, and fails the first time a user signs. Fixing it properly means keeping `@noble/hashes` external in the SDK's tsup config, or aliasing `crypto` to a stub exposing `webcrypto: globalThis.crypto`, and then testing a real signature in MetaMask.

This is SDK-side work and belongs in its own PR. Until it lands, this one cannot go green.

## Test plan

Baseline captured on `master` (`45a5369`): **4365 passing, 1 skipped, 0 failing** across 232 files.

After the port and merge: **4471 passing, 1 skipped, 0 failing** across 243 files — +106, the five new packages' suites (`sats-connect` 7, `snap` 70, `wallet-providers` 6, `sats-connect-react` 5, `checkout` 10) plus the Windows lint-guard test. No existing test changed result.

- [x] `pnpm run lint` — green
- [ ] `pnpm run build` — **fails on the snap's SES evaluation, see above**
- [x] `pnpm -r typecheck` — green, all 8 workspace members
- [x] `pnpm run test:unit` — 4471 passed, 0 failing
- [x] `pnpm -C examples/integration-playground build` — succeeds, checkout route handler builds as a dynamic route
- [x] `release.mjs` dry runs — fan-out scope verified for every target
- [x] Snap release hook — verified it accepts a consistent build and rejects version drift, a corrupted shasum, and a planned-version mismatch
- [x] Snap manifest — `mm-snap manifest` exits 0 when valid, 1 on a corrupted shasum
- [ ] **Manual, cannot be automated:** the wallet flows. Run `pnpm -C examples/integration-playground dev` and click through with UniSat, OKX, Leather, Phantom, Xverse, and MetaMask Flask installed. CI proves the packages compile and bundle; it cannot prove they work against real extensions.

## Known limitation

`transpilePackages` in the example's `next.config.mjs` is verified against packages linked from source. A consumer installing from npm exercises a different resolution path that has not been tested.

- File an issue for `checkout`'s two deferred defects above.
- Deduplicate `smoke-dist.mjs` across `ts-sdk` (257 lines), `boltz-swap` (118) and the new shared `config/` copy.
- Archive `arkade-os/packages` once a release from here proves the flow.

## Summary by CodeRabbit

* **New Features**
  * Added Lightning Checkout with hosted payment UI, invoice QR codes, payment tracking, claims, webhooks, and VSS key support.
  * Added Sats Connect wallet integration, React wallet tools, transaction signing, balances, Lightning payments, and network switching.
  * Added UniSat, OKX, Leather, Phantom, and MetaMask Snap wallet support.
  * Added an integration playground for checkout and wallet experiences.

* **Documentation**
  * Expanded checkout, Snap, release, and integration setup guides.

* **Tests**
  * Added export, build, validation, security, and cross-platform checks.


Kukks:
Merged `master` in (`45a5369` — the branch was 528 commits behind) and pushed `f30be0c2`. Six conflicts resolved, one blocker fixed, **three disproved**, and three breakages the merge surfaced that had nothing to do with the port's own code.

## The four carried blockers: one was real

I checked each at `73e8bcde` — the SHA the last review is stamped with — through the contents API rather than the PR diff.

### #1 `signMultiple` crash on a short response — already fixed, disproved

> Not touched in this increment. The response-length guard is still missing.

`GET /repos/arkade-os/ts-sdk/contents/packages/sats-connect/src/identity/SatsConnectIdentity.ts?ref=73e8bcde`:

```
228:        if (!Array.isArray(responses) || responses.length !== requests.length) {
231:                    `signed PSBT(s) for ${requests.length} request(s). Refusing to continue.`,
236:                throw new Error(`Wallet returned no signed PSBT at index ${i}.`);
```

Three guards, not one: the array check, the length check, and a per-entry `typeof entry?.psbtBase64 !== "string"` check before any decode. It landed in `cb0a09f5`/`b52892e8` — two commits *before* the reviewed head.

### #2 checkout README parity — already fixed, disproved

> confirm `packages/checkout/README.md` (not just the playground README) carries equivalent language, or add it.

It does, at `packages/checkout/README.md:13-26`, under the heading **"The route handlers have no authentication"**: *"`create`, `claim` and `webhook` ship with no authentication, no rate limiting, and… **Mount these behind your own authentication, rate limiting, and origin allowlist before exposing them publicly.** Treat the handlers as internal plumbing, not as a public API."*

### #4 unbounded peer range — already fixed, disproved

> Adding `<1.0.0` … remains needed.

All three, read at `73e8bcde` through the contents API:

```
packages/wallet-providers/package.json:    ">=0.4.15 <1.0.0"
packages/sats-connect/package.json:        ">=0.4.15 <1.0.0"
packages/sats-connect-react/package.json:  ">=0.4.15 <1.0.0"
```

### #3 `updateCheckout` read-modify-write race — real, fixed

The only one that still held. `getCheckout` → spread → `setCheckout` with nothing serializing it, while `status`, `claim` and `webhook` all update one record concurrently. The later write erases the earlier field, which is how `txid` disappears behind a `status: paid`.

Now serialized per checkout id. Two things stated plainly rather than papered over:

- The chain is **per-process**. That is total for the in-memory store, which is process-local anyway. It is **not** total for `@vercel/kv`, which offers no compare-and-set here, so two instances can still lose a write. Documented on the function.
- The queued promise swallows; the returned one does not. A caller still sees its own failure, but one failed update cannot strand every later update for that id.

**Mutations** (`packages/checkout/test/storage.test.ts`, 3 new tests):

| # | mutation | result |
|---|---|---|
| 1 | serialization removed (the original code) | **killed** — `expected { id: 'checkout-race', status: 'paid' } to deeply equal { id, status, txid }` |
| 2 | lock keyed globally instead of per id | **killed** — same, plus `expected undefined to be 'f0'` |
| 3 | tail no longer swallows rejections | **killed** — the stranding test |
| 4 | tail never recorded, so nothing chains | **killed** — both concurrency tests |
| 5 | lock map entry never deleted | **survived** — memory hygiene, no observable behaviour. Not worth contorting a test for. |

Mutation 3 initially **survived**, and it was my test's fault, not the code's: it drove the failure with a *missing* checkout id, and a missing checkout returns early rather than throwing, so the chain never rejected and the case was never exercised. Failing the read instead makes it load-bearing. That round also showed the `.then(onFulfilled, onRejected)` second argument was dead code — the tail already absorbs — so it is gone rather than tested around.

### #5, N2 — the ordering assertion

Fixed as arkana suggested. `expect(seen).toContain("snap_getEntropy")` then the unconditional ordering check, instead of `if (seen.includes(...))` — which stopped checking in exactly the case it exists to catch.

## Three things the merge broke, none of them the port's fault

All three were invisible while the branch sat on a July base. All are consequences of master moving.

1. **The snap declared no `prettier`.** `mm-snap build` needs it (`Cannot find module 'prettier/plugins/babel'`); it was being hoisted from a root dependency master has since dropped. Now declared at `3.6.2`.
2. **The five ported packages and the playground lint with `prettier --check`, and master deleted the root `.prettierrc` they relied on.** Prettier fell back to its defaults — 80 columns, two-space indent — and rejected every file in all six. They now lint with **biome**, like the rest of the monorepo, which accepts all 44 files with zero changes. Prettier is now only a snap build dependency.
3. **`config/smoke-dist.mjs` was not biome-clean**, so the Lint job would have failed on it.

Worth noting for anyone reading the history: **CI last ran on `b52892e8` (29 July). It never ran on `73e8bcde`**, so the head carrying the sign-confirmation dialog has never been tested. The first two items above would have failed that run.

## Conflicts, and the one resolution worth checking

Six files: `.github/workflows/ci.yml`, `CONTRIBUTING.md`, both `packages/ts-sdk/scripts/check-*.mjs`, `pnpm-lock.yaml`, `scripts/release.mjs`.

The two lint-guard scripts were **comment-only** conflicts: master landed the same path-separator fix independently, so `eed03d1f` is now redundant with upstream. Took master's wording.

`scripts/release.mjs` is the one to look at. Master grew `dependsOnSdk` / `bumpFlag` / `excludeFromAll` / `excludedFromRelease`, and its `primarySelection("sdk")` returns `ALL_KEYS`. That was *equivalent* to "sdk plus its `workspace:*` dependents" while boltz-swap and swap were the only other packages — and it stops being equivalent the moment the five ported packages join, at which point every SDK release would drag all five along. Since the whole point of the port's release design is that the web packages depend on the SDK through peer ranges and `workspace:^` and are deliberately **not** fan-out dependents, I resolved it as:

```js
if (target === "sdk") return ["sdk", ...DEPENDENT_PACKAGES.map((p) => p.key)];
```

which reproduces master's behaviour exactly on master's package set and preserves the port's on the new ones. A plain "take theirs" here would have published five packages on every SDK patch.

`pnpm-lock.yaml` was regenerated from master's rather than hand-merged.

## Gate

`pnpm run lint` exit 0. `pnpm run test:unit` exit 0:

| | baseline `45a5369` | this head `f30be0c2` |
|---|---|---|
| test files | 232 | 243 |
| tests | 4365 passed, 1 skipped | 4471 passed, 1 skipped |
| failing | 0 | 0 |

The eleven new files and 106 new tests are the ported packages' own suites (`sats-connect` 2/7, `snap` 4/70, `wallet-providers` 1/6, `sats-connect-react` 1/5, `checkout` 2/10) plus `lint-guards-path-separators.test.ts` in ts-sdk.

## Still blocking: `pnpm -r build` fails on the snap, in SES

**This is the one thing I could not resolve, and it is new.** `mm-snap build` now fails at the SES evaluation step:

```
✖ Error: Failed to evaluate snap bundle in SES.
  SES_UNCAUGHT_EXCEPTION: [TypeError: Cannot read properties of undefined (reading 'browser')
```

Traced as far as I can take it:

- The SDK's built chunk carries `import * as nc from 'crypto'`, used only for `nc.webcrypto` — the Node fallback of a `@noble`-style `randomBytes`, inlined into `packages/ts-sdk/dist/chunk-AOJLUYU4.js` by tsup.
- `mm-snap` therefore polyfills `crypto` with `crypto-browserify`, which pulls in `readable-stream`.
- `readable-stream` reads a **bare** `process.browser` (not `globalThis.process.browser`, which the same bundle uses safely elsewhere). Under SES lockdown `process` is undefined, and that is the TypeError.

What I ruled out:

- **Not the ported code.** `packages/ts-sdk/package.json` dependencies are byte-identical between the PR base `acde309c` and `master` — nothing new was added.
- **Not the new `snaps-sdk` UI import.** `73e8bcde` added `import { divider, heading, panel, text } from "@metamask/snaps-sdk"` to `wallet.ts`; stubbing that import out and rebuilding fails identically.
- **Not the two known SES fixes**, #581 and #602 — both merged before this PR's base, so both are already in.

So it is SDK-side and needs its own change: make that `crypto` fallback resolve `globalThis.crypto` (or mark it external / alias it in the snap's bundler config) so `crypto-browserify` is never pulled in. That is a separate PR against `packages/ts-sdk`, not something to bury in a port.

I did **not** commit the regenerated `snap.manifest.json` shasum. `mm-snap build` rewrote it during diagnosis, but the build never passed SES validation, and CONTRIBUTING is explicit that a shasum from an unvalidated build breaks installation for every existing user. It is reverted to `ymjLDokY16ut…`.

## On the PROTOCOL-CRITICAL flag

Unchanged, and I am not the one who can clear it. This still needs a protocol engineer walking the sign / `mergeAndValidate` paths across `SatsConnectIdentity`, `ArkadeWallet` and the snap's `wallet.ts`. Nothing here reduces that.

The claim I would most expect to be wrong is the SES diagnosis — the causal chain is read off the bundle and the stack, and I have not confirmed it by making the `crypto` import go away and watching the build pass.

Kukks:
Follow-up on the SES build failure — I traced it to the exact line, found a one-line "fix" that makes CI green, and am **not** proposing it, because it breaks signing.

## Root cause, confirmed

`packages/ts-sdk/dist/chunk-AOJLUYU4.js`, with the bundler's own provenance comment two lines below it:

```js
import * as nc from 'crypto';
...
var crypto2 = nc && typeof nc === "object" && "webcrypto" in nc ? nc.webcrypto
            : nc && typeof nc === "object" && "randomBytes" in nc ? nc
            : void 0;

// ../../node_modules/@noble/hashes/esm/utils.js
```

So it is `@noble/hashes/esm/utils.js`, **inlined** into the SDK bundle by tsup rather than left external, and `import * as nc from 'crypto'` is its Node fallback for `randomBytes`.

From there:

1. `packages/snap/snap.config.ts` sets `polyfills: { buffer: true, crypto: true, stream: true }`.
2. webpack therefore satisfies that import with `crypto-browserify`, which drags in `readable-stream`.
3. `readable-stream` reads a **bare** `process.browser` — not `globalThis.process.browser`, which the same bundle uses safely elsewhere.
4. Under SES lockdown `process` is undefined. `TypeError: Cannot read properties of undefined (reading 'browser')`.

Reproduced on CI (Linux) as well as locally, so it is not a Windows artifact.

## The tempting fix, and why it is wrong

Dropping `crypto` and `stream` from the snap's `polyfills` **works**:

```
- Evaluating the snap bundle.
ℹ Snap bundle evaluated successfully.
```

Green build, one line changed. It is also a trap. Here is `randomBytes` as it appears in that same inlined chunk:

```js
function randomBytes(bytesLength = 32) {
  if (crypto2 && typeof crypto2.getRandomValues === "function") { ... }
  if (crypto2 && typeof crypto2.randomBytes === "function") { ... }
  throw new Error("crypto.getRandomValues must be defined");
}
```

**There is no `globalThis.crypto` branch.** With the polyfill removed, `nc` resolves to an empty module, `crypto2` is `undefined`, and `randomBytes` throws — it does not fall back to the WebCrypto the snap environment provides. BIP340 schnorr signing draws its aux randomness through this path, so the change would produce a snap that builds, evaluates, passes CI, and then fails the first time a user signs.

That is the worst shape a change can have, so I have reverted it and pushed nothing.

## What a real fix looks like

Someone with the SDK's build config in hand should pick one of:

- **Keep `@noble/hashes` external in `packages/ts-sdk/tsup.config.ts`** so its `crypto` import is resolved by the consumer's bundler rather than baked into the SDK's chunk, and let the snap's webpack resolve it against the browser condition (where `@noble/hashes` uses `globalThis.crypto`).
- **Alias `crypto` in the snap's webpack config to a stub exposing `webcrypto: globalThis.crypto`**, which satisfies the first branch of `crypto2` without pulling in `crypto-browserify` or `readable-stream`.

Either wants a real MetaMask run afterwards — a build that evaluates proves nothing about `randomBytes` at signing time, which is precisely the lesson above.

## Correction to my earlier comment

Two things I stated with more confidence than I had earned:

- I attributed the shim to "a `@noble`-style `randomBytes` Node fallback". It is specifically `@noble/hashes/esm/utils.js`, and the bundler comment says so — I had not found that when I wrote it.
- Between the two comments I re-checked and briefly could **not** reproduce the `crypto` import in the SDK dist, which would have made the whole diagnosis wrong. It turned out I had `pnpm install`-ed a different branch's workspace into `node_modules`, so the SDK was building against the wrong dependency set. Reinstalling for this branch reproduces it. Recording the false negative because anyone re-checking this on a shared worktree will hit the same thing.

PR #656, opened 2026-07-29
fix(core): make the contract pipeline safe for non-wallet-owned contract types

pietro909:
Three related core issues that surface once a contract type the wallet doesn't own outright is registered with the `ContractManager`. The `vhtlc` crash has no trigger on `master` today, since nothing registers one yet. The balance-scope half does have a trigger: `arkade` contracts are registrable right now through public API — see §3.

### 1. Annotating a `vhtlc` contract crashes

`deriveContractTapscripts` casts every contract's script to `DefaultVtxo.Script | DelegateVtxo.Script` and calls `.forfeit()` on it. `VHTLC.Script` has no `.forfeit()`, so annotating a registered `vhtlc` contract's VTXO threw `script.forfeit is not a function` out of `ContractWatcher.emitVtxoEvent` / `ContractManager.annotateVtxos`.

Fixed by implementing the existing `TapscriptDeriving` capability on `VHTLCContractHandler` — the annotation pipeline already checks for it (that's how `arkade` program-compiled contracts are handled).

It returns `claim()` as both the forfeit and intent leaf. That choice is deliberate and documented in the handler: VHTLC has no single owner+server forfeit leaf, and since `Wallet.getVtxos()` / `ContractManager.getContractsWithVtxos()` don't filter by contract type, a live `vhtlc` VTXO is a candidate input to the generic settle/recover path — which never supplies a preimage witness. `claim()` makes any such accidental inclusion fail closed with an incomplete witness, rather than risking a leaf the wallet could fully sign.

### 2. Escrow counted as spendable balance

`getBalance()` summed `getVtxos()` across every contract type, so a `vhtlc` contract's VTXOs would be reported as spendable — funds the user can't actually spend unilaterally.

Adds an optional `type` scope to `GetVtxosFilter` (mirroring `ContractManager`'s existing `getContracts({type})` filter) and has `getBalance()` request only `default`/`delegate` — the same scope `getWalletScripts()` already uses for the wallet's own receiving addresses.

`getVtxos()` is unchanged for callers that want the full set: this is a balance-accounting decision, not a tracking one, and recovery/inspection paths still need to see escrow.

### 3. Behavior change: `arkade` contracts also leave wallet balance

The `["default", "delegate"]` scope excludes more than `vhtlc`. `arkade` is the other registered non-wallet type, and unlike `vhtlc` it is registrable today: `ArkadeContract.register()` is public API. So this is an **observable behavior change**, not a latent fix — a registered arkade contract's VTXOs stop counting toward `wallet.getBalance()`.

That exclusion is intended, for the same reason as escrow: an arkade contract is program-gated rather than unilaterally spendable by the wallet key, and `ArkadeContract.getBalance()` already exists for per-contract accounting. Registration still means everything else it meant — repository persistence, watcher events, visibility to `wallet.getVtxos()` and transaction history.

Both docstrings that promised the old behavior (`ArkadeContract.register()` and `ArkadeContractHandler`) said registration made a contract "counted in repository-backed balances". Corrected, and the decision is pinned by tests so it can't flip back unnoticed.

### 4. Service worker parity

`ServiceWorkerWallet` is a second implementation of `IWallet` and callers hold the two interchangeably, so the scope had to land in both or the same call would return different answers depending on which one you got.

`handleGetBalance` read via `getVtxosFromRepo()`, which queried `manager.getContracts()` unfiltered — so escrow still counted as spendable balance there, while the comment above that read claimed the two implementations agree. `handleGetVtxos` separately dropped the new `filter.type` (the wire type is already `GetVtxosFilter` and the client forwards it; only the handler discarded it).

`getVtxosFromRepo()` now takes a contract-type scope. One subtlety: it reads the wallet's primary-address bucket separately from the contract buckets as a legacy-row safety net, so that bucket needs its own gate — it's the wallet's own `default` contract and must not appear in, say, a `type: "vhtlc"` read. The other three call sites (transaction history, pending-tx recovery, cache refresh) stay unscoped, matching core, where `getTransactionHistory()` and `pendingRecoveryOutpoints()` also use an unscoped `getContractsWithVtxos()`.

`ExpoWallet` needed no change — its `getBalance`/`getVtxos` are one-line delegations to the core `Wallet`.

### Tests

- 4 regression tests pinning the annotation crash path, plus coverage that `default`/`delegate`/`boarding` stay forfeit-based.
- 3 tests for the vhtlc balance scope: escrow excluded from every balance bucket, still visible through unfiltered `getVtxos()`, and `type` filtering both ways. Verified these fail against `master` (escrow reads as 50,000 spendable sats).
- 3 tests pinning the arkade decision: excluded from balance, still tracked by `getVtxos()`, scopeable by type. The pair is self-validating — one asserts the VTXO is present and worth 25,000, so the `total: 0` assertion can't pass vacuously.
- 4 service worker parity tests, including the primary-address leak. Verified 3 of the 4 fail with the source fix stashed; the fourth is a non-regression guard that correctly passes either way. The shared `setupHandler` mock now honors the `type` filter the way every real `ContractRepository` does — a mock that ignored it would have hidden this bug.

Full `ts-sdk` suite green (2147), `boltz-swap` unaffected (487).

### Known gap, not addressed here — tracked in #657

Balance is now scoped, but coin selection is not: `sendBitcoin` uses `getVtxos({ withRecoverable: false })` and `settle` uses `getVtxos({ withRecoverable: true })`, both unscoped by type, as do `VtxoManager`'s auto-renewal and `recoverVtxos()`. So a `vhtlc` or `arkade` VTXO can still be selected as an input despite never appearing in balance.

For `vhtlc` the `claim()` leaf in §1 makes that fail closed — but round-wide, not per-input: one escrow input fails the entire settlement, which would block auto-renewal of the wallet's own funds. For `arkade` the derived leaf is the *collaborative* one, explicitly chosen to mirror `forfeit()`, so it may sign fully rather than fail closed. Scoping the selection paths is a larger semantic change with open sub-decisions, so it is tracked separately in #657.

---

Extracted from #612, which is where the first two bugs were found. That PR stays open and will rebase onto these.

### DIFF
--- a/packages/ts-sdk/src/arkade/contract.ts
+++ b/packages/ts-sdk/src/arkade/contract.ts
@@ -456,9 +456,16 @@ export class ArkadeContract<P extends Program = Program> {
     /**
      * Persist this contract through the wallet's {@link IContractManager} so it
      * is tracked like any other contract type: stored in the contract
-     * repository, watched for VTXO events, counted in repository-backed
-     * balances, and re-derivable offline via the `"arkade"` contract handler.
-     * Idempotent — re-registering the same script is a no-op.
+     * repository, watched for VTXO events, visible to `wallet.getVtxos()` and
+     * the transaction history, and re-derivable offline via the `"arkade"`
+     * contract handler. Idempotent — re-registering the same script is a no-op.
+     *
+     * Registering does NOT add this contract's VTXOs to `wallet.getBalance()`.
+     * That balance is scoped to the contract types the wallet owns outright
+     * (`default`/`delegate`) — the same scope `getWalletScripts()` uses — and an
+     * arkade contract is program-gated rather than unilaterally spendable by the
+     * wallet key. Use {@link ArkadeContract.getBalance} for this contract's own
+     * balance.
      */
     async register(options?: {
         label?: string;
--- a/packages/ts-sdk/src/contracts/handlers/arkade.ts
+++ b/packages/ts-sdk/src/contracts/handlers/arkade.ts
@@ -149,8 +149,13 @@ function pathsFor(
  * signer keys) as string params, so any contract created via
  * `arkade.contract(program, args)` participates in the standard contract
  * pipeline: ContractManager persistence and validation, watcher events,
- * repository-backed balances, and offline script re-derivation. Rebuild a
+ * repository-backed VTXO tracking, and offline script re-derivation. Rebuild a
  * callable contract from a stored row with `ArkadeContract.fromContract`.
+ *
+ * Tracked is not the same as counted: `wallet.getBalance()` is scoped to the
+ * wallet-owned types (`default`/`delegate`), so these VTXOs are visible to
+ * `wallet.getVtxos()` but excluded from wallet balance. See
+ * {@link ArkadeContract.register}.
  */
 export const ArkadeContractHandler: ContractHandler<ArkadeContractParams, ArkadeProgramScript> &
     TapscriptDeriving<ArkadeProgramScript> = {
--- a/packages/ts-sdk/src/contracts/handlers/vhtlc.ts
+++ b/packages/ts-sdk/src/contracts/handlers/vhtlc.ts
@@ -1,7 +1,14 @@
 import { hex } from "@scure/base";
 import { VHTLC } from "../../script/vhtlc";
 import { RelativeTimelock } from "../../script/tapscript";
-import { Contract, ContractHandler, PathContext, PathSelection } from "../types";
+import {
+    Contract,
+    ContractHandler,
+    DerivedContractTapscripts,
+    PathContext,
+    PathSelection,
+    TapscriptDeriving,
+} from "../types";
 import { isCltvSatisfied, isCsvSpendable, resolveRole } from "./helpers";
 import { sequenceToTimelock, timelockToSequence } from "../../utils/timelock";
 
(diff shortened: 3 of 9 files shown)


pietro909:
Added the positive multi-type assertion — `["vhtlc", "default"]` now pins that an array is membership across the listed types, rather than an intersection or only the first entry.

packages/ts-sdk/src/arkade/contract.ts
```diff
@@ -456,9 +456,16 @@ export class ArkadeContract<P extends Program = Program> {
...
      * Persist this contract through the wallet's {@link IContractManager} so it
      * is tracked like any other contract type: stored in the contract
-     * repository, watched for VTXO events, counted in repository-backed
-     * balances, and re-derivable offline via the `"arkade"` contract handler.
-     * Idempotent — re-registering the same script is a no-op.
+     * repository, watched for VTXO events, visible to `wallet.getVtxos()` and
+     * the transaction history, and re-derivable offline via the `"arkade"`
+     * contract handler. Idempotent — re-registering the same script is a no-op.
+     *
+     * Registering does NOT add this contract's VTXOs to `wallet.getBalance()`.
```
Kukks, line 463: I dont think I agree here. If a contract handler states it can be spent, it should be added to tghe balance. Same applies with the batches. the handler should dictate this, not hardwired specifics.

PR #660, opened 2026-07-30
fix(wallet): batch getTransactionHistory's createdAt lookups

Kukks:
`Wallet.getTransactionHistory()` asks the indexer for one outpoint at a time. Because `buildTransactionHistory` awaits `getTxCreatedAt` **inside** its per-VTXO loop, those requests run strictly sequentially — and repeat for every VTXO sharing an `arkTxId`.

## Measured

From a browser HAR against mutinynet, a wallet with roughly 120 transactions opening an activity view:

| | |
|---|---|
| `/v1/indexer/vtxos` requests | **379** — 92% of the page's total traffic |
| carrying exactly **one** outpoint | **348** |
| distinct txids behind those 348 | **123** → ~2.8x duplication |
| per request | median 181ms, p90 210ms, max 1,830ms |
| total | **23.9s**, one sequential burst |

Because nothing runs concurrently, coalescing in-flight requests cannot help here — only knowing the txids up front can.

## Why this is more than a latency issue

Cost grows linearly with wallet history. The application that reported it bounds the read with a 60s timeout, so once a wallet's history is roughly 2.5x this size the call exceeds it and the history view fails outright — and stays failed, worsening with every transaction. Heaviest users hit it first.

## The fix

`WalletMessageHandler.getTransactionHistory` (the service-worker path) has always batched this. That implementation is mirrored into `Wallet`:

- fill timestamps from the VTXOs already in hand — earliest per txid, so ordering stays stable — which needs no network at all;
- collect only the spent `arkTxId`s that remain unknown;
- fetch those in chunks of `BATCH_SIZE = 100`, the same size the service-worker path uses;
- `getTxCreatedAt` becomes a map lookup.

Two deliberate differences from the mirrored version:

- **Retryable-error tolerance is preserved.** The per-txid version tolerated an unavailable indexer by leaving `createdAt` undefined so history still built from repository VTXOs; the service-worker version has no such catch. Keeping it means the batch path degrades the same way it always did, and terminal failures still propagate.
- **`knownTxids` is a `Set`,** so the "is this `arkTxId` one of our own VTXOs?" check is O(1). A literal port would leave an O(n²) scan, which matters precisely on the large wallets this fix targets.

## Tests

Five, in `test/wallet/transactionHistoryBatching.test.ts`. Two are **discriminating** regression guards — verified by reverting the source change and re-running, where they fail with:

```
expected [ …(30) ] to have a length of 1 but got 30
expected [ Array(205) ] to deeply equal [ 100, 100, 5 ]
```

The other three pin behaviour that already held and would otherwise be easy to regress: no re-request for a txid shared by several VTXOs, no request at all when every timestamp is known locally, and history still building when the indexer batch fails retryably (using a real `FetchError`, since an ad-hoc `Error` with the same `name` is not what `isRetryableProviderError` accepts).

## Verification

Rebased onto current `master` (`df02985e`) and re-run there, not just on the base I started from:

- `tsc --noEmit` clean
- `pnpm -r lint` clean — prettier, `virtualStatus` guard, provider-boundary guard
- `packages/ts-sdk`: **149 files / 2198 tests** passed, 1 skipped
- `packages/boltz-swap`: **21 files / 525 tests** passed

## Not verified

The end-to-end reduction against a live indexer. The tests prove the request shape and count; confirming the real-world win wants a fresh HAR after this ships, where the same page should drop from ~379 requests to under 40.

## Summary by CodeRabbit

* **Performance**
  * Improved transaction history loading by reducing the number of network requests needed to retrieve transaction timestamps.
  * Added batching for large transaction histories to support more efficient loading.

* **Reliability**
  * Transaction history remains available when retryable timestamp lookup requests fail.

* **Tests**
  * Added coverage for batching, deduplication, request limits, locally available timestamps, and retryable failures.

### DIFF
--- a/packages/ts-sdk/src/wallet/wallet.ts
+++ b/packages/ts-sdk/src/wallet/wallet.ts
@@ -1191,17 +1191,74 @@ export class ReadonlyWallet implements IReadonlyWallet {
 
         const { boardingTxs, commitmentsToIgnore } = await this.getBoardingTxs();
 
-        const getTxCreatedAt = (txid: string) =>
-            getNormalizedVtxos(this.indexerProvider, { outpoints: [{ txid, vout: 0 }] })
-                .then((res) => res.vtxos[0]?.createdAt.getTime())
-                // Best-effort createdAt enrichment: when the indexer is
-                // unavailable, leave it undefined (history still builds from
-                // repository VTXOs) rather than failing the whole read. Terminal
-                // failures still propagate.
-                .catch((err) => {
-                    if (isRetryableProviderError(err)) return undefined;
-                    throw err;
-                });
+        // Timestamps we already hold, earliest per txid so history ordering is
+        // stable. Free: no indexer round trip.
+        const vtxoCreatedAt = new Map<string, number>();
+        for (const vtxo of allVtxos) {
+            const existing = vtxoCreatedAt.get(vtxo.txid);
+            const ts = vtxo.createdAt.getTime();
+            if (existing === undefined || ts < existing) {
+                vtxoCreatedAt.set(vtxo.txid, ts);
+            }
+        }
+
+        // Pre-fetch whatever is left in BATCHED indexer calls.
+        // buildTransactionHistory needs these for spent offchain virtual outputs
+        // with no change output (arkTxId set, but no virtual output has
+        // txid === arkTxId).
+        //
+        // This used to be one indexer request per txid, and because
+        // buildTransactionHistory awaits getTxCreatedAt inside its per-VTXO
+        // loop, they ran strictly sequentially and repeated for every VTXO
+        // sharing an arkTxId. MEASURED in a browser HAR against mutinynet: 348
+        // single-outpoint requests covering only 123 distinct txids — 92% of the
+        // page's total traffic — at a median 181ms each, 23.9s in one burst.
+        // Cost grew linearly with wallet history, so a large wallet could exceed
+        // a caller's timeout and fail the read outright.
+        //
+        // The service-worker path (WalletMessageHandler.getTransactionHistory)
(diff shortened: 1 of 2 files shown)


pietro909:
Same fix as #661 which changes `buildTransactionHistory` to take a batched `resolveTxCreatedAt(txids[])` called once, and moves the batching into a shared `fetchVtxoCreatedAtByTxid()` in `wallet/vtxo.ts` so **the copy** in `WalletMessageHandler` **is deleted rather than mirrored** a second time, and both paths get the same behaviour from one place.

Two smaller things it also picks up:

- the chunk size is a named `OUTPOINT_QUERY_CHUNK_SIZE = 64` next to `SCRIPT_QUERY_CHUNK_SIZE`, budgeted against the same URL ceiling;
- the query passes an explicit `pageSize`. The provider omits `page.size` when unset, so at `BATCH_SIZE = 100` a server default below 100 short-pages and silently drops timestamps into the `createdAt + 1` fallback — a latent issue in the service-worker code this mirrors.

Two things from here are now in #661:

- `knownTxids` as a `Set` — our collector still had the O(n²) `vtxos.some(...)` scan, which matters on exactly the wallets this targets;
- the wallet-level request-count test. We had the resolver covered by unit tests but nothing asserting `Wallet` actually reaches it, so there are now two tests driving a real wallet against a recording indexer: 69 spent VTXOs → `[64, 5]` outpoint requests instead of 69 singles, and 5 VTXOs sharing an `arkTxId` → one request. Verified discriminating by reverting to a per-txid resolver.

## Proposal: close this in favour of #661 
The measurement and the write-up in this PR are the better artifact of the two — happy to carry the HAR numbers into #661's description if that's useful. Either way, worth a fresh HAR after it ships.

PR #661, opened 2026-07-30
perf(history): batch tx-time indexer reads

pietro909:
`buildTransactionHistory` resolved each spent-with-no-change ark txid with its own sequential `/v1/indexer/vtxos` call — one round trip per txid. Uncovered by ArkLabsHQ/coinflip#112: 123 txids ≈ 22s of history load, growing linearly.

- collect the needed txids up front and resolve them once through a shared `fetchVtxoCreatedAtByTxid`, chunked at a new `OUTPOINT_QUERY_CHUNK_SIZE` (64, ~5 KB of query string)
- both the plain `Wallet` and the service-worker path now use the same resolver; the worker's hand-rolled batch loop is gone, and it inherits the wallet's best-effort semantics (retryable indexer failure → partial map + fallback timestamps, terminal errors propagate)
- `pageSize` is set explicitly so a low server-side page default can't silently short-page a chunk

123 requests → 2 for the coinflip case; their app-side `txTimeCache` workaround can be dropped once this ships.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction history timestamps for spent offchain transactions through batched lookups.
  * Added best-effort fallback behavior when timestamp information is unavailable.
  * Improved handling of large transaction ID batches, duplicate entries, and temporary service failures.
  * Ensured terminal service errors are surfaced appropriately while recoverable failures can return partial results.

* **Tests**
  * Added coverage for batched timestamp resolution, chunking, fallback behavior, partial results, and error handling.

### DIFF
--- a/packages/ts-sdk/src/contracts/constants.ts
+++ b/packages/ts-sdk/src/contracts/constants.ts
@@ -27,3 +27,10 @@ export const DEFAULT_PAGE_SIZE = 500;
  * onto different budgets for the same URL constraint.
  */
 export const SCRIPT_QUERY_CHUNK_SIZE = 32;
+
+/**
+ * Maximum outpoints per `getVtxos` query string. ~79 encoded bytes each, so 64
+ * is ~5 KB — real margin under the common 8 KB request-line ceiling, against
+ * the same unpublished `arkd` limit that budgets scripts at 32.
+ */
+export const OUTPOINT_QUERY_CHUNK_SIZE = 64;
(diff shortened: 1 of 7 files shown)


pietro909:
@arkana-ai-bot I've addressed point 1 and the first nit. 

Pushing back on:

-  `"No invariant enforcing OUTPOINT_QUERY_CHUNK_SIZE ≤ DEFAULT_PAGE_SIZE"` - the invariant is already enforced, by a test, see `vtxo-query-chunking.test.ts`
- `Test assertion for resolver return values depends on sort stability with identical timestamps` - Array.prototype.sort stability has been spec-mandated since ES2019

PR #666, opened 2026-07-31
Add @arkade-os/swap: Arkade Intents maker-side swap package

tiero:
Extract the swap client from arkade-os/wallet (commit 60b0834,
src/lib/swap/) into a standalone, framework-free package: offer covenant
binding + TLV codec, market discovery/validation, swap record store, and
restore-from-chain. Persistence goes through an injected SwapStorage
interface; registry URL, logger, and fetch are parameters instead of
wallet imports.

The two covenant program JSONs are byte-identical to the wallet's
(they are hashed into address derivation); only filenames, symbols
(bancoPrograms -> swapPrograms), and prose are renamed. The address
golden test guards the derivation.

Deferred with ponytail markers: preFeeDisplayRate (display-only), the
quote snapshot fields, storage remove(). MIGRATION.md documents the
wallet-side changes and the follow-up prompt.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_014k24sV12VwT3PX5d3GSNYT

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated swap package supporting atomic BTC and asset swaps.
  * Added offer creation, encoding, validation, cancellation, and deterministic address generation.
  * Added market discovery, quote calculation, caching, and plan validation.
  * Added persistent swap storage with in-memory and browser database options.
  * Added wallet restore support for recovering swap history from transactions.

* **Documentation**
  * Added package documentation, usage examples, and migration guidance.


tiero:
Thanks — worked through all 8. Six addressed, one declined with reasoning, two were already fixed in commits you hadn't seen.

**Fixed**

- **§1 · cancelled-swap mislabel** — went with your option (b), but made it a real contract rather than a footnote. `restoreAssetSwaps` now carries a "Caller contract" section stating that the `fulfilled` guess is never revisited once persisted, why `fulfilled` is the chosen default (a maker-initiated cancel normally has its tx locally, so a spend we can't see is more likely a fill), and the two ways out: feed live spend events into `updateAssetSwap`, or drop the record so it leaves `existingIds` and gets re-decided against a fuller `txs`.

  I considered (a). A `presumed-fulfilled` status only helps if the caller excludes it from `existingIds` — so it's the same caller contract, encoded in a public type that the wallet would then have to map in its UI. Not worth widening `AssetSwapStatus` for in a behaviour-preserving extraction. Worth revisiting before 1.0 if a restore-only consumer shows up.

- **§4 · body-read failure** — real bug, fixed. A throwing `clone().text()` took the live response down with it, so the caller lost the quote it had actually received, surfaced as an internal error rather than a network one. Cache write is best-effort now.

- **§5 · direction guard** — both sites read `Boolean(x) === Boolean(y)`.

- **§6 · key prefix** — fixed, and applied to the ark signer key too, which had the same shape (CodeRabbit flagged that one separately). Both now go through one helper that passes a 32-byte x-only key through, strips a 33-byte compressed key only when the prefix is `0x02`/`0x03`, and throws otherwise. Full point validation would need a curve dependency the package doesn't otherwise carry; the length+prefix check catches the misconfiguration case you describe.

- **§7 · `useCache: false`** and **§8 · atomic `clear()`** — both already landed in `6ee670b`, before this review; you were reading `4820f0a`. `clear()` spans all three stores in one transaction, and there's a test seeding a fresh cache, forcing the refetch, and asserting the stale fallback still works with `useCache: false`.

**Declined**

- **§2 · unknown TLV tags** — keeping the strict reject, and I've recorded why at the throw site. This payload binds a covenant: a record we can't interpret shouldn't be silently dropped and the offer then displayed or cancelled as though understood. There's also a migration hazard — the wallet's decoder rejects unknown tags today, so a lenient package would accept offers its only consumer rejects.

  The forward-compat concern is real, I just don't think one client should resolve it unilaterally. The clean fix is at the spec level: give the offer format an explicit ignorable-field rule (Lightning's odd/even "it's ok to be odd" is the obvious prior art) and then both sides implement it deliberately. Happy to do that as a follow-up if whoever owns the offer spec wants it — note the current tags don't follow that convention, since `0x02` and `0x08` are even and required.

- **§3 · first-VTXO selection** — leaving as-is. Adding a logger parameter to `cancelOffer` is public API surface that no caller would pass, and the behaviour is already documented on the function. The genuine fix is for callers to pass `fundingTxid`, which every `AssetSwap` record carries.

Also confirming your byte-identity finding independently: both program JSONs are sha256-identical to the wallet's originals, and the golden address test passes unmodified. The internal `banco-*` name fields are deliberately untouched — they hash into address derivation, so renaming them would invalidate every live swap address.

59 tests, `tsc`, lint, ESM+CJS+types build, and a Node smoke round-trip all green at `d8d51b5`.

---

tiero:
This pass read `225ec17`/`6ee670b`, but there are two commits after those — `11fb87c` and `d8d51b5` — which is where §1, §4, §5 and §6 were fixed. Line refs at current HEAD (`7b12db0`):

| Item | Verdict here | Actual state at HEAD |
| --- | --- | --- |
| §1 mislabel | still open (blocker) | **Fixed** — `restore.ts:70` opens a "Caller contract" section on the `restoreAssetSwaps` JSDoc |
| §4 body-read | still open | **Fixed** — `markets.ts:42` wraps the cache write, comment at :42–44 |
| §5 direction guard | still open | **Fixed** — `offer.ts:218` and `:260` both read `Boolean(x) === Boolean(y)` |
| §6 key prefix | partially improved | **Fixed** — `offer.ts:144` throws unless the prefix is `0x02`/`0x03` |

On §1 specifically: I took your option (b), and deliberately put it on `restoreAssetSwaps` rather than the module header, since that's where a consumer wiring up a cold-start restore will actually read it. It states that the `fulfilled` guess is never revisited once persisted, why `fulfilled` is the default (a maker-initiated cancel normally has its tx locally, so an unseen spend is more likely a fill), and gives both escape hatches — route live spend events through `updateAssetSwap`, or drop the record so it leaves `existingIds` and gets re-decided against a fuller `txs`.

I considered (a) and rejected it: a `presumed-fulfilled` status only helps if the caller excludes it from `existingIds`, so it's the same caller contract, just encoded in a public type the wallet would then have to map in its UI. Not worth widening `AssetSwapStatus` in a behaviour-preserving extraction — worth revisiting before 1.0 if a restore-only consumer materialises.

**N6 — good catch, fixed in `7b12db0`.** This was the one genuinely new item and it's a real trap: `initDatabase`'s `contains()` guard reads like it makes store additions self-healing, but `onupgradeneeded` only fires on a version *increase*, so a store added without a bump is silently missing for anyone whose database is already open at that version. Harmless today since nothing has shipped. `DB_VERSION` now carries a comment saying exactly that.

§3 stays as-is, as noted last round: a logger parameter on `cancelOffer` is public API no caller would pass, and the fix callers actually need — passing `fundingTxid` — is already available on every `AssetSwap`. §2 is settled; thanks for marking it a team decision.

Thanks for N1–N5 confirmations. Green at `7b12db0`: 59 tests, `tsc`, lint, ESM+CJS+types build, Node smoke round-trip. Program JSONs still sha256-identical to the wallet's, golden address test unmodified.

---

PR #673, opened 2026-08-04
feat: enforce agreed amounts on swap funding and claims

pietro909:
- `claimArk` and `claimVHTLC` claim every spendable VTXO at the lockup script, not only the first
- all three claim paths check the locked total against the agreed amount before claiming; a partially claimed lockup sweeps the remainder
- accepted renegotiation quotes are persisted (`acceptedQuoteAmount`) and floor later renegotiations
- `createChainSwap` requires the response to echo the locally computed server lock amount
- `createSubmarineSwap` bounds `expectedAmount` to the invoice plus the advertised fee schedule; nothing is persisted or funded on a mismatch

## Summary by CodeRabbit

* **New Features**
  * Added support for claiming multiple VTXOs in a single offchain transaction.
  * Added persistence and tracking of renegotiated quote amounts.
  * Added support for recoverable chain-claim paths and persisted claim transaction IDs.

* **Bug Fixes**
  * Improved validation of lockup and funding amounts against agreed amounts and invoice fees.
  * Added retries while waiting for sufficient funds to appear.
  * Added clear errors when received amounts are below or above acceptable limits.

### DIFF
--- a/packages/boltz-swap/src/arkade-swaps.ts
+++ b/packages/boltz-swap/src/arkade-swaps.ts
@@ -24,6 +24,7 @@ import {
     VHTLC,
     ArkInfo,
     isRecoverable,
+    hasTerminalSpend,
     ArkTxInput,
     Identity,
     VirtualCoin,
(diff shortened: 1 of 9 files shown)


pietro909:
@arkana-ai-bot I'm pushing back on *Issue 1*:
>  assertSubmarineExpectedAmount fee-fetch failure blocks all submarine swaps 

The fee schedule and the swap quote come from the same counterparty: a server that wants to overcharge just fails the GET and serves the POST, which defeats the point of the defensive check.  The  cost of fail-closed is a retry; the cost of fail-open is an unbounded funding amount.

PR #675, opened 2026-08-04
feat: bind recipient addresses to wallet network and operator

pietro909:
## What

Recipient Arkade addresses are now validated against the wallet's own context before any funds move. A new `assertRecipientArkAddress` (in `wallet/utils.ts`, reusing `classifyAgainstSignerSet`) checks:

- the bech32 prefix matches the wallet's network (`ark` vs `tark`), and
- the embedded server key is the wallet's current operator signer, or a deprecated signer whose rotation cutoff has not passed (`CURRENT`/`MIGRATABLE`/`DUE_NOW` accepted; `UNKNOWN_SIGNER`/`EXPIRED` rejected).

Previously both fields were decoded and discarded, so an honest paste of a testnet or foreign-operator address was accepted — producing a VTXO at an output key the recipient's wallet (watching a different operator's indexer) would never see. No escape hatch is added: cross-operator transfer is not a defined feature.

## Where it applies

- `send` — via `validateRecipients`, which now takes the wallet context.
- `sendBitcoin` — the selected-VTXOs path, which decoded the address directly.
- `settle` — the output classification loop, restructured so a binding failure surfaces as its own error instead of falling through to the onchain-address branch. Context is built lazily, only when an offchain output is present.

Payment-rail classification (`arkTarget`/`isValidArkAddress`) stays format-only by design: the rails execute through `wallet.send`, so enforcement at the validator covers them without changing public predicates.

Sweep of the remaining `ArkAddress.decode` sites in the SDK: all are the wallet's own address (`ramps`, `asset-manager`, `vtxo-manager`, settle/change paths), format-only checks, server-response validation (`wallet/validation.ts`), or the delegate path whose address comes from the configured delegate provider rather than user input.

## Compatibility

`validateRecipients` is module-internal (not exported from the package index); its only caller is `Wallet._sendImpl`. The rotation-window case is covered: an address issued just before a signer rotation stays payable until the advertised cutoff.

## Tests

11 new tests in `test/recipient-address-binding.test.ts`: validator-level policy (prefix, unknown operator, current key, deprecated pre/post-cutoff, no-cutoff sentinel) and wallet-level wiring for `send`, `sendBitcoin` and `settle`. Rejection tests verified to fail with the guard removed while accept tests stay green. Full unit suites, lint and build green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to ensure recipient Arkade addresses match the wallet’s network and recognized signer keys.
  * Rejected addresses using unknown or expired signer keys.
  * Accepted addresses signed by current or valid deprecated signer keys.
  * Applied validation across Bitcoin sends, regular sends, and settlement transactions.
  * Prevented invalid recipient addresses from proceeding to payment processing.
  * Improved handling of Arkade and on-chain settlement outputs.

### DIFF
--- a/packages/ts-sdk/src/wallet/signerRotation.ts
+++ b/packages/ts-sdk/src/wallet/signerRotation.ts
@@ -59,8 +59,12 @@ export interface SignerSet {
      * Deprecated signers keyed by x-only hex, mapped to their cutoff. The cutoff
      * is always a bigint (arkd advertises it non-nullable); `0n` means "no cutoff
      * advertised" (→ `DUE_NOW`).
+     *
+     * Read-only: the wallet hands out its live `_deprecatedSigners` here rather
+     * than a copy, and that map drives coin selection and the pendingRecovery
+     * balance bucket. A holder must not write through it.
      */
-    deprecated: Map<string, bigint>;
+    deprecated: ReadonlyMap<string, bigint>;
 }
 
 /**
--- a/packages/ts-sdk/src/wallet/utils.ts
+++ b/packages/ts-sdk/src/wallet/utils.ts
@@ -212,8 +212,7 @@ type ValidatedRecipient = Required<Omit<Recipient, "extensions">> & {
 };
 
 /**
- * Wallet-side context a recipient Arkade address must match: the wallet's
- * bech32 prefix and the operator signer set. An address failing either check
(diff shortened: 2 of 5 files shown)


pietro909:
@arkana-ai-bot about F4: Declining. The preceding commit deliberately trimmed comments.

PR #679, opened 2026-08-05
feat(swap): add @arkade-os/swap — maker-side Arkade Intents swaps

pietro909:
Adds `@arkade-os/swap`: the maker-side Arkade Intents swap implementation, extracted out of `arkade-os/wallet` so integrations other than the wallet app can consume it.

Four layers — `offer` (the swap covenant, TLV codec, `createOffer`/`cancelOffer`), `markets` (solver-registry discovery, cached, with plan validation), `store` (persisted `AssetSwap` records over a repository), and `restore` (rebuilding records from chain after a wallet restore).

The extraction is behaviour-preserving: the compiled program JSONs are sha256-identical to the wallet's originals and the golden address test came across unmodified, so derived addresses are pinned across the move.

The second commit wires the package into CI (type-check, dist smoke test, publish-shape check) and the release pipeline (`swap` release target, `@arkade-os/swap/<version>` tags), and generalises the existing boltz-swap special-casing into a `dependsOnSdk` flag. Version `0.0.1` → `0.1.0`.

**Note for the next SDK release:** it will now also release `@arkade-os/swap`, for the same reason it already releases `@arkade-os/boltz-swap` — pnpm rewrites `workspace:*` to an exact version at publish time, so a dependent left unreleased stays pinned to the previous SDK. `pnpm run release:dry-run` shows the full plan first.

## Relationship to #667

This replaces the extraction half of #667. That PR is now stacked on top of this one and carries only the RFQ taker layer and the onchain corridor — the net-new protocol code, which is easier to review against a standard of its own.

The 15 original extraction commits, including both bot review passes and the responses to them, are preserved on `archive/swap-extraction-full-history` (tip `ee38780`).

## Summary by CodeRabbit

* **New Features**
  * Added the `@arkade-os/swap` package for creating, encoding, cancelling, and restoring asset-swap offers.
  * Added market discovery, quote validation, swap tracking, and IndexedDB persistence.
  * Added support for restoring swap history after wallet recovery.

* **Documentation**
  * Added package documentation and migration guidance for adopting the new swap functionality.
  * Updated release instructions and monorepo documentation to include the swap package.

* **Tests**
  * Added comprehensive coverage for offers, markets, persistence, restoration, and distribution packaging.


tiero:
Context on the "8 still open" list from the #666/#667 review lineage this extraction descends from — most of these were adjudicated to completion there, and the recorded rationale survived into this branch. Checked against head `128c735`:

- **1 (empty `swapAddress` → cancel falls back to current server key):** a recorded deliberate decision, with the revisit condition in the file — `restore.ts:221-222`: *"ponytail: empty address makes cancel fall back to the current server key; store the funded address if server-key rotations become real."* Server-key rotation is not a thing today; when it becomes one, the marker says exactly what to change.
- **2 (fulfilled/cancelled mislabel permanent after restore):** documented as an explicit caller contract, not an oversight — `restore.ts:73-86` spells out that the `fulfilled` guess is never revisited and that callers must either feed live spend events or drop the record to force a re-decide. A restore-only integration accepting that contract is the design.
- **3 (`cancelOffer` falls back to `vtxos[0]`):** adjudicated in the prior round — the fix a caller actually needs is passing `fundingTxid`, which every persisted `AssetSwap` already carries; the fallback exists for the offline-offer case where there is exactly one deposit. Not silent wrong-VTXO selection when used per the API.
- **8 (all-or-nothing vtxo lookup):** acknowledged optimization in the prior round, as this pass itself notes.
- **4, 5, 6 (namespace cross-reference comment, exported BTC sentinel, IndexedDB transaction-lifetime warning):** comment/constant hygiene nits — fine as a small follow-up, not merge material.
- **7 (zero-length variable-width TLV field reaches `AssetId.fromBytes`):** the one item with substance — `!v` doesn't catch an empty `Uint8Array`, so a malformed offer yields a misleading downstream error instead of a clean decode rejection. Robustness-of-error-message only (the decode still fails), so worth a two-line guard whenever convenient, not a blocker.

Full prior-review history, including both bots' passes and the responses, is preserved on `archive/swap-extraction-full-history` per the PR description.

---

PR #682, opened 2026-08-06
swap: client-unilateral refund leaf, matching the reference solver

Kukks:
## Summary

The reference solver (`arkade-os/lightning-swap-service`, see [PR #8](https://github.com/arkade-os/lightning-swap-service/pull/8)) now ships a fourth leaf on the shared lightning-send/onchain-send covenant: `refundUnilateral` — the maker's own key behind a CSV delay, needing neither the Ark server nor the emulator once it matures. This closes the maker's only non-cooperative recourse gap: previously the sole refund path needed both the Ark server *and* the external emulator to co-sign.

`@arkade-os/swap` derived only the three-leaf shape, so a trader using it would compute a different `lockup_address` than the solver quotes and refuse to fund every swap once the solver starts serving four-leaf quotes. This PR closes that gap.

- New leaf in `swap-lightning-send.program.json` (shared by both the lightning-send and onchain-send corridors — one program, confirmed unchanged by the existing "aliases the Arkade lockup" golden test).
- `lightningSendVtxoScript`/`deriveOnchainSend` take `clientRefundPubkey` + `clientRefundDelay` (required — every RFQ-family request this package builds needs the leaf; there's no legacy shape to fall back to).
- `lightningSendRequest`/`onchainSendRequest` carry `client_refund_pubkey`.
- New `unilateralRefundWithoutReceiverDelay(claimDelay)`, one 512s tier past `unilateralClaimDelay`, matching the solver's own three-tier delay ladder.
- `requestLightningSend`/`requestOnchainSend` generate the key per call and now return it (`clientRefundPubkey`/`clientRefundPrivateKey`) — they were generating it internally with no way for a caller to retrieve it, which would have silently discarded the maker's only non-cooperative recourse. Same bug, same fix, as the equivalent function on the solver side.

Golden `scriptPubKey` re-pinned after adding the leaf. Verified byte-identical against the solver's own `CovenantSwapScript` for the same inputs via a throwaway cross-repo script (not part of either test suite) — two independent implementations, same output.

Based on `feat/arkade-swap` (not `master`), matching #667/#681 — this is where the RFQ/onchain client code lives; `@arkade-os/swap` hasn't been published yet, so there are no live consumers to break.

## Test plan
- [x] `pnpm --filter @arkade-os/swap typecheck/lint/build` — clean
- [x] `pnpm --filter @arkade-os/swap test:unit` — 116/116 passing
- [x] Full monorepo pre-commit hook (`pnpm -r lint && pnpm -r test:unit`) — passing
- [x] Cross-repo byte-identical check against the reference solver's `CovenantSwapScript`

### DIFF
--- a/packages/swap/src/index.ts
+++ b/packages/swap/src/index.ts
@@ -53,6 +53,7 @@ export {
     requestLightningSend,
     rfqPair,
     unilateralClaimDelay,
+    unilateralRefundWithoutReceiverDelay,
     verifyLockupAddress,
     type InvoiceFacts,
     type RelaySocket,
--- a/packages/swap/src/rfq.ts
+++ b/packages/swap/src/rfq.ts
@@ -33,6 +33,7 @@
  */
 import { hex } from "@scure/base";
 import { ripemd160 } from "@noble/hashes/legacy.js";
+import { schnorr } from "@noble/curves/secp256k1.js";
 import {
     ArkAddress,
     RestArkProvider,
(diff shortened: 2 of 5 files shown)


Kukks:
Superseded by #683.

This PR added `refundUnilateral` (client alone, CSV-gated) to `lightningSendVtxoScript`'s hand-rolled JSON artifact. #683 replaces that artifact entirely — `lightningSendVtxoScript` now builds `VHTLC.Script` from `@arkade-os/sdk` directly, which already includes this exact leaf (`unilateralRefundWithoutReceiver`, one of VHTLC's own base six) plus six more (`refund`, `refundWithoutReceiver`, `unilateralRefund`, `nonInteractiveClaim`, `nonInteractiveRefund`) on top. Everything here is a strict subset of #683 via a superseded mechanism, so closing rather than carrying two competing implementations of the same function forward.

PR #683, opened 2026-08-06
feat(vhtlc): non-interactive claim/refund leaves; packages/swap adopts VHTLC.Script

Kukks:
## Summary

- Ports `nonInteractiveClaim` from #613 into `vhtlc.ts`: `server` + a covenant-tweaked emulator co-signer can push the receiver's claim, pinned to a pre-committed destination, without the receiver being online.
- Adds `nonInteractiveRefund`, the symmetric leaf on the refund side: `server` + `receiver` + a covenant-tweaked emulator co-signer, immediate (no timelock), pinned to the sender's pre-committed destination. Every other refund-side leaf in VHTLC (`refund`, `refundWithoutReceiver`, `unilateralRefund`, `unilateralRefundWithoutReceiver`) requires the sender's own signature — if that key is ever permanently lost, none of them are reachable. This leaf is the exception: server + receiver can release the refund the moment they agree the swap has failed, with no sender signature needed and no timelock to wait out. Trade-off worth being explicit about: unlike a timelock-gated design, this depends on the receiver's active cooperation, not just infrastructure uptime — if the sender's key is lost and the receiver is also unavailable, there's no remaining path.
- `packages/swap`'s `lightningSendVtxoScript` now builds `VHTLC.Script` directly instead of a hand-rolled `program.json` artifact compiled via `arkade.parseArtifact` — one implementation, not two that can drift apart. `swap-lightning-send.program.json` is removed as dead code.
- Fixes `packages/swap/package.json`'s `@arkade-os/sdk` dependency to `workspace:*`, matching `packages/boltz-swap`'s existing convention. It was pinned to the published registry version (`0.4.56`) and silently never linking to this monorepo's own `packages/ts-sdk` — meaning local SDK changes were invisible to `packages/swap` until an actual npm publish.
- Bumps `@arkade-os/sdk` to `0.4.57` for the new exported surface.
- Supersedes #682 (closed) — that PR's `refundUnilateral` leaf is now `unilateralRefundWithoutReceiver`, one of `VHTLC.Script`'s own base six.

## Self-review pass, before this PR was ever looked at by anyone else

Same process as the consuming repo's PR (`lightning-swap-service#8`): four independent adversarial passes over the actual diff, every finding verified against the real code myself (including running probes against the live code, not just reading it) before acting. Two real, critical issues came out of it:

1. **`packages/swap`'s trader library couldn't complete a swap against the real solver at all.** `lightningSendRequest`/`onchainSendRequest` emitted `profile.sender_pubkey`; the reference solver's schema requires `profile.client_refund_pubkey` (confirmed against `docs/rfq-protocol.md` and the solver's own Zod schema directly) and is `.strict()` on the profile object — so every request failed on both the missing required field and the unrecognized one. The two names describe the same key; they just diverged independently on each side of the wire. Fixed in `49a0569b`.
2. **A same-length, wrong-witness-version destination script was silently accepted and baked into the covenant with no error**, on `nonInteractiveClaim`/`nonInteractiveRefund` — the latter being the sender's documented last-resort recovery path when their own key is already lost. `validateOptions`/`enforcePayTo` checked `receiverPkScript`/`senderPkScript` were exactly 34 bytes but never that they actually started with the P2TR `0x51 0x20` prefix; a same-length OP_RETURN script (exactly what `ArkAddress.subdustPkScript` produces) passed silently. A near-identical `enforcePayTo` already exists in this repo (`test/e2e/utils.ts`, explicitly "mirrors the Go `enforcePayTo` helper") and does check the prefix — this one didn't match it. Fixed in `8bd63e26`, along with an adjacent missing-field guard inconsistency in the same validation blocks (a partially-populated leaf crashed with a raw `TypeError` instead of the intended clean error).

Both new/strengthened test assertions were verified as real regression guards, not just passing by construction: reverted each fix locally, confirmed the corresponding test fails with the predicted symptom, restored, confirmed the working tree diff came back clean.

**Flagging, not fixing — an existing tension worth an explicit call:** `packages/swap/package.json`'s `@arkade-os/sdk` pin (`workspace:*`, changed from `0.4.56` in this same PR) reverts commit `7d94e46b`, whose own message states that `workspace:*` breaks `pnpm add "github:arkade-os/ts-sdk#...&path:/packages/swap"` (`ERR_PNPM_WORKSPACE_PKG_NOT_FOUND`) — i.e. this PR's own fix for local-monorepo linking re-breaks `packages/swap`'s installability as a git dependency, the same way `packages/ts-sdk` is installed by `lightning-swap-service` today. Both needs are real (this PR needs unpublished `ts-sdk` changes; some future consumer may want to git-install `packages/swap` directly); reconciling them isn't a coding fix, it's a call for whoever owns the publish workflow.

## A new HTLC construct: `VHTLC.ScriptV2`

Adds `VHTLC.ScriptV2` alongside the existing `VHTLC.Script`: same leaf ladder, same options shape, built with a preimage condition that gates on an exact 32-byte witness item (`OP_SIZE 32 OP_EQUALVERIFY` ahead of the hash check) rather than any-length input — the standard shape real-world HTLC scripts (BOLT3's, for one) use. `VHTLC.Script` is unchanged and stays available; the two share one internal builder (`BaseScript`) so they can never drift apart on anything but that one condition, rather than being hand-duplicated copies.

`packages/swap`'s `lightningSendVtxoScript` and `onchainHtlc.ts` (the L1 side of the onchain corridor) both move to the new construct — this is what solvers derive for `arkade:BTC<->lightning:BTC` and the onchain BTC corridor going forward. Golden tests updated to the freshly-derived bytes for both, plus explicit coverage that all three preimage-gated leaves (`claim`, `unilateralClaim`, `nonInteractiveClaim`) pick up the new condition together, and that every leaf not gated on the preimage stays byte-identical to `VHTLC.Script`'s.

Also rebased onto `feat/arkade-swap`'s current tip (picked up #681's spendability-gate work — clean rebase, no conflicts, no overlap with anything this PR touches).

## External review follow-ups

Two items from a second review pass, both explicit decisions rather than code changes:

- **`packages/swap/package.json` stays at `0.0.1`.** Not an oversight: the package isn't published, so `0.0.1` is being treated as unstable/pre-release rather than a version this PR's breaking changes need to bump.
- **`VHTLC.ScriptV2`'s preimage condition has no go-sdk counterpart** — `go-sdk/vhtlc/utils.go`'s `makePreimageConditionScript` still emits the unguarded shape, so a Go-based trader would derive a different (V1) address for the same VHTLC params. Confirmed. Same class of gap as the pre-existing `nonInteractiveRefund`/go-sdk mismatch noted above, and impact is latent, not active — this corridor is TS-only via `lightning-swap-service` today. Being coordinated on the go-sdk side outside this PR.

## Test plan

- [x] `pnpm --filter @arkade-os/sdk typecheck` / `pnpm --filter @arkade-os/swap typecheck` — clean
- [x] `pnpm -r lint` / `pnpm -r test:unit` (full monorepo, via the pre-commit hook on every commit) — clean, 2319 + 116 tests passing across `@arkade-os/sdk` and `@arkade-os/swap`
- [x] New unit coverage for both leaves in `packages/ts-sdk/test/vhtlc.test.ts` (byte-shape assertions incl. independently-recomputed covenant-tweaked keys, address changes when either leaf's params change, base six leaves stay byte-identical when neither optional leaf is supplied, P2TR-prefix and missing-field rejection)
- [x] `packages/swap/test/rfq.test.ts` re-pinned golden `pkScript` for the new 8-leaf shape; per-leaf byte assertions for all eight
- [x] Cross-repo verification against a consumer (`lightning-swap-service`, pinned to this branch via git+subdirectory dependency): built `CovenantSwapScript` and `lightningSendVtxoScript` from identical params and diffed the output directly — `pkScript` and all seven spendable leaves came back byte-identical
- [x] `pnpm -r lint` clean (prettier + the package's own vtxo/provider-boundary guards)
- [x] `VHTLC.ScriptV2`: verified as a genuine regression guard, not passing by construction — reverted the new preimage condition locally, confirmed the version-comparison tests fail with the predicted symptom, restored
- [x] Live regtest: `lightning-swap-service`'s onchain-send E2E (real Taproot script-path spend through the new claim leaf, broadcast and accepted by a real regtest Bitcoin node) and relay E2E (Arkade covenant claim via `ScriptV2`) both re-run end to end against the new construct — both pass
- [x] CI green: `Lint, Test & Build` + all 5 integration jobs (two transient failures on this run — a 1ms timing assertion in `rate-gate.test.ts` and a `settlement.test.ts` timeout, both unrelated to this change and both passed clean on re-run)


Kukks:
Heads-up for whoever reviews/merges this, found while researching `arkade-os/wallet`'s `feat/arkade-swap` branch as part of E2E-testing the sibling `lightning-swap-service` PR:

The wallet branch's `package.json` pins `@arkade-os/swap` to `github:arkade-os/ts-sdk#feat/arkade-swap&path:/packages/swap` — an **unpinned branch ref**, not a commit. Its `src/lib/lnSwap.ts` → `requestLnSend` calls `requestLightningSend` without supplying any sender/client-refund pubkey.

Once this PR merges into `feat/arkade-swap`, `lightningSendRequest`'s new required `senderPubkey` parameter means the wallet's next `pnpm install` will silently pull in a `packages/swap` its own call site no longer satisfies (a TypeScript build break at best, a runtime one if it's not caught first). Since the dependency floats on the branch rather than a commit, there's no version bump to force a coordinated update.

Not blocking this PR — just flagging so the wallet side can be pinned/updated in step rather than breaking unexpectedly on the next `pnpm install` there.

pietro909:
Two issues in the onchain corridor, both in the amount-handling path. Neither is in `vhtlc.ts` — the VHTLC leaves themselves check out against go-sdk's `vhtlc/noninteractive.go` opcode for opcode, and the P2TR guard added in 8bd63e26 closes the earlier gap.

## 1. The L1 fill amount is never checked — the maker can be induced to publish `P` for a token fill

`packages/swap/src/onchainHtlc.ts:339,371`

`awaitOnchainFill(chain, htlc, minConfirmations)` returns the largest UTXO paying the HTLC script at the required depth. `claimOnchainFill` then gates only on the time window and broadcasts. Nothing in the package compares the fill against `quote.to_amount` — `OnchainHtlc` doesn't even carry an expected amount, so a caller has nothing to compare against unless it threads the quote through itself.

Concretely, on `arkade:BTC->onchain:BTC`:

1. maker funds the Arkade lockup with `from_amount` (say 100 000 sats);
2. solver funds the L1 HTLC with 400 sats instead of the quoted 99 000;
3. `awaitOnchainFill` reports it as filled — 400 > `ONCHAIN_DUST_SATS`, depth is satisfied;
4. `claimOnchainFill` broadcasts, and `P` is now public in the claim witness;
5. solver claims the full 100 000 sat lockup through the VHTLC `claim` leaf (receiver + server — no maker signature needed).

Maker pays 100 000, receives 400. Every guardrail on this path (`claim_window_too_short`, `timelock_order`, `confirmations_out_of_range`, the 90-minute claim margin) is about *time*; none is about *value*.

The invariant already exists in the data model — `store.ts` documents `AssetSwap.toAmount` as *"the fill pays >= this"* — it just isn't enforced on this corridor. Suggested fix: make `minAmount` a required parameter of `awaitOnchainFill` (skip UTXOs below it rather than returning the largest of a bad set), and re-assert it in `claimOnchainFill` so a caller that recovers a UTXO from `classifyOnchainHtlc` can't bypass the check. The README example at line 178 currently claims blind and would need the same argument.

## 2. No amount gate on any quote

`packages/swap/src/rfq.ts:971` and `rfq.ts:713`

`requestOnchainSend` returns `fundAmount: quote.from_amount` verbatim, and `assertFundable` never looks at an amount. So when the maker pins `amountSide: "from"` with `amount: 100_000`, nothing checks the solver honoured that side, and when it pins `"to"` there is no ceiling on `from_amount` at all — the maker asks for a price and funds whatever comes back. `offerTermsFromQuote` has the mirror gap on the arkade↔arkade corridor: `quote.to_amount` becomes the offer's `wantAmount` with no floor.

This is the one place where the module's own trust model isn't applied. `docs/rfq-protocol.md` §4.2 makes `from_amount`/`to_amount` binding fields carrying the solver's fee in the spread, and §6 says the client trusts binding fields *and derives locally* — for addresses that's `verifyLockupAddress`, but for amounts there is no equivalent. Minimum fix: assert the pinned side matches what was requested, and take a caller-supplied `maxFromAmount` / `minToAmount` for the other side (which finding 1 can then reuse as its `minAmount`).

Related, lower severity, same root cause — `rfq.ts:695`: `requestLightningSend` funds `params.invoice.amountSats` and ignores `quote.from_amount` entirely. That's correct only while the spread is zero, which the protocol doc calls out as today's accident (§13: *"`to_amount` (and `from_amount`; today they are equal — zero spread)"*). The first solver that charges a spread gets an underfunded lockup, won't pay the invoice, and the maker's sats sit until `refund_locktime`.

Build, unit tests and lint are green on 8bd63e26 (`packages/swap` tests need `pnpm run build` first — they resolve `@arkade-os/sdk` through `dist/`, so a vitest-only run fails 3 `rfq.test.ts` cases against a stale build).

pietro909:
@arkana-ai-bot ignore release issues and breaking public API issues for `@arkade-os/swap` in this PR. The package hasn't been released yet.

pietro909 (approved):

Kukks:
Follow-up on the two open items from this review:

- **Package version**: staying at `0.0.1` — explicit decision, not an oversight. The package isn't published yet, so `0.0.1` is being treated as unstable/pre-release rather than a version this PR's changes need to bump. Noted in the PR description.
- **go-sdk `VHTLC.ScriptV2` parity** (Finding C): confirmed real, latent (this corridor is TS-only via `lightning-swap-service` today, so no active impact). Being coordinated on the go-sdk side separately, outside this PR — not filing a tracking issue from here since that coordination is already in motion elsewhere.

Merged as 5327fe19.

PR #684, opened 2026-08-07
feat(swap): register offer covenants with the contract manager

pietro909:
Phases 1–2 of ContractManager integration, stacked on #683.

- `createOffer` registers the covenant as an `"arkade"` contract with `metadata.genericallySpendable: false`, so the deposit is watched and stays out of generic coin selection. The cancel leaf has no timelock, so an unmarked deposit is forfeitable by `send`/`settle`/background renewal.
- `cancelOffer` passes the contract manager to `Arkade.connect`, so a registered offer resolves its VTXOs from the repository.

Two deviations from the plan, both because `createOffer` derives and the consumer funds — there is no in-package funding site:

- registration happens before funding, matching NArk and boltz-swap (register on create);
- it throws rather than being best-effort: nothing is at stake pre-funding, whereas the same failure after `wallet.send` strands an unwatched deposit.

Fixed while writing it: without `network` on `Arkade.connect` the row's address derives against the SDK default (`ark1…` row for a `tark1…` deposit).

## Test plan

- [x] e2e merge gate through the real `createOffer`/registration path, no hand-built rows. Flipping the marker to `true` reproduces the hazard: the deposit enters `getSpendableVtxos()`, an unrelated `wallet.send` **succeeds**, and cancel then fails with "no spendable VTXO at the swap address".
- [x] every new assertion probed by reverting the fix and confirming failure
- [x] cancel still works with the deposit gated (explicit-input path stays open)
- [x] unit: swap 124, `@arkade-os/sdk` 2315, boltz-swap 609; lint clean

## Note to reviewers

The whole `packages/swap` package isn't released yet: exclude API breaking changes and version numbering from the analysis.

## Summary by CodeRabbit

* **New Features**
  * Added support for non-interactive claim and refund paths in swap transactions.
  * Swap requests now securely bind sender keys and receiver payout destinations.
  * Offers are registered before funding, enabling reliable recovery and cancellation.
  * Escrowed deposits remain recoverable while being excluded from ordinary spendable balances.

* **Bug Fixes**
  * Improved cancellation handling when restoring offers.
  * Added validation to prevent mismatched swap scripts and invalid payout destinations.

* **Tests**
  * Expanded coverage for registration, recovery, cancellation, and end-to-end swap behavior.

### DIFF
--- a/packages/swap/src/offer.ts
+++ b/packages/swap/src/offer.ts
@@ -303,7 +303,7 @@ async function registerOfferContract(
     // funded would leave the real deposit unwatched and unmarked, which is the
     // failure this whole registration exists to prevent
     if (hex.encode(contract.pkScript) !== hex.encode(expectedPkScript)) {
-        throw new Error("registered covenant does not match the offer's swapPkScript");
+        throw new Error("derived covenant does not match the offer's swapPkScript");
     }
     await contract.register({
         label: OFFER_CONTRACT_LABEL,
@@ -455,6 +455,9 @@ export async function cancelOffer(
         // instead of a direct indexer query; the indexer above stays as the
         // fallback for offers created before registration existed
         contractManager: await wallet.getContractManager(),
+        // no `network`, unlike registerOfferContract: the row lookup is by
+        // script and the payout script comes from wallet.getAddress(), so the
+        // client's network (which only shapes address derivation) is unused here
     });
 
     // Rebuild the contract with the offer's own keys (not the client's) so the
--- a/packages/swap/src/offer.ts
+++ b/packages/swap/src/offer.ts
@@ -253,6 +253,64 @@ export function decodeOffer(data: Uint8Array): Offer {
     };
 }
 
+// ── Contract registration ────────────────────────────────────────────────────
+
+/** Label for a registered offer covenant. A script-level string, deliberately
+ * not the swap id: identical offers share one address and `createContract` is
+ * first-writer-wins, so the second deposit would inherit the first's label. */
+export const OFFER_CONTRACT_LABEL = "Arkade swap offer";
+
+/** `metadata.kind` for a registered offer covenant — what this script *is*,
+ * which is the only kind of fact a shared script row can carry truthfully.
+ * Per-offer identity (swap id, `offerHex`, `fundingTxid`) stays in `AssetSwap`. */
+export const OFFER_CONTRACT_KIND = "asset-swap-offer";
+
+/**
+ * Register an offer's covenant as an `"arkade"` contract, so the deposit is
+ * watched, survives restarts and re-derives offline — and, critically, so the
+ * wallet knows the funds are escrowed.
+ *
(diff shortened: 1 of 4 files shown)


pietro909:
Re: @arkana-ai-bot's review — a mapping of each finding to where it was addressed. Context first: after the rebase, this PR's diff is only the offer-registration work (`offer.ts` + tests); the VHTLC/RFQ commits the review covered now live in the base branch `feat/arkade-swap`. Findings against that code are fixed in **#692** (targeting the base branch) rather than widening this diff.

| Finding | Resolution |
| --- | --- |
| **#1** golden script change / deployment ordering | **#692** — pre-release "Breaking changes" section in the swap README (CHANGELOG.md is managed externally) covering both corridors' address change and the coordinated-deployment requirement; both golden tests now cross-reference it. A version mismatch refuses quotes at the compare-only address check — it does not lose funds. |
| **#2** `client_refund_pubkey` | Fixed (`d2bfc18b`), confirmed in the review. |
| **#3** `senderPrivateKey` persistence | **Deferred by design.** This is a persistence-model decision, not a patch; it is recorded in the integration plan's RFQ-secret-persistence section and gates Phase 3b (RFQ VHTLC registration), where the swap record and the key must be persisted together. |
| **#4** `hex.decode` guards | **#692** — both `receiver_pk_script` sites decode through a helper that raises a solver-blaming diagnostic on malformed hex. |
| **#5** preimage length check | Closed by the review itself (`SIZE 32 EQUALVERIFY` in both corridors). |
| **#6** `PUSHCURRENTINPUTINDEX` as output index | **#692** — documented on `enforcePayTo`. The answer is stronger than a protocol citation: the pairing is not an assumption about round assembly, the covenant *imposes* it on the spender. Whatever index the input occupies, the output at that same index must pay the destination ≥ the input value or the ArkadeScript fails and the server never co-signs; a tx with no output at that index makes the leaf unsatisfiable, not fooled. Index alignment is a liveness obligation on the solver (the only party building these spends — the SDK's own refund deliberately uses the interactive leaf, see `refund.ts`), never a safety assumption. |
| **#7/#8** removed exports / new required params | **#692** — migration notes in the same README section. |
| **N1** BIP68 overflow in the refund tiers | **#692** — confirmed real and fixed: the `unilateralClaimDelay` cap now sits two 512s steps below BIP68's ceiling so both stacked refund tiers always encode, with unit tests at the new maximum (the full eight-leaf contract compiles) and just past it. |
| **N2** duplicate `getInfo` in `registerOfferContract` | Partially addressed here (`serverPubkey`/`network` are threaded; the pkScript guard makes the second fetch harmless). Fully eliminating it needs `Arkade.connect` to accept pre-resolved info — a core API change, noted as a follow-up. |
| `cancelOffer` network omission | Comment added (`59e66fca`) explaining why it is deliberate: the row lookup is by script and the payout script comes from `wallet.getAddress()`, so the client's network is unused there, unlike `registerOfferContract`. |

PR #686, opened 2026-08-07
fix(batch): bound the server-supplied batchExpiry before committing to the sweep leaf

pietro909:
The sweep leaf of the shared batch output is derived from `BatchStartedEvent.batchExpiry`, so `validateVtxoTxGraph` can only prove the tree is consistent with whatever value the round supplied — the value itself never leaves the leaf hash, and there was no floor on it.

`assertValidBatchExpiry` (new, `wallet/batchExpiry.ts`) checks it against a per-network policy and returns the timelock to encode:

- seconds-typed required off regtest, mirroring arkd's own rule (`config.go`: block-typed vtxo tree expiry is regtest-only)
- a wall-clock floor — 24h by default, ~10 blocks on regtest, which legitimately runs block-typed expiries (`ARKD_VTXO_TREE_EXPIRY=180`)
- equality with `vtxoTreeExpiry` when the operator advertises it

It runs at `onBatchStarted` before `confirmRegistration`, so a rejected round is never confirmed and no signing state is initialized. Rejections are `ServerResponseMismatchError`, matching the existing terminal-mismatch taxonomy.

Wired into all three batch handlers — arkade, `Wallet`, and the boltz-swap VHTLC one, which carried the same unbounded copy — plus the hand-rolled e2e handler.

`ArkInfo.vtxoTreeExpiry` is parsed as optional (`undefined`, never `0n`, when unadvertised — arkd dropped it from `GetInfoResponse` in arkade-os/arkd#737) and carried through the info snapshot as an optional v1 field, so existing snapshots stay readable.

Configurable via `WalletConfig.minBatchExpirySeconds` and a policy override on the free-function handlers; the floor can be lowered but not disabled.

### Notes

- Follow-ups, not in this PR: re-adding `vtxo_tree_expiry` to arkd's `GetInfoResponse` (until then the equality check is inert), and the same bounding for the checkpoint unroll script and the unilateral/boarding exit delays — the validator is shaped for that reuse.
- `NArk` has the same unbounded derivation at `BatchSession.cs:53`; worth mirroring there.

### Testing

`test/batch-expiry-validation.test.ts` covers the policy defaults, a boundary table (1 / 511 / 512 / 86016 / 86528 / 604672 on mainnet; 1 / 10 / 100 / 180 on regtest), and asserts the handler neither confirms nor initializes signing on rejection. Snapshot round-trip cases cover advertised, unadvertised, and pre-existing v1 snapshots.

Full unit suite green: 2299 ts-sdk, 610 boltz-swap, 67 swap. Lint and build clean. Regtest/e2e not run locally.

## Summary by CodeRabbit

- **New Features**
  - Added configurable batch-expiry policies with network-specific defaults and minimum expiry settings.
  - Added validation for server-advertised expiries, including seconds- and block-based values.
  - Exposed Ark server VTXO tree expiry information through provider, wallet, and snapshot APIs.
  - Added wallet configuration for custom minimum batch-expiry durations.

- **Bug Fixes**
  - Invalid or mismatched expiries are rejected before registration confirmation or signing.
  - Preserved compatibility with older wallet snapshots lacking expiry information.

- **Tests**
  - Added coverage for expiry validation, policy overrides, snapshot handling, and batch processing.

### DIFF
--- a/packages/boltz-swap/src/batch.ts
+++ b/packages/boltz-swap/src/batch.ts
@@ -65,9 +65,13 @@ export function createVHTLCBatchHandler(
 
             // Bound the expiry before confirming, so a rejected round is never
             // confirmed to the operator.
+            const info = await arkProvider.getInfo();
             const timelock = assertValidBatchExpiry(
                 event.batchExpiry,
-                resolveBatchExpiryPolicy(network, batchExpiryPolicy),
+                resolveBatchExpiryPolicy(network, {
+                    advertisedVtxoTreeExpiry: info.vtxoTreeExpiry,
+                    ...batchExpiryPolicy,
+                }),
             );
 
             await arkProvider.confirmRegistration(intentId);
--- a/packages/boltz-swap/src/batch.ts
+++ b/packages/boltz-swap/src/batch.ts
@@ -19,6 +19,9 @@ import {
     buildForfeitTx,
     ArkTxInput,
     getSequence,
+    assertValidBatchExpiry,
+    resolveBatchExpiryPolicy,
+    type BatchExpiryPolicy,
 } from "@arkade-os/sdk";
 import { sha256 } from "@noble/hashes/sha2.js";
 import { base64, hex } from "@scure/base";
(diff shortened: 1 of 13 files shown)


pietro909:
@arkana-ai-bot 

#3: `node_modules/bip68/index.js` - `encode` throws on non-multiples and on out-of-range values, so nothing truncates and nothing reaches tree validation

#4: "The floor can be lowered but not disabled."  doesn't exist,  `wallet/index.ts:107-114` says "Lowering it below the default relaxes a fund-safety bound

`dotnet-sdk` and `go-sdk` issues will be tracked separately, descope them from this PR.

PR #687, opened 2026-08-07
fix(checkpoint): bound the server-supplied checkpoint exit delay before committing to it

pietro909:
## Summary

Stacked on #686 (`fix/batch-expiry-validation`) — fixes the follow-up that PR's own description flagged as not-yet-covered: *"the same bounding for the checkpoint unroll script... the validator is shaped for that reuse."* This is `[P720-2]` from the security review: the checkpoint sweep script is entirely server-defined, same unbounded-server-timelock class as the batch sweep expiry Critical (`[P720-1]`, fixed by #686).

- Every offchain send/claim builds a checkpoint output whose server-claim leaf comes from `ArkInfo.checkpointTapscript`, checked only for decodability — no floor on the timelock, no check on the embedded pubkey. A malicious operator can ship a 1-block checkpoint script and sweep an in-flight checkpoint before it settles into a batch, stranding the recipient's vtxo.
- `assertValidServerUnrollScript` (new, `wallet/checkpointExitDelay.ts`) decodes and validates: a wall-clock floor (24h default — matches arkd's own `defaultCheckpointExitDelay`; ~2 blocks on regtest, tuned to clear this repo's own regtest envs, which range from 5 to 20 blocks across packages), seconds-typed required off regtest (mirrors arkd), and the embedded pubkey pinned to the forfeit key (arkd always builds this script from its own `forfeitPubkey`).
- The floor/type-check core is shared with `batchExpiry.ts` via a new `timelockPolicy.ts` extraction — no change to `batchExpiry.ts`'s existing public API.
- Wired into all **six** commit-time decode sites (the finding's "setup and rotation" undersold it):
  - `wallet.ts`: `Wallet.create()`, `rotateServerSigner()`, and the arkadeCash sweep path — pinned against the wallet's own `forfeitPubkey` at rotation/sweep (a value trusted before the current response, same asymmetry the existing server-signer rotation/deprecation flow relies on); self-consistency against the same response's `forfeitPubkey` at first-contact construction (same TOFU limitation as pinning `arkServerPublicKey` itself — the timelock floor is the real defense there).
  - `boltz-swap/src/utils/vhtlc.ts`: all three VHTLC claim/refund functions — self-consistency only, since these are stateless free functions with no persistent wallet state to pin against (same trust model #686 already used for the VHTLC batch handler's `vtxoTreeExpiry` check).
  - `checkpointUnrollCandidates()` (read-only recovery-matching helper) is untouched — gating it would break recovery of legitimately pre-existing checkpoints built before this fix ships.

## Test plan

- [x] New `test/checkpoint-exit-delay-validation.test.ts` (ts-sdk): policy defaults, boundary table (1 block / just-under-floor / just-over-floor / mainnet default, on both mainnet and regtest, including this repo's actual `ARKD_CHECKPOINT_EXIT_DELAY` env values), pubkey-mismatch rejection, decode-failure rejection.
- [x] New `Wallet.create()` rejection tests in `ark-info-snapshot.test.ts` (sub-floor timelock, wrong pubkey) asserting no snapshot is cached on rejection.
- [x] New rejection tests in boltz-swap's `vhtlc-claim-checkpoint.test.ts` covering `claimVHTLCwithOffchainTx`.
- [x] Fixed ~150 existing tests broken by the new validation (a shared hardcoded `checkpointTapscript` fixture across ~13 files no longer satisfied the floor/pubkey checks) and two `serverSignerRotation.test.ts` tests whose premises collided with the new pubkey pin.
- [x] `pnpm run build && pnpm run lint` clean.
- [x] `pnpm run test:unit` — 100% pass across all three packages (2319 ts-sdk, 614 boltz-swap, plus `swap`).
- [x] Regtest/e2e not run locally (no live regtest node in this environment).

### DIFF
--- a/packages/boltz-swap/src/utils/vhtlc.ts
+++ b/packages/boltz-swap/src/utils/vhtlc.ts
@@ -3,6 +3,8 @@ import {
     ArkProvider,
     ArkTxInput,
     assertSubmittedArkTxid,
+    assertValidServerUnrollScript,
+    resolveCheckpointExitDelayPolicy,
     Batch,
     buildOffchainTx,
     getSequence,
@@ -10,13 +12,13 @@ import {
     matchServerCheckpoints,
     Intent,
     getNetwork,
+    Network,
     networks,
     NetworkName,
     Recipient,
     VHTLC,
     VtxoScript,
     VtxoTaprootTree,
-    CSVMultisigTapscript,
     combineTapscriptSigs,
     Transaction,
     scriptFromTapLeafScript,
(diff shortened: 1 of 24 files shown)


pietro909:
@arkana-ai-bot cross-repo exposure is tracked elsewhere. The last commit d974af6 addresses the Blocker and the Medium priority findings.

PR #690, opened 2026-08-07
feat(swap): RfqSwapManager — drive funded RFQ swaps to resolution automatically

Kukks:
Stacked on `feat/arkade-swap`.

## What this adds

`packages/swap` has every building block a funded RFQ swap needs — `requestLightningSend`/`requestOnchainSend` quote and gate funding, `awaitOnchainFill`/`claimOnchainFill` take the L1 fill, `refundIfUnresolved`/`pushRefundWithoutReceiver` take the lockup back — but calling the right one at the right moment, for every swap in flight, across restarts, was left to the caller. `RfqSwapManager` is that missing piece: it monitors a set of funded swaps and drives each through to resolution automatically, via injected callbacks for every action that moves money (so no key material ever reaches this class — the same split `onchainHtlc.ts` makes with its `sign` callback), an injected `saveSwap` for persistence, and both an event surface and a `waitForSwapCompletion` promise escape hatch.

It also adds the primitive that makes that possible without a counterparty in the loop (`readLockupFate`), registers each lockup with the wallet's `ContractManager` so the indexer **pushes** its funding and its spend, and corrects a wrong claim about swept lockups that a previous commit on this branch introduced.

## The manager never asks the solver

**`RfqSwapManager` holds no `RfqTransport` at all.** Every fact it acts on is read from chain — the Arkade lockup through `readLockupFate`, the L1 half through `classifyOnchainHtlc`. Resolution is not something a solver reports; it is something the trader derives.

Two reasons, neither stylistic:

- **Liveness.** A solver that stops answering must not be able to degrade a trader's view of its own swap. Every deadline that matters here is a consensus timelock — `refundLocktime`, `htlcLocktime` — and none of them care whether a status route is up.
- **Trust.** `settled`/`refunded` from a solver is self-reported. The chain fact underneath it cannot be forged. The lockup's claim leaf can only be spent by revealing `P`, and the only legitimate way the solver obtains `P` is by completing its side. So a spend witness that **hashes to** the quote's `payment_hash` is proof of settlement. Every other leaf is a refund: `nonInteractiveRefund` is covenant-pinned to the trader's own address via `enforcePayTo`, and the four remaining refund leaves all require the trader's own signature. "Spent, but not by a hash-verified claim" therefore means the money went back to the trader.

### `readLockupFate`

New in `refund.ts`, following the same narrow-seam style as `RefundIndexer` / `RestoreIndexer`:

```ts
export type LockupSpendIndexer = Pick<RestIndexerProvider, "getVtxos" | "getVirtualTxs">;

export type LockupFate =
    | { fate: "open" }                           // still funded
    | { fate: "claimed"; preimage: Uint8Array }  // hash-verified
    | { fate: "returned" }                       // spent, provably not a claim
    | { fate: "unknown" };                       // nothing learned
```

Three details it encodes, each checked against SDK source rather than assumed:

- **The preimage lives in Ark's proprietary `ConditionWitness` PSBT field, not in `finalScriptWitness`.** A condition witness is attached *after* signing, precisely because it is not part of what is signed. Both sources are searched as one set — reading only the raw witness stack would miss every real settlement.
- **`spentBy` is the empty string, not absent, when there is nothing to name**, and when it *is* set it names the **checkpoint** transaction, which is the one carrying the lockup's `tapLeafScript`. The ark transaction spends the checkpoint, not the lockup.
- **`spentBy` alone is not a spend test.** The wire contract permits `isSpent: true` with an empty `spentBy`, which is why the SDK's own `hasTerminalSpend` unions all three facts. `readLockupFate` takes the same union.

**A matching witness *shape* is never proof; only a matching *hash* is.** Anything the indexer cannot answer comes back `unknown`, which is deliberately **not** `returned`.

## The lockup is now a registered contract

The manager used to hand-roll its watching layer: a timer, and an indexer read per swap per tick. `ContractManager` already provides that, and provides more — a registered lockup is in the wallet's own contract set rather than known only to this class.

**The blocker was the handler.** `ContractManager.upsertContractRow` derives the script from a row's params and refuses any row whose supplied `script` disagrees. The `vhtlc` handler builds `VHTLC.Script` (V1); `packages/swap` builds `VHTLC.ScriptV2` — the covenant script with `nonInteractiveClaim`/`nonInteractiveRefund`. The derived pkScript never matched, so registering a swap lockup failed outright.

### New `vhtlc-v2` contract handler (`packages/ts-sdk`)

A separate type rather than a flag on `vhtlc`, because a handler's `type` names the script class it derives, as every other registered type does (`default` → `DefaultVtxo.Script`, `vhtlc` → `VHTLC.Script`, `boarding`, `arkade`). The two **cannot collide** on one row: their preimage conditions differ in every tree, so no parameters make their pkScripts equal (pinned by a test). A use-named type (`swap-lockup`) would have been worse — rows are keyed by script, so a second use registering the same ScriptV2 would fight over one row.

It serializes every ScriptV2 param the swap package passes, including both optional covenant leaves, and refuses a half-specified pair by name rather than deriving a silently different script.

**Which leaves it offers.** ScriptV2 has eight; a wallet holding ONE participant key can satisfy four. Offering a leaf whose signature the caller cannot produce turns a clean refusal into a transaction that is built and then rejected.

| leaf | needs | offered |
|---|---|---|
| `claim` | receiver + server, preimage | yes, to the receiver |
| `refund` | sender + receiver + server | no — needs the counterparty live |
| `refundWithoutReceiver` | sender + server, CLTV | yes, to the sender |
| `unilateralClaim` | receiver, CSV | yes, to the receiver |
| `unilateralRefund` | sender + receiver, CSV | no — needs the counterparty live |
| `unilateralRefundWithoutReceiver` | sender, CSV | yes, to the sender |
| `nonInteractiveClaim` | server + emulator | no — the wallet holds neither key |
| `nonInteractiveRefund` | server + receiver + emulator | no — the wallet holds neither key |

This is the same set the `vhtlc` handler offers, for the same reasons, so the selection discipline is identical — and a test pins the two handlers rung-for-rung across every role and timelock context so they cannot drift. **The V1 handler is untouched.**

Unlike `vhtlc`, `vhtlc-v2` is **not generically spendable**. A live lockup is escrow: letting generic selection (send, settle, renewal, offboard) pick it would move it out from under a swap the counterparty is still entitled to complete. `vhtlc` answers `true` only as documented back-compat; a type with no deployed rows inherits no such obligation. Recovery is unaffected — every way out of a lockup names its outpoints explicitly.

### Push where it helps, poll because it must

`RfqSwapManagerDeps.contracts` is **optional**, mirroring `chain?: ChainSource`: a caller with no wallet, or one that only wants the timer, gets an unchanged manager. Supplying it (prefer `await wallet.getContractManager()`, as `createOffer` does) registers each lockup and subscribes to `onContractEvent`.

**The subscription is a latency optimization and never a source of truth.** An event names a script; the reaction is to run the ordinary pass for the swap at that script, and that pass re-reads the lockup through `readLockupFate` exactly as the timer's pass does. So an event that is missed, duplicated, reordered or outright **forged** can only cost or save latency — it can never change what the manager believes, and can never on its own cause a claim or a refund. `poll()` stays armed as the failsafe rather than being replaced: every deadline that moves money is an absolute timelock that passes whether or not an event ever arrives. A test pins this directly — a `vtxo_spent` over a lockup the indexer still reports as open leaves the swap exactly where it was.

Registration is best-effort in the same spirit: it buys latency and a contract row, it decides nothing, so a failure is reported and retried next pass and never aborts the pass it is part of. Two cases are refused outright instead, because no retry can fix them: a record with no covenant, and a covenant whose script is not the one that was funded (rows are keyed by script, so registering the wrong one would leave the real lockup unwatched while reporting success — the failure `registerOfferContract` already guards against).

A terminal swap's row is **retired, not deleted** — the row is what keeps the lockup's VTXOs annotatable and its history readable. Only a row that was actually written, so a swap whose registration was refused does not then report a phantom "not found" failure on top of having settled. `removeSwap` leaves the row alone: it says this manager stops driving the swap, not that the lockup is done.

## Correction: a swept lockup needs recovery, not a refund push

`064147a3` (earlier on this branch) merged the spendable and recoverable indexer reads so a swept lockup stops reading as "nothing to refund". **That part was right. Its justification was not:**

> "It is still the trader's money and `refundWithoutReceiver` still takes it back (that leaf needs the trader's own key plus the server, neither of which a sweep removes)."

The SDK's own predicates say otherwise, and they are mutually exclusive by construction (`wallet/vtxo.ts`):

```ts
canSpendOffchain  = !hasTerminalSpend && !(isSwept || isPastExpiry)
canRecoverOnchain = !hasTerminalSpend &&  (isSwept || isPastExpiry)
```

`canRecoverOnchain` is documented as "must be recovered into a fresh batch **rather than spent offchain**". A sweep removes the leaf from the live tree; holding the sender key does not help. `packages/boltz-swap` splits on exactly this rather than working around it — `settleRefundWithoutReceiver` routes a live VTXO through an offchain tx and a recoverable one through `joinBatch`, because "a swept (recoverable) VTXO is no longer a live leaf, so it can only be reclaimed by re-registering it into a batch".

That merge therefore had a second consequence nobody wanted: because `pushRefundWithoutReceiver` puts every input into **one** aggregate offchain transaction, a single swept output would take the live ones down with it, and `refundIfUnresolved` would then retry that doomed push until `attemptDeadline` — burning the very window the trader needed in order to recover the outputs and finish the refund properly.

Corrected, and made actionable rather than something a caller discovers as a server rejection:

- `pushRefundWithoutReceiver` refuses up front with a new `LockupNeedsRecoveryError`, naming the outpoints. The **whole** push, not a filtered subset — silently dropping the swept ones would report success over money that never moved.
- `refundIfUnresolved` returns a `needs_recovery` outcome instead of retrying. Unlike a median-time-past refusal this is not "not yet" but "not this way".
- `RfqSwapManager` reports the error **unwrapped**, so a caller can `instanceof` it, and deliberately *does* keep retrying — recovery is something the caller can perform while the window is still open, after which the next pass simply succeeds.

`findLockupVtxos` still returns both sets, still tagged: the money being visible is the point of the original fix. What is corrected is the claim that visible meant refundable.

## Recovering a swept lockup: registration is what makes it possible

A swept lockup cannot be spent offchain (see the correction above), so the question is what *can* take it back. The answer already exists in the SDK, and finding that took correcting a wrong turn worth recording.

**`IVtxoManager.recoverVtxos()` is the remedy.** It re-registers swept outputs into a fresh batch — the same round `packages/boltz-swap` reaches through its own `joinBatch` — and it reads the wallet's **registered-contract** snapshot:

```
recoverVtxos()
  → wallet.getVtxos({ withRecoverable: true })        (vtxo-manager.ts:1099)
    → contractSnapshot()                               (wallet.ts:1198)
      → contractManager.getContractsWithVtxos()        (wallet.ts:1315-1318)
```

Note this is the **ungated** `getVtxos`, not `getSpendableVtxos` — so `isGenericallySpendable: false` does not exclude the lockup here, though it still (correctly) excludes it from `renewVtxos`, which goes through `getSpendableVtxos` (`vtxo-manager.ts:1270`, gate at `wallet.ts:1210-1216`).

**But that chain was broken, by this PR, in a way nothing reported.** Registering the lockup persisted a row and watched it, while its VTXOs were silently never synced:

`deriveContractTapscripts` falls back to `script.forfeit()` for any handler that is not `TapscriptDeriving`, and **no VHTLC script version has a `forfeit()`** — `VtxoScript` does not define one and `VHTLC.BaseScript` does not add one. Annotation threw `legacy.forfeit is not a function`, `annotatableIn` caught it into an annotation failure, and `fetchContractVtxosBulk` filtered the lockup's VTXOs out before they were persisted. Symptoms: a permanently `degraded` sync state and a lockup reading as zero balance. The money path was never affected — `RfqSwapManager` reads `readLockupFate` from the indexer directly, never from the wallet repository — but registration was buying substantially less than advertised.

Fixed by implementing `deriveTapscripts` on the `vhtlc-v2` handler, which is exactly what the `TapscriptDeriving` capability exists for. The leaf is `refundWithoutReceiver` for both the forfeit and intent roles: the only collaborative path a wallet holding the sender key can satisfy, and the same conclusion `selectPath` reaches.

Reproduced before fixing and pinned two ways — directly on `deriveContractTapscripts`, and end-to-end through a real `ContractManager`, where removing the fix turns `expect(vtxos).toHaveLength(1)` into `[]`.

**So no batch machinery is duplicated into `packages/swap`.** Porting `joinBatch` + `createVHTLCBatchHandler` (~330 lines of security-critical batch state machine, a fourth copy in this monorepo) was scoped and then dropped once the real blocker turned out to be a missing 6-line annotation method.

Two caveats a caller must hold, both now stated on `LockupNeedsRecoveryError` rather than left to be discovered:

- **The wallet must hold the lockup's `sender` key**, since recovery settles through `refundWithoutReceiver`.
- **`refundLocktime` must have matured.** That leaf carries a CLTV which `recoverVtxos` does not check, and because it sweeps every recoverable output into one settlement, recovering early can fail the whole batch rather than just this output. `packages/boltz-swap` encodes the same rule as "pre-CLTV recoverable → skipped". Making `recoverVtxos` CLTV-aware is a real follow-up and is **not** in this PR.

**Renewal proper** (keeping a lockup alive by re-locking it to the *same* VHTLC script before expiry, rather than recovering it after) is still unbuilt, and still needs a per-contract batch handler along the lines of `createVHTLCBatchHandler`. `vhtlc-v2` answering `false` to `isGenericallySpendable` correctly keeps it out of generic `renewVtxos`, which would otherwise settle the lockup to the wallet's own address and break the swap.

## Unchanged from the original design

- **`nextOnchainAction` for the onchain corridor's claim window.** `classifyOnchainHtlc` reports `claimable` right up until median-time-past reaches `refundLocktime`, while `claimOnchainFill` refuses starting `ONCHAIN_CLAIM_MARGIN_SECONDS` before that point (broadcasting into the counterparty's live refund window risks losing the race with `P` already revealed). A manager driving straight off `OnchainHtlcPhase` would spend that entire margin throwing `claim_window_closed` and never fall back. `nextOnchainAction` applies the margin itself.
- **No manager-level retry backoff.** The poll interval already is the retry cadence and `REFUND_MTP_LAG_SECONDS` already is the deadline.
- **`readLockupFate`'s witness read and hash verification.** `vtxo_spent` says an output was spent, not by which leaf or with what preimage — the settled-vs-refunded distinction is only decidable by reading the spend witness. Registering contracts does not replace this and must not.
- **`findLockupVtxos` stays an authoritative indexer read**, not the contract manager's VTXO state. That would be the weaker answer: it serves the wallet repository, which a degraded sync hands back stale (`getSyncState()` reports `degraded` and returns cached rows rather than failing), and its height-based expiry test needs a chain tip this module does not have. It also still works standalone for callers using `refundIfUnresolved` without a manager.
- Resolution is still decided **before** the refund gate; the per-swap in-progress lock, the dirty/save flush, waiter-settles-after-persist ordering, the event model and the public `poll()` all stand.

## API

- `RfqSwapManagerDeps` is `{ indexer: LockupSpendIndexer; chain?: ChainSource; contracts?: SwapContractRegistry }`.
- `RfqSwap` records gain an optional `lockup: { script: VHTLC.ScriptV2; address: string }` — the covenant and the funded address, both things a caller already holds. Still a live record, not a serialization format.
- New in `packages/swap`: `SwapContractRegistry`, `RfqSwapLockup`, `SWAP_LOCKUP_CONTRACT_TYPE`/`_LABEL`/`_KIND`, `LockupNeedsRecoveryError`, and a `needs_recovery` member of `RefundOutcome`.
- New in `packages/ts-sdk`: `VHTLCV2ContractHandler`, `VHTLCV2ContractParams`, registered as `vhtlc-v2`. It implements `TapscriptDeriving`, without which its VTXOs cannot be annotated at all.

`@arkade-os/swap` is excluded from the release cycle (#689) and `RfqSwapManager` is new in this PR, so nothing published speaks the old shape. The `packages/ts-sdk` change is purely additive: a new handler type alongside the existing five, with the `vhtlc` handler untouched.

## Test plan

Revert-verified — every safety branch broken in turn, each producing exactly the predicted failure and nothing else, then restored and re-confirmed green:

| mutation | observed failure |
|---|---|
| `vhtlc-v2` reports `isGenericallySpendable: true` | `is never generically spendable` — expected true to be false |
| half-specified covenant leaf returns undefined instead of throwing | `refuses a half-specified covenant leaf by name` — expected [Function] to throw |
| `createScript` builds `VHTLC.Script` (V1) | 2 tests, incl. `cannot collide with a vhtlc row` — the two pkScripts became equal |
| handler offers `refund` to the sender | `never offers a leaf this wallet cannot sign` — the leaf appeared in the offered set |
| a `vtxo_spent` event sets state directly instead of running a pass | `re-reads the chain rather than believing the event` — expected 1 to be greater than 1 (the indexer was never asked) |
| drop the funded-script guard on registration | `refuses to register a covenant that is not the funded script` — a row was written for the wrong script |
| a registration failure aborts the pass | `still drives the swap when registration fails` — the refund never ran |
| `stop()` leaves the subscription open | `drops the subscription on stop()` — expected 1 to be 0 |
| never retire a terminal swap's row | `retires the contract once the swap is over` — nothing retired |
| retire on `removeSwap` too | `leaves the contract alone when a swap is merely removed` — a row was retired |
| retire a row that was never written | `does not try to retire a row it never managed to write` — a settled swap reported 2 failures instead of 1 |
| drop the swept-output guard in `pushRefundWithoutReceiver` | **4 tests**, incl. the push being submitted instead of refused |
| `refundIfUnresolved` retries instead of returning `needs_recovery` | `reports a swept lockup instead of retrying` — **test timed out at 5s** (the doomed-retry loop, demonstrated) |
| remove `deriveTapscripts` from the `vhtlc-v2` handler | 2 tests: `legacy.forfeit is not a function`, and end-to-end `expected [] to have a length of 1` — the lockup's VTXOs silently dropped |

One mutation did **not** behave as predicted and is reported rather than hidden: removing the `byLockupScript` cleanup on `removeSwap` left `stops reacting to events for a removed swap` green, because `pollSwap`'s own `monitored` check also catches it. Removing *both* does fail the test, so the guards are genuine defence-in-depth and the test is not vacuous — this is now stated in the code rather than implied.

Full gate, real exit codes:

| step | baseline (`064147a3`) | this branch |
|---|---|---|
| `pnpm lint` (3 pkgs) | pass | pass |
| `pnpm --filter @arkade-os/swap typecheck` | pass | pass |
| ts-sdk unit | 2368 passed, 1 skipped (155 files) | **2385 passed, 1 skipped (156 files)** |
| swap unit | 191 passed (13 files) | **212 passed (13 files)** |
| boltz-swap unit | 614 passed (23 files) | unchanged |

**0 failing at baseline, 0 failing now; +38 tests, no regressions.**

`packages/swap/test` and `packages/ts-sdk/test/**/*.test.ts` are excluded from their tsconfigs, so the new test files were typechecked separately and are clean. Two pre-existing type errors were found that way and are noted rather than silently absorbed: `test/contracts/helpers.ts`'s subscription mock (untouched, out of scope), and `refund.test.ts`'s `VTXOS` fixture, which `064147a3` broke by making `LockupVtxo.recoverable` required without updating it — fixed here, since the file was already being edited.


packages/swap/src/swapManager.ts
```diff
@@ -0,0 +1,764 @@
...
+    }
+
+    private async pollSwap(swap: RfqSwap): Promise<void> {
+        if (this.inProgress.has(swap.rfqId)) return;
+        if (!this.monitored.has(swap.rfqId)) return;
+        this.inProgress.add(swap.rfqId);
+        try {
+            await this.runPass(swap);
+        } finally {
+            this.inProgress.delete(swap.rfqId);
```
pietro909: `inProgress` is cleared before `saveSwap`, waiter settlement, and finalize. While a slow `saveSwap` is pending, another timer tick or public poll() can re-enter the same swap because it is still in monitored.

For a just-refunded swap, that can call `refundArkade` a second time before finalization. It can also create out-of-order saves.

Keep the lock until after save + waiter settlement + finalization.

packages/swap/src/swapManager.ts
```diff
@@ -0,0 +1,764 @@
...
+    }
+
+    private emitAction(swap: RfqSwap, action: RfqSwapActionName): void {
+        notify(this.actionExecutedListeners, (listener) => listener(swap, action));
+    }
+
+    private async save(swap: RfqSwap): Promise<void> {
+        if (!this.callbacks) return;
+        try {
+            await this.callbacks.saveSwap(swap);
```
pietro909, line 1010: save() swallows a saveSwap failure, and line 509 has already cleared the dirty flag, so the write is never retried. The pass then settles waiters and finalizes anyway: for a terminal swap, finalize removes it from monitored, so there is no next poll

storage stays pending forever while the caller observed "completed". That breaks the invariant stated at 510-512, and after a restart (which rebuilds from storage) the money-moving path runs again.

Keep the dirty flag until the save succeeds, and skip waiter settlement/finalization on failure so the next tick retries. 

 Note :the retry pass must skip runPass for an already-terminal swap otherwise it re-drives refundArkade, the same  double-call as the lock issue above.

PR #698, opened 2026-08-08
fix(vtxo-manager): price the intent fee in renewVtxos and recoverVtxos

### LINKED ISSUE #696: vtxo-manager: renewVtxos and recoverVtxos build zero-fee intents, rejected by any fee-charging operator
`IVtxoManager.renewVtxos()` and `recoverVtxos()` both build their settlement output as the gross sum of the inputs they selected, so the intent fee they imply (`inputs - outputs`) is always zero. Against an operator with a non-zero `intentFee` policy the server rejects the intent outright and the call can never succeed:

```
INTENT_INSUFFICIENT_FEE (31): got 0 min expected 3679
```

That is arkd v0.9.15 on the arkade-regtest default policy (`ARK_OFFCHAIN_INPUT_FEE="amount * 0.01"`), renewing a single 367,896-sat VTXO — 1% of it is 3,679. The practical effect is that a wallet's float is never renewed at all against such an operator, and the VTXOs eventually walk into expiry.

The SDK already handles this correctly everywhere else. `runPeriodicSettle` even names the failure:

```ts
// Fetch server intent-fee config so each input/output can be priced.
// Without this, settle sends `outputAmount = sum(inputs)` and the
// server rejects with INTENT_INSUFFICIENT_FEE whenever the operator
// charges non-zero intent fees.
```

Both it and the no-argument `Wallet.settle()` branch price every input and the output through `Estimator`. Only the two `VtxoManager` entry points were left gross:

- `renewVtxos` does `const totalAmount = vtxos.reduce((sum, vtxo) => sum + vtxo.value, 0)` and passes `amount: BigInt(totalAmount)` as its only output.
- `recoverVtxos` takes `totalAmount` straight from `getRecoverableWithSubdust` and passes it through unchanged.

Neither method takes a fee argument, so a caller cannot correct this from outside — the only workaround is to stop calling them and rebuild the selection on top of `Wallet.settle({ inputs, outputs })` with `Estimator`, which is what we ended up doing for renewal in a downstream service. That works for renewal because `getExpiringVtxos` is public. It does not work as well for `recoverVtxos`: `getRecoverableWithSubdust` is not exported, so the subdust-inclusion decision can't be reproduced from the public surface, and reimplementing it guesswork-style in front of an untimelocked all-or-nothing sweep isn't something we wanted to ship. So recovery is still broken for us.

Both are present on master (39e6212) as well as the released 0.4.57, so this isn't a recent regression.

The fix looks like the arithmetic `_settleImpl`'s no-params branch already does: price each input with `evalOffchainInput(toOffchainInputFeeParams(v))`, skip any input whose fee meets or exceeds its own value, subtract `evalOffchainOutput` from the total, and check dust against the net rather than the gross. Happy to open a PR if that's the shape you'd want, or to split the `recoverVtxos` half out if you'd rather handle the subdust path yourselves.

One smaller thing worth deciding alongside it: with a fee in play, a flat renewal threshold that exceeds the batch lifetime causes a renewal treadmill. On a stack with a 6144-second batch expiry against `DEFAULT_RENEWAL_CONFIG.thresholdMs` of 3 days, every renewal mints a coin that is instantly due again, so the float is re-settled on every pass and the fees eat it. Capping the effective threshold at half the coin's own batch lifetime avoids that (a freshly renewed coin always has more than half its life left) and is a no-op wherever the flat threshold is already the smaller number.

Kukks:
`IVtxoManager.renewVtxos()` and `recoverVtxos()` both build their settlement output as the gross sum of the inputs they selected, so the fee the intent implies — `sum(inputs) - sum(outputs)` — is always exactly zero. Against an operator with a non-zero `intentFee` policy the server rejects the intent outright and the call can never succeed. Fixes #696.

The arithmetic already existed a few hundred lines down in the same file. `runPeriodicSettle` prices every input and the output through `Estimator`, and its comment names this exact failure ("Without this, settle sends `outputAmount = sum(inputs)` and the server rejects with INTENT_INSUFFICIENT_FEE whenever the operator charges non-zero intent fees"). The no-argument `Wallet.settle()` branch does the same. Only these two entry points were left gross, and neither takes a fee argument, so a caller can't correct it from outside.

Both now price their inputs and output through two small shared helpers, `priceSettlementInputs` and `deductOffchainOutputFee`. Three things came along with that, because the fee deduction is what creates them:

An input whose own fee meets or exceeds its value is dropped — settling it takes more out of the output than it puts in. That filter runs *before* `capSettlementBatch` rather than after, so an uneconomic input can't occupy a slot a viable one behind it could have used; that's the hazard the comment in `runPeriodicSettle` already warns about ("skipping uneconomic inputs and continuing past the cap avoids an uneconomic prefix permanently starving valid VTXOs behind it"). `capSettlementBatch` itself is untouched, and its documented gross-value bound stays correct — the post-fee output is strictly smaller, so a batch that fits on gross still fits.

Dust is now judged on the net output rather than the gross input sum. `recoverVtxos` had no dust check at all, and `getRecoverableWithSubdust` decides subdust inclusion on gross value, so a batch could clear dust gross and land under it once the fees came off. Subdust still consolidates normally whenever each coin is worth more than its own fee, which is the case under any percentage policy.

The renewal throws keep the wording the `vtxo_received` subscription already classifies as benign, so a pass with nothing worth renewing stays quiet instead of turning into a reported failure.

One choice worth flagging. `deductOffchainOutputFee` evaluates the output fee on the pre-deduction subtotal, not on the amount finally committed — matching what the two working paths already do. Under a percentage output rate `r` that implies `r * subtotal` where the server asks only `r * subtotal * (1 - r)`, so it over-pays by `r^2 * subtotal` (99 sats on a 990,000-sat subtotal at 1%). Solving the fixed point would recover those satoshis, but the failure mode of getting it wrong is landing a satoshi *under* the minimum, which is the rejection this whole change exists to prevent — so I kept the conservative form and pinned the exact magnitude in a test rather than leaving it incidental. Happy to switch to an exact solve if you'd rather, though I'd suggest doing it to all four call sites at once.

I deliberately did not touch the `vtxoMaxAmount` ceiling checks. `_settleImpl` tests the projected post-output-fee amount while `runPeriodicSettle` and `capSettlementBatch` test the pre-fee running total; the latter two are conservative rather than wrong (they under-batch, never over-batch, so they can't cause a rejection), and tightening them makes settlements larger, which is the riskier direction. It's a real inconsistency and worth a follow-up, but not inside a fee fix — credit to `arkana-ai-bot` for spotting that `_settleImpl`'s version is the more correct one.

One note on scope, stated carefully because it's easy to overstate. Neither method applies a per-contract spendability gate: both select from `getVtxos({ withRecoverable: true })`, and `getExpiringAndRecoverableVtxos` pulls `canRecoverOnchain` coins into renewal explicitly, so an escrow output past its batch expiry is a candidate for renewal as well as recovery. On `master` today that is latent rather than live, at least for VHTLC: `VHTLCContractHandler` exposes only `createScript`, so `deriveContractTapscripts` (`wallet/utils.ts:83`) falls through to its `legacy.forfeit()` branch, and `VHTLC.Script` — which extends `VtxoScript` — implements no `forfeit`. Annotation throws, nothing in that chain catches it, and those VTXOs never reach the snapshot selection reads. The exposure arrives with the work that fixes annotation by adding `deriveTapscripts` for covenant handlers, which makes those VTXOs visible and selectable in the same change. So this PR neither creates nor closes that hazard; it's out of scope here and worth its own issue. Recovery's all-or-nothing sweep having no CLTV awareness is a separate matter and holds regardless.

Verified against a live arkade-regtest stack (arkd v0.9.15, `offchainInput: "amount * 0.01"`). Submitting the gross-output intent that `master` builds reproduces the bug exactly:

```
INTENT_INSUFFICIENT_FEE (31): got 0 min expected 990
```

and `renewVtxos()` on this branch is accepted, taking a 99,000-sat VTXO to 98,010 — a fee of exactly 990, the server's minimum to the satoshi. Incidentally, arkd's own `ark redeem-notes` CLI fails the same way on that stack, which is why the e2e `beforeEachFaucet` helper can't fund against a fee-charging operator; I funded through onchain boarding instead.

Eighteen unit tests cover zero fee (unchanged behaviour), flat and percentage input fees, flat and percentage output fees, both percentages together with the over-payment pinned exactly, an input whose fee exceeds its value, dust against the net, the subdust paths on recovery, and the preview agreeing with the sweep. Reverting the source while keeping the tests fails 12 of the original 14 — the two that still pass are precisely the zero-fee "unchanged behaviour" cases — and reverting the preview to the gross sum fails its two. Gate before and after: `lint`, all three `typecheck`s and `build` clean; ts-sdk unit went 2326 passed / 1 skipped to 2344 passed / 1 skipped (+18, no pre-existing test changed), `swap` 67 and `boltz-swap` 614 both unmoved.

Three things landed from @arkana-ai-bot's review. `getRecoverableBalance` still answered with the gross sum while `recoverVtxos` settles that minus fees, and both are reachable from the same service-worker call — so a UI previewing with one and executing with the other showed two different numbers under any non-zero policy. It now prices exactly as the sweep does, including `vtxoCount`. The batch caps are still not applied to the preview: they defer overflow to the next cycle rather than reducing what is recoverable, and matching them would mean re-running the subdust decision over a capped subset. That gap predates the fee pricing and is noted where it lives rather than fixed here. Worth naming the shape, because it is the same rule the covenant-spendability work applies elsewhere — a preview exists to say what the operation would do, so it has to agree with it; fees and spendability are two dimensions of one idea.

`priceSettlementInputs` was also being called twice per entry point, once to filter and once for the subtotal after capping, where the second call could never drop anything and just re-ran the fee evaluation to re-sum values it already had. It now returns each survivor's net and `subtotalOf` totals whichever survived, throwing on an input it has no net for rather than counting it as zero — silently skipping one understates the output, which overpays the operator out of the user's own funds. And the capping block's error named the size caps even when fee filtering alone emptied the batch with no cap binding; those want different responses from an operator, so they are now two messages.

Two more went on after that, and how they were found is the interesting part. I backported this fix to `feat/arkade-swap` as #701, because that is the branch the team works on and the one downstream pins, and asking @arkana-ai-bot to review the backport turned up a defect in *this* PR that its own review pass had not caught. Worth stating plainly: it took a second reviewer pass, on a second PR, to find a bug in the first one.

The defect is that this PR priced `recoverable` net while leaving `getRecoverableBalance`'s companion `subdust` field summing gross `value`. That puts the two on different bases, and for a wallet whose recoverable set is entirely subdust — precisely the wallet the field exists to describe — `subdust` could come back larger than `recoverable`. Note this is a defect the fee pricing *introduced*: before it, both fields were gross and therefore consistent. Subdust is still classified on gross value, which is what makes a coin subdust and what `getRecoverableWithSubdust` judged inclusion on, but it is now reported net. I totalled it through the existing `subtotalOf` rather than a `net.get(...) ?? 0n` lookup: a `?? 0n` fallback silently understates `subdust` if an unpriced input ever reaches it, and `subtotalOf` already throws on exactly that case for the same reason this PR gives elsewhere — on a money path an input silently counting as zero is the failure mode worth preventing. Everything in `payable` is priced by construction, so the throw is unreachable today; it is there so it stays unreachable, and the reasoning is in the commit message so it does not get "simplified" back.

That fix then needed its own claim narrowed, which is the part most likely to be got wrong later. My first comment called `subdust` a component of `recoverable`. It is not, quite: `subdust` is net of the per-input fees but **not** of the output fee, because that one is charged on the batch as a whole and does not decompose per coin. So `subdust <= recoverable` holds when no output fee is configured or the non-subdust coins cover it, but not in general. @arkana-ai-bot supplied a counter-example and I reproduced it rather than reasoning about it: two coins at 600 and 500 with a 1% input fee and a flat 200 output fee gives `subdust` 1089 against `recoverable` 889. Rather than clamp — which would invent a number that answers to nothing and would still mislead on mixed sets, where the non-subdust coins absorb the output fee — the public TSDoc now says what the field actually is, and a test pins the case so a future refactor cannot quietly break it in the other direction. A documented invariant with no test is one refactor away from being wrong.

Gate after those two commits, against a baseline recorded on this branch's previous head: `lint`, `build` and all three `typecheck`s clean; ts-sdk unit 2344 → 2347 passed / 1 skipped across the same 154 files, with no pre-existing test moved; `swap` 67 and `boltz-swap` 614 unmoved. Reverting the subdust pricing while keeping the tests fails all three new ones with `expected 1100n to be 1089n`, and leaves the thirteen pre-existing subdust tests passing, since those run at zero fee where gross and net coincide.

One note on what was actually reviewed, since a green check does not always mean a review happened. On this PR both are real: @arkana-ai-bot's passes caught the defects above, and CodeRabbit did post a review here. On #701 it did not — its check passes carrying *"Review skipped: reviews are disabled for this base branch"*, which renders identically to a completed review in the checks list. So the backport has had one reviewer, not two, and that is worth knowing before reading its green ticks as agreement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved wallet recovery and renewal calculations to account for settlement and output fees.
  * Automatically excludes uneconomic transaction outputs and prevents settlements that would fall below the minimum dust threshold.
  * Recovery balance reporting now reflects the net recoverable value after fees.
  * Recovery can handle fee-affordable amounts below the standard dust threshold.
  * Renewal clearly reports when no eligible outputs remain after fees.
  * Fee-aware previews now match final transaction amounts.
  * Zero-fee behavior remains unchanged.

### DIFF
--- a/packages/ts-sdk/src/wallet/vtxo-manager.ts
+++ b/packages/ts-sdk/src/wallet/vtxo-manager.ts
@@ -306,6 +306,73 @@ export function capSettlementBatch<T extends { value: number }>(
     return batch;
 }
 
+/**
+ * Price a settlement's VTXO inputs against the operator's intent-fee policy,
+ * dropping any input that cannot pay for itself.
+ *
+ * An intent's fee IS `sum(inputs) - sum(outputs)`, so a single-output settlement
+ * asking for the gross input sum offers exactly zero and the server rejects the
+ * whole intent with `INTENT_INSUFFICIENT_FEE` wherever the operator prices
+ * offchain inputs at all. `subtotal` is what the inputs are worth once each has
+ * paid its own way; the caller still owes {@link deductOffchainOutputFee} on top.
+ *
+ * An input whose fee meets or exceeds its value is skipped rather than settled —
+ * it would take more out of the output than it puts in. Callers filter with this
+ * BEFORE {@link capSettlementBatch} so an uneconomic input cannot occupy a slot
+ * that a viable one behind it could have used.
+ */
+function priceSettlementInputs<T extends NormalizedExtendedVirtualCoin>(
+    vtxos: T[],
+    estimator: Estimator,
+): { payable: T[]; subtotal: bigint } {
+    const payable: T[] = [];
+    let subtotal = 0n;
+    for (const vtxo of vtxos) {
+        const inputFee = estimator.evalOffchainInput(toOffchainInputFeeParams(vtxo));
+        if (inputFee.satoshis >= vtxo.value) {
+            continue;
+        }
+        payable.push(vtxo);
+        subtotal += BigInt(vtxo.value - inputFee.satoshis);
+    }
+    return { payable, subtotal };
+}
+
+/**
+ * Take the operator's offchain output fee off a settlement's single output.
+ *
+ * The fee is evaluated on `subtotal` — the amount before this deduction — rather
+ * than on the amount finally committed, matching what
+ * {@link VtxoManager.runPeriodicSettle} and the no-argument `Wallet.settle()`
+ * branch already do. Under a percentage rate `r` that implies a fee of
+ * `r * subtotal` where the server asks only `r * subtotal * (1 - r)`, i.e. an
+ * over-payment of `r^2 * subtotal`. Deliberate: paying over is always accepted,
+ * whereas solving the fixed point exactly to recover those satoshis risks
+ * landing one under the minimum — which is the rejection this pricing exists to
+ * avoid.
+ *
+ * Can return a value below dust, or below zero, when a flat output fee outweighs
+ * the inputs; every caller checks the result against the dust threshold.
+ */
+function deductOffchainOutputFee(
+    subtotal: bigint,
+    estimator: Estimator,
+    arkAddress: string,
+): bigint {
+    // Mirror the estimator's own early return rather than reaching it through
+    // `ArkAddress.decode`: with no output program the script is never read, and
+    // an operator that doesn't price outputs shouldn't make a decodable address
+    // a precondition for renewing or recovering.
+    if (!estimator.config.offchainOutput) {
+        return subtotal;
+    }
+    const outputFee = estimator.evalOffchainOutput({
+        amount: subtotal,
+        script: hex.encode(ArkAddress.decode(arkAddress).pkScript),
+    });
+    return subtotal - BigInt(outputFee.satoshis);
+}
+
 /** Default renewal threshold in seconds (3 days). */
 export const DEFAULT_THRESHOLD_SECONDS = 259_200;
 
@@ -1106,7 +1173,7 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
 
         // Filter recoverable virtual outputs and handle subdust logic
         const now = await fetchTimeHeight(this.wallet);
-        let { vtxosToRecover, totalAmount } = getRecoverableWithSubdust(allVtxos, dustAmount, now);
+        let { vtxosToRecover } = getRecoverableWithSubdust(allVtxos, dustAmount, now);
 
         if (vtxosToRecover.length === 0) {
             throw new Error("No recoverable VTXOs found");
@@ -1124,21 +1191,32 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
         // overflow is recovered next cycle.
         const info = await this.getInfoProvider()?.getInfo();
         const vtxoMaxAmount = info?.vtxoMaxAmount ?? -1n;
-        const capped = capSettlementBatch(byValueDescending(vtxosToRecover), vtxoMaxAmount);
+
+        // Price the batch before capping so a VTXO that cannot pay its own intent
+        // fee is dropped here rather than occupying a slot in the capped batch
+        // ahead of one that can. Subdust inputs survive this whenever they are
+        // still worth more than their own fee, which is what keeps consolidating
+        // them possible. `info` is undefined only when no ark provider is wired,
+        // which prices as zero and leaves the gross behaviour untouched.
+        const estimator = new Estimator(info?.fees.intentFee ?? {});
+        const payable = priceSettlementInputs(vtxosToRecover, estimator).payable;
+        const capped = capSettlementBatch(byValueDescending(payable), vtxoMaxAmount);
         if (capped.length < vtxosToRecover.length) {
             const recoverableCount = vtxosToRecover.length;
-            ({ vtxosToRecover, totalAmount } = getRecoverableWithSubdust(capped, dustAmount, now));
+            ({ vtxosToRecover } = getRecoverableWithSubdust(capped, dustAmount, now));
             if (vtxosToRecover.length === 0) {
                 // Recoverable VTXOs exist, but the highest-value subset that
-                // fits in one settlement stays below dust, so submitting it
-                // would be rejected and the next cycle would pick the same
-                // prefix. Distinct from the "none recoverable" case above so
-                // operators can tell a stuck-but-funded wallet from an empty
-                // one. Recovery resumes once the prefix accumulates dust.
+                // fits in one settlement — and can pay its own intent fee —
+                // stays below dust, so submitting it would be rejected and the
+                // next cycle would pick the same prefix. Distinct from the "none
+                // recoverable" case above so operators can tell a stuck-but-funded
+                // wallet from an empty one. Recovery resumes once the prefix
+                // accumulates dust.
                 throw new Error(
                     `Capped recovery batch (highest-value subset of ${recoverableCount} ` +
                         `recoverable VTXOs within the ${MAX_VTXOS_PER_SETTLEMENT}-input and ` +
-                        `${vtxoMaxAmount}-sat limits) is below the dust threshold ${dustAmount}`,
+                        `${vtxoMaxAmount}-sat limits, net of intent fees) is below the ` +
+                        `dust threshold ${dustAmount}`,
                 );
             }
         }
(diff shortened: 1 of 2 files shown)


Kukks:
Two commits added here that came out of reviewing the backport rather than this PR, and one of them fixes a real defect in what this PR was about to merge — so worth flagging before it lands.

I backported this onto `feat/arkade-swap` as #701, since that is the branch the team works on and the one downstream consumers pin. Asking @arkana-ai-bot to review the backport surfaced a bug in *this* PR that the review pass here had not caught: `getRecoverableBalance` prices `recoverable` net of intent fees, but its companion `subdust` field was still summing gross `value`. Different bases, and since `subdust` describes part of what `recoverable` reports, a wallet whose recoverable set is entirely subdust — exactly the wallet the field exists for — could report `subdust` larger than `recoverable`. This is a defect the fee pricing introduced; before it both fields were gross and consistent.

`f7528c1e` fixes it. Subdust stays classified on gross value, which is what makes a coin subdust and what `getRecoverableWithSubdust` judged inclusion on, and is now reported net. I used the existing `subtotalOf` rather than `net.get(...) ?? 0n` deliberately: a `?? 0n` fallback silently understates `subdust` if an unpriced input ever reaches it, and `subtotalOf` already throws on that case for the same reason this PR gives elsewhere — on a money path, an input quietly counting as zero is the failure worth preventing. It is unreachable today since everything in `payable` is priced by construction; it is there so it stays unreachable, and that reasoning is in the commit message so a future reader does not simplify it back.

`ab1e857d` then narrows a claim I got wrong in the first fix's comment, which is the part I would expect someone to trip over later. I called `subdust` a component of `recoverable`. It is not, quite — it is net of the per-input fees but not of the output fee, which is charged on the batch as a whole and does not decompose per coin. @arkana-ai-bot gave a counter-example and I reproduced it rather than reasoning about it: 600 and 500 at a 1% input fee with a flat 200 output fee gives `subdust` 1089 against `recoverable` 889. I did not clamp it — that would invent a number that answers to nothing, and would still mislead on mixed sets where the non-subdust coins absorb the output fee. The public TSDoc now states what the field actually is, and a test pins the case.

Both cherry-picked from #701 with no conflicts, which is consistent with the only divergence between the branches being `selectExpiringVtxos`'s accessor line — I checked that rather than assuming it, and confirmed the resulting file differs from #701's by exactly that one line plus the mock stubs.

Gate against a baseline recorded on this branch's previous head (`f80b47f8`): lint, build and all three typechecks clean; ts-sdk unit 2344 → 2347 passed / 1 skipped over the same 154 files, nothing pre-existing moved; `swap` 67 and `boltz-swap` 614 unmoved. Reverting the subdust pricing while keeping the tests fails all three new ones with `expected 1100n to be 1089n` and leaves the thirteen pre-existing subdust tests green, since those run at zero fee.

PR #701, opened 2026-08-08
fix(vtxo-manager): price the intent fee in renewVtxos and recoverVtxos

Kukks:
This is the `feat/arkade-swap` companion to #698, which carries the same fix onto `master`. Both target #696.

`IVtxoManager.renewVtxos()` and `recoverVtxos()` build their settlement output as the gross sum of the inputs they selected, so the fee the intent implies — `sum(inputs) - sum(outputs)` — is always exactly zero. Against an operator with a non-zero `intentFee` policy the server rejects the intent outright and neither call can ever succeed. The bug is present here in the same form it is on `master`: `packages/ts-sdk/src/wallet/vtxo-manager.ts` differs between the two branches by exactly one line, and it is not one of the lines that computes the amount.

I am opening this separately rather than waiting for #698 to merge because `feat/arkade-swap` is the branch the team works on day to day and the one downstream consumers pin, so a fix that only lands on `master` does not reach them for however long the two branches stay apart. #698 stays on `master` since that is where releases are cut.

The first two commits are #698's, unmodified. Both entry points now price their inputs and their output through two shared helpers, `priceSettlementInputs` and `deductOffchainOutputFee`, using the arithmetic that already existed a few hundred lines down the same file — `runPeriodicSettle` and the no-argument `Wallet.settle()` branch both do this, and `runPeriodicSettle`'s comment names this exact failure. Only these two entry points were left gross, and neither takes a fee argument, so a caller cannot correct it from outside. Three things come along with the deduction: an input whose own fee meets or exceeds its value is dropped (and dropped *before* `capSettlementBatch`, so it cannot occupy a slot a viable input behind it could have used); dust is judged on the net output rather than the gross input sum; and `getRecoverableBalance` prices identically to the sweep, so a UI that previews with one and executes with the other stops showing two different numbers. #698's description has the full derivation and I would rather not duplicate it here.

The one thing worth reading carefully on *this* branch is where the pricing lands relative to the gated selection accessor. `feat/arkade-swap` routes the renewal read through `getSpendableVtxos({ withRecoverable: true })` where `master` uses `getVtxos` — that is the single-line divergence, and it sits in `selectExpiringVtxos`. The pricing runs downstream of it: `renewVtxos` selects through the gate, revalidates, and only then prices whatever survived, which is the ordering you want, since pricing a coin the gate is going to reject is wasted work and the gate is the stricter filter. Recovery is unaffected — `recoverVtxos` and `getRecoverableBalance` both still read the ungated `getVtxos` on this branch exactly as they do on `master`, so the pricing lands in the identical place. Whether recovery *should* also be gated is a separate question that was raised and dropped in #700; nothing here changes that either way.

The last two commits came out of review and are **not** in #698, so they should be ported there too. `getRecoverableBalance` returns both a `recoverable` total and a `subdust` figure, and #698 priced the first net while leaving the second summing gross `value` — which puts them on different bases and, for a wallet whose recoverable set is entirely subdust, could report `subdust` larger than `recoverable`. Subdust is still *classified* on gross value, which is what makes a coin subdust and what `getRecoverableWithSubdust` judged inclusion on, but it is now *reported* net, totalled through the existing `subtotalOf` so an unpriced input throws rather than silently contributing zero.

That fix then needed its own claim narrowed, which is worth stating plainly because it is a real limit on the number: `subdust` is net of the per-input fees but *not* of the output fee, because that one is charged on the batch as a whole and does not decompose per coin. So `subdust <= recoverable` holds when no output fee is configured or the non-subdust coins cover it, but not in general — a flat output fee against a wholly subdust wallet still leaves `subdust` the larger. Rather than clamp it, which would invent a number that answers to nothing and would still mislead on mixed sets, the public TSDoc now says what the field actually is (what the subdust coins are worth once each has paid its own way) and a test pins the case. The same round moved a comment up to the recovery capping condition to say that it fires when *either* fee pricing or a size cap narrows the batch, since the comparison is against the pre-pricing count deliberately and reads as cap-only otherwise.

One tradeoff carried over from #698 and worth restating, because it costs the user satoshis. `deductOffchainOutputFee` evaluates the output fee on the pre-deduction subtotal rather than on the amount finally committed, matching what the two already-working paths do. Under a percentage output rate `r` that means paying `r * subtotal` where the server asks only `r * subtotal * (1 - r)`, an over-payment of `r² * subtotal` — 99 sats on a 990,000-sat subtotal at 1%. Solving the fixed point would recover those, but the failure mode of getting it wrong is landing a satoshi *under* the minimum, which is precisely the rejection this change exists to prevent, so I kept the conservative form and pinned the exact magnitude in a test rather than leaving it incidental. If you would rather have the exact solve, I would suggest doing it to all four call sites in one go instead of just these two.

One review suggestion I did not take: making the `getAddress()` call in `getRecoverableBalance` conditional on the operator actually pricing outputs. It is `this.arkAddress.encode()` — a local bech32m encode over an in-memory address, no I/O — and skipping it would mean testing `estimator.config.offchainOutput` at the call site, which duplicates `deductOffchainOutputFee`'s own early return and leaks its internals to every caller. The abstraction seems worth more than the encode.

On evidence, and I want to be precise about what I did and did not re-run. Both of #698's commits cherry-picked onto `feat/arkade-swap` with no conflicts, and the resulting `vtxo-manager.ts` differed from #698's head by exactly the `getSpendableVtxos` line above before the review commits went on top. Gate, measured on pristine `feat/arkade-swap` first and again after: `lint`, `build` and all three `typecheck`s clean on both sides; ts-sdk unit went 2396 passed / 1 skipped to 2417 passed / 1 skipped across the same 156 files, which is 18 new tests from #698 plus 3 from the review round, with no pre-existing test moved; `swap` held at 232 and `boltz-swap` at 614, both unmoved. Reverting the pricing to the gross output while keeping the tests fails 15 of those 18 — the three that still pass are exactly the zero-fee "unchanged behaviour" cases, one per entry point plus the preview — and the flagship failure reads `expected 8000n to be 7920n`, the gross sum offering a zero fee, which is the bug itself. Reverting just the subdust change fails its tests with `expected 1100n to be 1089n` while leaving all eleven pre-existing subdust tests passing, since those run at zero fee where gross and net coincide.

What I did not do is re-run the live regtest verification. #698 confirmed the fix against an arkade-regtest stack (arkd v0.9.15, `offchainInput: "amount * 0.01"`), reproducing `INTENT_INSUFFICIENT_FEE (31): got 0 min expected 990` on the gross intent and getting a 99,000-sat VTXO renewed to 98,010 — a fee of exactly 990, the server's minimum to the satoshi. The recovery path here is byte-identical to what was exercised there, so that result carries over. The renewal path is not quite: it reaches the same pricing code through `getSpendableVtxos` rather than `getVtxos`, so the selection read in front of the fix is one this branch covers by unit test rather than on a live fee-charging stack. This repo's own regtest integration jobs do run against this branch and pass, but their operator charges no intent fee, so they exercise the zero-fee no-op rather than the priced path.

Naming the limit of that squarely, because it is the thing most likely to be wrong and it becomes invisible once this is merged and the reasoning is gone: the unit tests mock `getSpendableVtxos` to return the same set as `getVtxos`. That is what lets them pass identically on both branches — but it also means they cannot detect a divergence between what the gate admits and what the pricing then sees. If the gated accessor ever returns a different set from the ungated one in a way that matters to fee pricing, nothing in this test suite would notice. What the backport proves is that the arithmetic is right and that it sits downstream of the gate; it does not prove the gate and the pricing agree on a real wallet.

One note on what was actually reviewed, since a green check does not always mean a review happened. @arkana-ai-bot's reviews on this PR are real, and are what caught the `subdust` defect and then the overstated invariant in my own first fix for it. CodeRabbit's check, by contrast, passes carrying *"Review skipped: reviews are disabled for this base branch"* — no review took place, and it renders identically to a completed one in the checks list. #698 did get a real CodeRabbit review, so the difference is this PR's base branch rather than anything about the change. Net effect: this PR has had one reviewer, not two.

### DIFF
--- a/packages/ts-sdk/src/wallet/vtxo-manager.ts
+++ b/packages/ts-sdk/src/wallet/vtxo-manager.ts
@@ -306,6 +306,73 @@ export function capSettlementBatch<T extends { value: number }>(
     return batch;
 }
 
+/**
+ * Price a settlement's VTXO inputs against the operator's intent-fee policy,
+ * dropping any input that cannot pay for itself.
+ *
+ * An intent's fee IS `sum(inputs) - sum(outputs)`, so a single-output settlement
+ * asking for the gross input sum offers exactly zero and the server rejects the
+ * whole intent with `INTENT_INSUFFICIENT_FEE` wherever the operator prices
+ * offchain inputs at all. `subtotal` is what the inputs are worth once each has
+ * paid its own way; the caller still owes {@link deductOffchainOutputFee} on top.
+ *
+ * An input whose fee meets or exceeds its value is skipped rather than settled —
+ * it would take more out of the output than it puts in. Callers filter with this
+ * BEFORE {@link capSettlementBatch} so an uneconomic input cannot occupy a slot
+ * that a viable one behind it could have used.
+ */
+function priceSettlementInputs<T extends NormalizedExtendedVirtualCoin>(
+    vtxos: T[],
+    estimator: Estimator,
+): { payable: T[]; subtotal: bigint } {
+    const payable: T[] = [];
+    let subtotal = 0n;
+    for (const vtxo of vtxos) {
+        const inputFee = estimator.evalOffchainInput(toOffchainInputFeeParams(vtxo));
+        if (inputFee.satoshis >= vtxo.value) {
+            continue;
+        }
+        payable.push(vtxo);
+        subtotal += BigInt(vtxo.value - inputFee.satoshis);
+    }
+    return { payable, subtotal };
+}
+
+/**
+ * Take the operator's offchain output fee off a settlement's single output.
+ *
+ * The fee is evaluated on `subtotal` — the amount before this deduction — rather
+ * than on the amount finally committed, matching what
+ * {@link VtxoManager.runPeriodicSettle} and the no-argument `Wallet.settle()`
+ * branch already do. Under a percentage rate `r` that implies a fee of
+ * `r * subtotal` where the server asks only `r * subtotal * (1 - r)`, i.e. an
+ * over-payment of `r^2 * subtotal`. Deliberate: paying over is always accepted,
+ * whereas solving the fixed point exactly to recover those satoshis risks
+ * landing one under the minimum — which is the rejection this pricing exists to
+ * avoid.
+ *
+ * Can return a value below dust, or below zero, when a flat output fee outweighs
+ * the inputs; every caller checks the result against the dust threshold.
+ */
+function deductOffchainOutputFee(
+    subtotal: bigint,
+    estimator: Estimator,
+    arkAddress: string,
+): bigint {
+    // Mirror the estimator's own early return rather than reaching it through
+    // `ArkAddress.decode`: with no output program the script is never read, and
+    // an operator that doesn't price outputs shouldn't make a decodable address
+    // a precondition for renewing or recovering.
+    if (!estimator.config.offchainOutput) {
+        return subtotal;
+    }
+    const outputFee = estimator.evalOffchainOutput({
+        amount: subtotal,
+        script: hex.encode(ArkAddress.decode(arkAddress).pkScript),
+    });
+    return subtotal - BigInt(outputFee.satoshis);
+}
+
 /** Default renewal threshold in seconds (3 days). */
 export const DEFAULT_THRESHOLD_SECONDS = 259_200;
 
@@ -1106,7 +1173,7 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
 
         // Filter recoverable virtual outputs and handle subdust logic
         const now = await fetchTimeHeight(this.wallet);
-        let { vtxosToRecover, totalAmount } = getRecoverableWithSubdust(allVtxos, dustAmount, now);
+        let { vtxosToRecover } = getRecoverableWithSubdust(allVtxos, dustAmount, now);
 
         if (vtxosToRecover.length === 0) {
             throw new Error("No recoverable VTXOs found");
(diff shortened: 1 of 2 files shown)


Kukks:
Thanks — the `subdust`/`recoverable` finding is right and I've fixed it in 9a4ac6c6.

Confirmed it by reading the code rather than taking it on trust: `recoverable` comes out of `deductOffchainOutputFee(subtotalOf(payable, net), ...)` while `subdustAmount` was reducing over `BigInt(v.value)`. Since `subdust` reports a component of `recoverable`, that put the two on different bases, and for a wallet whose recoverable set is *entirely* subdust — which is exactly the wallet the field exists to describe — `subdust` came back larger than `recoverable`. Worth noting this is a defect in the change itself rather than a pre-existing one: before the pricing both fields were gross and therefore consistent.

I kept the classification on gross value, since that is what makes a coin subdust and what `getRecoverableWithSubdust` judged inclusion on, and changed only the reporting to net. One deviation from your suggested patch: I totalled it through the existing `subtotalOf` instead of `net.get(...) ?? 0n`. A `?? 0n` fallback would silently understate `subdust` if an unpriced input ever reached it, and `subtotalOf` already throws on exactly that case for the same reason — an input silently counting as zero is the failure mode the helper was written to prevent. Everything in `payable` is priced by construction, so the throw is unreachable today; it is there so it stays unreachable.

Two tests pin it: an all-subdust recoverable set where `subdust` and `recoverable` must agree exactly, and a mixed set where `subdust` is the net contribution of just the subdust coins. Reverting to the gross sum fails both with `expected 1100n to be 1089n`, and leaves the eleven pre-existing subdust tests green, since those run at zero fee where gross and net coincide.

On the capping condition: agreed it is not a bug, and I'd rather not restructure the control flow inside a fee fix. I moved the explanation up to the condition itself so the "either filter fired" reading is visible before someone tries to simplify it, rather than three lines further in next to `cappedAway`.

On the unconditional `getAddress()`, I'm going to leave it. It resolves to `this.arkAddress.encode()` — a local bech32m encode over an in-memory address, no I/O — and avoiding it means testing `estimator.config.offchainOutput` at the call site, which duplicates `deductOffchainOutputFee`'s own early return and leaks its internals to every caller. Happy to be argued out of that if you think the encode is worth the coupling.

Gate after the fix, against a baseline recorded on pristine `feat/arkade-swap`: lint, build and all three typechecks clean; ts-sdk unit 2396 → 2416 passed / 1 skipped over the same 156 files (18 from the backport, 2 for this fix, nothing pre-existing moved); `swap` 232 and `boltz-swap` 614 both unmoved.

One thing to flag beyond this PR: #698 on `master` carries the identical `subdust` defect — I checked its head commit directly and the gross reduce is there verbatim. This fix should be ported across before that one merges.

Kukks:
Good catch, and you're right that the comment overstated it. Fixed in 8afcbb66.

I reproduced your counter-example before changing anything rather than taking the arithmetic on trust: two coins at 600 and 500, 1% input fee, flat 200 output fee gives `subdust` 1089 against `recoverable` 889, exactly as you had it. So "a component of `recoverable`" was wrong — `subdust` nets off the per-input fees but not the output fee, which is charged on the batch as a whole and genuinely does not decompose per coin.

I took option 2. Clamping would invent a number that answers to nothing, and as you note it would still mislead on mixed sets where the non-subdust coins absorb the output fee — a clamped `subdust` would look precise while being arbitrary. The honest reading is what the subdust coins are worth once each has paid its own way, and that is now what it says.

Two things beyond the comment you flagged. I put the explanation on the public method's TSDoc as well as in the body, since `subdust` is what a UI renders and the caller is the one who needs to know it is not a slice of the total; that doc now also states that both figures are net of intent fees and that the batch caps are not applied to the preview. And I pinned the edge case with a test (`does not attribute the batch-level output fee to subdust`) rather than leaving it as prose — you were right that neither of the previous two exercised an output fee, and a documented invariant with no test is one refactor away from being silently wrong in the other direction.

Gate: lint, build and all three typechecks clean; ts-sdk unit 2417 passed / 1 skipped over 156 files, up from the 2396 baseline on pristine `feat/arkade-swap` — 18 from the backport plus 3 from this review round, nothing pre-existing moved. `swap` 232 and `boltz-swap` 614 unmoved.

PR #709, opened 2026-08-10
feat(swap): name the refund this wallet cannot make

pietro909:
Second of the two PRs in `plans/arkade-swap-leftover-5-6.md` (item 6). Independent of
#708 — different files, and its own retirement is `RfqSwapManager.retireContract`, which
is already on this base — so the two can merge in either order. Reviewing #708 first is
still the intended order: it establishes "a state with money behind it stays watched",
which this one has to not violate.

## What

`RfqSwapState` gains `needs_counterparty`, with `blockedReason` beside `failure` on the
record. A swap enters it when no local refund is possible at all:

- `refundArkade` throws the new `RefundNotLocallyPossibleError` — caught *before* the
  retry branch, so it costs one call instead of ~1440 passes of `onSwapFailed` ending in
  `failed`, a label that would claim an action failed when none was ever possible.
- Nothing is wired to act (`enableAutoActions: false`, or no callbacks) and the window
  has passed. This is the loudest gap and needs no key material to see: the swap used to
  sit `pending` forever — monitored, never acted on, never reported.
- The new optional `canRefundArkade` probe refuses. It runs every pass, including before
  the window opens, so a swap nobody can refund says so while the solver can still act.

**Not terminal, and not a dead end.** `RFQ_SWAP_TERMINAL_STATES` is unchanged, so the
lockup contract stays watched and a later solver claim still arrives as `settled`; the
probe answering `ok` returns the swap to `pending` and resumes the normal drive.

## Composing the three causes

The causes reached callers in three shapes — an `undefined` return, two plain throws —
across two functions, with nothing composing record → identity. Added
`senderIdentityForSwapRecord(wallet, record)`, which `refundArkade` should wire to:

- `no-secrets` (the `undefined`, which a lower-level wiring would skip — it becomes a
  `TypeError` at the push site, which the manager would then treat as retryable)
- `unreadable-secrets`
- `foreign-descriptor` — now thrown typed by `senderIdentityForRfqSecrets` itself, at the
  site that discovers it.

`rfqSecretsOfRecord` keeps its documented totality: it is called while iterating swap
history, where a throw on one record aborts the loop. Its `undefined`-return tests are
unmodified, and so are the two `rejects.toThrow(/…/)` message assertions the typed error
now satisfies as a subclass.

## Two interactions worth a look in review

- A refusal the *push* reported is not re-issued every pass (`refundRefused`) — it cannot
  start working on its own, and re-issuing it is the grind this state removes. Only the
  probe retracts it; without a probe the swap learns again on the next process start,
  when the wallet that can sign may have been restored.
- The L1 half is a different key: before the window an onchain-send swap's live claim
  keeps the label; after it the refusal wins, and `setOnchainState` stops the
  re-asserted-from-chain `claimed` from flipping a blocked swap every pass. The claim
  itself always runs.

## Tests

`test/swapManager.test.ts` (8 new, incl. the item-5 interaction asserted from this side:
`setContractWatchState` is never called for a blocked swap) and `test/secrets.test.ts`
(one per cause, asserting the error *and* its `reason`). Mutation-checked: adding
`needs_counterparty` to the terminal set, and removing the typed-error branch, each fail
4 tests.

- `pnpm -r test:unit` — green (swap: 292).
- No e2e: a filled RFQ swap needs a solver counterparty, which is the standing coverage
  gap for this corridor. The unit tests are the coverage.

## Summary by CodeRabbit

- **New Features**
  - Swaps that cannot currently be refunded locally now show a `needs_counterparty` status with a clear blocking reason.
  - Refund capability is rechecked automatically, allowing swaps to return to normal processing when recovery becomes possible.
  - Monitoring and on-chain claim tracking continue while a refund is blocked.

- **Bug Fixes**
  - Prevented repeated failed refund attempts and misleading terminal failures when automatic refunds are unavailable.
  - Improved handling of missing, unreadable, or unsupported refund information.

### DIFF
--- a/packages/swap/README.md
+++ b/packages/swap/README.md
@@ -381,7 +381,10 @@ The package is pre-release; these notes replace a changelog for consumers tracki
   `senderPrivateKey` is gone from both return types; caller-owned onchain preimages live inside
   `secrets` and must be persisted with the record. `pushRefundWithoutReceiver` /
   `refundIfUnresolved` take `sender: Identity` instead of `senderPrivateKey: Uint8Array` — build
-  it with `senderIdentityForRfqSecrets`. `AssetSwap` gains `signingDescriptor?`,
+  it from the record with `senderIdentityForSwapRecord`, which is what keeps a wallet that cannot
+  sign reporting `RefundNotLocallyPossibleError` rather than a `TypeError` at the push site;
+  `senderIdentityForRfqSecrets` is for callers that already hold resolved secrets. `AssetSwap`
+  gains `signingDescriptor?`,
   `preimageHex?`, and complete stored-arm `fallbackSecrets?`. Landed while the package is
   unpublished and consumer-free, which is the whole window for doing it: after a consumer ships,
   the same change becomes a secret migration across every deployed wallet.
--- a/packages/swap/README.md
+++ b/packages/swap/README.md
@@ -328,9 +328,21 @@ await saveSwap({ ...record, ...rfqSecretsToRecord(swap.secrets) });
 const secrets = rfqSecretsOfRecord(record);
 if (secrets) {
     const preimage = await preimageForRfqSecrets(wallet, secrets);
-    const sender = await senderIdentityForRfqSecrets(wallet, secrets);
 }
+
+// For a refund, take the composition instead of the guard: it turns all three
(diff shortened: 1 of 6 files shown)


pietro909:
@coderabbitai verify your comments against commits 3711be and 0463802, they have been all addressed.

Issue #712, opened 2026-08-10
swap: no client API for the lightning:BTC → arkade:BTC receive route


tiero:
`@arkade-os/swap` covers three of the four cross-corridor directions. Lightning **receive** has no client-side entry point on any branch.

Surfaced while documenting the route: the Arkade Intents docs cannot show working code for it, so the integration page currently has readers assemble the request from the wire shape by hand.

## What exists

On `feat/arkade-swap`, `packages/swap/src/rfq.ts`:

| | Pair constant | Request builder | Orchestrator |
|---|---|---|---|
| `arkade:BTC → lightning:BTC` | `LIGHTNING_SEND_PAIR` (:114) | `lightningSendRequest` | `requestLightningSend` |
| `arkade:BTC → onchain:BTC` | `ONCHAIN_SEND_PAIR` (:116) | `onchainSendRequest` | `requestOnchainSend` |
| `onchain:BTC → arkade:BTC` | `ONCHAIN_RECEIVE_PAIR` (:118) | `onchainReceiveRequest` (:827) | — |
| **`lightning:BTC → arkade:BTC`** | **none** | **none** | **none** |

The constants stop at `LIGHTNING_BTC` (:107) — there is no `LIGHTNING_RECEIVE_PAIR`.

Notably the *hard* part is already shipped: `sealClaimPacket` in `claimPacket.ts` produces the `claim_packet` this profile needs, and `secrets.ts` has `newPreimage` / `derivePreimage` / `paymentHashOf`. What's missing is the request, the quote handling, and the lockup derivation for the direction where **the solver funds and the client claims**.

## What the solver already serves

The reference RFQ service implements this profile today, and its e2e drives a receive `rfq_request` over a real relay. The wire contract:

- `amount` is **required** — unlike the BOLT11 send profile, nothing else implies it, because the solver mints the invoice
- `profile`: `payment_hash`, `payout_address`, `payout_pubkey`, `claim_packet`
- quote returns `from_amount`, `to_amount`, `solver_pubkey`, `valid_until`, `refund_locktime`

So this is a client-side gap only.

## Why the claim path matters here

The send direction ends with the solver claiming. Receive ends with **the client claiming**, and that claim is what reveals the preimage the solver needs to settle Lightning — so an unclaimed receive strands both sides until the timeouts fire.

[...]

PR #717, opened 2026-08-11
feat(swap): gate the receive claim on the funded lockup value

pietro909:
PR 2 of the issue #712 receive stack (`plans/issue-712-arkade-swap-receive.md`). Stacked on #716 — retarget to `feat/arkade-swap` when that merges.

Deriving the lockup script locally is what protects every other corridor, and it protects nothing here: the solver can fund the correctly derived script with dust, and the claim publishes `P` regardless, which is what lets it settle the payer's Lightning HTLC in full.

- `pushClaim` takes a **required** `expectedAmount` and refuses with `LockupAmountMismatchError` (carrying expected/locked) when the live outputs sum below it. Required rather than optional so the records that most need the guard cannot skip it.
- Summed across every live output — funding in several is legitimate; `>=`, so overfunding passes. The existing swept-output refusal still runs first, which is what leaves this a plain sum over live outputs.
- The gate runs **before signing**: `P` rides the PSBT to the Ark server at submit, so a check that waits for the transaction to land has already leaked the secret.
- Skipped only when `partiallyClaimed` is set — this lockup already carries a claim of ours, so `P` is public and blocking the remainder would only strand our own money.
- `requestLightningReceive` / `requestOnchainReceive` return `expectedAmount = quote.to_amount`, documented must-persist: captured at claim time instead, the comparand would be whatever the solver funded.

Together with #716 this closes the value chain — `invoice.amountSats === from_amount` (what the payer is asked for), `to_amount === requestedAmount` (already shipped as `assertQuotedAmount`), `Σ funded >= to_amount` (here).

### DIFF
--- a/packages/swap/README.md
+++ b/packages/swap/README.md
@@ -321,7 +321,13 @@ window that can run out is the hold invoice's, and the claim window is measured 
 which is the deadline to show a payer, not `valid_until`. The optional `maxPayAmount` caps `from_amount`
 (`price_too_high`). The trader-side
 completion lands in `claim.ts`: `claimReceiveLockup` waits for the solver's funding and pushes the
-collaborative claim with the swap's own `P` and receiver key (covclaimd optional). Until covclaimd's
+collaborative claim with the swap's own `P` and receiver key (covclaimd optional). Both request
+flows return `expectedAmount` (the quote's `to_amount`) — persist it: `pushClaim` requires it and
+refuses, with `LockupAmountMismatchError`, to publish `P` for a lockup funded below it. Matching the
+`pkScript` is not enough on this leg, since a solver that funds the correctly derived script with
+dust still settles the payer's HTLC in full once `P` is out. The gate sums every live output and
+runs before signing — `P` reaches the Ark server at submit — and is skipped only for a lockup we
+have already partially claimed (`partiallyClaimed`), where `P` is public anyway. Until covclaimd's
 reference vectors are cross-checked, the `sealClaimPacket` test vector is pinned from this
 implementation and marked provisional (`TODO(claim-packet-vectors)`).
 
--- a/packages/swap/src/claim.ts
+++ b/packages/swap/src/claim.ts
@@ -62,6 +62,8 @@ export type ClaimArkProvider = RefundArkProvider;
  * HTLC in full.
  */
 export class LockupAmountMismatchError extends Error {
+    readonly name = "LockupAmountMismatchError";
+    readonly reason = "amount_mismatch";
     readonly expectedAmount: number;
     readonly lockedAmount: number;
     constructor(expectedAmount: number, lockedAmount: number) {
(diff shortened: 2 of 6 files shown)


pietro909:
Thanks — follow-ups pushed in `7dd16089`. Notes on what I took, what I reshaped, and the two I'm leaving.

**#2 (`expectedAmount === 0`) — taken, with a wider guard.** The framing that convinced me isn't "zero shouldn't happen" but that `undefined` and `NaN` fail the same way: `locked < undefined` and `locked < NaN` are both false, so any unusable number leaves the gate a silent no-op — precisely the failure mode "required rather than optional" was meant to rule out, and precisely what a JS caller or a record rehydrated from storage without the field would hand it. `assertExpectedAmount` now rejects anything non-finite or `<= 0`. It sits behind `partiallyClaimed` alongside the gate, so a record written before the field existed can still release its remainder — the bypass exists to avoid stranding our own money, and it shouldn't newly depend on a field that record never carried.

**#4 (piecemeal funding) — taken, but as a shape fix rather than a test.** Adding a test would have documented the problem instead of removing it. `claimReceiveLockup` *owns* the poll loop, so a transient shortfall is something it can wait out; handing back a `LockupAmountMismatchError` made a transient state indistinguishable at the call site from a dust-funded lockup, and a caller treating it as terminal abandons a swap that would have settled a poll later — on the onchain leg, that means walking down the L1 refund path for nothing. `awaitLockupFunding` now takes `minValue` and waits for the sum; only a lockup still short at the `deadline` throws, where the shortfall is attributable to the solver rather than the indexer. Swept outputs still return immediately, since what restores that sum is recovery, not more funding. Covered by tests for the wait, the deadline-short case, and the swept early return.

**#1 (name collision) — not taking.** The observation is right, and it is the only duplicated error name across the two packages. But the suggested remedy would break this package's own convention: every error in `packages/swap` extends plain `Error` (`LockupNeedsRecoveryError`, `RefundNotLocallyPossibleError`, `SwapRefusal`, `AddressMismatch`). The `SwapError`/options-object shape is `boltz-swap`'s, and the two packages are independent integrations that deliberately don't share an error base — adopting it here would import a hierarchy this package doesn't otherwise have. On the practical risk: nothing in either package dispatches on `.name` for these, `instanceof` works as you note, and an import alias resolves the collision for anyone consuming both. The names describe the same condition in both places, which seems like the right outcome rather than something to disambiguate.

**#3 (`partiallyClaimed` is caller-asserted) — agreed on substance, deferring the doc.** The flag has no callers yet: `swapManager` has no receive wiring at all, so there is currently nothing whose bookkeeping the note would constrain. The obligation you're describing is real and load-bearing, and it belongs next to the code that decides to set the flag — I'd rather write it there, in the wiring PR, than state it here where it can drift from whatever that code actually does. Say the word if you'd prefer it on the parameter now as well.

PR #719, opened 2026-08-11
feat(sdk): reuse the core offchain-tx primitive for the receive claim

pietro909:
PR 3 of the issue #712 receive stack (`plans/issue-712-arkade-swap-receive.md`). Stacked on #717; retarget as the stack merges.

`claim.ts` hand-rolled build → sign → submit → match → finalize because the core primitive that already does it was never exported.

`packages/ts-sdk`:
- export `signAndSubmitOffchainTx`
- add `claimWithPreimageIdentity(identity, preimage)` — the condition-witness-after-signing decorator, generic so it returns the identity type it was given
- opt-in `verifyServerSignatures: { serverPubkey }` on `submitOffchainTx`, forwarded by `signAndSubmitOffchainTx`. Checked before `finalizeTx`, per input, and — unlike the boltz version this generalizes — against the leaf the **local** build spends rather than one assumed-shared leaf read off the server's own copy. A checkpoint signed on another leaf of the same contract keeps its txid, so `matchServerCheckpoints` accepts it; only the local-leaf comparison catches it. Fails closed when `finalArkTx` is absent and verification was requested. Compressed or x-only server key both accepted.
- Default-off: omitting the option leaves every existing caller, including the wallet send path, bit-identical.

`packages/swap`: `pushClaim` now calls the primitive with `claimWithPreimageIdentity` and `verifyServerSignatures`, taking the server key off the covenant it already holds (`script.options.server`) rather than adding a parameter. The value gate from #717 stays exactly where it was, before signing.

What the check buys, precisely: it does **not** protect `P`, which reaches the server at submit, and it does not stand between us and a false terminal state. It converts "report `claimed`, watch nothing land, learn the truth when the solver's refund fires hours later" into an immediate failure.

`boltz-swap` is untouched — it is in production and this buys its users nothing on its own; folding its duplicate onto the export is a separate follow-up.

### DIFF
--- a/packages/swap/src/claim.ts
+++ b/packages/swap/src/claim.ts
@@ -11,9 +11,10 @@
  *
  * - The claim leaf carries a **preimage condition**, so the preimage rides
  *   the PSBT's `ConditionWitness` field — attached AFTER signing, on the ark
- *   transaction and every checkpoint (the condition is not part of the
- *   signed payload; attaching it first invalidates the signature the server
- *   then rejects as `INVALID_SIGNATURE`).
+ *   transaction and every checkpoint (`claimWithPreimageIdentity` in core owns
+ *   that ordering; the condition is not part of the signed payload, and
+ *   attaching it first invalidates the signature the server then rejects as
+ *   `INVALID_SIGNATURE`).
  * - The leaf is NOT covenant-pinned, so one aggregate output pays the whole
  *   balance wherever the trader says — the `nonInteractiveClaim` pin is the
  *   offline path's, not this one's. The destination is therefore a required
@@ -24,19 +25,15 @@
  * `secrets` (`senderIdentityForRfqSecrets`), a trader that is online never
  * depends on it.
  */
-import { base64, hex } from "@scure/base";
+import { hex } from "@scure/base";
 import { ripemd160 } from "@noble/hashes/legacy.js";
 import { sha256 } from "@noble/hashes/sha2.js";
 import {
     CSVMultisigTapscript,
-    ConditionWitness,
     type Identity,
-    Transaction,
     type VHTLC,
-    assertSubmittedArkTxid,
-    buildOffchainTx,
-    matchServerCheckpoints,
-    setArkPsbtField,
+    claimWithPreimageIdentity,
+    signAndSubmitOffchainTx,
 } from "@arkade-os/sdk";
 
 import {
@@ -98,30 +95,6 @@ const assertFiniteAmount = (value: number, reason: string, label: string): void 
 };
 
 /**
- * A signer that reveals a preimage when spending a condition leaf.
- *
- * The ordering encoded here is the whole point: the condition witness is NOT
- * part of what is signed, so attaching it before signing leaves a signature
- * over a PSBT that no longer matches once the field is present. Decorate per
- * claim, never wallet-wide.
- */
-const claimIdentity = (identity: Identity, preimage: Uint8Array): Identity => ({
-    ...identity,
-    sign: async (tx: Transaction, inputIndexes?: number[]): Promise<Transaction> => {
-        // Clone-and-round-trip so the caller's transaction is never mutated
-        // and the signed result is a fresh object we can add a field to.
-        const signed = Transaction.fromPSBT(
-            (await identity.sign(tx.clone(), inputIndexes)).toPSBT(),
-        );
-        const indexes = inputIndexes ?? Array.from({ length: signed.inputsLength }, (_, i) => i);
-        for (const index of indexes) {
-            setArkPsbtField(signed, index, ConditionWitness, [preimage]);
-        }
-        return signed;
-    },
-});
-
-/**
  * Build, sign, and push the collaborative claim of a receive-corridor lockup:
  * move every funded output at the lockup to the trader's own destination,
  * revealing `P` in the witness — which is also what settles the trader's side
@@ -141,6 +114,11 @@ const claimIdentity = (identity: Identity, preimage: Uint8Array): Identity => ({
  * single non-live input would take the live ones down with it. That refusal
  * comes first, which is what leaves the value gate a plain sum over live
  * outputs.
+ *
+ * The server's countersignature is verified before finalizing, per input and
+ * against that input's own leaf. It does not protect `P` — that reached the
+ * server at submit — but it turns "reported claimed, nothing landed, the
+ * solver refunds hours later" into an immediate failure.
  */
 export async function pushClaim(
     ark: ClaimArkProvider,
@@ -208,10 +186,15 @@ export async function pushClaim(
 
     const leaf = input.script.claim();
     const tapTree = input.script.encode();
-    const signer = claimIdentity(input.receiver, input.preimage);
 
-    const { arkTx, checkpoints } = buildOffchainTx(
-        input.vtxos.map((vtxo) => ({
+    // The core primitive owns build → sign → submit → match → finalize; this
+    // module supplies only what is swap-specific: the claim leaf, the preimage
+    // signer, and the server key to check the response against — the covenant
+    // already carries it, so it is not a caller obligation.
+    const arkTxid = await signAndSubmitOffchainTx({
+        identity: claimWithPreimageIdentity(input.receiver, input.preimage),
+        provider: ark,
+        inputs: input.vtxos.map((vtxo) => ({
             txid: vtxo.txid,
             vout: vtxo.vout,
             value: vtxo.value,
@@ -220,29 +203,11 @@ export async function pushClaim(
         })),
         // One aggregate output: unlike the covenant refund, this leaf inspects
         // nothing about the output set.
-        [{ script: input.destinationPkScript, amount: BigInt(locked) }],
+        outputs: [{ script: input.destinationPkScript, amount: BigInt(locked) }],
         serverUnrollScript,
-    );
-
-    // No index list: every input spends the same claim leaf, so all are
-    // signed — and the claim identity attaches `P` after signing.
-    const signedArkTx = await signer.sign(arkTx);
-    const submitted = await ark.submitTx(
-        base64.encode(signedArkTx.toPSBT()),
-        checkpoints.map((c) => base64.encode(c.toPSBT())),
-    );
-    assertSubmittedArkTxid(submitted, signedArkTx, "claim");
-
-    // Only checkpoints we built ourselves get signed — the server's response
-    // is matched against the local set first — and the preimage rides the
-    // checkpoint signatures exactly as it rides the ark transaction's.
-    const matched = matchServerCheckpoints(submitted.signedCheckpointTxs, checkpoints, "claim");
-    const finalCheckpoints = await Promise.all(
-        matched.map(async ({ server }) => base64.encode((await signer.sign(server, [0])).toPSBT())),
-    );
-
-    await ark.finalizeTx(submitted.arkTxid, finalCheckpoints);
-    return { arkTxid: submitted.arkTxid, amount: locked };
+        verifyServerSignatures: { serverPubkey: input.script.options.server },
+    });
+    return { arkTxid, amount: locked };
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/ts-sdk/src/utils/arkTransaction.ts
```diff
@@ -525,6 +525,61 @@ export function assertSubmittedArkTxid(
...
+): void {
+    const leaf = localTx.getInput(inputIndex).tapLeafScript?.[0];
+    if (!leaf) {
+        throw new Error(
+            `${context}: input ${inputIndex} carries no spend leaf to verify the server signature against`,
+        );
+    }
+    try {
+        verifyTapscriptSignatures(
+            serverTx,
```
pietro909, line 568: **assertServerSignedLeaf verifies the server's signature against the SERVER's own copy of the transaction, so the sighash inputs that the txid does not commit to (witnessUtxo script/amount) are attacker-chosen and the check is trivially bypassable.**

`verifyTapscriptSignatures(serverTx, ...)` collects `prevoutScripts`/`prevoutAmounts` from `serverTx.getInput(i).witnessUtxo` and computes `serverTx.preimageWitnessV1(...)` from them. Only the leaf script is pinned to the local build; `witnessUtxo` is a PSBT field that `assertSubmittedArkTxid`'s txid comparison does not cover. I confirmed this empirically in the review worktree: a fake arkd that rebuilds `finalArkTx` and each checkpoint with the same txid but `witnessUtxo.amount + 1_000_000n` and signs THAT (also dropping the user's signature entirely) passes both loops and `finalizeTx` is called — `signAndSubmitOffchainTx` resolves with the arkTxid. User-visible consequence on the path this PR enables it for (`pushClaim`): a buggy or hostile arkd returns a countersignature that is worthless against the real VTXO amounts, `pushClaim` returns `{arkTxid, amount}`, the trader's app records the receive swap as claimed, nothing ever confirms, and the solver reclaims the lockup at `refund_locktime` — exactly the 'reported claimed, nothing landed, the solver refunds hours later' outcome the doc comment on claim.ts:117-121 says this check converts into an immediate failure. Fix direction: take prevout scripts/amounts (and ideally the whole verification target) from `localTx`, the way the leaf already is.

<sub>CONFIRMED · automated high-effort review</sub>

packages/swap/src/claim.ts
```diff
@@ -208,10 +186,15 @@ export async function pushClaim(
...
     const tapTree = input.script.encode();
-    const signer = claimIdentity(input.receiver, input.preimage);
 
-    const { arkTx, checkpoints } = buildOffchainTx(
-        input.vtxos.map((vtxo) => ({
+    // The core primitive owns build → sign → submit → match → finalize; this
+    // module supplies only what is swap-specific: the claim leaf, the preimage
+    // signer, and the server key to check the response against — the covenant
+    // already carries it, so it is not a caller obligation.
+    const arkTxid = await signAndSubmitOffchainTx({
```
pietro909, line 194: **The claim's value gate is only guarded against non-finite `expectedAmount`, never against `expectedAmount <= 0`, so a zero agreed amount silently deletes the gate and `P` is published for a dust-funded lockup.**

`assertFiniteAmount` (claim.ts:166, added in the base commit 07f5ba8d and therefore outside this PR's hunks — anchored here at line 194, the call that actually publishes `P`) accepts 0 and negatives: `Number.isFinite(0) === true`. The gate is then `if (locked < input.expectedAmount)`, and `locked` is always >= 1 for a funded lockup, so `locked < 0` is never true and `LockupAmountMismatchError` can never fire. `expectedAmount` is `quote.to_amount`, and nothing forces it positive: `assertQuotedAmount` (rfq.ts:1189) only rejects `to_amount > from_amount` and only pins the side the caller named, so with `amountSide: "from"` a solver may legally quote `to_amount: 0`; `assertReceivable`'s only price ceiling is the OPTIONAL `maxPayAmount`, which caps `from_amount`, not the rate. The same value arrives as 0 from any persisted record whose missing numeric column was migration-filled with 0 — the exact "record predates the field" case the guard's own doc says "must not proceed". Concrete consequence: trader accepts a `to_amount: 0` quote (or restores such a record), pays the solver's hold invoice for `from_amount`, the solver funds the lockup with 1 sat, `pushClaim` sails past the gate and hands the preimage to the Ark server at submit, and the solver settles the payer's full Lightning HTLC with `P` — precisely the dust-funding attack `LockupAmountMismatchError` exists to stop.

<sub>CONFIRMED · automated high-effort review</sub>

packages/ts-sdk/src/utils/arkTransaction.ts
```diff
@@ -684,6 +783,41 @@ export async function signAndSubmitOffchainTx(params: {
...
+ *
+ * For any condition-leaf spend (VHTLC claims on both swap directions); the
+ * generic signature keeps the decorated identity's own type, so the result is
+ * still whatever was passed in.
+ */
+export function claimWithPreimageIdentity<
+    T extends { sign(tx: Transaction, inputIndexes?: number[]): Promise<Transaction> },
+>(identity: T, preimage: Uint8Array): T {
+    return {
+        ...identity,
```
pietro909, line 808: **`claimWithPreimageIdentity` builds the decorated signer with object spread, which drops prototype methods and copies own-property `signMultiple` undecorated — yet the generic signature promises the caller gets its own type back.**

`{ ...identity }` copies only own enumerable properties. `SingleKey` (identity/singleKey.ts:28) declares `sign`, `xOnlyPublicKey`, `compressedPublicKey`, `signerSession`, `signMessage` and `toHex` as class methods on the prototype and holds `private key` as an instance field, so the returned object is `{ key: <raw private key bytes>, sign: <wrapper> }` — the private key escapes onto a plain object, and every other `Identity` member is gone. The function is now an exported core API whose doc claims "the generic signature keeps the decorated identity's own type, so the result is still whatever was passed in", and `T` is returned unchanged, so TypeScript reports no error. A consumer writing the documented "any condition-leaf spend" flow — `const claimer = claimWithPreimageIdentity(wallet.identity, preimage); const pk = await claimer.xOnlyPublicKey();` — gets `TypeError: claimer.xOnlyPublicKey is not a function` at runtime. Sharper still for batch-signing wallets: `isBatchSignableIdentity` (identity/index.ts:70) tests `"signMultiple" in identity`, and an identity that carries `signMultiple` as an own property (the natural shape for an injected browser-wallet adapter) has it copied across UNDECORATED. Any core path that takes that fast path signs the ark tx and checkpoints without ever attaching `ConditionWitness`, so the preimage never reaches the leaf and arkd rejects the spend as unsatisfiable after the full network round-trip.

<sub>CONFIRMED · automated high-effort review</sub>

packages/ts-sdk/src/utils/arkTransaction.ts
```diff
@@ -631,6 +691,43 @@ export async function submitOffchainTx(
     // been matched to a local one.
     const matched = matchServerCheckpoints(signedCheckpointTxs, offchainTx.checkpoints, "submitTx");
 
+    const verify = options?.verifyServerSignatures;
+    if (verify) {
+        // Fail closed: `finalArkTx` is optional in the wire type and
+        // `assertSubmittedArkTxid` skips it when absent, but a check that was
+        // asked for and cannot be run is a failed check, not a passed one.
+        if (response.finalArkTx === undefined) {
```
pietro909, line 699: **The new fail-closed gate deletes the deliberate tolerance for an absent `finalArkTx` and converts it into an unrecoverable abort between `submitTx` (where `P` is disclosed) and `finalizeTx`.**

`OffchainTxSubmitProvider` declares `finalArkTx?: string` (line 413) and `assertSubmittedArkTxid` explicitly documents and implements the tolerance — `if (response.finalArkTx === undefined) return;` (line 519). The old `pushClaim` never touched the field at all. Now any provider implementation that omits it — the interface permits it, and only `RestArkProvider` is known to forward `data.finalArkTx` — makes `pushClaim` throw `ServerResponseMismatchError` *after* `submitTx` has already handed `P` to the Ark server and *before* `finalizeTx` is ever called. The claim never lands; the trader retries and hits the identical deterministic failure every time (nothing about the response changes), so the lockup stays unspent until `refund_locktime`, at which point the solver reclaims it — while already holding the public `P` needed to settle the payer's Lightning HTLC in full. The trader loses the swapped amount to a well-behaved-but-terse server, and `packages/swap/src/claim.ts:208` opts every receive claim into this unconditionally.

<sub>PLAUSIBLE · automated high-effort review</sub>

packages/swap/src/claim.ts
```diff
@@ -220,29 +203,11 @@ export async function pushClaim(
...
-    // is matched against the local set first — and the preimage rides the
-    // checkpoint signatures exactly as it rides the ark transaction's.
-    const matched = matchServerCheckpoints(submitted.signedCheckpointTxs, checkpoints, "claim");
-    const finalCheckpoints = await Promise.all(
-        matched.map(async ({ server }) => base64.encode((await signer.sign(server, [0])).toPSBT())),
-    );
-
-    await ark.finalizeTx(submitted.arkTxid, finalCheckpoints);
-    return { arkTxid: submitted.arkTxid, amount: locked };
+        verifyServerSignatures: { serverPubkey: input.script.options.server },
```
pietro909, line 208: **pushClaim now has post-submit failure modes that claimReceiveLockup cannot retry, because findLockupVtxos returns nothing once the lockup is spent.**

Every throw from the newly enabled `verifyServerSignatures` fires inside `submitOffchainTx` after `provider.submitTx` returned, i.e. after `P` is disclosed and the VTXOs are consumed. `claimReceiveLockup` (claim.ts:253) is the documented one-call composition and retries through `awaitLockupFunding` → `findLockupVtxos` (packages/swap/src/refund.ts:264), which queries the indexer with `spendableOnly` and `recoverableOnly` only — a spent lockup yields an empty array. So a caller that retries after a verification failure polls every `pollMs` until `deadline` and then gets `Error('the lockup never appeared at the covenant script')` with `reason: 'lockup_timeout'`, a terminal and actively misleading error for a claim whose sats already landed at `destinationPkScript`. Contrast the pre-submit refusals (`LockupAmountMismatchError`, `LockupNeedsRecoveryError`), which claim.ts:246-251 explicitly documents as safe to retry — the new check breaks that invariant without the doc or the retry path acknowledging it.

<sub>PLAUSIBLE · automated high-effort review</sub>

pietro909 (commented):
High-effort automated review of this PR. The lead item is that the new server-cosignature check verifies against the server's own copy of the transaction, which makes it bypassable — details inline.

packages/ts-sdk/src/utils/arkTransaction.ts
```diff
@@ -684,6 +783,41 @@ export async function signAndSubmitOffchainTx(params: {
...
+    return {
+        ...identity,
+        sign: async (tx: Transaction, inputIndexes?: number[]): Promise<Transaction> => {
+            // Clone-and-round-trip so the caller's transaction is never mutated
+            // and the signed result is a fresh object we can add a field to.
+            const signed = Transaction.fromPSBT(
+                (await identity.sign(tx.clone(), inputIndexes)).toPSBT(),
+            );
+            const indexes =
+                inputIndexes ?? Array.from({ length: signed.inputsLength }, (_, i) => i);
```
pietro909, line 816: **With `inputIndexes` omitted the decorator stamps the raw preimage onto EVERY input's `ConditionWitness`, which the newly generalized core doc invites for mixed-input spends.**

`indexes = inputIndexes ?? Array.from({ length: signed.inputsLength }, (_, i) => i)` attaches the preimage to all inputs, and `signAndSubmitOffchainTx`'s `signArkTx` calls `params.identity.sign(arkTx)` with no index list (arkTransaction.ts:782). That was safe while this was `claim.ts`-private and every input spent the same VHTLC claim leaf, but the promoted doc now advertises it "for any condition-leaf spend", and the old warning "decorate per claim" was softened to "per spend". A caller that follows the doc and builds one offchain tx mixing a VHTLC claim input with an ordinary VTXO (for example to sweep the rest of a wallet in the same transaction) gets `ConditionWitness = [preimage]` written onto the ordinary input too. arkd rejects a condition witness on a leaf that carries no condition, so the whole aggregate transaction fails — and because the rejection happens at `submitTx`, the preimage has already been disclosed to the server on a transaction that will never land, which for the receive corridor is the one ordering the module is built to avoid.

<sub>PLAUSIBLE · automated high-effort review</sub>

PR #720, opened 2026-08-11
feat(swap): make lockup registration retryable on its own

pietro909:
Stacked on #719.

`LockupRegistrationFailed` gains `readonly script` beside `address`, so a caller can retry `registerLockupContract(contracts, script, address)` without re-quoting. Additive; applies to all four corridors.

On the receive legs retrying the request cannot work: it re-quotes, and the solver refuses a second live quote on the same payment hash. The error's class comment and `requestLightningReceive`'s doc comment now state the layered recovery — retry registration, else start over with a fresh preimage and `rfq_id`, and never publish the abandoned invoice, since nothing watches that lockup.

`requestLightningReceive`'s doc also carries the other two obligations: persist `secrets` and `expectedAmount` before publishing the invoice, and stay online to claim (covclaimd cannot claim this covenant today).

Tests: registration failure surfaces both `address` and the covenant that derives it, and returns no invoice; two successive `requestLightningReceive` calls on one wallet derive different payment hashes.

PR 4 of the plan for #712.

### DIFF
--- a/packages/swap/src/rfq.ts
+++ b/packages/swap/src/rfq.ts
@@ -1446,15 +1446,26 @@ export function deriveLightningReceive(input: {
  * ({@link claimReceiveLockup} in `claim.ts`), or via covclaimd if it is
  * offline.
  *
- * The same obligations as {@link requestOnchainSend}: persist `secrets`
- * BEFORE paying — the preimage and the payout key re-derive from it (or, on a
- * non-HD wallet, are carried by it).
+ * Three obligations, all of them before the invoice is handed to a payer:
+ *
+ * 1. Persist `secrets` and `expectedAmount`. The preimage and the payout key
+ *    re-derive from `secrets` (or, on a non-HD wallet, are carried by it), and
+ *    without `expectedAmount` the claim has nothing to compare the funded
+ *    value against.
+ * 2. Stay online. covclaimd cannot claim this covenant today, so the offline
+ *    path the claim packet exists for does not run yet: an unclaimed lockup is
+ *    reclaimed by the solver at `refund_locktime` and the payer refunded.
+ * 3. On {@link LockupRegistrationFailed}, call this function again once the
+ *    store is working. The failed attempt is inert — it returned no invoice,
+ *    so nothing can be paid into the lockup nobody is watching — and the new
+ *    call derives its own preimage and `rfq_id`. Re-registering `error.script`
+ *    does NOT resume it: the invoice and `secrets` were never handed back.
  *
  * The invoice is the solver's, so it is verified here against the trader's own
  * `H` and the quote ({@link verifyReceiveInvoice}) before it is returned —
- * nothing publishable comes back from a failed check. Pay before
- * `invoiceExpiresAt`: the hold-invoice window is minutes, not the quote's
- * `valid_until`.
+ * nothing publishable comes back from a failed check, registration included.
+ * Pay before `invoiceExpiresAt`: the hold-invoice window is minutes, not the
+ * quote's `valid_until`.
  */
 export async function requestLightningReceive(
     wallet: IWallet,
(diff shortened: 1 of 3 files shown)


packages/swap/src/rfq.ts
```diff
@@ -1446,15 +1446,26 @@ export function deriveLightningReceive(input: {
...
+ * Three obligations, all of them before the invoice is handed to a payer:
+ *
+ * 1. Persist `secrets` and `expectedAmount`. The preimage and the payout key
+ *    re-derive from `secrets` (or, on a non-HD wallet, are carried by it), and
+ *    without `expectedAmount` the claim has nothing to compare the funded
+ *    value against.
+ * 2. Stay online. covclaimd cannot claim this covenant today, so the offline
+ *    path the claim packet exists for does not run yet: an unclaimed lockup is
+ *    reclaimed by the solver at `refund_locktime` and the payer refunded.
+ * 3. On {@link LockupRegistrationFailed}, call this function again once the
```
pietro909, line 1458: **The newly documented `LockupRegistrationFailed` retry contract misses contract-manager initialization failure, because `await wallet.getContractManager()` is evaluated outside `registerLockupContract`'s try/catch.**

The new obligation 3 tells callers to key their recovery on `LockupRegistrationFailed`. But every call site evaluates the writer as an argument: `await registerLockupContract(await wallet.getContractManager(), derived.script, derived.address)` (rfq.ts:1570; identically at 810, 1164, 1778 — the true anchor is line 1570, outside this PR's hunks, so this is anchored to the doc line that makes the promise). `Wallet.getContractManager()` (ts-sdk wallet.ts:1963) lazily runs `initializeContractManager()`, which reads the repositories, and it rethrows on failure (it explicitly clears `_contractManagerInitializing` "so subsequent calls can retry"). That rejection propagates before `registerLockupContract` is ever entered, so the try/catch that produces `LockupRegistrationFailed` never runs. Consequence: when the trader's local store is down — the single most likely instance of the "your own store failed" class this error was invented to name, and one that is equally inert and equally safe to retry — `requestLightningReceive` throws a bare `Error`. An integrator that implemented exactly the documented handler falls into its generic catch, where the safe advice is "unknown state, do not retry": the swap is abandoned instead of retried, or worse the caller misreads it as a quote-level refusal and walks away from a corridor that was fine.

<sub>CONFIRMED · automated high-effort review</sub>

packages/swap/src/rfq.ts
```diff
@@ -1446,15 +1446,26 @@ export function deriveLightningReceive(input: {
...
+ *    store is working. The failed attempt is inert — it returned no invoice,
+ *    so nothing can be paid into the lockup nobody is watching — and the new
+ *    call derives its own preimage and `rfq_id`. Re-registering `error.script`
```
pietro909, line 1461: **Obligation 3 claims a retry "derives its own preimage and `rfq_id`", but `rfqId` is a caller-supplied parameter, so the natural retry reuses the old id with a fresh payment hash.**

`requestLightningReceive` starts with `const rfqId = params.rfqId ?? newRfqId()` (rfq.ts:1489, just below this doc block). The documented recovery is "call this function again", and the natural way to do that is to re-invoke with the same `params` object. A caller that supplies `params.rfqId` — the parameter exists precisely so an integrator can correlate a quote with its own record id — therefore re-quotes under the SAME `rfq_id` while `deriveSwapSecrets`/`preimageForRfqSecrets` produce a NEW preimage and a new `H` (which the companion test at rfqReceive.test.ts asserts). The solver, keyed on `rfq_id`, returns its already-issued quote and hold invoice, which are bound to the OLD payment hash; `verifyReceiveInvoice` then throws a payment-hash mismatch. Consequence: the caller loops on the documented recovery and never succeeds — every retry fails on an invoice-vs-`H` mismatch that reads like a hostile solver — until it happens to clear `rfqId`, which nothing in the doc tells it to do.

<sub>CONFIRMED · automated high-effort review</sub>

pietro909 (commented):
High-effort automated review of this PR. Both items concern the newly documented `LockupRegistrationFailed` retry contract versus what the code does — details inline.

PR #723, opened 2026-08-11
feat(swap): let the trader register its own claim packet, and prove the sealing

Kukks:
`sealClaimPacket` seals `P` to covclaimd so a trader can go offline after funding — but the packet could only *reach* covclaimd by being handed to the solver in the `rfq_request`. So a trader's ability to go offline depended on the solver having wired a covclaimd, which the trader can neither see nor verify. The failure is silent in the worst way: a solver that accepts the packet and never forwards it is indistinguishable from one that forwarded it to a covclaimd that never claimed.

covclaimd's reveal endpoint is a **registration**, not a claim instruction — it stores the packet against the swap's script and claims when it sees the funding transaction. So a trader can register **before** the solver funds, which is exactly the window in which a trader who intends to go offline is still online.

Registering here *and* the solver revealing is belt and braces, not a double spend: covclaimd keys registrations by pkScript, so the second replaces the first, and a claim already made leaves nothing to spend.

## The provider is lifted, not rewritten

From `packages/boltz-swap/src/covclaimd-provider.ts` (#613), which already had the better shape — bytes at the boundary rather than hex strings, and the pubkeys cached so a swap doesn't refetch them. Only the docs are new.

**If #613 lands, these two copies should become one.** It currently lives in `boltz-swap` on an unmerged branch, which is why this is a copy rather than an import.

## The valuable half is the test

`claimPacket.ts` carried a standing TODO:

> byte-exactness against covclaimd's reference implementation is pinned here only by our own generated vectors — confirm against covclaimd's before production use

Self-consistency and "covclaimd can open this" are different claims, and only the second matters — a packet covclaimd cannot decrypt fails at the moment a trader has gone offline trusting it.

So the TODO is closed by a live test, not more vectors. Seal with this package, derive the taptree with this package, register with a **real covclaimd**, assert acceptance. covclaimd decrypts *before* it registers, so a 200 means the ECIES construction, the wire layout, the preimage-to-condition binding and the taptree derivation are all correct **together**.

The negative control is asserted alongside it, because acceptance proves nothing if a packet sealed to the wrong key is accepted too:

| case | result |
| --- | --- |
| sealed by this package, taptree derived by this package | **accepted** |
| sealed to the wrong key | **refused** |

Verified against live `covclaimd:v0.0.1-rc.3` + arkd. The test skips loudly (not silently) when neither is reachable — a silent skip would let the TODO reopen without anyone noticing.

## Known limit, documented rather than fixed

covclaimd's registry is in memory, capped at 10,000, expiring after 15 minutes. So a covclaimd **restart drops every pending registration silently**, and a swap funded more than the TTL after registering is never claimed.

The durable fix is to carry the packet **on the lockup transaction**, so covclaimd rebuilds intent from the chain rather than from memory. That needs a covclaimd change — `Match()` currently consults only the in-memory registry and `Filter()` is empty — and is tracked separately. This PR documents the limit in the module rather than pretending it isn't there.

None of this is load-bearing for correctness: the trader holds the covenant's `receiver` key and can always claim the lockup itself (`claim.ts`). covclaimd is the convenience for going offline.

## Test plan

- `test/e2e/covclaimd.test.ts` against a live stack — **2/2 pass**
- Package suite — **22 files / 341 tests**
- `pnpm typecheck` clean, `pnpm lint` clean

### DIFF
--- a/packages/swap/src/claimPacket.ts
+++ b/packages/swap/src/claimPacket.ts
@@ -8,9 +8,13 @@
  * - AES-256-GCM with the ephemeral pubkey as additional data;
  * - wire layout `ephPub(33) ‖ nonce(12) ‖ ciphertext`, base64.
  *
- * TODO(claim-packet-vectors): byte-exactness against covclaimd's reference
- * implementation is pinned here only by our own generated vectors — confirm
- * against covclaimd's before production use (see the package README).
+ * Byte-exactness against covclaimd is pinned by `test/e2e/covclaimd.test.ts`,
+ * which seals with this function and registers the result with a LIVE
+ * covclaimd. Acceptance is the assertion: covclaimd decrypts before it will
+ * register, so a wrong key dies at the AEAD tag with a 400 — and that negative
+ * case is asserted too, because otherwise acceptance would prove nothing. The
+ * unit vectors below remain as a fast self-consistency check; they were never
+ * evidence that covclaimd can open what this produces.
  */
 import { base64 } from "@scure/base";
 import { secp256k1 } from "@noble/curves/secp256k1.js";
--- a/packages/swap/src/covclaimd.ts
+++ b/packages/swap/src/covclaimd.ts
@@ -1,0 +1,93 @@
+/**
+ * covclaimd, from the TRADER's side.
+ *
+ * Lifted from `packages/boltz-swap/src/covclaimd-provider.ts` (arkade-os/ts-sdk#613)
+ * rather than written again: that implementation already had the shape this one
+ * wants — bytes rather than hex strings at the boundary, and the pubkeys cached
+ * so a swap does not re-fetch them per call. Only the docs below are new. If
+ * #613 lands, these two should become one module rather than two copies.
+ *
+ * WHY THE TRADER CALLS THIS AT ALL. `claimPacket.ts` seals `P` to covclaimd so
+ * a trader can go offline after funding, but until now the packet reached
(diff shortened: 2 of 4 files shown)


Kukks:
Closing: the trader shouldn't need to register at all.

covclaimd already reads the sealed packet **off the funding transaction** — `pkg/preimage/plugin.go` parses the extension from the tx, and `packet.go` defines the format (`PacketType 0x04`, TLV ciphertext + arkadeScript). The correct flow is:

```
client seals P, hands the packet to the solver
solver funds the HTLC and commits the packet in the ark extension
covclaimd detects it on-chain, decrypts, claims via the emulator non-interactive path
```

No registration, no HTTP call, no TTL. This PR had the trader calling `/v1/reveal` itself to route around the solver not attaching it — solving the symptom rather than the cause. The fix belongs in the solver, and is being built there.

**Two things here are worth salvaging rather than losing:**

1. **The live sealing verification.** `claimPacket.ts` carried a TODO admitting byte-exactness against covclaimd was pinned only by vectors this package generated for itself — self-consistency, which says nothing about whether covclaimd can open what it produces. The test in this branch closes that against a real covclaimd, with the negative control (wrong key → refused) that makes acceptance mean something. That result holds regardless of who delivers the packet, and it should be reopened on its own.

2. **The observation about #613.** `CovclaimdProvider` lives in `packages/boltz-swap` on an unmerged branch. Whatever path wins, it shouldn't end up duplicated across packages.

Branch kept at `feat/swap-covclaimd-reveal` so the test can be lifted out.

PR #724, opened 2026-08-11
feat(sdk): let sendBitcoin carry extension packets

Kukks:
The two send paths had drifted into a false choice:

| API | choose inputs | commit an extension |
| --- | --- | --- |
| `send(Recipient)` | ❌ | ✅ `extensions?: [{type, payload}]` |
| `sendBitcoin({ selectedVtxos })` | ✅ | ❌ |

A caller needing both had to give one up — and giving up input selection isn't a real option for anything spending contract-gated coins. So this is a missing half rather than a preference.

## The case that surfaced it

A swap solver funding an HTLC lockup must do two things at once: pick coins whose batch outlives the swap (otherwise it funds a lockup the counterparty can't claim and the service may not renew), and commit the sealed claim packet that lets covclaimd claim on the receiver's behalf.

covclaimd already reads that packet **off the funding transaction** — `pkg/preimage/plugin.go` parses the extension, `packet.go` defines the format. But the solver couldn't put it there, so the packet had to travel out of band via covclaimd's HTTP registration instead. That path keeps registrations **in memory**, capped at 10,000, expiring after 15 minutes — so a covclaimd restart silently drops every pending claim.

An on-chain registration is recoverable by anyone reading the transaction. That's the difference this unlocks.

## The change

`extensions` — the same shape `Recipient` already accepts — threaded into both `sendBitcoin` paths: the `selectedVtxos` branch builds the OP_RETURN output the way the recipient path does, and the generic fallback forwards to `send`.

**One ordering detail worth review attention.** `changeVout` is now read *before* the extension output is appended. It indexes the change output as `outputs.length - 1`, so appending the extension first would have made it point at the OP_RETURN instead — silently retargeting change metadata. The reason is in the code so it doesn't get "tidied" back.

## Test plan

- `pnpm test:unit` — **158 files / 2480 tests, 0 failures**
- `pnpm typecheck` clean

No unit test is added here, and I'd rather say why than pad it: nothing in the suite currently covers `extensions` on *any* send path, so there's no pattern to extend, and a mocked `sendBitcoin` would mostly assert the mock. The real verification is the consumer — a solver funding a lockup with the packet committed, and covclaimd claiming from what it reads on chain. That runs on regtest as the next step in this chain, and proves this change end to end rather than in isolation.

### DIFF
--- a/packages/ts-sdk/src/wallet/index.ts
+++ b/packages/ts-sdk/src/wallet/index.ts
@@ -397,6 +397,29 @@ export interface SendBitcoinParams {
      * @see IReadonlyWallet.getSpendableVtxos
      */
     selectedVtxos?: ExtendedVirtualCoin[];
+
+    /**
+     * Extension packets to commit in the transaction's OP_RETURN, the same
+     * shape `Recipient.extensions` takes on {@link IWallet.send}.
+     *
+     * Here because the two send paths had drifted apart: `send` could carry an
+     * extension but not choose its inputs, and `sendBitcoin({ selectedVtxos })`
+     * could choose its inputs but not carry an extension — so a caller that
+     * needed both had to give one up. Giving up input selection is not a real
+     * option for anything spending contract-gated coins, which is what made
+     * this the missing half rather than a preference.
+     *
+     * The concrete case: a swap solver funding an HTLC lockup has to pick coins
+     * whose batch outlives the swap AND commit the claim packet that lets
+     * covclaimd claim on the receiver's behalf. Without this it had to route
+     * the packet out of band and hope, which is strictly worse — an out-of-band
+     * registration lives in someone's memory, while an on-chain one is
+     * recoverable by anyone reading the transaction.
+     */
+    extensions?: Array<{
+        type: number;
+        payload: Uint8Array;
+    }>;
 }
 
 /**
--- a/packages/ts-sdk/src/wallet/wallet.ts
+++ b/packages/ts-sdk/src/wallet/wallet.ts
@@ -3471,10 +3471,26 @@ export class Wallet extends ReadonlyWallet implements IWallet, HDWalletCapable {
                     });
                 }
 
+                // `changeVout` is read BEFORE the extension output is appended,
+                // deliberately: it indexes the change, and appending after would
+                // make the extension the last output and silently retarget it.
+                const changeVout = selected.changeAmount > 0n ? outputs.length - 1 : 0;
+
+                if (params.extensions && params.extensions.length > 0) {
+                    outputs.push(
+                        Extension.create(
+                            params.extensions.map((ext) => ({
+                                type: () => ext.type,
+                                serialize: () => ext.payload,
+                            })),
+                        ).txOut(),
+                    );
+                }
+
                 return this._submitOffchainSpend(selected.inputs, outputs, {
                     sentAmount: params.amount,
                     changeAmount: selected.changeAmount,
-                    changeVout: selected.changeAmount > 0n ? outputs.length - 1 : 0,
+                    changeVout,
                     offchainTapscript,
                     serverPubKey,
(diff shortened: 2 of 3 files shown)


Kukks:
Thanks — all three addressed. Two are fixed, one I'm pushing back on.

## 1. Asset packet — fixed by refusing (the important one)

You're right, and it's worse than "not a regression". This branch has **never** emitted an asset packet, so an asset carried in on a selected VTXO was already being dropped silently: spent, not re-declared, nothing said. `extensions` didn't cause it, but it did make it worse in a specific way — an extension output now *exists*, so the transaction looks well-formed while still omitting the asset packet that had to be in it.

It now **refuses, with or without extensions**. The silent drop predates this PR and a loud failure beats a quiet loss either way, so guarding only the `extensions` combination would have left the original hole open.

I considered building the packet instead and decided against it: `_sendImpl` derives asset routing from recipients and a change receiver that this path doesn't have, and inventing that mapping in a BTC-only helper is exactly how asset accounting goes quietly wrong. Callers moving assets want `send()`, and the error says so.

## 2. Test coverage — added, and it bites

Fair point that "no existing pattern" is a reason to start one. Four tests now, and I checked the `changeVout` one isn't decorative by **regressing the code**: move the extension push above the `changeVout` read and it fails with `expected 2 to be 1` — change metadata bound to the OP_RETURN. Restore, and it passes. The invariant is genuinely pinned rather than merely described.

## 3. OP_RETURN size — not taken, the premise is out of date

**Bitcoin Core v30 removed the datacarrier size limit.** 80 bytes is pre-v30 standardness, not current policy, so there's no relay cap to validate against.

It would have been the wrong bound even when it applied: asset packets and the swap package's offer packets already exceed 80 bytes, and the claim packet this feature exists for (ciphertext + arkadeScript) is larger still — enforcing it would have rejected precisely the callers it was meant to protect. These are also Arkade *virtual* transactions, which aren't broadcast under L1 relay policy unless a vtxo is unilaterally exited.

Happy to add a bound if there's a real one to enforce, but it should come from Arkade's own tx-size constraints rather than a Bitcoin policy that no longer exists.

---

SDK suite **159 files / 2484 tests**, typecheck and repo-wide lint clean.

Kukks:
Closing — superseded by a better shape, per maintainer direction.

This added `extensions` to `sendBitcoin` because that was the path with coin selection. The right fix is the other direction: **add coin selection to `send`**, and use `send` everywhere.

That dissolves the problem this PR worked around rather than patching it:

- **The asset issue in review point 2 stops needing a guard.** `_sendImpl` already builds `createAssetPacket(...)` alongside custom extension packets. A caller that gets selection *there* gets correct asset handling for free — instead of the refusal this PR added, which was only ever the safe answer to a path that could not do it properly.
- **One send path instead of two partial ones.** Today `send` can carry an extension but not choose inputs, and `sendBitcoin({ selectedVtxos })` can choose inputs but not carry an extension. Merging capability into `send` removes the split rather than making both halves bigger.
- **The `changeVout` ordering hazard goes away** with the second output-building path it lived in.

What's worth carrying over into the replacement:

1. **The `changeVout` test.** `_sendImpl` has the same shape of hazard — change indexed by position, extension appended after — so the invariant is still worth pinning there. It's verified to bite: regress the ordering and it fails with `expected 2 to be 1`.
2. **The review's point 3 resolution.** The 80-byte OP_RETURN cap is not a real constraint — Bitcoin Core v30 removed the datacarrier size limit, and these are Arkade virtual transactions in any case. No size guard is needed on the new path either.
3. **The motivating case**, unchanged: a solver funding an HTLC lockup must pick coins whose batch outlives the swap *and* commit the client's sealed claim packet, so covclaimd can claim from the chain rather than from an in-memory registration that expires in 15 minutes and is lost on restart.

Branch kept at `feat/sendbitcoin-extensions`.

PR #730, opened 2026-08-12
feat(wallet): let send carry a taptree and a caller's coin selection

Kukks:
covclaimd can claim a preimage-gated output straight from the chain, but only if it can recognise the contract, and an Arkade address commits to nothing but the taproot output key. Its `matchOutput` requires `po.TaprootTapTree`; a funding transaction built through this SDK carried none on any output, so the daemon parsed the committed claim packet and then declined, silently. The funded output just sat there unclaimable.

Funding that lockup also needs the funder's own coins. `send` could build the asset packet and carry extension packets but always chose its own inputs; `sendBitcoin({ selectedVtxos })` could take a caller's inputs but built its own outputs and emitted no asset packet, so an asset riding on a selected VTXO was dropped without a word. A VHTLC lockup wants both at once — the coins have to outlive the swap's refund horizon, which generic selection cannot know, since it prefers the soonest-expiring batch and would hand the contract exactly the batch it must not inherit.

So `Recipient.tapTree` publishes the leaf set on that output's `PSBT_OUT_TAP_TREE`, and `send({ recipients, selectedVtxos })` spends exactly what it is given. The taptree is refused unless it derives the recipient address's taproot key. That check is the point of the feature rather than a nicety: a well-formed tree for the wrong contract produces a transaction that looks entirely correct and is unclaimable, and its wrongness surfaces only in whatever later tries to read it. One hash of leaves already in hand turns that into an error at the call that caused it. It compares against `pkScript` rather than the output script, because a sub-dust recipient is paid to the `RETURN` form of the same key and the tree describes the key either way.

Explicit selection is never quietly topped up. A shortfall in sats, a shortfall in any asset carried on those coins, and change too small to carry the asset change are three separate errors, each naming what was short. Assets arriving on the named coins are re-declared on the way out, which is what `sendBitcoin({ selectedVtxos })` was missing. The two call forms are told apart by the presence of `recipients`, not by argument count — a single recipient produces one argument in both forms — and an empty `selectedVtxos` array is an error rather than a fallback to generic selection, since a caller that meant to name inputs and named none is not asking the wallet to choose. `IWallet`, the Expo wallet and the service-worker wallet all carry the overload so they don't drift; the worker payload omits `selectedVtxos` entirely when unset, so an older worker sees byte-identical traffic.

The honest cost. Under `selectedVtxos`, `send` reads none of the wallet's own outputs, which means the past-cutoff deprecated-signer filter that generic selection applies does not run — name such a coin and the operator declines to co-sign at submit. That matches what `sendBitcoin({ selectedVtxos })` already does on this branch, and it is the same bargain as `settle({ inputs })`: you named it, you get it. The other thing worth flagging is that this commit was written on top of `feat/arkade-swap` and cherry-picked here, so one line did not survive: a fire-and-forget `logUngatedInputs` diagnostic that logs why a named input might be excluded. That helper and the exclusion machinery under it live only in that branch, and backporting them to carry a log line seemed a worse trade than dropping it — nothing on master logs anything on the equivalent `sendBitcoin` path either. It comes back for free whenever that lineage lands. The `_sendImpl` conflict resolved toward master's `getVtxos` plus `pendingRecoveryOutpoints` rather than the branch's `getSpendableVtxos` accessor, which master does not have.

Tests go 2378 to 2396: six on the taptree check, eight on caller-selected vtxos, four on the argument dispatch; swap and boltz-swap are unchanged at 67 and 614. Typecheck across the three packages, prettier and the commit hook's full suite are clean. Negative control on the refusal: deleting the derivation comparison so any decodable tree is accepted fails exactly one test, `rejects a well-formed taptree belonging to a different contract`, with `expected [Function] to throw an error` — while `rejects bytes that are not a taptree at all` keeps passing off the decode, which is the separation I wanted.

`arkade-os/lightning-swap-service` branch `feat/commit-claim-packet` is the solver half of this and is consuming the SDK from a local tarball; it cannot have green CI until this lands.

Not verified here: the end-to-end regtest run — a solver funding a lockup with a committed claim packet and taptree, and released covclaimd v0.0.1-rc.3 claiming it from the chain with no reveal call — was three consecutive green runs on the original branch's base, not on this rebase onto master. The code in the two is identical apart from the dropped log line and the selection accessor above, so I would be surprised if it diverged, but I have not re-run it here. Happy to rename `tapTree` or `selectedVtxos`, or to split the taptree and the selection into two PRs, if you would rather review them apart.

## Summary by CodeRabbit

- **New Features**
  - Added support for sending with a parameter object, including explicitly selected transaction outputs.
  - Added optional tap-tree data for recipients, with validation against their addresses.
  - Preserved recipient-based sending while enabling automatic or caller-selected input handling.
  - Exposed the new sending options through the SDK’s public types.

- **Bug Fixes**
  - Improved validation for insufficient funds, invalid selections, malformed tap trees, and dust-level change.


Kukks:
First CI run had `Integration (ts-sdk / exit-providers-rotation)` red on `unilateralExitDeepChain` — `timeout in waitFor` at the hop loop, waiting for the vtxo count to advance after a `send`. It passed on re-run with no code change, and I think it is environmental rather than mine: the failure is a timeout after `send` had already returned, not an error out of `send`, and the default (no `selectedVtxos`) path is the same code as before — the only new unconditional check on it is `changeAmount < 0`, which cannot fire there because `selectVirtualCoins` throws `Insufficient funds` while selecting rather than under-selecting. Flagging it since the red run stays in the history.

Kukks:
Thanks — all three taken, in `861c21a4`.

The depth call was right, and I measured it rather than reasoning about it. `TapTreeCoder.decode` does return depth (a tree encoded `[1,2,2]` decodes back as `[1,2,2]`), but `VtxoScript.decode` maps straight to `leaf.script` and rebuilds through `assembleBtcdTaprootTree`, so the shape is forced to arkd's canonical one and the depths never matter. Concretely, over the same three leaves:

```
shape (A,(B,C))  true key 51208f4e1edea88a8ca91677a3fcbeaa41261a70ef815f66b609d5a9e75fcb343067
shape ((A,B),C)  true key 5120e853c205dc36261f245830eb0fff77e36f69ec6c7f911fb221068d8a2f9420a8
VtxoScript.decode derives 5120e853c2... for BOTH
```

So `((A,B),C)` — which is what btcd assembles for that leaf order — round-trips, and `(A,(B,C))` is refused despite committing to its own address, exactly as you described. One detail worth recording: `VtxoScript.encode()` emits depth `1` for every leaf, which is not a structurally valid serialization for three leaves at all; it round-trips only because the reader ignores the field. So the real constraint is not "uniform depths" but "the shape btcd would build", and the only producer guaranteed to satisfy that is `VtxoScript.encode()`. `Recipient.tapTree` now says so and the mismatch error ends with `Expected VtxoScript.encode() form: leaf depths are ignored and the tree is rebuilt in arkd's canonical shape.`

Thirteen tests added for the selection paths. I tried to write the accounting ones so the assertion proves the arithmetic rather than just that something threw: two coins of 100 against a demand of 250 has to report a gap of 50, which is only reachable by summing both, since one coin alone would say 150; and one coin of 100 against two recipients wanting 60 each reports 20, which only a running subtraction sees. Negative controls, each restored after: dropping the bitcoin-shortfall guard fails 1 test, the asset-shortfall guard 3, the asset-change dust guard 2, the empty-selection guard 1. What surfaces without each is worse than the guard — `Cannot read properties of undefined`, `invalid asset id length: got 32 bytes`, and `Insufficient funds` respectively, that last one because the fallback tries to top up from a selection set that is empty by construction.

On the dead guard I think the premise only holds for TypeScript. `[Recipient, ...Recipient[]]` rules out `send()` at compile time, but this ships as JavaScript and `asSendParams` returns `{ recipients: [] }` for an empty argument list rather than throwing — the `params.recipients.length === 0` branch is only reached when there is one argument carrying `recipients`. So `_sendImpl` does see zero, from any JS consumer calling `send()`. I kept it with a line naming who reaches it, and added a test; removing the guard makes that test fail with an unrelated fetch error instead of "At least one receiver is required", which is the outcome I would rather not ship. Happy to drop it anyway if you would rather the SDK not carry runtime guards for cases the types forbid — it is your call and a one-line revert.

Left the triplicated dispatch alone as you said.

Tests 2383 to 2396; typecheck, prettier and the commit hook's three packages clean.

Kukks:
Correcting my earlier note on the red integration job, because I was more confident than the evidence supports.

`Integration (ts-sdk / exit-providers-rotation)` has now failed three times on this branch and passed twice — all of them `unilateralExitDeepChain` timing out in the same `waitFor`, at the vtxo count after a `send`. It failed on the doc-only trim commit too, which carries no behaviour change, and master has passed this job seven times including a control re-run I triggered a few minutes ago in the same window. So my "environmental" read is not established: what is true is that the only runs in that window were this branch's, so there was nothing to compare against when I said it.

What I can show is that the default `send` path — no `selectedVtxos`, no `tapTree` — is behaviourally identical to master. A whitespace-and-comment-normalised diff of the whole of `wallet.ts` leaves exactly one new unconditional statement on that path, `if (changeAmount < 0) throw`, and it is unreachable there: `btcAmountToSelect` is `Σ max(amount, dust) ≥ Σ amount = totalBtcOutput`, and `selectVirtualCoins` throws `Insufficient funds` rather than under-selecting, so `totalBtcSelected ≥ totalBtcOutput`. The `utils.ts` changes are inert without a `tapTree` beyond adding a `tapTree: undefined` own property to validated recipients, and `index.ts` is types only. The e2e suite imports `../../src`, so the build is not in the picture either. On master the test finishes in 12.4s against a 30s-per-hop budget, so it is not running close to its limit and this is not my change making it slower.

That leaves me without a mechanism, which is why I am flagging it rather than calling it. If you have seen `unilateralExitDeepChain` flake before, that would settle it; if you have not, it is worth a look independently of this PR, because a 30s `waitFor` on an indexer-observed count is the kind of assertion that goes intermittent under load. All eight checks are green on `861c21a4` as of now.

louisinger (approved):

louisinger (approved):

PR #736, opened 2026-08-12
fix(swap): one teardown path per pool ownership in the Nostr transport

tiero:
Every negotiation ends with the browser logging `WebSocket is already in CLOSING or CLOSED state`. Cause: `close()` ends the subscription and then closes the owned pool — and `pool.close()` ends every subscription on those relays again, so the second CLOSE frame lands on a socket that is already closing.

One teardown path per ownership now: an owned pool tears down through `pool.close()` alone; a shared pool (the caller's, possibly serving other swaps) keeps the subscription-only close.

Context: the wallet currently works around this at its call site, and the tempting wallet-side "fix" — dropping the `close()` call — would leak the owned pool and its socket on every negotiation. This removes the reason to touch that call site at all. Transport suite 9/9 green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qyyk1wuSFmcVXYz2aLeJqx)_

### DIFF
--- a/packages/swap/src/nostr.ts
+++ b/packages/swap/src/nostr.ts
@@ -255,11 +255,16 @@ export const nostrRfqTransport = (options: NostrRfqOptions): RfqTransport => {
 
         async close() {
             closed = true;
-            subscription.close();
             waiters.clear();
             // Only tear down a pool this transport created; a shared one belongs
-            // to the caller and may still be serving other swaps.
+            // to the caller and may still be serving other swaps. On an owned
+            // pool, `pool.close()` already ends every subscription on those
+            // relays — closing the subscription first as well sends a CLOSE
+            // frame on a socket the pool is about to close, and the browser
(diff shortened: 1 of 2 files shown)


tiero:
Addressed the required review item in 4372fe1: the suite now covers the owned-pool teardown. The new test spies on `SimplePool.prototype` so the default-pool path runs off the network, constructs the transport without a `pool` option, and asserts `pool.close(relays)` is called while `subscription.close()` is not — the exact double-close this PR removes. Suite is 10/10 green.

The two noted follow-ups (rejecting pending waiters on `close()`, and removing the wallet's call-site workaround once this lands) are out of scope here as the review says; happy to take them in a follow-up PR.

Earlier CI failures on this PR were infrastructure flakes (a Docker Hub pull timeout, then two unrelated integration jobs timing out waiting for a batch); each retrigger has moved the flake to a different job while this PR's own suite stays green.

---

pietro909 (approved):

PR #738, opened 2026-08-12
Move contract key provisioning into the SDK, asked for by leg

### LINKED ISSUE #737: swap: the wallet must provide the sender/refund key — static or fresh HD — instead of the package minting random keys for non-HD wallets
> [!NOTE]
> **Rescoped.** This issue originally targeted `boltz-swap`, which is being deprecated/deleted. The scope is now the **core SDK** and the **`@arkade-os/swap` package on `feat/arkade-swap`** (the branch the package is published from). The boltz-swap analysis is kept at the bottom as background only — no work is planned there.

## Problem

On `feat/arkade-swap`, when the wallet is not an HD wallet, the swap package generates a **random in-memory private key** and uses it as the VHTLC sender/refund key — then persists it as plaintext hex in the swap repository. The swap package should not care about wallet type: the wallet is passed in and has full identity capabilities. It should ask the wallet for a key, and the *wallet* decides what that means — its static key, or a freshly allocated HD one.

The chain of responsibility breaks in two places (permalinks pinned to `a3fc6e0`):

**1. The wallet abdicates for static wallets.** [`Wallet.getNextSigningDescriptor()`](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/ts-sdk/src/wallet/wallet.ts#L2540-L2544) returns `undefined` unless the descriptor provider is an `HDDescriptorProvider` — even though [`StaticDescriptorProvider.getNextSigningDescriptor()`](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/ts-sdk/src/identity/staticDescriptorProvider.ts#L27-L29) exists and correctly answers `tr(pubkey)`. A static wallet *can* answer "here is my key" through the same method; the wallet just refuses to route to it.

**2. The swap package compensates by minting keys.** Because the wallet won't answer, [`deriveSwapSecrets`](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/secrets.ts#L108-L113) comes back `undefined` for static/`auto` wallets and every RFQ entry point falls into [`randomSwapSecrets()`](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/secrets.ts#L119-L137):

- [`senderPrivateKey: schnorr.utils.randomSecretKey()`](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/secrets.ts#L134) — a signing key created outside any wallet key management, held in plugin memory,
- persisted plaintext as [`fallbackSecrets.senderPrivateKeyHex`](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/store.ts#L20-L28) in the swap repository,
- at 4 call sites: [rfq.ts#L741](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/rfq.ts#L741), [#L1098-L1103](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/rfq.ts#L1098-L1103), [#L1518](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/rfq.ts#L1518), [#L1726](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/rfq.ts#L1726).

Consequences: the key exists only in memory until the record is saved (crash = funded HTLC with no refund key); a device backup or copied profile leaks a spendable private key; nothing about the fallback arm is recoverable from seed; and [`senderIdentityForRfqSecrets` probes `isHDWalletCapable`](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/secrets.ts#L263) — wallet-type branching in package code, plus structural sniffing (`isDeterministicSigner`) to defend against `signerForDescriptor`'s silent fallback to the baseline identity for descriptors it cannot derive.

The whole point of the `Identity` abstraction is that consumers never need raw signers or key material out of the wallet. The derived arm honors that; the fallback arm gives up on it for exactly the wallets that were supposed to be the simple case.

## Why the random arm exists (the constraint a fix must honor)

The derived preimage is `sha256(sign_det(sha256(TAG ‖ xonly ‖ u32le(index))))` with [`index` pinned to 0](https://github.com/arkade-os/ts-sdk/blob/a3fc6e0/packages/swap/src/secrets.ts#L62-L68) — safe **only because each HD swap allocates a fresh descriptor**. A static wallet answers with the same key every time, so two swaps would derive the *identical preimage*, and one solver learning its own preimage would learn the other swap's. The random arm was the escape hatch for that. The fix must separate the two concerns: the **sender key** can and should always come from the wallet; the **preimage** needs per-swap uniqueness handled explicitly.

### 1. Core: the wallet always answers `getNextSigningDescriptor()`

- Route `Wallet.getNextSigningDescriptor()` to the configured descriptor provider whatever it is: `HDDescriptorProvider` → allocate fresh (unchanged); `StaticDescriptorProvider` → its `tr(pubkey)` descriptor. `undefined` remains only for wallets with no provider at all.
- `advanceSigningDescriptorWatermark` on a static wallet: no-op for its own descriptor (nothing to advance), keep throwing for foreign ones.
- `signerForDescriptor` must stop silently returning the baseline identity for a descriptor it cannot derive — that identity "signs happily with the wrong key" (the package's own docstring). Verify the returned signer's pubkey against the descriptor's key and fail loudly (typed error) on mismatch. That check is pure public-key equality — no wallet-type probing anywhere.

### 2. Swap package: never mint a key, never probe wallet type

- `deriveSwapSecrets` gets a descriptor for every `Wallet` — HD wallets a fresh one, static wallets their static one. The `?? randomSwapSecrets()` arms disappear from the RFQ entry points.
- Sender identity is always `signerForDescriptor(record.signingDescriptor)`; the `isHDWalletCapable` probe in `senderIdentityForRfqSecrets` goes away (replaced by the pubkey-equality check on what comes back).
- **Preimage per arm** (onchain sends only):
  - HD: derived, unchanged (fresh descriptor ⇒ unique).
  - Static: make the derivation index per-swap instead of pinned 0 — `buildPreimageMessage(xonly, index)` already takes the index; store the (public) index on the record. Requires `SingleKey` to gain `signSchnorrDeterministic` (aux_rand = 0), which it trivially can since it holds the key.
  - Custom/`auto` identities that cannot sign deterministically: store the **preimage** on the record — a per-swap secret, strictly less bad than a private key, and the only residual stored secret in the design. Never a private key.
- Keep *reading* existing `fallbackSecrets` records (v1, `senderPrivateKeyHex`) so already-persisted swaps stay refundable; stop *producing* them.

### 3. Documentation / separation of concerns

- State the ownership rule in `AGENTS.md` and the core README, mirroring the contracts-subsystem ownership section: **"Wallets own key material and derivation policy. Packages/plugins request a descriptor from the wallet and get back an `Identity`; they never probe wallet type and never generate signing keys. `isHDWalletCapable` in package code is a smell."**
- TSDoc on `getNextSigningDescriptor` documenting the static-wallet answer explicitly (it is a feature, not an absence), and on `signerForDescriptor` documenting the fail-loud contract.
- `@arkade-os/swap` README: per-arm table of where the sender key and preimage come from and what is recoverable from seed vs repository.

## Reference implementation (NArk, `arkade-os/dotnet-sdk`)

NArk's swap service never branches on wallet type: every wallet implements [`IArkadeAddressProvider.GetNextSigningDescriptor()`](https://github.com/arkade-os/dotnet-sdk/blob/c8273bc/NArk.Abstractions/Wallets/IArkadeAddressProvider.cs#L20-L23) with its own policy, and [`SwapsManagementService` calls it unconditionally](https://github.com/arkade-os/dotnet-sdk/blob/c8273bc/NArk.Swaps/Services/SwapsManagementService.cs#L234-L237). Preimages are wallet-derived ([`DerivePreimageAsync`](https://github.com/arkade-os/dotnet-sdk/blob/c8273bc/NArk.Swaps/Services/SwapsManagementService.cs#L149-L159) — which the `secrets.ts` scheme already mirrors). This issue brings the ts-sdk core and swap package to the same contract. A separate upstream issue will be filed on `dotnet-sdk` for the chain-swap ephemeral BTC-leg key data-loss risk (and to check NArk's own static-wallet preimage-index behavior against the uniqueness constraint above).

---

<details>
<summary>Background: original boltz-swap analysis (deprecated package — no work planned)</summary>

The same two patterns exist in `boltz-swap` on `master` (`3c2ee19`): `isHDWalletCapable` probes with plugin-side static fallbacks in `currentSigningDescriptor` / `swapSigner` / `swapOwnerKeys` (arkade-swaps.ts L3781–L3822), and a `secp256k1.utils.randomSecretKey()` ephemeral for the chain-swap BTC leg (L2890–L2908) persisted plaintext (`BoltzChainSwap.ephemeralKey`) and explicitly unrecoverable on restore (`ephemeralKey: ""`, L4066–L4077; related #515, #680). Since the package is being deleted, these are documented here only as the origin of the pattern.

</details>

tiero:
Closes #737. Retargeted to `master` now that `feat/arkade-swap` is merged (#741).

## The problem

When the wallet could not allocate an HD descriptor, `@arkade-os/swap` generated a **random in-memory private key**, used it as the VHTLC sender/refund key, and persisted it as plaintext hex in the swap repository. The swap package should not care what kind of wallet it was handed.

`Wallet.getNextSigningDescriptor()` refused to answer for static wallets (returning `undefined` even though `StaticDescriptorProvider` knows the answer is `tr(pubkey)`), so `deriveSwapSecrets()` fell through to `randomSwapSecrets()` at four RFQ entry points. That key existed only in memory until the record was saved — a crash after funding left an unrefundable lockup — and a copied profile leaked a spendable private key.

## The change

**`packages/swap/src/secrets.ts` is deleted.** The swap package no longer names, derives, or stores keys. It asks the SDK for what the leg needs:

```ts
// a leg we fund — all it needs is the key that refunds it
const { pubkey, descriptor } = await provisionRefundKey(wallet);

// a leg we claim — the key that receives it, and the P that unlocks it
const { pubkey, descriptor, preimage, paymentHash, mustPersistPreimage } =
    await provisionClaimSecret(wallet);

// later
const signer = await contractSigner(wallet, descriptor);
const p = await contractPreimage(wallet, descriptor, storedPreimage);
```

How the wallet answers is invisible to the caller: HD allocates a fresh descriptor, static answers with its identity key. No consumer branches on wallet type, and no consumer generates key material.

**Core wallet:** `getNextSigningDescriptor()` answers for every wallet; `signerForDescriptor()` throws a typed `ForeignDescriptorError` for a key the wallet does not hold instead of silently substituting the baseline identity; `Wallet` and `ServiceWorkerWallet` route through one shared `resolveDescriptorSigner` so the two sides of the bus cannot answer differently for one descriptor.

**Preimages** split on the **descriptor's shape, not the wallet's type**. An HD child belongs to one artifact, so `sha256(sign_det(...))` is safe and nothing secret is at rest. A static descriptor is the same key every time, so deriving would give every swap the *identical* preimage — one counterparty learning its own would learn all of them. Those get a random per-swap preimage, and `mustPersistPreimage` says so.

What stays in the swap package is only swap lifecycle: `RefundNotLocallyPossibleError` / `senderIdentityForSwapRecord` (`refundBlocked.ts`), and `swapSecretsToRecord` next to the `AssetSwap` shape it maps onto. **No swap record can carry a private key** — `fallbackSecrets` and the `derivable` discriminant are gone rather than kept readable (0.0.1–0.0.3 shipped before any consumer; see the discussion below).

## Reference implementation

NArk's swap service never branches on wallet type. Two upstream issues came out of the comparison: arkade-os/dotnet-sdk#184 (chain-swap BTC-leg ephemeral key unrecoverable after storage loss) and arkade-os/dotnet-sdk#185 (NArk's static wallets derive the identical preimage for **every** swap — the same defect this fixes, in the reference implementation). The cross-SDK preimage vectors moved into the SDK with the scheme and still pass byte-for-byte.

## Defects found in review and fixed here

Two audit rounds (correctness, security, cross-package compatibility, restore/persistence, then a dedicated service-worker pass) ran against intermediate commits. The confirmed ones:

- **A funded leg bound to a key the wallet never proved it could sign for.** `provisionRefundKey` returned the descriptor's leaf key without asking for a signer; deriving that key is pure parsing and succeeds for any well-formed descriptor. `provisionClaimSecret` caught it only incidentally via `derivePreimage`, and not at all when the caller supplied its own P — so both funding paths were unverified. The pubkey now comes from a signer the wallet produced, moving the failure to before there is a quote.
- **Fund-stranding behind the service worker.** `ServiceWorkerWallet.signerForDescriptor` was not updated for the new bare descriptor. A seed identity claims a pathless `tr(pubkey)` by raw-pubkey match, so it returned a `DescriptorIdentity` whose `xOnlyPublicKey()` was *correct* — passing the key check — and whose every signature threw `"Descriptor must specify a full derivation path"`.
- **Reachable preimage collision.** A custom `DescriptorProvider` may contractually return a constant descriptor; if HD-shaped, two swaps would share a preimage and payment hash. That allocation branch is gone.
- **`ExpoWallet` forwarded none of the descriptor surface,** so an Expo wallet on `walletMode: 'hd'` bound every artifact to its baseline identity key while its receive addresses rotated.
- **`auto`-mode wallets refused their own keys** for every HD index except 0 (the baseline-key alias masked it).

`contractSigner`'s equality check is documented for what it can and cannot catch: it catches a wallet substituting its baseline identity, but not one returning a descriptor-scoped identity over the same descriptor, whose pubkey is read back out of the descriptor string. `resolveDescriptorSigner`'s `isOurs` test is what rules that out.

## Key-material sweep

The pattern is not widespread. Legitimate randomness (MuSig2 session keys, HTLC preimages, the ClaimPacket ECDH ephemeral, ArkadeCash bearer notes) is correctly random. `SingleKey.fromRandomBytes()` is left alone — it is a key holder and that is its documented path. Two items remain, out of scope: `serializeSigningIdentity` ships raw private-key hex across the service-worker boundary including a duck-typed `toHex()` fallback for *any* identity, and all six runnable examples teach `SingleKey.fromRandomBytes()` as the way to build a wallet. `boltz-swap` is untouched — it is being deprecated.

## Testing

`pnpm run build`, `pnpm run lint`, and all three unit suites pass: ts-sdk 2555, swap 360, boltz-swap 614 (+ typecheck).

New regression coverage for paths that had none: `serviceWorkerContractSecrets.test.ts` (12 tests) pins that page-side and worker-side wallets resolve the same key for one descriptor, that the returned signer can **actually sign** rather than merely carry the right pubkey, that allocation over the bus never reissues an index or repeats a preimage, and that a claim preimage re-derives after a page reload.

On CI, `Integration (swap-rfq / all)` and `Integration (swap / all)` both pass against the live regtest stack — these exercise the refactored corridors end to end. `Integration (ts-sdk / exit-providers-rotation)` failed once on an earlier head with `indexer service not ready: wallet is locked or syncing` and intents never landing in a batch: regtest infrastructure, in a job unrelated to key provisioning.

## Overlap with #740

#740 solves the same half of #737 inside the swap package (a `BaselineSwapSecrets` arm falling back to `wallet.identity`, explicitly "no change to core `wallet.ts`"). Both independently derived the same preimage-collision reasoning, which is reassuring, but the two conflict directly in `secrets.ts`, `rfq.ts`, `index.ts` and three test files and cannot both merge. #740's persistence half (`rfqRecord.ts`, DB v1→2) is disjoint from this and unaffected either way.

## Summary by CodeRabbit

* **New Features**
  * Added wallet-managed key and preimage provisioning for swap transactions.
  * Added descriptor-based signer recovery, restoration, and adoption.
  * Expanded SDK support for contract signing, preimage recovery, and descriptor utilities.

* **Improvements**
  * Simplified swap record persistence to store descriptors and only non-recoverable preimages.
  * Added typed refund failure reasons for unavailable signing capabilities or inaccessible descriptors.
  * Removed legacy swap secret and fallback-secret APIs.


packages/swap/src/store.ts
```diff
@@ -17,16 +19,6 @@ export type AssetSwapStatus =
  * restore layers share one spelling instead of re-typing the literal. */
 export const BTC_ASSET_ID = "btc";
 
-export interface AssetSwapFallbackSecretsV1 {
```
pietro909, line 20: The description says *"Legacy `fallbackSecrets` records stay readable so existing swaps remain refundable"* — that is not true of this branch. This type is deleted here and nothing reads the field anywhere: `git grep fallbackSecrets packages/swap` hits only the README line describing it as gone.

Concretely, a record written by 0.0.1–0.0.3 (all three are published, and they wrote this shape for every static wallet) now carries no `signingDescriptor`, so `senderIdentityForSwapRecord` refuses it with `no-secrets` and that swap loses its refund key.

Failing loudly is the right failure mode, so this only matters if such records exist. If no consumer is on 0.0.1–0.0.3 yet, the fix is one line of the description plus the README notes — not code. If any consumer does have records, see the note on `refundBlocked.ts` for the read-only branch that restores them.

packages/swap/src/refundBlocked.ts
```diff
@@ -0,0 +1,59 @@
...
+ * **Wire `refundArkade` here, not to `contractSigner` directly.** The SDK
+ * answers about a descriptor; only this knows that a record without one is a
+ * permanent refusal rather than a `TypeError` at the push site — which the
+ * manager would treat as retryable.
+ */
+export async function senderIdentityForSwapRecord(
+    wallet: IWallet,
+    record: { signingDescriptor?: string },
+): Promise<Identity> {
+    if (!record.signingDescriptor) {
```
pietro909, line 57: If you decide 0.0.1–0.0.3 records should stay refundable, this is the whole fix — and nothing starts writing the old shape again:

```ts
if (!record.signingDescriptor) {
    // read-only: re-declare AssetSwapFallbackSecretsV1 in store.ts, never written
    const legacy = record.fallbackSecrets;
    if (legacy?.version === 1 && legacy.type === "stored") {
        return SingleKey.fromPrivateKey(hex.decode(legacy.senderPrivateKeyHex));
    }
    throw new RefundNotLocallyPossibleError("no-secrets", "...");
}
```

The `contractSigner` path below is untouched, and "one secrets shape, no stored private keys anywhere" still holds for everything this branch *writes* — only the reader comes back.

Worth noting it is not just refunds: claiming needs `legacy.preimageHex` too, wherever the record → P read ends up living after `ffa75b25`.

packages/swap/README.md
```diff
@@ -487,17 +512,26 @@ after heavy swap use.
 
 The package is pre-release; these notes replace a changelog for consumers tracking the branch.
 
+- **`randomSwapSecrets` and the stored-secrets arm are gone; the wallet is the only key source.**
+  `deriveSwapSecrets(wallet, { preimage })` answers for every wallet (HD → fresh descriptor,
```
pietro909: These migration notes went stale in `ffa75b25`: `deriveSwapSecrets` and `SwapSecrets` no longer exist anywhere in `src` (`git grep deriveSwapSecrets packages/*/src` is empty). The API a consumer now meets is `provisionRefundKey` / `provisionClaimSecret` / `ProvisionedKey` / `ProvisionedClaimSecret` from `@arkade-os/sdk`, plus `swapSecretsToRecord` here in `store.ts`.

Since this section explicitly stands in for a changelog for consumers tracking the branch, it is the one place that rename has to land — a consumer upgrading from 0.0.3 meets both refactors as a single breaking step.

pietro909 (commented):
Three back-compat / doc notes from reading this against `@arkade-os/swap@0.0.3` (the current `feat/arkade-swap` head, `a3fc6e06`). No objection to the design — "the wallet owns key policy, packages ask and never mint" is the right contract, and `ForeignDescriptorError` closes a genuinely nasty silent-wrong-key path.

The substantive one is the first: **the description and the code disagree about `fallbackSecrets`.** Whichever way that is resolved, they should agree before merge.

Context for the third: a follow-up is planned to close the other half of #737 — a public per-swap salt so static wallets *derive* their preimage instead of storing one, which removes the last secret at rest. It stacks cleanly on this; nothing below blocks it.

tiero:
Thanks — all three land. Two are fixed in `36ad22c`; the `fallbackSecrets` one is a deliberate call, so here is the reasoning rather than a silent change.

**1. Description vs code on `fallbackSecrets` — you're right that they disagreed, and the code is what's correct.** The "stays readable" line was from the first revision of this PR and survived into the description after `b460ec5` deleted the reader. The description has since been rewritten; the read-only branch is **not** coming back.

The call is that 0.0.1–0.0.3 have no consumers. They were published as release-plumbing commits on this feature branch, not adopted — so this is the window where the change costs nothing, and after a consumer ships it becomes a secret migration across every deployed wallet. If that assumption is wrong for anyone, say so and I'll take your `refundBlocked.ts` branch as written — it is the right shape, and you're right that it isn't only refunds: the claim path needs `legacy.preimageHex` too, which after `ffa75b25` means `contractPreimage(wallet, descriptor, stored)`'s `stored` argument.

What the branch does instead is fail loudly and specifically: such a record has no `signingDescriptor`, so `senderIdentityForSwapRecord` refuses it as `no-secrets` rather than mis-signing with the wrong key.

**2. Stale migration notes — fixed.** Genuinely stale, and worse than you saw: besides the prose, a runnable example still imported `rfqSecretsToRecord` and `preimageForRfqSecrets`. The notes now name the actual surface (`provisionRefundKey` / `provisionClaimSecret` / `contractSigner` / `contractPreimage` from the SDK, `swapSecretsToRecord` from `store`), list every removed symbol so a 0.0.3 consumer meets both refactors as one step, and state the record change explicitly.

**3. The per-swap salt follow-up** — that removes the last secret at rest and I'd rather it landed than this stored-preimage arm persisted. `mustPersistPreimage` is deliberately a boolean the wallet computes, not something the caller infers, so flipping the static case from "store P" to "derive P from a public salt" is a change inside `provisionClaimSecret` plus a record field, with no consumer edit.

One thing worth flagging since you're reading this area: **#740 overlaps this PR heavily.** It solves the same half of #737 in the swap package (a `BaselineSwapSecrets` arm falling back to `wallet.identity`, explicitly "no change to core `wallet.ts`") while this one moves provisioning into the SDK and deletes `packages/swap/src/secrets.ts`. Both independently derived the same preimage-collision reasoning, which is reassuring, but they conflict directly in `secrets.ts`, `rfq.ts`, `index.ts` and three test files and cannot both merge. #740's persistence half (`rfqRecord.ts`, DB v1→2) is disjoint from this and unaffected either way.

---

packages/swap/src/refundBlocked.ts
pietro909: Dropping this one. Confirmed there are no records anywhere from 0.0.1–0.0.3, so the reader would have nothing to read — refusing a record with no `signingDescriptor` as `no-secrets` is the right terminal behaviour, and the claim-path note goes with it.

pietro909:
Check my comment in https://github.com/arkade-os/ts-sdk/pull/743#issuecomment-5273156678 and see :
> What https://github.com/arkade-os/ts-sdk/pull/738 should take from here: the cannot-sign check. Its provisionRefundKey calls wallet.identity.xOnlyPublicKey() unconditionally and returns a descriptor nobody can sign with, so the failure lands as a TypeError at push time; signingIdentityOf's intent belongs in contractSigner as a typed refusal.

Can you act on it? Is it relevant?

tiero:
Relevant, and done in `8792f74` — the diagnosis was exactly right.

`provisionRefundKey` verified that the descriptor's key was *ours* but never that we can **sign** with it, so a watch-only identity (or a remote signer that isn't attached) passed the pubkey check — it really is our key — and the covenant bound it. The failure then landed as a `TypeError` deep inside the push, where `RfqSwapManager` reads it as retryable and grinds for the whole refund window.

`signingIdentityOf`'s intent now lives in `contractSigner`, as you suggested:

```ts
if (!isSigningIdentity(signer)) throw new WalletCannotSignError(descriptor);
```

Checking all four members (`sign`, `signMessage`, `signerSession`, `xOnlyPublicKey`) for #743's reason — `signerSession` is the one an interactive refund needs and the one a half-identity is missing. Both provisioning entry points inherit it, so a wallet that could never spend a leg finds out **before** it funds one rather than at push time.

On the swap side it maps to a third `RefundBlockedReason`, `unsignable-wallet`, separate from `foreign-descriptor`, because the remedy differs: "attach your signer" vs "restore the other wallet" — sending a user after a seed they already have is its own failure mode.

The one deliberate divergence from #743: where it mints and stores a private key for the cannot-sign case, this refuses. Minted keys are what this branch exists to remove, and a wallet that cannot sign cannot own a swap's refund key — making the application the silent custodian of one is the situation #737 opened about. Failing at provisioning is the honest version.

Fallout worth flagging for review: a genuine `ReadonlyWallet` can no longer create a swap. One test fixture was exactly that, and two others carried identities too thin to sign; all three now hold real signing identities. I think refusing is correct — a readonly wallet could never refund the lockup it funds — but it is a behaviour change, so say so if you disagree.

Also dropped the `fallbackSecrets` reader question per your other note — confirmed no 0.0.1–0.0.3 records exist, so `no-secrets` is the right terminal state.

---

packages/ts-sdk/src/wallet/contractSecrets.ts
```diff
@@ -0,0 +1,316 @@
...
+ * Know its limit. It catches a wallet that substitutes its *baseline*
+ * identity, which is the failure this exists for. It cannot catch one that
+ * returns a descriptor-scoped identity built over the same descriptor, since
+ * such an identity reads its pubkey back out of the descriptor string and the
+ * comparison becomes a tautology. What rules that case out is
+ * {@link resolveDescriptorSigner}'s `isOurs` test, which both shipped wallets
+ * resolve through — so this is the backstop for a hand-rolled `IWallet`, not
+ * the primary guarantee.
+ */
+export async function contractSigner(wallet: IWallet, descriptor: string): Promise<Identity> {
```
pietro909, line 193: This should preserve the typed foreign-descriptor boundary. `wallet.signerForDescriptor()` throws `ForeignDescriptorError`, but this backstop can still throw a plain `Error` when the wallet returns the wrong key, and `deriveDescriptorLeafPubKey()` can also throw parser errors before the equality check.

Callers need to distinguish "wrong wallet / foreign descriptor" from "this wallet cannot sign" or an operational signer failure. Leaking a generic error here makes that branch depend on message/cause shape and can turn a terminal wrong-wallet condition into a retryable operational failure.

Suggestion: import `ForeignDescriptorError` and wrap both the descriptor-parse failure and the mismatch path in it, while keeping `WalletCannotSignError` for the "ours, but cannot sign" case.

pietro909 (commented):
One stability/security note inline.

tiero:
Fixed in `65cedac` — and following it downstream turned up a worse instance of the same problem than the one you pointed at.

Both leaks you named are closed. `deriveDescriptorLeafPubKey` is wrapped, so an unreadable descriptor raises `ForeignDescriptorError` (with the parser error as `cause`), and the mismatch branch raises it too instead of a bare `Error`. `WalletCannotSignError` stays for "ours, but cannot sign".

The part worth a second look is where that generic error was landing. `refundBlocked` mapped *everything* that wasn't a `WalletCannotSignError` to `foreign-descriptor` — so a signer that merely failed to answer (dropped transport, remote signer mid-restart) was recorded as a wrong-wallet verdict. Since every `RefundNotLocallyPossibleError` is terminal to `RfqSwapManager`, that abandoned a perfectly refundable swap for the rest of its window on a transient fault. Exactly the inversion you describe, one layer down.

So the boundary now has three outcomes rather than two:

```ts
if (cause instanceof WalletCannotSignError) → "unsignable-wallet"   // terminal
if (cause instanceof ForeignDescriptorError) → "foreign-descriptor" // terminal
throw cause;                                                        // operational → retryable
```

The `await signer.xOnlyPublicKey()` call sits deliberately outside the wrapped region for the same reason: a signer that cannot answer must never be mistaken for a key belonging to someone else.

Two tests pin it — an unreadable descriptor classifies as `foreign-descriptor`, and a signer whose `xOnlyPublicKey()` rejects comes back out as the *same object*, not wrapped in a typed refusal.

Unit suites green: ts-sdk 2556, swap 366, boltz-swap 614.

Unrelated CI note: the `Integration (ts-sdk / settlement-delegation)` failure reported on `8792f74` was a cancellation, not a real failure — its whole log is teardown with no test output, from my push superseding the run. The earlier `exit-providers-rotation` failure did run, and failed on `indexer service not ready: wallet is locked or syncing` with intents never landing in a batch — regtest infrastructure. In both runs `Integration (swap-rfq / all)` and `Integration (swap / all)`, which exercise this diff, passed.

---

packages/ts-sdk/src/wallet/contractSecrets.ts
```diff
@@ -0,0 +1,343 @@
...
+ * Throws for a descriptor shared across artifacts and no stored preimage —
+ * deriving there would hand two artifacts one preimage, and the caller was
+ * told to persist it (`mustPersistPreimage`).
+ */
+export async function contractPreimage(
+    wallet: IWallet,
+    descriptor: string,
+    stored?: Uint8Array,
+): Promise<Uint8Array> {
+    if (stored) return stored;
```
pietro909: **Review finding (stability, verified):** `contractPreimage` returns a stored preimage verbatim with no 32-byte check, dropping the `decodeHex32` restore-time validation the deleted `rfqSecretsOfRecord` enforced.

A truncated or corrupted `preimageHex` on a record (a field-mapped backend trimming a column, a partial write) now restores silently where master threw `preimageHex must be 32 bytes` at read time. The damage is bounded — `claim.ts` and `onchainHtlc.ts` cross-check the preimage hash before signing, so the claim fails loudly at build time rather than on-chain — but the corruption is diagnosed only when the user attempts to claim, with timeout margin already spent, instead of at restore where the record could be flagged.

Suggestion: `if (stored) { if (stored.length !== 32) throw new Error(...); return stored; }`

packages/ts-sdk/src/wallet/contractSecrets.ts
```diff
@@ -0,0 +1,343 @@
...
+ * Whether `value` is a complete {@link Identity} rather than the read-only
+ * half of one.
+ *
+ * All four members are checked because all four are load-bearing downstream —
+ * `signerSession` in particular, which an interactive refund needs and which a
+ * watch-only identity lacks. A partial identity passes a pubkey check happily
+ * and then fails as a `TypeError` deep inside a push, where the swap layer
+ * reads it as retryable and grinds against it for the whole refund window.
+ */
+function isSigningIdentity(value: unknown): value is Identity {
```
pietro909: **Review finding (simplification, verified):** `isSigningIdentity` re-implements the pre-existing `isSigningCapable` guard (`serviceWorker/wallet.ts:282`), leaving two private copies of the "is this a full `Identity`" structural check.

Both probe `sign`/`signMessage`/`signerSession` (this one adds `xOnlyPublicKey`). When `Identity` gains or loses a load-bearing member, one guard gets updated and the other silently lets a watch-only identity through to fail as a `TypeError` mid-push — the exact failure this guard's own doc comment warns about.

Suggestion: promote one shared guard into `src/identity` next to the `Identity`/`ReadonlyIdentity` types and use it from both sites.

packages/ts-sdk/src/wallet/contractSecrets.ts
```diff
@@ -0,0 +1,343 @@
...
+/**
+ * A descriptor for one artifact. HD wallets allocate a fresh index; static
+ * wallets answer with their identity key, which is their whole policy.
+ */
+async function provisionDescriptor(wallet: IWallet): Promise<string> {
+    const allocated = isHDAllocationCapable(wallet)
+        ? await wallet.getNextSigningDescriptor()
+        : undefined;
+    // A wallet that cannot allocate still provides a key it holds.
+    return allocated ?? normalizeToDescriptor(hex.encode(await wallet.identity.xOnlyPublicKey()));
```
pietro909: **Review finding (simplification, verified):** the static-policy descriptor spelling `normalizeToDescriptor(hex.encode(await identity.xOnlyPublicKey()))` is duplicated between here and `Wallet.getNextSigningDescriptor` (`wallet.ts`), with a third `tr(${hex.encode(...)})` spelling in `Wallet`'s address path.

All three sites must forever produce byte-identical descriptors for one identity, or a static wallet's stored descriptor stops round-tripping through `signerForDescriptor`/`contractSigner`'s pubkey-equality check — a signable record refused as foreign.

Suggestion: a single exported helper, e.g. `identityDescriptor(identity)` in `src/identity/descriptor.ts`, keeps the policy in one place — which is what AGENTS.md's "policy lives inside the wallet" rule asks for.

tiero:
`5441e37` closes the review round. Build clean, lint clean, 3539 unit tests green (2559 ts-sdk / 366 swap / 614 boltz-swap), one new.

**The one that would have misled a consumer.** `pushRefundWithoutReceiver`'s doc told callers to build the sender signer with `contractSigner`, which skips exactly the classification `senderIdentityForSwapRecord` exists to add — a permanent refusal would reach `RfqSwapManager` untyped and get retried for the rest of the refund window. Now points at the guard.

**Also fixed**

- README: the provisioning example redeclared `pubkey`/`descriptor` in one scope, so it did not compile if copied; and it still listed the deleted fallback arm as a refund failure mode. Both now match the three reasons `senderIdentityForSwapRecord` can actually report.
- `ExpoWallet` declares `HDWalletCapable`/`HDAllocationCapable`, so its five delegated methods are checked against the interfaces at build time instead of failing inside a structural probe at runtime.
- `signerForDescriptor` documents that the identity it returns must really sign; a watch-only one carrying the right key is `WalletCannotSignError`, a distinct failure with a distinct remedy (attach the signer, don't restore another seed). Recorded in `AGENTS.md` and pinned by a service-worker test that funds nothing — thanks for that suggestion, it was the cheapest way to nail the fund-stranding shape.
- Tests: the foreign descriptor is materialized through the provider rather than by substituting the wildcard — `materializeDescriptorAt`'s own docstring warns against "ad-hoc string substitution", and a substitution that misses leaves the template ranged while the test still reads as though it named a key. Plus `mustPersistPreimage` asserted directly, the preimage-per-corridor expectation pinned so the leak check cannot silently narrow to the `rfqId`, and `hex.encode` over Node-only `Buffer`.

**Already fixed in `a9b16e8`, before these reviews ran** — all three of @pietro909's findings, which is also what arkana's change request was tracking: the stored-preimage 32-byte check in `contractPreimage`, the duplicate signing guard (now one `isSigningIdentity` in `src/identity/index.ts`, imported by both `contractSecrets.ts` and `serviceWorker/wallet.ts:1604`), and the three descriptor spellings (now `identityDescriptor()`). CodeRabbit's re-post of the duplicate-guard finding is stale — there is no local copy left to remove.

**Not changed, deliberately:** `adoptContractDescriptor`'s probe gating. The gap is real but only for a hand-rolled wallet implementing `HDAllocationCapable` without `HDWalletCapable`; every wallet shipped here implements both, and the current code already fails in the safe direction (propagate rather than silently advance a watermark). Recorded in the docstring rather than adding speculative branching.

Still open and outside this PR: `serializeSigningIdentity` shipping raw private-key hex across the service-worker boundary. Worth its own issue.

---

packages/ts-sdk/src/identity/index.ts
```diff
@@ -113,3 +115,28 @@ export { isHDCapableIdentity } from "./hdCapableIdentity";
...
+export function isSigningIdentity(value: unknown): value is Identity {
+    if (typeof value !== "object" || value === null) return false;
+    const v = value as Partial<Identity>;
+    return (
+        typeof v.sign === "function" &&
+        typeof v.signMessage === "function" &&
+        typeof v.signerSession === "function" &&
+        typeof v.xOnlyPublicKey === "function"
+    );
+}
```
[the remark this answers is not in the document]
tiero: Already done in `a9b16e8` — this review ran against `a7c079b`, before that commit landed.

There is no local guard left to remove. `serviceWorker/wallet.ts` imports the shared one at line 30 and calls it at line 1604:

```
packages/ts-sdk/src/wallet/serviceWorker/wallet.ts:30:    isSigningIdentity,
packages/ts-sdk/src/wallet/serviceWorker/wallet.ts:1604:        if (!isSigningIdentity(options.identity)) {
packages/ts-sdk/src/identity/index.ts:133:export function isSigningIdentity(value: unknown): value is Identity {
packages/ts-sdk/src/wallet/contractSecrets.ts:24:import { Identity, ReadonlyIdentity, isSigningIdentity } from "../identity";
```

Lines 133-142 of that file are the import block, not a guard. Promoting the four-member check into `src/identity` next to the `Identity` type — including `xOnlyPublicKey`, which the old three-member `isSigningCapable` missed — is exactly what `a9b16e8` did, so the divergence risk this describes is closed.

---

tiero:
@arkana-ai-bot Pass 2 ran against `a9b16e8`, so it missed `5441e37` — which already closed the interface-boundary half of Finding #2 (`HDWalletCapable.signerForDescriptor` now documents the four-member requirement, with the rule recorded in `AGENTS.md` and pinned by a new service-worker test) and the `adoptContractDescriptor` note.

`fcb18ff` closes what genuinely remained, both comment-only:

- **Finding #2, the last piece.** `IWallet.identity` now says it must be a real signer rather than a `ReadonlyIdentity` that structurally fits, and names `signerSession` as the member a watch-only identity lacks. That was the gap — a custom wallet author satisfies the property type and learns the rule from a runtime `WalletCannotSignError`.
- **Finding #1.** The tautology limit is now inline at the equality check, not only in the JSDoc block above it. Fair point that it belongs at the line someone would otherwise trust.

On the `adoptContractDescriptor` note — I'm leaving the code as is and keeping it documentation. The gap only reaches a hand-rolled wallet implementing `HDAllocationCapable` without `HDWalletCapable`; every wallet shipped here implements both, and as you note `advanceSigningDescriptorWatermark` does its own `isOurs` check, so the current code already fails in the safe direction. Adding a branch for a shape that does not exist is the kind of speculative defence I'd rather not carry.

Build clean, lint clean, 3539 unit tests green.

---

PR #743, opened 2026-08-12
swap: use the wallet's own key when it cannot allocate a descriptor

tiero:
A wallet that cannot allocate an HD descriptor is not a wallet without a key, but `deriveSwapSecrets` treated the two as the same case and generated a random private key. That key exists nowhere else: the application has to persist it before funding, and losing it strands the lockup until the refund window opens. Every single-key and static wallet took that path on every swap — including the reference wallet, whose `ServiceWorkerWallet` carries a perfectly good signing identity and just returns no descriptor.

**The wallet's own key is the answer to "give me a static public key".** Wallet type stops mattering for the sender key: the arm is chosen by whether an identity can sign, not by whether it is HD.

## The three arms

| Wallet | Arm | What the record stores |
|---|---|---|
| Can allocate (HD) | `signingDescriptor` | a public descriptor |
| Can sign, cannot allocate | **`identityKey: true`** (new) | nothing secret |
| Cannot sign (readonly, remote signer) | `fallbackSecrets` | the raw key — persist it |

The identity arm is precedent, not invention: the Boltz integration shipped it for years, resolving a swap's signer by descriptor and **falling back to `wallet.identity`**. HD stays preferred where available — a fresh descriptor per swap keeps swaps unlinkable and lets the preimage derive from the seed.

## The preimage is the sharp edge

`derivePreimage` is a function of the signing key alone. Reusing one key would produce the **same P for every swap**, and a solver holding one lockup's preimage could claim another's. So the identity arm never derives one: preimage-bearing flows (`requestOnchainSend`, both receives) attach a fresh random 32-byte preimage per swap, `preimageForRfqSecrets` throws rather than derive, and `preimageHex` on the record becomes mandatory — which the flows warn about at call time. A lightning send carries no preimage at all, so that path stores literally nothing.

`signingIdentityOf` checks the full `Identity` contract (`sign`, `signMessage`, `signerSession`, `xOnlyPublicKey`), matching the check `ServiceWorkerWallet`'s own constructor makes. An interactive refund needs `signerSession`; a half-identity would surface as a `TypeError` deep inside the push instead of the typed `RefundNotLocallyPossibleError` that keeps `RfqSwapManager` from grinding for the whole refund window.

## Surface

- `IdentitySwapSecrets` and `isIdentitySwapSecrets` are exported; `SwapSecrets` becomes a three-way union, so a consumer that reads `.signingDescriptor` off a derivable arm now needs a narrowing check.
- `AssetSwap` gains `identityKey?: true` alongside `signingDescriptor?`. Both are public; `rfqSecretsToRecord` / `rfqSecretsOfRecord` round-trip it.
- No wire, address, or covenant change — the sender key changes only in where it comes from.

## Verification

400 unit tests pass (the e2e suites need a regtest server). New coverage: the arm engages for a key-bearing non-HD wallet and not over an available allocation; the sender pubkey is the wallet's own key; a record restores on a fresh wallet instance built from the same key; a wallet that cannot sign gets the typed refusal; deriving a preimage from the reused key is refused; two swaps on one wallet share a sender key but never a preimage or payment hash.

Supersedes the interim wallet-side workaround (arkade-os/wallet#910, closed) — with this, the wallet stores no secret to begin with.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qyyk1wuSFmcVXYz2aLeJqx)_

## Summary by CodeRabbit

* **New Features**
  * Added support for swaps using a wallet’s identity key when signing descriptors are unavailable.
  * Added fresh, explicitly stored preimages for preimage-based swap flows.
  * Improved support for static, automatic, and single-key wallets.
  * Added reliable restoration and refund handling for identity-backed swaps.
* **Bug Fixes**
  * Prevented reuse of preimages when the same sender key is used across swaps.
  * Added validation when required signing identities or stored preimages are unavailable.


pietro909:
Read against #738, which solves the same half of #737. They conflict in **all seven files this touches** (`secrets.ts` as modify/delete — #738 deletes it), and #743 conflicts with #740 in four more, so exactly one of the three can land in this surface.

Where all three agree, independently: never derive P from a reused key — per-swap random preimage, mandatory persist. That part is settled by convergence.

Why I'd land #738:

- **Migration timing.** Both PRs change the record shape and the salt follow-up changes it again; with no records out there today all of it is free, but shipping the marker first turns the next two shapes into migrations.
- **`identityKey: true` records *that* an arm was used, not *which key*.** `signingIdentityOf(wallet)` returns whatever identity the wallet happens to have, so a repo restored on another seed signs with the wrong key silently — #738 stores the pubkey and refuses on inequality. Fixable here: store the pubkey.
- **Minted keys survive.** `randomSwapSecrets` / `fallbackSecrets` keep plaintext keys at rest for cannot-sign wallets, and the arm count goes to three; #738 reaches zero and one shape.
- **Expo / service-worker wallets take the identity arm** — working, but one key across every swap and no indices for a restore scan; #738 + `36ad22c` makes them actually allocate.
- **Ownership direction.** `AGENTS.md:20-31` on master — not #738's own section — says shared wallet behavior gets promoted into `ts-sdk` rather than copied outward.

What #738 should take from here: the cannot-sign check. Its `provisionRefundKey` calls `wallet.identity.xOnlyPublicKey()` unconditionally and returns a descriptor nobody can sign with, so the failure lands as a `TypeError` at push time; `signingIdentityOf`'s intent belongs in `contractSigner` as a typed refusal.

The real argument for this PR: it ships as `@arkade-os/swap` 0.0.4 against the published SDK, while #738 needs an `@arkade-os/sdk` release and `packages/swap` still pins `workspace:*`. If the wallet team is blocked this week that's a release-plumbing problem, not a design verdict.

pietro909:
Superseded by #738

PR #744, opened 2026-08-12
Derive static wallets' preimage from a public per-swap salt

pietro909:
**Stacked on #738** — review that first; this branch contains its commits. Retarget to `master` when it merges.

Closes the remainder of #737. #737 §2 proposed both halves — the wallet provides the sender key, and the preimage becomes per-swap derivable. #738 shipped the first and dropped the second after its "reachable preimage collision" finding. This is the second half, designed so that finding does not apply.

## The problem

After #738 a static (`SingleKey`) wallet still stores its preimage in the clear:

| Wallet | Sender key | Preimage | Secret at rest |
|---|---|---|---|
| HD | fresh child descriptor | `sha256(sign_det(TAG ‖ xonly ‖ u32le(0)))` | none |
| static | identity key, via `tr(pubkey)` | **`randomBytes(32)` → `preimageHex`** | **P** |

The HD arm gets per-swap uniqueness from a fresh key, which is what lets it pin the message index. A static wallet has one key, so deriving at a pinned index would hand every swap the *identical* preimage — one counterparty learning its own would learn all of them.

## The change

Uniqueness comes from the message instead. A second, salted derivation, with the existing one left byte-identical:

```
v1 (HD, unchanged)  msg = "Arkade-RFQ-Preimage-v1"             ‖ xonly(32) ‖ u32le(0)
v2 (salted, new)    msg = "Arkade-Contract-Preimage-Salted-v1" ‖ xonly(32) ‖ salt(32)

P = sha256(signSchnorrDeterministic(sha256(msg)))
```

`salt` is 32 random bytes minted per swap and persisted **in the clear** as `preimageSaltHex`. Knowing it yields nothing without the seed — that is the whole difference from the `preimageHex` it replaces. The record goes from carrying a per-swap *secret* to a per-swap *public* value, which is exactly what `signingDescriptor` already is on the HD arm.

`SingleKey` and `SeedIdentity` gain `signSchnorrDeterministic` (aux_rand = 0). `DeterministicSigner` already existed in the SDK from #738, so no new interface.

**Why the #738 finding does not apply here.** That finding is about deriving from the descriptor: `isPerArtifactDescriptor` tests *shape*, not freshness, so a custom `DescriptorProvider` returning a constant HD-shaped descriptor would collide. Here uniqueness comes from a value minted per swap by the code that stores it, not from any claim about the descriptor. A constant descriptor — static, custom-provider, or otherwise — still yields distinct preimages. The shape test only picks the arm, and a wrong answer from it is no longer a collision.

**Three arms, and the probe is the use.** `provisionClaimSecret` resolves: caller-supplied P → per-artifact descriptor → salted. It does not ask `isDeterministicSigner` and then act on the answer — `DescriptorIdentity` always exposes the method and only refuses at call time. It mints the salt and *actually derives*; only if that throws does it fall back to a stored random P. The v1 HD arm still raises loudly, because an HD descriptor whose wallet cannot sign deterministically is a broken wallet, not a fallback case.

The three `mustPersistPreimage` warnings in `rfq.ts` stop firing for static wallets with no code change, which is the observable outcome.

## Corridor-generic tag

`Arkade-Contract-Preimage-Salted-v1`, where v1 is RFQ-scoped, and the asymmetry is deliberate. The v1 tags *must* be per-corridor (`Arkade-RFQ-Preimage-v1` here, NArk's `Arkade-Boltz-Preimage-v1` there) because v1 pins its index, leaving the tag as the only separation between two corridors reaching the same key. v2 mints a salt per swap, so no two swaps share a message within a corridor or across two — the salt carries the separation and the tag names the layer.

arkade-os/dotnet-sdk#185 is the same defect in NArk, on the Boltz corridor. If this argument holds, both SDKs converge on one salted tag and NArk adds no `Arkade-Boltz-Preimage-Salted-v1`. Nothing is pinned or stored yet, so the tag can still be renamed on agreement.

## Record and repository

`AssetSwap` gains `preimageSaltHex?`; `AssetSwapRepository.version` goes 1 → 2 so external implementors recompile — a field-mapped backend that drops the salt leaves the swap unclaimable, exactly as one dropping `preimageHex` does. `DB_VERSION` is unchanged: swaps are stored whole, so an added optional field needs no migration and no backfill. Nothing already written needs rewriting either — the field is optional, and any row carrying a stored `preimageHex` or an HD descriptor keeps resolving through precedence steps 1 and 2 forever.

Two additions worth looking at:

- **`SwapSecretsProjection`** — the mapper's output as a named type that `AssetSwap` embeds, rather than three fields restated per record type. It costs nothing now and makes a record that omits one a compile error later — which is what keeps #740's `RfqSwapOrigin` honest when it rebases, since today it restates those fields behind a comment with nothing checking it.
- **`preimageForSwapRecord`** — #738 deleted `rfqSecretsOfRecord` with no replacement, so nothing in `packages/swap` reads a preimage back out of a record. This is that reader, and it verifies against `paymentHash`. The salted arm has two inputs that can be wrong (key and salt) where the HD arm had one, and a caller who forgets to pass the salt gets a *wrong* preimage from a wallet that can derive, not an error — which otherwise surfaces as an opaque script failure at claim time.

`contractPreimage`'s third parameter becomes `{ stored?, salt? }`. A break, but it is new in #738 and unreleased.

## Testing

`pnpm run build`, `pnpm run lint`, all three unit suites: ts-sdk 2572, swap 371, boltz-swap 614. **Plus the `ts-sdk` regtest e2e**, run against a live stack — the full `vhtlc` suite is green, including the four pre-existing tests.

### Against a real covenant (`test/e2e/vhtlc.test.ts`)

**`should claim with a preimage re-derived from a static wallet's seed and salt`** is the property end to end. Provision on a static wallet → bind a VHTLC to `hash160(P)` with the provisioned key as receiver → fund it → discard everything but `signingDescriptor`, `preimageSaltHex` and `paymentHash` → rebuild the wallet from the key alone → re-derive → claim. The server accepting the claim is the assertion: a covenant funded before the restore, unlocked by a preimage that existed nowhere in between.

It also pins the hash-form seam, which is the part unit tests structurally cannot reach — the record commits to `sha256(P)` while the covenant pins `ripemd160(sha256(P))`, so a derived P has to satisfy a form the record never names. Both are asserted against the real script.

**I verified the test can fail.** Tampering the `ConditionWitness` preimage gets `INVALID_SIGNATURE (18)` from arkd; the passing run is not a no-op.

A second case funds nothing and just shows two swaps on one static wallet producing unrelated covenants — same key and descriptor, different addresses and preimages.

The swap package's `preimageForSwapRecord` is not exercised on-chain: it is a thin composition over `contractPreimage` plus a hash comparison, unit-covered on every branch, and the swap RFQ regtest profile stubs its solver (`"Nothing here fills the swap — a fill needs a solver that pays"`), so it could not prove more there than the ts-sdk stack already does.

### Unit tests worth looking at

- **Distinctness** — three swaps on one static wallet produce three distinct salts, preimages and payment hashes off one descriptor. The regression test for the collision the design exists to avoid.
- **Determinism** — two independently constructed wallets on one key agree, given only the record's public fields.
- **Cross-arm separation** — the same key through both derivations yields different P, plus a message-level test that the two forms cannot collide.
- **The aux_rand trap** — `signSchnorrDeterministic` must not route through `signMessage`, whose schnorr branch draws a random aux_rand. That would still verify, so only an explicit comparison against a zero-aux signature catches it. Asserted on both identities, plus that `SeedIdentity`'s two deterministic paths agree.
- **Fallback** — covered for both a signer missing the method and one that throws at call time; they take different code paths, and neither may leave a salt on the record.
- **Cross-SDK vectors** — v1 unchanged byte-for-byte; v2 pinned as literals, keyed off raw private keys so a reimplementation needs no derivation path.
- **Behind the service worker** — a static SW wallet now takes the salted arm too, and page-side and worker-side derive one preimage for one descriptor. #738's test asserting the old stored-preimage behaviour is updated rather than deleted.

## Note on #740 and #743

**Neither is a rival to this PR; both are rivals to #738.** #740 defers the preimage half explicitly and its baseline arm stores P permanently. #743 arrives at this PR's preimage reasoning independently — never derive P from a reused key, attach a random one per swap, mandate `preimageHex` — and so keeps exactly the secret at rest that this removes. Both make the same key-provisioning choice against #738: a **marker** (`senderKey: "baseline"`, `identityKey: true`) resolved by handing back `wallet.identity` unverified, where #738 stores a real descriptor and checks `equalBytes(signer.xOnlyPublicKey(), deriveDescriptorLeafPubKey(descriptor))`. A marker records *that* an arm was used; a descriptor records *which key*, and only the second survives a restore onto the wrong wallet.

Both also conflict irreconcilably with #738 — all three rewrite `packages/swap/src/secrets.ts`, which #738 deletes. #740's *persistence* half is orthogonal and conflicts with nothing: `git merge-tree pr738 pr740` puts every conflict in the key-provisioning surface and none in `rfqRecord.ts`, `indexedDbRepository.ts` or `repository.test.ts`.

Suggested order: **#738 → this → #740's persistence rebased**, with #743 closed in #738's favour. That way `RfqSwapOrigin.preimageIndex` — the slot #740 reserves for a monotonic-counter derivation — never ships, and `preimageSaltHex` takes its place. A counter needs durable state a static wallet lacks, and rebuilding it as `1 + max(index)` restarts at 0 on a wiped repository and re-derives a live swap's P.

**One thing #738 should take from #743 first:** `provisionRefundKey` calls `wallet.identity.xOnlyPublicKey()` unconditionally, so a remote-signer or readonly identity gets a descriptor nobody can sign with and fails as a `TypeError` at push time. #743's `signingIdentityOf` catches that up front; it belongs in `contractSigner` as a typed refusal. This PR does not change that path either way.

---

## Summary by CodeRabbit

* **New Features**
  * Added salted preimage derivation for shared swap descriptors, with public salt persistence and payment-hash verification.
  * Added deterministic Schnorr signing for supported identities and keys.
  * Added swap-secret recovery utilities and public SDK exports.
* **Bug Fixes**
  * Improved secret recovery validation for missing, invalid, or tampered swap data.
* **Documentation**
  * Updated swap secret provisioning, persistence, migration, and recovery guidance.
* **Chores**
  * Updated swap repository schema versions.

### DIFF
--- a/packages/swap/src/indexedDbRepository.ts
+++ b/packages/swap/src/indexedDbRepository.ts
@@ -47,7 +47,7 @@ const txDone = (tx: IDBTransaction): Promise<void> =>
 /** Browser backend over the SDK's shared IndexedDB manager — the same
  * infrastructure the wallet already uses for its Boltz swap repository. */
 export class IndexedDbAssetSwapRepository implements AssetSwapRepository {
-    readonly version = 1 as const;
+    readonly version = 2 as const;
     // the promise, not the resolved database: openDatabase bumps a refcount on
     // every call including cache hits, while dispose closes once, so two
     // concurrent first calls would strand the refcount above zero and leak the
--- a/packages/swap/src/repository.ts
+++ b/packages/swap/src/repository.ts
@@ -30,12 +30,13 @@ export const marketsCacheKey = (network: string, registry: string) =>
  * subset queries.
  */
 export interface AssetSwapRepository extends AsyncDisposable {
-    readonly version: 1;
+    readonly version: 2;
 
     /** Insert or replace a swap by id. Store the record whole: `preimageHex`
-     * is secret-bearing — the only claim secret of a swap whose descriptor is
-     * not per-swap — and a field-mapped backend that drops it leaves the swap
-     * unclaimable. */
+     * and `preimageSaltHex` both leave the swap unclaimable if a field-mapped
+     * backend drops them — the first is the only claim secret of a swap whose
+     * signer cannot derive, the second the public input every other static
+     * wallet's preimage derives from. */
     saveSwap(swap: AssetSwap): Promise<void>;
     /** All stored swaps, in no particular order — `getAssetSwaps` is the
      * canonical newest-first read. */
@@ -53,7 +54,7 @@ export interface AssetSwapRepository extends AsyncDisposable {
 }
 
 export class InMemoryAssetSwapRepository implements AssetSwapRepository {
-    readonly version = 1 as const;
+    readonly version = 2 as const;
     private readonly swaps = new Map<string, AssetSwap>();
     private readonly scanned = new Set<string>();
     private readonly markets = new Map<string, MarketsCacheEntry>();
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,10 +71,15 @@ keys and never branch on wallet *type* — they ask the wallet and use what come
   wallet is HD, or calls `randomSecretKey()` for a signing/refund/claim key, is a defect: the
   wallet's identity key is the fallback, never a minted one.
 - Per-artifact secrets (swap preimages) key off the **descriptor's shape**, not the wallet's type:
-  an HD child descriptor is unique per artifact and can derive secrets deterministically; a bare
-  `tr(pubkey)` repeats across artifacts, so derived-from-key secrets would collide and must be
-  stored per artifact instead (see `packages/ts-sdk/src/wallet/contractSecrets.ts` and
-  `swapSecretsToRecord` in `packages/swap/src/store.ts`).
+  an HD child descriptor is unique per artifact and derives secrets from the key alone; a bare
+  `tr(pubkey)` repeats across artifacts, so that derivation would collide and the uniqueness comes
+  from a public per-artifact **salt** stored with the artifact instead. Only a signer that cannot
+  sign deterministically at all falls back to a stored secret. See
+  `packages/ts-sdk/src/wallet/contractSecrets.ts`.
+- The ban is on **key material and on anything that signs** — not on public per-artifact values. A
+  package may mint and store public material such as a derivation salt: it is an input the wallet
+  needs back, not a capability, and knowing it grants nothing without the seed. Apply the same test
+  to anything new: could a reader of the repository spend with it?
 
 ## Local Scratch Files
 
--- a/packages/swap/README.md
+++ b/packages/swap/README.md
@@ -436,34 +436,42 @@ descriptor, which is public, and `contractSigner(wallet, descriptor)` recovers t
 
 What each swap stores, and what is recoverable:
 
-| Wallet answers with | Spending key          | Preimage (when the leg needs one) | Secret at rest    |
-| ------------------- | --------------------- | --------------------------------- | ----------------- |
-| fresh HD descriptor | re-derives from seed  | derives deterministically         | none              |
-| static `tr(pubkey)` | the wallet's identity | random, stored on the record      | the preimage only |
+| Wallet answers with       | Spending key          | Preimage (when the leg needs one)     | Secret at rest    |
+| ------------------------- | --------------------- | ------------------------------------- | ----------------- |
+| fresh HD descriptor       | re-derives from seed  | derives deterministically             | none              |
+| static `tr(pubkey)`       | the wallet's identity | derives from a public per-swap salt   | none              |
+| a signer that cannot sign |                       |                                       |                   |
+| deterministically         | the wallet's identity | random, stored on the record          | the preimage only |
 
-The preimage split follows the **descriptor's shape**, not the wallet's type: an HD child
-descriptor is unique to its swap, so `sha256(sign_det(...))` is safe; a static descriptor is the
-same key for every swap, so a derived preimage would repeat across swaps — one solver learning its
-own preimage would learn every other swap's — and a per-swap random preimage is stored instead.
-`mustPersistPreimage` says which you got. A stored preimage is the one secret at rest in the
-design, and it is never a private key.
+The preimage split follows the **descriptor's shape**, not the wallet's type. An HD child
+descriptor is unique to its swap, so `sha256(sign_det(...))` over the key alone is safe. A static
+descriptor is the same key for every swap, so that derivation would repeat across swaps — one
+solver learning its own preimage would learn every other swap's — and the uniqueness has to come
+from the message instead: the SDK mints 32 random bytes per swap, signs a **salted** message, and
+stores the salt in the clear.
+
+**The salt is not a secret.** Knowing it yields nothing without the seed, which is the whole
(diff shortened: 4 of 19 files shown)


pietro909:
Thanks — one correction on the closure pass, then what changed.

**Findings 2 and 5 were already fixed in the commit this review ran against.** The review's `commit_id` is `73eb4de8`, and both fixes landed in `73eb4de8` itself ("tighten three edges found in review of #744"), so the prior-finding section graded against the parent:

- **Finding 2** — `packages/swap/src/store.ts:342` compares `hex.encode(sha256(preimage)) !== record.paymentHash.toLowerCase()`. Covered by "accepts an uppercase payment hash" in `swapSecretsRecord.test.ts`.
- **Finding 5** — `contractPreimage` length-checks `opts.stored` before returning it (`contractSecrets.ts:363`), throwing on anything but 32 bytes. The zero-length case is the one that motivated it: an empty array is truthy.
- **Finding 1** — the arm-3 catch no longer absorbs everything: `ForeignDescriptorError` and `WalletCannotSignError` rethrow, so only "cannot sign deterministically" is swallowed.

**Actionable item 2 is now done.** The residual transient-throw risk is documented at the `DeterministicSigner.signSchnorrDeterministic` declaration (`contractSecrets.ts:386`), which is where an external signer implementor meets the requirement — a throw that will not repeat costs the artifact a stored random preimage for its whole life, and nothing surfaces the downgrade to the caller.

Findings 3 and 4 are follow-ups, as you suggest. On 3 specifically: the mapper guard is unreachable today (`swapSecretsToRecord` only ever sees `provisionClaimSecret` output, whose arms are exclusive), so the real fix is making `ProvisionedClaimSecret` a discriminated union over the three arms rather than adding a runtime check — worth its own PR.

Rebased onto master now that #738 has merged; conflicts resolved in favour of keeping master's shared `isSigningIdentity` (the branch's local copy is gone) and master's stored-preimage length check, folded into the new `opts` signature. Build, lint and the full unit suite are green (3,594 tests).

tiero:
ACK for me @Kukks if you think the sal approach is right way

packages/swap/src/indexedDbRepository.ts
```diff
@@ -47,7 +47,7 @@ const txDone = (tx: IDBTransaction): Promise<void> =>
 /** Browser backend over the SDK's shared IndexedDB manager — the same
  * infrastructure the wallet already uses for its Boltz swap repository. */
 export class IndexedDbAssetSwapRepository implements AssetSwapRepository {
-    readonly version = 1 as const;
```
Kukks, line 50: why?

packages/swap/src/repository.ts
```diff
@@ -30,12 +30,13 @@ export const marketsCacheKey = (network: string, registry: string) =>
  * subset queries.
  */
 export interface AssetSwapRepository extends AsyncDisposable {
-    readonly version: 1;
```
Kukks, line 33: why

PR #746, opened 2026-08-12
feat(swap): persist RFQ swaps

pietro909:
Extracts the RFQ persistence half of #740, as planned in
https://github.com/arkade-os/ts-sdk/pull/740#issuecomment-5272993247. #740's other half — the swap
package minting its own signing keys — is superseded by #738, which moves key provisioning into the
SDK and deletes `packages/swap/src/secrets.ts` outright.

Rebased onto master after #738, #744 (salted preimage), #750 (corridor handlers) and #758 (the SQLite
and Realm backends) landed, and after `eda45298` released the pre-RFQ shape as
`@arkade-os/swap@0.0.5`.

## This PR's typecheck is red until #759 merges into its branch

Not a defect, and not something a rebase fixes. #758 landed the SQLite and Realm `AssetSwapRepository`
backends while the three RFQ methods here were in flight, so on the merged tree both are
`implements AssetSwapRepository` with `version = 2` and no RFQ methods — two `TS2416`s, and a
`TS2420` for the missing methods once the version matches. Neither branch conflicts textually with the
other; the incompatibility exists only in the merge, which is exactly what the interface `version`
marker is for.

**#759 carries those two backends' RFQ half and is based on this branch.** Merge it here first, and
this PR goes green before it goes to master — so master is never broken by the ordering. The
unit suite passes on this branch as it stands; it is the typecheck of those two files that fails.

## What lands

`AssetSwapRepository` gains `saveRfqSwap` / `getAllRfqSwaps` / `removeRfqSwap`, and the IndexedDB
backend a matching `rfqSwaps` object store — `DB_VERSION` 1 → 2, interface `version` **2 → 3**. One
storage seam, as `repository.ts` already intends. `0.0.5` shipped `version: 2` meaning "swaps, scan
cursor, markets, with `preimageSaltHex` on the record", so the RFQ methods have to be 3: an
implementor built against the published shape would otherwise satisfy a `version: 2` assertion while
missing `saveRfqSwap`, which is exactly what the marker exists to prevent.

`RfqSwap` is a live record, not a storage format — it holds derived `Uint8Array`s and a
`VHTLC.ScriptV2` class instance, and IndexedDB's structured clone strips prototypes. So `rfqRecord.ts`
adds the projection: `RfqSwapOrigin` (the immutable request-time facts) plus the manager's mutable
state, with `createRfqSwapRecord` / `updateRfqSwapRecord` to write it and `rebuildRfqSwap` to get the
live record back at boot.

**No covenant is stored.** Every RFQ lockup registers a contract row before its address can be
funded, and that row already holds the tree parameters, keyed by the script they derive. So the
rebuild takes the parameters from its caller — `lockupContractParams` reads the row — rather than
keeping a second copy to drift.

### The record shape

`RfqSwapOrigin` is `kind`, `lockupAddress`, `amount?` and one opaque `profile`. A corridor's own keys
live in the profile under two of its own:

- `profile.signer` — `signingDescriptor`, which wallet key signs this leg. Independent of any
  hashlock: it feeds `senderIdentityForSwapRecord` on the refund side, and a corridor that settles
  with no preimage at all still has a leg to sign.
- `profile.hashlock` — `paymentHash`, plus preimage material **only on legs we claim**.
  `lightning_send` carries a hash it can never open (P belongs to the payee) and its descriptor is a
  *refund* key.

Write both with `rfqSecretsProfile(secrets, paymentHash?)`; read them back with `rfqSignerOf` (any
leg) and `rfqClaimSecretOf` (claim legs only, `undefined` on `lightning_send`). Both readers answer
`undefined` only when the corridor has no such half and **throw** on a half that is present and
unusable — `preimageForSwapRecord` verifies only when the projection carries a `paymentHash`, so a
partial one would claim with an unverified preimage rather than fail.

That shape is the fix for a funds bug the rebase created: a static wallet's claim secret is a public
per-swap **salt**, the record hand-listed the secrets fields, and the salt reached no backend — a
lightning-receive swap round-tripped with a preimage that could not be re-derived, discovered at claim
time. A corridor with no hashlock now stores none rather than a fabricated one.

## The rebuild checks itself

`rebuildRfqSwap` decodes the record's `lockupAddress` and refuses when its pkScript disagrees with the
covenant rebuilt from the caller's parameters. The two reach the record independently — the row was
written from the covenant, the address came verbatim from the entry point — so the wrong row, or one a
field-mapped backend read back short a key, throws at restore instead of surfacing weeks later as a
refund that cannot be signed. Each corridor's `hydrate` refuses the same way for its own half: no
usable `paymentHash`, no `expectedAmount`, no L1 keys, no checkable `minConfirmations`, or L1 inputs
that do not reproduce `htlcAddress`.

## Also in this PR

- **A repository whose connection went away now reopens.** The manager closes the database on
  `versionchange`; the repository cached the closed handle, so every later transaction threw
  `InvalidStateError` with no path back. Unreachable before this PR, since the version never
  increased. Forgetting the connection recovers from an external delete or an eviction, and reports a
  database another tab upgraded past as `VersionError` — which names the reload.
- **A blocked upgrade fails instead of hanging** (`packages/ts-sdk`). A blocked open settles neither
  way, so the first repository call hung with no error and no timeout. It now rejects after 10s with
  the bookkeeping cleared so a retry is possible — and the request handlers are gated on that, since
  rejecting does not cancel the open: a late success would leak an untracked connection whose
  `versionchange` handler deletes its successor's cache entry.
- **`initDatabase` takes the migration arguments** it was being handed and dropping.
- **`MIGRATION.md` §3c** tells `arkade-os/wallet` how to adopt this: the three methods, how to fill
  `profile` per corridor, which reader answers on which leg, that non-hashlock corridors write no
  hashlock, that a partial profile throws at boot and at read, and that **no backfill exists** — RFQ
  swaps started by the released wallet were never persisted, so refunds for in-flight pre-upgrade
  swaps stay manual.

## Compatibility

**Additive for every consumer, breaking only for an implementor of `AssetSwapRepository` — of which
there are none.** The implementors are `InMemoryAssetSwapRepository`,
`IndexedDbAssetSwapRepository` and two test mocks; the wallet *constructs* the IndexedDB one. Existing
databases upgrade in place: the new store is added and the three original ones are untouched.

Two things to know:

- **Anyone who ran this branch before it merged should clear their `rfqSwaps` store.** Prerelease rows
  are in the pre-fix shape, and a static-wallet claim row that stored neither `preimageHex` nor
  `preimageSaltHex` cannot produce P at all.
- **`DB_VERSION` 2 is a one-way door.** Once a browser's database has upgraded, rolling the app back
  to `@arkade-os/swap@0.0.5` opens it at version 1 and fails `VersionError` across the whole swap
  store, not just the RFQ half. Forward is clean.

## Commits

- `fix(swap): renumber the repository interface for a released version 2`
- `fix(swap): move an rfq record's signer and hashlock into its corridor`
- `fix(swap): reopen an rfq repository whose connection was closed`
- `fix(sdk): fail a blocked database upgrade instead of hanging`
- `docs(swap): tell the wallet how to adopt rfq persistence`
- `test(swap): keep the rfq repository tests in their own backend matrix` — a rebase resolution, not a
  change of behaviour: rebasing onto #758 dropped the RFQ repository tests into that PR's
  SQLite-only `describe`, where `create` is not in scope, so five of them failed on a `ReferenceError`
  rather than on anything they assert. They now sit in a matrix of the backends that implement the
  methods; #759 folds them into the full four-backend one.

on top of the RFQ persistence commits themselves (records, the corridor handler registry, the
onchain-send leg, and the restore-time refusals).

## Test plan

`build` and `lint` clean; `test:unit` **3714 passed / 1 skipped** (ts-sdk 2625, swap 475, boltz-swap
614). `typecheck` reports the two backends above and nothing else — see the section at the top. The
new exports are verified present in the built `dist/index.d.ts`, and
`RfqSwapOrigin` there shows no top-level `paymentHash`, `signingDescriptor`, `preimageHex` or
`preimageSaltHex` — a stray one is how the old shape would come back.

Worth looking at specifically:

- **A static wallet's receive record hands P back** — request → record → `rfqClaimSecretOf` →
  `preimageForSwapRecord`, asserting the salt is stored, no preimage is, and the recovered P hashes to
  the stored `paymentHash`. Confirmed failing before the fix (`not-derivable`), which is the funds bug.
- **A dropped `paymentHash` throws** rather than reading back unverified, with companions for a
  63-character hash and a non-hex one, plus `rfqSignerOf` over an emptied `signer`.
- **A send-leg record has a signer and a payment hash and no preimage at all**, on both wallets, with
  `rfqClaimSecretOf` answering `undefined` — the property that keeps the three concerns apart.
- **No record carries a secret VALUE** — walks every string leaf and asserts none is the preimage or
  the wallet's key, rather than allowlisting names: the onchain leg's `claimKey`/`refundKey` are public
  64-hex values, so a "64 hex means secret" rule would reject a correct record.
- **A hashlock-free corridor round trip**, through the registry's test seam: usable `profile.signer`,
  no `profile.hashlock`, no preimage anywhere, and `rfqClaimSecretOf` **returning** `undefined` rather
  than throwing.
- **Mutation test over the projection, per leg** — perturbs each tree parameter in turn and fails if
  any one does not change the rebuilt `lockupPkScript`.
- **v1 → v2 migration** — seeds a genuine v1-shaped database, opens it through the repository at v2,
  and asserts swaps, scanned txids and the markets cache all survive while the new store is usable
  immediately.
- **The two connection tests**, each confirmed failing without its fix: recovery after an external
  delete, and `VersionError` (not `InvalidStateError`) repeatably when another tab upgraded past this
  bundle.
- **The blocked-upgrade tests**, likewise: the rejection names the database, and the connection the
  late open produced is closed. The second is the only assertion found that fails without the
  `settled` guard — every indirect probe is blind, because anything used to observe the leaked
  connection fires `versionchange` at it and closes it.

## Not verified

- The upgrade against a **real browser** rather than `fake-indexeddb`, which simulates neither a
  second tab nor eviction. Both connection fixes are reasoned, with unit coverage, not field-proven.
- The SQLite and Realm backends do not carry RFQ records on this branch — #759 does, and its tests
  are what exercise an RFQ record through their JSON boundary.
- The repository's `close`-event arm specifically has **no** unit coverage: `fake-indexeddb` never
  dispatches that event. Its `versionchange` arm is covered.
- **The swap package's tests are not typechecked** (`tsconfig.json` excludes `test`, as all three
  packages do), so the record-literal sweep in this PR was manual and the compiler will not catch the
  next field added to `RfqSwapRecord` or to a corridor profile.
- `RfqSwapCommon.paymentHash` still assumes every corridor has a hashlock. This PR removes that
  assumption from the stored record only; the live type follows when the first non-hashlock corridor
  lands.
- Nothing here has been exercised against a live solver.

## Left for follow-up

Nothing prunes retired records. `shouldRetainRfqSwap` and `removeRfqSwap` are the primitives and the
consumer sweeps with them at boot; that is stated in the README and MIGRATION.md rather than wired,
since nothing in this package owns a boot path.

## Summary by CodeRabbit

- **New Features**
  - Added persistent RFQ swap records with creation, updates, restoration, removal, and 30-day retention for completed swaps.
  - Added recovery from saved contract parameters with validation and missing-contract error handling.
  - Added access to Lightning and on-chain swap parameters, signer details, and claim-secret data.
  - Added safer persistence that excludes private keys and sensitive covenant data.

- **Breaking Changes**
  - Upgraded repository storage to version 3 with RFQ swap support and migration guidance.

- **Bug Fixes**
  - Added validation for malformed records, inconsistent contract data, and unsupported networks.
  - Improved recovery from blocked IndexedDB upgrades.


packages/swap/src/rfqRecord.ts
```diff
@@ -0,0 +1,310 @@
...
+/**
+ * Rebuild the live record. Pure: everything it needs is on the record.
+ *
+ * Hand the result to {@link RfqSwapManager.start}. The covenant is rebuilt the
+ * way it was made, and then checked against the address that was actually
+ * funded — see {@link assertRebuildMatchesLockup} — so the `lockupPkScript` it
+ * produces is the one the funded lockup is keyed by.
+ */
+export function rebuildRfqSwap(record: RfqSwapRecord): PersistableRfqSwap {
+    const script = rfqSwapCovenant(record);
```
tiero: **The self-check misses the two fields that authorize spending.**

`assertRebuildMatchesLockup` only exercises the tree parameters, because those are the only inputs to the covenant. `signingDescriptor` and `preimageHex` are on the record too, and neither is validated here or carried onto the rebuilt swap.

That leaves two ways to lose money that this function reports as success:

- Drop `signingDescriptor` and the rebuild passes. The failure arrives later as a refund that cannot be signed — the exact outcome the PR description says this check prevents.
- Drop `preimageHex` on a `lightning_receive` funded by a static wallet and the rebuild passes. That field is the only copy of P, so the claim is impossible and the solver reclaims at `refund_locktime`.

Two lines, reusing the SDK helper that already decided whether to store P:

```ts
required(record.signingDescriptor, "signingDescriptor");
// A static descriptor repeats across swaps, so P was minted and stored.
// An HD child re-derives it.
if (record.kind === "lightning_receive" && !isPerArtifactDescriptor(record.signingDescriptor)) {
    required(record.preimageHex, "preimageHex");
}
```

The tests do not cover this today, and cannot as written: the mutation loop asserts that changing a field changes `pkScript`, so it can only ever contain tree parameters. `signingDescriptor` and `preimageHex` appear only in the "carried through" assertions (lines 229, 230, 250). Those prove that a stored field comes back, not that a missing one is caught.

---

packages/swap/src/rfqRecord.ts
```diff
@@ -0,0 +1,310 @@
...
+/**
+ * Every later write. The origin half is carried through untouched.
+ *
+ * The mutable half is REPLACED, not merged. `managerState` omits a key the live
+ * swap no longer carries, so spreading it over the old record could only ever
+ * set these fields, never clear them. The manager clears them on purpose: it
+ * deletes `blockedReason` when a swap leaves `needs_counterparty`, precisely
+ * because a stale `blockedReason` reads as a live refusal.
+ */
+export function updateRfqSwapRecord(
```
tiero, line 216: **These two functions are one function.**

`createRfqSwapRecord` is `{ ...origin, ...managerState(swap) }`. `updateRfqSwapRecord` strips the four mutable keys and then does the same spread. Run the update version against a bare `RfqSwapOrigin` and the strip is a no-op, so it already covers the create case. One export would do.

There is also a maintenance cost in the split. `managerState`'s conditional keys and the destructure list here are two hand-written copies of "which fields are mutable". Add a fifth mutable field and only one of them gets updated — and the failure mode is a stale field that never clears, which is what the doc comment above says the strip exists to prevent.

---

packages/swap/src/repository.ts
```diff
@@ -41,6 +42,20 @@ export interface AssetSwapRepository extends AsyncDisposable {
...
+     * Insert or replace a monitored RFQ swap by `rfqId`.
+     *
+     * Store the record WHOLE. Every field is a covenant tree parameter or the
+     * manager's own state, and a field-mapped backend that drops one round-trips
+     * a record whose covenant `rebuildRfqSwap` cannot reproduce — which surfaces
+     * as a refund that cannot be signed, long after the write.
+     */
+    saveRfqSwap(record: RfqSwapRecord): Promise<void>;
+    /** Every stored RFQ swap record, in no particular order. */
+    getAllRfqSwaps(): Promise<RfqSwapRecord[]>;
```
tiero, line 65: **No way to read one record, but the update path needs one.**

`updateRfqSwapRecord(record, swap)` takes the previous record. The manager's callback is `saveSwap: (swap: RfqSwap) => Promise<void>`, which hands over only the live swap. So on every save the consumer has to get the old record from somewhere, and the only read here is `getAllRfqSwaps()`.

That means a full table scan per save, or the consumer quietly keeping its own map. Neither is written down.

Two ways out, either is fine:

- add `getRfqSwap(rfqId)` here, or
- put the origin on the live swap so `saveSwap` is self-sufficient — which also collapses the create/update pair.

---

packages/swap/src/rfqRecord.ts
```diff
@@ -0,0 +1,310 @@
...
+
+/**
+ * Whether a retired swap's record should be kept.
+ *
+ * `needs_counterparty` is deliberately not terminal and is never dropped: the
+ * money is still at the lockup, the counterparty's move is still what ends the
+ * swap, and the refusal is re-checked every pass; restoring the right wallet
+ * returns it to `pending`.
+ *
+ * @param now Current time in **unix seconds**; the same unit as
```
tiero, line 316: The seconds-versus-milliseconds warning is now in three places, down to the same "~43 minutes" figure: the `RFQ_SWAP_RETENTION_SECONDS` doc above, this `@param`, and the README.

Pick one home and point at it from the other two. Three copies of a number is three chances for two of them to go stale.

---

packages/swap/src/rfqRecord.ts
```diff
@@ -0,0 +1,310 @@
...
+    // ── Receive-leg tree parameters ──────────────────────────────────────────
+    /** VHTLC's `receiver` — the trader's own key on this leg. */
+    payoutPubkey?: string;
+    /** `nonInteractiveClaim`'s pinned destination. */
+    payoutPkScript?: string;
+    /** From `profile.solver_refund_pk_script` — the one tree parameter nothing
+     * else on the wire determines. */
+    solverRefundPkScript?: string;
+    /** Where the claim pays; kept for display and for rebuilding the claim. */
+    payoutAddress?: string;
```
tiero: The comment says this is kept "for display and for rebuilding the claim", but `rebuildRfqSwap` never reads it and `RfqSwapCommon` has no field for it. A consumer that needs it has to read it off the record itself.

Drop the second half, or say who actually uses it.

---

packages/swap/src/rfq.ts
```diff
@@ -773,7 +791,9 @@ export async function requestLightningSend(
 
     const serverPubkey = xOnly(hex.decode(info.signerPubkey), "ark signer key");
     const network = getNetwork(info.network as NetworkName);
-    const script = lightningSendVtxoScript({
+    // Named rather than inlined so the exact inputs the covenant was built from
```
tiero, line 819: This comment, and its near-twin at line 1433, explain that a variable was given a name. The `treeParams` doc on the return type already says why the caller gets these. I would drop both.

The rest of this file is the right shape, for what it is worth: naming the literal is the smallest change that could work, and `Parameters<typeof lightningSendVtxoScript>[0]` means the type cannot drift from the builder.

---

tiero (commented):
Storage seam and the self-checking rebuild are right; `rfq.ts` is minimal.

One gap worth blocking on, on `rebuildRfqSwap`: the check covers only tree parameters, so a dropped `signingDescriptor` or `preimageHex` rebuilds cleanly and fails later — an unsignable refund, or a receive leg with no copy of P. Two lines to fix. The mutation test can't catch it: it asserts `pkScript` changes, so it only ever holds tree parameters.

The rest is shape, in the inline comments: create/update are one function; the update path needs a record `getAllRfqSwaps()` can't fetch cheaply; nothing in `src/` calls any of this and the README never shows the flow.

That last point is why I'd swap the follow-up priorities — `rfqSwapOriginFrom` removes the error-prone step, the create/update pair doesn't.

tiero:
Every RFQ swap already registers a contract row: `params: VHTLCV2ContractHandler.serializeParams(script.options)`, keyed by its own script.

That's the same tree parameter set `RfqSwapOrigin` re-encodes by hand. `createScript(params)` already rebuilds it, and keying by script already makes it self-checking — `rfqRegister.test.ts` asserts exactly that. It sits behind `ContractRepository`: four backends plus a migration framework.

So `RfqSwapOrigin`'s tree half, `rfqSwapCovenant` and `assertRebuildMatchesLockup` look like duplicates of production-tested code.

What does need a home here: `rfqId`, manager state, `expectedAmount`, `signingDescriptor`, `preimageHex`. Not the covenant.

I take the point that `rebuildRfqSwap` is pure on purpose. Could it take the params as an argument, so the caller picks between the contract row and its own copy?

How much of this can come from the contract repository? @pietro909

Kukks:
Flagging a design problem in the record shape, which I wrote — most of `RfqSwapOrigin` duplicates what the contract manager already persists.

`registerLockupContract` writes the whole covenant:

```ts
await contracts.createContract({
    params: VHTLCV2ContractHandler.serializeParams(script.options),
    script: hex.encode(script.pkScript),
    address,
    ...
})
```

and `rfqRegister.test.ts` already pins the round trip — `VHTLCV2ContractHandler.createScript(row.params)` rebuilds a script whose `pkScript` equals `row.script`. So the contract row, keyed by pkScript, is a complete and durable description of the covenant.

`lockupContract.ts`'s own docstring states the split this was supposed to follow:

> The row carries script-level facts only. Per-swap identity and key material stay in the swap record: rows are keyed by script and first-writer-wins, so anything per-swap written here is stale from the second swap onward.

The record inverted that — it stores script-level facts too, and `rebuildRfqSwap` re-derives a covenant the contract manager can already hand back.

### What is actually redundant

| Field | Already on the contract row |
|---|---|
| `solverPubkey`, `serverPubkey`, `emulatorPubkey`, `refundLocktime`, `claimDelay`, `senderPubkey`, `payoutPubkey`, `refundPkScript`, `receiverPkScript`, `solverRefundPkScript`, `payoutPkScript` | yes — all in `params` |
| `lockupAddress` | yes — `address` |

### What genuinely needs the swap record

- `rfqId`, `state`, `createdAt`/`updatedAt`, `refundArkTxid`/`claimArkTxid`, `failure`, `blockedReason` — per-swap identity and manager state, nowhere on chain
- `signingDescriptor` / `preimageHex` — key material; the row is explicitly script-level only
- `expectedAmount` and the quoted amounts — quote facts, not covenant facts
- `kind` — both legs are `vhtlc-v2`, so the row cannot tell a send from a receive

### `paymentHash` is weaker than it looks

The covenant commits to `hash160(P)` (`Script.encode(["HASH160", preimageHash, "EQUAL"])`, 20 bytes, the conventional HTLC shape), so `sha256(P)` is genuinely not recoverable from it. But `readLockupFate` only needs to prove a witness item is the preimage, and `hash160(candidate) === row.params.preimageHash` proves exactly what `sha256(candidate) === paymentHash` proves. So `paymentHash` is a convenience for correlating the swap to the payer's BOLT11 invoice in history — not something the fate check requires.

### The counter-argument, stated plainly

Rebuilding from the contract row makes `getContracts` an input to resume, where the current design is a pure function of the record. I argued for purity when I wrote it, and I now think that was wrong: `registerLockupContract` **throws** before the caller ever holds an address to fund, so there is no state in which the contract store is unavailable but the swap record is. The dependency was never avoidable — storing the parameters twice only bought a second thing to keep in sync.

### Prior art

NArk goes further and stores the preimage on the HTLC contract itself. The row here has a `metadata` field, so that is expressible. The "first-writer-wins" caveat in the docstring does not bite for lockups — each swap's VHTLC has a unique `preimageHash`, so the script is per-swap by construction (which is what #744's preimage uniqueness protects).

Happy to do the rework if the direction is agreed: shrink the record to identity, manager state, quote facts and key material, and resolve the covenant through the contract manager.

Kukks:
Opened #748 with this as a diff, stacked on this branch — one commit, 401/401 green. Take it, argue it, or close it; the counter-argument to press on is that a swap can no longer be resumed from its record alone.

pietro909:
Agreed, and it went further than the tree half. Pushed in 57d9da5.

`RfqSwapOrigin` lost 11 fields — `solverPubkey`, `emulatorPubkey`, `serverPubkey`, `refundLocktime`, `claimDelay`, `senderPubkey`, `refundPkScript`, `receiverPkScript`, `payoutPubkey`, `payoutPkScript`, `solverRefundPkScript`. `rfqSwapCovenant` and `assertRebuildMatchesLockup` are gone with them, along with the `bytes()`/`required()` scaffolding that existed only to name a dropped tree parameter. `refundLocktime` also went: the rebuild reads it off `script.options`, since the covenant binds it.

Yes to the params as an argument — that is the shape:

```ts
const params = await lockupContractParams(await wallet.getContractManager(), record.lockupAddress);
const swap = rebuildRfqSwap(record, params);
```

`rebuildRfqSwap` stays pure and synchronous. `lockupContractParams` reads the row (by script — the row’s key) and throws `LockupContractMissing` when this wallet has none: a cleared contract store, or a record from elsewhere. A consumer that would rather not depend on the contract store keeps its own `VHTLCV2ContractHandler.serializeParams(script.options)` and passes that instead.

One correction to the list of what needs a home here: `paymentHash` stays too, and not as convenience. The covenant binds `ripemd160(sha256(P))` — one-way over the hash the record carries — so no rebuild recovers it. Same for `kind`. So the record is now `rfqId`, `kind`, `paymentHash`, `lockupAddress`, `signingDescriptor`, `preimageHex?`, `expectedAmount?`, `payoutAddress?`, `amount?`, plus manager state.

I kept one check, in a weaker form than the old one: the params are verified against the record’s `lockupAddress` before a swap is handed back. Not the self-consistency check `upsertContractRow` already enforces at write time — this catches the wrong row, or a params key a field-mapped backend dropped on the way *out*.

On tests, the "mutating any tree parameter changes the script" battery is gone; that property belongs to the row’s serialize/deserialize round trip and is covered in `vhtlcV2-handler.test.ts`. In its place `rfqRegister.test.ts` runs the loop end to end through a real `ContractManager` and repository: entry point writes the row, the row round-trips storage, and a record carrying no covenant rebuilds the funded script. The two "integrator copies this mapping" tests dropped from ~11 hand-written fields to 4-6.

Two things I deliberately left out of this PR:

- `contractManager.ts:1953` and `:1986` call `handler.createScript(contract.params)` without comparing the result against `contract.script`. Write time is guarded, read time is not, and that is wanted for every contract type — not just `vhtlc-v2`.
- `treeParams` still comes back from the entry points, now documented as optional to persist. Removing it is a public API break and a separate call.

Kukks:
Heads-up so we do not converge on this twice in one day: taking a swing at persisting the onchain-send leg, on a branch stacked here. Shout if you are already on it.

The gap, if it is useful before the diff lands: `OnchainSendSwap` is excluded from `PersistableRfqSwap` because its L1 half cannot be reproduced, and that holds — `OnchainHtlc` carries `address`/`pkScript`/`leaves`/`controlBlocks`, all derived, but not the `claimKey`/`refundKey` that `onchainHtlcScript` takes as inputs. And unlike the Arkade covenant there is no contract row to lean on: the HTLC is Bitcoin L1, not an Arkade contract, so nothing else can rebuild it.

So the onchain record needs its own persisted params — `claimKey`, `refundKey`, `refundLocktime`, network, `minConfirmations` — as a nested object on the record, plus `funding`/`claimTxid` in the mutable half.

Which also settles `treeParams`, I think: it is dead weight on the two lightning corridors now that the contract row owns their covenant, and it is the shape the onchain corridor actually needs and does not have. Planning to drop it from lightning send/receive and expose the L1 params on `requestOnchainSend` instead.

Kukks:
One field worth considering while this is still open: should the record carry the **funding txid**?

It carries `refundArkTxid` and `claimArkTxid` — the two spends — but nothing identifying the transaction that funded the lockup.

### Where it bites

#753 adds `swapActivityResolver`, which groups a swap's transactions into one activity so a failed swap stops rendering as an unrelated send plus an unrelated receive. It takes `{ rfqId, kind, state, txids }`, and defers the join to this branch:

> A correlation helper that derives these from the record store and the funding lockup's VTXOs lands separately.

That helper cannot be a pure mapping over the record as it stands. `refundArkTxid` and `claimArkTxid` come straight off it, but the funding txid has to be recovered by querying the indexer for VTXOs at the lockup script and picking the one that funded it. So history rendering — the read path, over records that may be months old and long since terminal — acquires a network round trip per swap to name a transaction the wallet itself created.

### The case for storing it

It is known at the exact moment the caller has it: funding is what the caller does between the entry point returning an address and `addSwap`. One string, no derivation, no ambiguity.

It is also the one identifier the wallet's own history is keyed by, which makes it the natural correlation key rather than something reconstructed from script queries.

### The case against, which is this branch's own rule

Do not store what something else owns. That is what shrank the record onto the contract row, and I would rather not undo it a field at a time.

The distinction I would draw: the covenant is *owned and durably stored* by the contract manager, so the record pointing at it costs nothing. The funding txid is owned by the chain, and the local route to it is an indexer query keyed by script — not a lookup in a store the wallet already keeps. Those are different kinds of "something else owns it".

The counter I can see is that `spendTxidsOf`/`findLockupVtxos` already do exactly that query for the manager's own passes, so the cost may be near zero in practice and the helper could reuse a result the manager already has.

No strong view on where it lands — flagging it because it is cheap to add now and awkward to add after the store ships, and because the read path is the one place a network dependency is least expected.

tiero:
Yes—reviewing current head `279366b`:

- **a) It could be materially simpler.**
- **b) It reuses parts of the SDK storage system, but not as much as it could.**

I would not add the separate `rfqSwaps` repository/store. Every RFQ already requires a unique persisted contract row, and that row already provides:

- Cross-platform IndexedDB, SQLite, Realm, and memory backends.
- Covenant parameters and address.
- `metadata: Record<string, unknown>` specifically for integration data. [`[Contract](https://github.com/arkade-os/ts-sdk/blob/201368189f25c1e34d56d9cdfdef0b527cb89e5d/packages/ts-sdk/src/contracts/types.ts#L99-L139)`](https://github.com/arkade-os/ts-sdk/blob/201368189f25c1e34d56d9cdfdef0b527cb89e5d/packages/ts-sdk/src/contracts/types.ts#L99-L139)
- Filtered reads and updates through `IContractManager`. [`[updateContract](https://github.com/arkade-os/ts-sdk/blob/201368189f25c1e34d56d9cdfdef0b527cb89e5d/packages/ts-sdk/src/contracts/contractManager.ts#L417-L425)`](https://github.com/arkade-os/ts-sdk/blob/201368189f25c1e34d56d9cdfdef0b527cb89e5d/packages/ts-sdk/src/contracts/contractManager.ts#L417-L425)
- Metadata round-tripping in [SQLite](https://github.com/arkade-os/ts-sdk/blob/201368189f25c1e34d56d9cdfdef0b527cb89e5d/packages/ts-sdk/src/repositories/sqlite/contractRepository.ts#L112-L130) and [Realm](https://github.com/arkade-os/ts-sdk/blob/201368189f25c1e34d56d9cdfdef0b527cb89e5d/packages/ts-sdk/src/repositories/realm/contractRepository.ts#L82-L100) already. [SQLite](https://github.com/arkade-os/ts-sdk/blob/201368189f25c1e34d56d9cdfdef0b527cb89e5d/packages/ts-sdk/src/repositories/sqlite/contractRepository.ts#L112-L130), [Realm](https://github.com/arkade-os/ts-sdk/blob/201368189f25c1e34d56d9cdfdef0b527cb89e5d/packages/ts-sdk/src/repositories/realm/contractRepository.ts#L82-L100)

The PR already writes RFQ identification into contract metadata. [`[registerLockupContract](https://github.com/arkade-os/ts-sdk/blob/279366bf0623f4d6d3f5a4d0d856d9c5a6198154/packages/swap/src/lockupContract.ts#L104-L117)`](https://github.com/arkade-os/ts-sdk/blob/279366bf0623f4d6d3f5a4d0d856d9c5a6198154/packages/swap/src/lockupContract.ts#L104-L117) I would extend that metadata with a namespaced `rfqSwap` record and expose small helpers over `IContractManager`:

```ts
saveRfqSwap(contracts, record)
getAllRfqSwaps(contracts)
removeRfqSwap(contracts, record)
```

That would eliminate:

- The three new `AssetSwapRepository` methods. [[Current additions](https://github.com/arkade-os/ts-sdk/blob/279366bf0623f4d6d3f5a4d0d856d9c5a6198154/packages/swap/src/repository.ts#L55-L67)](https://github.com/arkade-os/ts-sdk/blob/279366bf0623f4d6d3f5a4d0d856d9c5a6198154/packages/swap/src/repository.ts#L55-L67)
- The new IndexedDB object store and database version bump. [[Store definition](https://github.com/arkade-os/ts-sdk/blob/279366bf0623f4d6d3f5a4d0d856d9c5a6198154/packages/swap/src/indexedDbRepository.ts#L11-L28)](https://github.com/arkade-os/ts-sdk/blob/279366bf0623f4d6d3f5a4d0d856d9c5a6198154/packages/swap/src/indexedDbRepository.ts#L11-L28)
- The required SQLite/Realm follow-up PR #759.
- Most of the blocked-upgrade work from this feature PR; that useful fix can land separately.
- A second record whose lifecycle must remain aligned with the contract row it already cannot recover without.

The helper must merge existing metadata because `updateContract` replaces nested metadata. The one-record-per-contract assumption is sound here only because each RFQ VHTLC has a unique preimage hash.

Further simplification findings:

`packages/swap/src/rfqCorridor.ts:L45-158: yagni: internal registry for three closed manager kinds, with casts, side-effect registration and test mutation. Use an exhaustive switch over a discriminated union.`

`packages/swap/src/rfq.ts:L684: delete: Lightning send/receive treeParams are now redundant because the contract row owns those parameters. Keep only onchain_send’s L1 htlcParams, which has no contract row.`

`packages/swap/src/rfqRecord.ts:L89-127: shrink: opaque Record<string, unknown> makes callers manually construct a fragile profile. Export a typed result-to-record factory so the wallet cannot omit signer, salt, amount, or L1 fields.`

I would keep the JSON-safe record projection, secret/salt handling, reconstruction checks, and on-chain HTLC parameters. Those solve real problems; the avoidable complexity is mainly the parallel persistence system and registry machinery.

`net: -300 lines possible.`

PR #748, opened 2026-08-13
refactor(swap): let the contract row own the covenant

Kukks:
Stacked on #746 — review that first; this is one commit on top of its branch.

Follow-up to [my comment there](https://github.com/arkade-os/ts-sdk/pull/746#issuecomment-5278877143), as a diff rather than an opinion. Take it, argue it, or close it.

## The problem

`RfqSwapOrigin` stores every covenant tree parameter and `rebuildRfqSwap` reconstructs the script from them. `registerLockupContract` already persists exactly those parameters:

```ts
await contracts.createContract({
    params: VHTLCV2ContractHandler.serializeParams(script.options),
    script: hex.encode(script.pkScript),
    address,
    ...
})
```

keyed by the pkScript, at request time, throwing before the caller holds an address to fund if it cannot. `rfqRegister.test.ts` already pins the round trip — `VHTLCV2ContractHandler.createScript(row.params)` rebuilds a script whose pkScript equals `row.script`.

`lockupContract.ts` states the division this restores:

> The row carries script-level facts only. Per-swap identity and key material stay in the swap record.

## The change

Eleven fields leave `RfqSwapOrigin`: `solverPubkey`, `serverPubkey`, `emulatorPubkey`, `claimDelay`, `refundLocktime`, `senderPubkey`, `refundPkScript`, `receiverPkScript`, `payoutPubkey`, `payoutPkScript`, `solverRefundPkScript`.

What stays is what the row cannot answer — the pointer back to it (`lockupScript`, `lockupAddress`), quote facts (`paymentHash`, `expectedAmount`, `payoutAddress`), key material (`signingDescriptor`, `preimageHex`), and `amount` for display.

`rebuildRfqSwap` becomes async, resolves the covenant through `getContracts({ script })`, and takes `refundLocktime` off the covenant itself. It **refuses when no row is registered** rather than inventing a covenant from stored parameters: a record whose contract was never registered, or whose store was cleared, describes a lockup nothing is watching, and that should be loud.

`assertRebuildMatchesLockup` still applies — `lockupAddress` is stored independently of the row, so the rebuild keeps checking itself.

## On the argument I made when I wrote this

I argued for a pure function of the record, to avoid an ambient dependency. That was wrong. `registerLockupContract` throws before funding, so there is no state in which the row is unavailable but the record is. The dependency was never avoidable; the second copy only bought something to keep in sync, free to disagree with the contract the wallet is watching.

## The cost, stated plainly

A swap can no longer be resumed from its record alone — `getContracts` is now an input. That is a real reduction in independence and the thing to push back on if you disagree. I think one source of truth for the covenant beats two that can disagree, and a record whose row is missing describes a lockup nothing is watching anyway.

## Tests

401/401, typecheck and lint clean.

Two tests changed shape rather than being deleted for convenience, and it is worth being explicit about which:

- The nine-parameter mutation test and the two "a record built from the result rebuilds the covenant" round trips existed because a consumer hand-mapped tree parameters into the record — the only place one could go missing. With the covenant resolved from the row there is no hand-mapping, so that failure mode is unreachable rather than untested.
- What replaces them: the record carries no covenant facts at all, and the rebuild refuses an unregistered lockup.

The consumer mapping in `rfqDerivedSecrets.test.ts` and `rfqReceive.test.ts` is still written longhand, because it is what an integrator copies — it is just much shorter now.


pietro909:
We converged on this independently — 57d9da5 landed on `feat/rfq-persistence-only` about an hour after you opened this, which is why it now shows conflicting. Same eleven fields leave `RfqSwapOrigin`, same list, and both of us ended up on the same `Number(script.options.refundLocktime)` line for the same reason.

Your write-up is better than mine in one place, and I have taken it: the covenant commits to `hash160(P)` — 20 bytes, the conventional HTLC shape — and `paymentHash` stays to correlate a swap to the payer's invoice, not for the fate check, which reads the covenant's own `preimageHash` against `hash160(candidate)`. That is now the doc on the field, in 29c73e7 and credited here.

Three places the landed version differs, so you can push back on any of them.

**The seam.** Yours makes `rebuildRfqSwap` async and resolves the row inside; the landed one stays sync over `params`, with `lockupContractParams(contracts, lockupAddress)` as a separate reader that throws a typed `LockupContractMissing`. This is [tiero's question](https://github.com/arkade-os/ts-sdk/pull/746#issuecomment-5278423825) answered as asked — "could it take the params as an argument, so the caller picks between the contract row and its own copy". Your safety property survives it: a record whose row is gone still refuses loudly rather than inventing a covenant, because the reader throws before the rebuild is reached. What is preserved is the ability for a consumer to keep its own `serializeParams` copy and pass that instead, which your signature removes. The cost is one extra line at the call site.

I do not think your recantation is wrong, incidentally — the dependency genuinely was not avoidable. I read the pure core plus an explicit reader as the same conclusion with the choice left open, rather than as a defence of the old design.

**The pointer.** You add `lockupScript` to the record; the landed version decodes `lockupAddress`, which already determines it. A stored second copy of a derivable value is the thing this PR argues against, and it is only caught by the `assertRebuildMatchesLockup` you kept.

**Tests.** You drop the mutation battery and both round trips as unreachable, which is defensible — but it leaves the row-to-record loop unproven. The landed version replaces them with the loop end to end through a real `ContractManager` and repository in `rfqRegister.test.ts`: entrypoint writes the row, the row round-trips storage, a record carrying no covenant rebuilds the funded script. 408 tests.

One thing to fix wherever this ends up: `packages/swap/arkade-os-swap-0.0.3.tgz`, 192KB, `create mode 100644`. Looks accidental.

Closing this as superseded seems right to me, but it is your call — if you want the async signature instead, argue it here and I will rebase to it rather than have you redo the work.

Kukks:
Closing as superseded — 57d9da5 is the better version and I am not arguing for the async signature.

**The seam.** Yours is better and I withdraw mine. A sync core over `params` plus `lockupContractParams` as an explicit reader answers tiero's question as asked, keeps the loud refusal, and leaves a consumer free to pass its own `serializeParams` copy — which my signature quietly removed. One extra line at the call site is the right price for that. I read your framing of the recantation the same way: the dependency was unavoidable, but where the caller resolves it is a separate choice and worth leaving open.

**The pointer.** You are right, and it is this PR's own argument used against it: `lockupScript` is derivable from `lockupAddress`, so storing both is exactly the second copy I spent the description objecting to — caught only by the self-check I kept. I did not notice I had reintroduced the pattern one field lower down.

**Tests.** Also right. "Unreachable rather than untested" was true of the hand-mapping, but it left the row-to-record loop with nothing covering it, and an end-to-end pass through a real `ContractManager` and repository is a stronger replacement than what I removed. That is the gap I should have filled instead of just deleting.

**The tarball is mine.** `packages/swap/arkade-os-swap-0.0.3.tgz` is a local build I packed to point a wallet branch at this work, swept in by an `add -A` because it is not gitignored. It never reached `feat/rfq-persistence-only` — only this branch — so closing removes it. Worth a `*.tgz` line in `packages/swap/.gitignore` regardless, since `pnpm pack` drops it exactly there and the next person will hit the same thing.

Nothing else here that is not already in 57d9da5 or 29c73e7.

PR #749, opened 2026-08-13
fix(swap): level the two-signature refund with claim, and give the solo refund real headroom

Kukks:
Tracks the reference solver's [lightning-swap-service#81](https://github.com/arkade-os/lightning-swap-service/pull/81), merged as `c904d44`. Without this, a trader on this package derives a different taproot tree than the solver quotes and refuses every swap.

## Why it propagates at all

These delays are **not carried on the wire**. The trader and the solver each derive them from the same operator `/v1/info`, which is deliberate — a wire-carried delay is a delay the solver gets to choose. The price is that the derivation itself *is* the wire format, so changing it on one side is a protocol break wearing the clothes of a one-file refactor.

## What changed

The ladder spaced all three unilateral leaves one 512s unit apart, as though they were interchangeable rungs. They are not — each times a **different party's** recourse:

| leaf | who can spend it |
| --- | --- |
| `unilateralClaim` | the receiver **alone**, holding the preimage |
| `unilateralRefund` | sender **and** receiver — needs both signatures |
| `unilateralRefundWithoutReceiver` | the sender **alone**, needing nobody |

Only the last is a solo path for the funder, so it is the only one whose timing can steal: a funder able to refund before the claimant can claim takes money from someone who holds the preimage and did nothing wrong.

**So the solo refund gets real headroom** — `SOLO_REFUND_HEADROOM_SECONDS`, 4096s, newly exported — sized for what a claimant actually has to do in that window with the server gone: an unroll broadcast per chain step, each waiting on a confirmation, then the CSV spend. One 512s tick was never enough for that.

**And claim sits level with the two-signature refund.** Separating them bought nothing, since neither party can spend a two-signature leaf alone, while spending 512s of the headroom that does matter.

`unilateralClaimDelay`'s BIP68 ceiling reserved two granularity steps for the stacked tiers; it now reserves the full headroom. This guard alone is **not** mirrored in the solver, which still rejects only above `0xffff * 512`: for an operator `unilateralExitDelay` in `(33549824, 33553920]` seconds the trader throws here while the solver quotes and then fails deeper in its own script build. Both refuse, at different seams — diagnosability, not fund risk, and the window is ~388 days wide so it is unreachable in practice.

## The property had no test on this side

Both refund tiers' CSV bytes were the one part of the tree nothing asserted. The golden test pinned the scriptPubKey and the claim leaf's sequence, but described the two refund tiers only in comments — so the ladder could drift silently, which is exactly how it drifted. There are now six property tests (`test/unilateralLadder.test.ts`, ladders from 512s to a week) asserting: the solo refund opens strictly after the claim; the gap is at least the headroom rather than a single tick; claim and the two-signature refund are level; every value lands on a BIP68 boundary; nothing opens before the operator's own exit delay; and the headroom equals the reference solver's. Plus explicit CSV pins on both refund leaves in the golden test.

## How the new goldens were produced

**Not from this package** — that would assert that my code equals my code, green and blind to the drift the test exists to catch.

They were generated from the reference solver's own `CovenantSwapScript` at `c904d44`, driven by the merged `deriveUnilateralDelays`. The generator was first run against the **pre-change** ladder (4096/4608/5120) as a negative control, and reproduced both currently-pinned scriptPubKeys exactly:

- send `51200370b2a6…` ✔
- receive `5120f683cdac…` ✔

Only then was its output for the new ladder taken as truth. The changed code in this PR then derived the same bytes independently:

| leg | old | new |
| --- | --- | --- |
| send | `51200370b2a6…` | `51209e4ec65c3dc94c8046e7ef50258b69a88d61720df5b87bd0069576859e22ab89` |
| receive | `5120f683cdac…` | `5120a7a178e3751648bfbcca49731d98d103c805812e2cb6d7a89563eacb3a5dc415` |

CSV sequences: `unilateralRefund` `03090040b275` (4608s) → `03080040b275` (4096s); `unilateralRefundWithoutReceiver` `030a0040b275` (5120s) → `03100040b275` (8192s).

## Verification

Baseline captured on `master` at `c8b5c3b`/`c8d114be` before any edit: `@arkade-os/swap` **22 files, 368 tests, exit 0**.

- `pnpm run test:unit` (all packages): ts-sdk **162 files / 2595 passed**, swap **23 files / 374 passed**, boltz-swap **23 files / 614 passed**. Swap delta is +1 file / +6 tests — exactly the new file, no regressions.
- `pnpm run lint`, `pnpm run build`, `pnpm -C packages/swap run smoke:dist`, `npm pack --dry-run`, and all three `typecheck`s: exit 0.
- Negative control on the new test: it fails on the old ladder for the right reasons — gap `1024` vs expected `4096`, refund one tick above claim, headroom constant undefined.

## Deployment

Coordinated, on the same terms as the previous address-moving change documented in the README: trader and solver derive the lockup independently and compare, so for a quote **not yet funded** a version mismatch refuses it at `verifyLockupAddress` rather than losing funds. Upgrade both sides before expecting fills.

**An in-flight lockup funded before the upgrade needs care** (raised in review, fixed in `ad262086`). The delays are neither quote fields nor persisted — `AssetSwap` keeps `swapPkScript`, not `claimDelay` — and `RfqSwap`'s doc tells callers to *rebuild* the script on restart from the quote's binding fields, which re-derives the delays under whatever ladder is compiled in. A trader who funded on `0.0.4`, upgraded and restarted therefore rebuilds a **new** address, and `refundIfUnresolved` finds no VTXOs there and returns `nothing_to_refund` — terminal-sounding, for money still locked at the old script with `refundLocktime` ticking. Drain in-flight lockups before upgrading, or rebuild the old script from the pre-upgrade delays by hand. Pre-existing gap that any address-moving change hits; persisting the delays is the real fix and is deliberately not attempted here.

Consumers vendoring a built tarball (the wallet vendors `arkade-os-swap-0.0.1-pr691-752ed57.tgz`) need a rebuild from this commit — that tarball still carries the old `+512`/`+1024` math.

## Not verified here

`SOLO_REFUND_HEADROOM_SECONDS` is reasoned, not measured — nothing has yet driven a unilateral exit to completion. It should be re-derived from a real exit rather than trusted indefinitely; the same caveat is recorded on the solver side.

## Summary by CodeRabbit

* **Breaking Changes**
  * Updated unilateral refund timing rules, including a 4,096-second solo-refund headroom.
  * Tightened maximum supported claim delays and validation for incompatible timing configurations.
  * Derived swap addresses and quotes may change; coordinated deployment is required, and mismatched quotes are rejected.

* **New Features**
  * Added the publicly available `SOLO_REFUND_HEADROOM_SECONDS` timing constant.

* **Bug Fixes**
  * Corrected refund-delay calculations to follow the revised claim and solo-refund schedule.

* **Tests**
  * Added coverage for timing order, alignment, minimum delays, and boundary conditions.

### DIFF
--- a/packages/swap/README.md
+++ b/packages/swap/README.md
@@ -521,10 +521,29 @@ The package is pre-release; these notes replace a changelog for consumers tracki
   never enough for a claimant to complete a unilateral exit in. The two-signature refund needs no
   separation at all, since neither party can spend that leaf alone. This tracks the reference
   solver's [lightning-swap-service#81](https://github.com/arkade-os/lightning-swap-service/pull/81);
-  the two derivations must agree exactly. **Deployment must be coordinated** on the same terms as
-  the entry below: a mismatch refuses every quote at `verifyLockupAddress` rather than losing funds.
-  `unilateralClaimDelay`'s BIP68 ceiling tightened to match (it now reserves the full headroom
-  rather than two steps).
+  the two derivations must produce **the same three delay values** for the same operator, which is
+  what keeps the derived addresses identical. **Deployment must be coordinated** on the same terms
+  as the entry below: for a quote not yet funded, a mismatch refuses it at `verifyLockupAddress`
+  rather than losing funds.
+
+  **An in-flight lockup funded before the upgrade needs care, and the entry below understates
+  this.** The delays are not quote fields and are not persisted on the swap record
+  (`AssetSwap` keeps `swapPkScript`, not `claimDelay`), and `RfqSwap`'s own doc tells callers to
+  *rebuild* the script on restart from the quote's binding fields — which re-derives the delays
+  under whatever ladder is compiled in. So a trader who funded on `0.0.4`, upgraded, and restarted
+  rebuilds a **new** address, and `refundIfUnresolved` finds no VTXOs there and returns
+  `nothing_to_refund` — a terminal-sounding answer for money still locked at the old script, with
+  `refundLocktime` still ticking. Until the delays are persisted and rebuilt from the stored value,
+  drain in-flight lockups before upgrading, or rebuild the old script from the pre-upgrade delays
+  by hand. This is a pre-existing gap that any address-moving change hits, not one this change
+  introduces.
+
+  `unilateralClaimDelay`'s BIP68 ceiling tightened to reserve the full headroom rather than two
+  steps. Note this guard alone is **not** mirrored in the reference solver, which still rejects
+  only above `0xffff * 512`: for an operator `unilateralExitDelay` in `(33549824, 33553920]`
+  seconds the trader throws here while the solver quotes and then fails deeper in its own script
+  build. Both refuse, at different seams with different messages, so it is a diagnosability wart
+  rather than a fund risk — and the window is unreachable in practice (~388 days).
 
 - **`secrets.ts` is gone; key provisioning moved into `@arkade-os/sdk`.** This package no longer
   derives, mints, or names keys. It asks the SDK for what the leg needs — `provisionRefundKey(wallet)`
--- a/packages/swap/README.md
+++ b/packages/swap/README.md
@@ -512,6 +512,20 @@ after heavy swap use.
 
 The package is pre-release; these notes replace a changelog for consumers tracking the branch.
 
+- **Every derived address changed again, in both corridors — the unilateral ladder was re-spaced.**
+  `unilateralRefundDelay` now sits **level with** `claimDelay` instead of one 512s step above it,
+  and `unilateralRefundWithoutReceiverDelay` sits `SOLO_REFUND_HEADROOM_SECONDS` (4096s, newly
+  exported) above it instead of two steps. The old ladder spaced all three leaves one step apart as
+  though they were interchangeable rungs; they are not. Only `unilateralRefundWithoutReceiver` is a
+  solo path for the funder, so it is the only one whose timing can steal, and one 512s tick was
+  never enough for a claimant to complete a unilateral exit in. The two-signature refund needs no
+  separation at all, since neither party can spend that leaf alone. This tracks the reference
+  solver's [lightning-swap-service#81](https://github.com/arkade-os/lightning-swap-service/pull/81);
+  the two derivations must agree exactly. **Deployment must be coordinated** on the same terms as
+  the entry below: a mismatch refuses every quote at `verifyLockupAddress` rather than losing funds.
+  `unilateralClaimDelay`'s BIP68 ceiling tightened to match (it now reserves the full headroom
+  rather than two steps).
+
 - **`secrets.ts` is gone; key provisioning moved into `@arkade-os/sdk`.** This package no longer
   derives, mints, or names keys. It asks the SDK for what the leg needs — `provisionRefundKey(wallet)`
   for a leg it funds, `provisionClaimSecret(wallet, { preimage? })` for one it claims — and
--- a/packages/swap/src/rfq.ts
+++ b/packages/swap/src/rfq.ts
@@ -551,6 +551,25 @@ export const relayTransport = (
 /** BIP68 sequence granularity; the delay derivation rounds up to it. */
 const SEQUENCE_GRANULARITY_SECONDS = 512;
 
+/**
+ * How long the sender's SOLO refund opens after the receiver's claim, seconds.
+ *
+ * This is the window in which a claimant holding the preimage must be able to
+ * finish taking their money before the funder could take it back. On a live
+ * Arkade server that is one collaborative spend; with the server gone it is a
+ * full unilateral exit — an unroll broadcast per chain step, each waiting on a
+ * confirmation, then the CSV spend.
+ *
+ * 4096s (eight 512s units, ~68 minutes) is sized for that worst case. It is
+ * REASONED, not measured, and it mirrors `SOLO_REFUND_HEADROOM_SECONDS` in the
+ * reference solver's `src/core/timelocks.ts` — the two must move together or a
+ * trader derives an address the solver never quoted.
+ *
+ * A multiple of the granularity on purpose: BIP68 would round anything else,
+ * making the encoded timelock differ from the number written here.
+ */
+export const SOLO_REFUND_HEADROOM_SECONDS = 8 * SEQUENCE_GRANULARITY_SECONDS;
+
 /** The solver's unilateral-claim delay, derived from the Ark server's reported
  * exit delay exactly as the reference solver derives it — both sides read the
  * SAME server, so the derivation (not a quote field) is what keeps the two
@@ -564,12 +583,15 @@ export const unilateralClaimDelay = (serverExitDelaySeconds: number): number => 
             `server exit delay must be at least ${SEQUENCE_GRANULARITY_SECONDS}s of seconds, got ${serverExitDelaySeconds}`,
         );
     }
-    // two granularity steps below BIP68's ceiling, not at it: the refund tiers
-    // add one and two steps on top of this value, and every tier must encode
-    if (serverExitDelaySeconds > (0xffff - 2) * SEQUENCE_GRANULARITY_SECONDS) {
+    // the headroom below BIP68's ceiling, not at it: the solo refund stacks
+    // SOLO_REFUND_HEADROOM_SECONDS on top of this value, and it must encode too
+    if (
+        serverExitDelaySeconds >
+        0xffff * SEQUENCE_GRANULARITY_SECONDS - SOLO_REFUND_HEADROOM_SECONDS
+    ) {
         throw new Error(
             `server exit delay ${serverExitDelaySeconds}s exceeds what BIP68 can encode ` +
-                `once the two refund tiers are stacked above it`,
+                `once the solo refund's headroom is stacked above it`,
         );
     }
     return (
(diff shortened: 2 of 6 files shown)


packages/swap/README.md
```diff
@@ -512,6 +512,20 @@ after heavy swap use.
...
+  `unilateralRefundDelay` now sits **level with** `claimDelay` instead of one 512s step above it,
+  and `unilateralRefundWithoutReceiverDelay` sits `SOLO_REFUND_HEADROOM_SECONDS` (4096s, newly
+  exported) above it instead of two steps. The old ladder spaced all three leaves one step apart as
+  though they were interchangeable rungs; they are not. Only `unilateralRefundWithoutReceiver` is a
+  solo path for the funder, so it is the only one whose timing can steal, and one 512s tick was
+  never enough for a claimant to complete a unilateral exit in. The two-signature refund needs no
+  separation at all, since neither party can spend that leaf alone. This tracks the reference
+  solver's [lightning-swap-service#81](https://github.com/arkade-os/lightning-swap-service/pull/81);
+  the two derivations must agree exactly. **Deployment must be coordinated** on the same terms as
+  the entry below: a mismatch refuses every quote at `verifyLockupAddress` rather than losing funds.
```
pietro909: **medium — "refuses every quote at `verifyLockupAddress` rather than losing funds" is only true for quotes not yet funded; an already-funded in-flight lockup silently reports as empty after the upgrade.**

The pre-funding claim holds — all four entrypoints do call `verifyLockupAddress` (`rfq.ts:812, 1027, 1437, 1658`). The gap is recovery.

`claimDelay` is not a quote field and is not persisted: `AssetSwap` (`store.ts:33-77`) stores `swapPkScript` but no delays, and `RfqSwap`'s own doc (`swapManager.ts:305-308`) instructs callers to *rebuild* the script on restart from the quote's binding fields — which re-derives `claimDelay` from `info.unilateralExitDelay` under the new ladder.

Concrete scenario: a trader funds a lightning-send lockup on 0.0.4, upgrades, restarts, and calls `refundIfUnresolved` with the rebuilt script. `refund.ts:674-675` looks up VTXOs at `input.script.pkScript`, finds nothing at the *new* address, and returns `{ outcome: "nothing_to_refund" }` — a reassuring terminal answer for money that is still locked at the old script, past a `refundLocktime` that keeps ticking. `swapManager.ts:983-992` does catch this mismatch, but only on the contract-registration path, not on the refund path.

Fix is either persisting the three delays (or just `claimDelay`) on the swap record and rebuilding from the stored value, or at minimum saying so in the migration note. Same wording and same gap as the earlier entry at README.md:558-562, so it's a repeat rather than a new class of problem.

packages/swap/src/rfq.ts
```diff
@@ -564,12 +583,15 @@ export const unilateralClaimDelay = (serverExitDelaySeconds: number): number =>
             `server exit delay must be at least ${SEQUENCE_GRANULARITY_SECONDS}s of seconds, got ${serverExitDelaySeconds}`,
         );
     }
-    // two granularity steps below BIP68's ceiling, not at it: the refund tiers
-    // add one and two steps on top of this value, and every tier must encode
-    if (serverExitDelaySeconds > (0xffff - 2) * SEQUENCE_GRANULARITY_SECONDS) {
+    // the headroom below BIP68's ceiling, not at it: the solo refund stacks
+    // SOLO_REFUND_HEADROOM_SECONDS on top of this value, and it must encode too
+    if (
```
pietro909, line 588: **low — the tightened BIP68 ceiling now diverges from the reference solver's, contradicting "the two derivations must agree exactly" in the README two lines up.**

This side rejects `serverExitDelaySeconds > 0xffff*512 - 4096`; the solver on `fix/ladder-solo-refund-headroom` left its own guard at `> 0xffff*512` and did not reserve the headroom. For an operator `unilateralExitDelay` in `(33549824, 33553920]` seconds the trader throws out of `unilateralClaimDelay` while the solver quotes happily and then fails deeper in its own script build.

Both sides fail, just at different seams and with different messages, so this is diagnosability rather than fund loss — and the window is ~388 days wide, so it's unreachable in practice. Worth either mirroring into the solver PR or narrowing the README's "agree exactly" phrasing to cover the delay values only.

packages/swap/test/unilateralLadder.test.ts
```diff
@@ -0,0 +1,103 @@
...
+                claim,
+                unilateralRefundDelay(claim),
+                unilateralRefundWithoutReceiverDelay(claim),
+            ]) {
+                expect(value).toBeGreaterThanOrEqual(exitDelay);
+            }
+        }
+    });
+
+    it("matches the reference solver's headroom exactly", () => {
```
pietro909: **low — this cannot detect solver drift; it restates the constant's own definition.**

`SOLO_REFUND_HEADROOM_SECONDS` is defined as `8 * SEQUENCE_GRANULARITY_SECONDS`, and the test asserts `toBe(8 * SEQUENCE_GRANULARITY_SECONDS)` against a locally re-declared `512`. That's a valid value pin against an accidental edit on *this* side, but it verifies nothing about the solver — despite the file header calling itself "the contract with the reference solver's `deriveUnilateralDelays`".

Same caveat applies to the two regenerated goldens (`test/rfq.test.ts:98`, `test/rfqReceive.test.ts:112`), whose titles say "byte-identical to the reference solver's script" while the expected value was recomputed from this code. I confirmed by hand that both sides agree today; nothing in CI would catch it if #81 landed with a different number. A shared fixture — or a comment recording that the pin was hand-verified against solver commit `b9fc3fe` — would close it.

pietro909 (commented):
Review of the ladder change. The core is sound — I confirmed the delay derivation matches solver commit `b9fc3fe` on [lightning-swap-service#81](https://github.com/arkade-os/lightning-swap-service/pull/81) exactly, the BIP68 ceiling arithmetic holds (`0xffff*512 - 4096` is itself a multiple of 512, so `ceil()` can never overshoot it), the new CSV byte pins are correct (`03080040b275` → 8 units → 4096s; `03100040b275` → 16 units → 8192s), and nothing enforces a `claim < refund` ordering — `vhtlc.ts:355-399` validates only `>0`, `%512==0`, `>=512` per tier, and arkd's `TapscriptsVtxoScript.Validate` only compares the *smallest* exit delay against its minimum. So levelling `unilateralRefund` with `unilateralClaim` is safe on both sides. Tests: 374 pass in `packages/swap`; only `test/e2e/*` fail on `ECONNREFUSED :7070` (no regtest server, pre-existing). No stale goldens anywhere else in the repo.

Three findings inline — one medium, two low.

packages/swap/README.md
Kukks: Confirmed, and the note was wrong as written. Walked the path you describe: `AssetSwap` (`store.ts:33-77`) carries `swapPkScript` but no delay fields, `swapManager.ts:305-308` does instruct callers to rebuild from the quote's binding fields, and `refund.ts:674-675` returns `nothing_to_refund` on an empty lookup. So a lockup funded on 0.0.4 and rebuilt post-upgrade is looked for at the new address and reported terminal while the money sits at the old script with `refundLocktime` still running.

Rewrote the entry in `ad262086`: the `verifyLockupAddress` claim is now explicitly scoped to quotes not yet funded, and the in-flight case gets its own paragraph naming the mechanism, the `nothing_to_refund` symptom, and the two workarounds (drain before upgrading, or rebuild the old script from the pre-upgrade delays).

Not persisting the delays here. As you say it is the same gap as the entry at README.md:558-562 rather than something this change introduces, and the fix touches the record schema plus the rebuild path in `swapManager` and `refund` — too much to bolt onto a timelock constant. Worth its own issue; happy to open one if you agree that's the right split.

packages/swap/src/rfq.ts
Kukks: Correct — checked the solver at `c904d44` (the #81 merge) and its guard is still `serverExitDelaySeconds > 0xffff * SEQUENCE_GRANULARITY_SECONDS` at `src/core/timelocks.ts:102`, reserving nothing for the tiers stacked above it. So the two guards do diverge, and "agree exactly" was the wrong claim to put two lines above a change that broke it.

Took the narrowing option in `ad262086`: the README now says the two derivations must produce the same three delay values, which is what actually has to hold for the addresses to match, and the ceiling divergence is documented alongside it with the window and the failure mode you describe — both sides refuse, at different seams and with different messages.

Left the solver's guard alone rather than mirroring it. It is unreachable at ~388 days, and the solver is a separate repo and release, so bundling a cosmetic guard change into it costs a coordinated deploy for no behavioural gain. Say the word if you'd rather have it mirrored and I'll raise it there instead.

packages/swap/test/unilateralLadder.test.ts
Kukks: Right about the test, and I've retitled it in `ad262086` to "pins the headroom constant against an accidental edit on this side", with a comment saying plainly that it restates the constant's definition, that re-declaring 512 locally is the only thing giving it value, and that agreement with the solver is not testable from inside this package at all.

On the goldens I'd push back on the mechanism, though not on the conclusion. They were not recomputed from this code — that would have been circular and green for the wrong reason. They came from the solver's own `CovenantSwapScript` at `b9fc3fe` (your sha; it merged as `c904d44`), driven by that commit's `deriveUnilateralDelays`. The check that makes that claim falsifiable: the generator was first run against the PRE-change ladder (4096/4608/5120) and reproduced the goldens that were pinned here at the time — `51200370b2a6…` and `5120f683cdac…`. Code that derived its expectations from this package could not have produced the old bytes.

But the substantive point stands, because none of that was anywhere a reader could check — it lived in the PR description, which is not where someone regenerating a fixture looks. Both golden tests now carry the provenance inline: which solver commit, that it was driven from the solver's builder, that the negative control reproduced the prior pin, and an instruction to regenerate the same way — reproduce the current pin first, then trust the new number.

Your shared-fixture suggestion is the stronger fix and I haven't done it; it needs a home neither repo currently has. Noted in the comment as what it would take if we want this checked mechanically rather than by procedure.

PR #752, opened 2026-08-13
feat(balance): add an escrow bucket to WalletBalance

Kukks:
Adds an `escrow` bucket to `WalletBalance` so a consumer can isolate contract-gated funds from the rest of the `total - available` gap.

Picks up the follow-up [arkade-os/wallet#913](https://github.com/arkade-os/wallet/pull/913) named in its own out-of-scope section:

> The BTC home-screen total is still wrong, for the same root cause. It cannot be fixed cleanly at the display layer today: `total - available` also folds in `recoverable`, `pendingRecovery` and intent-locked funds, and `WalletBalance` has no bucket isolating escrow. Showing `available` instead would erase swept and awaiting-recovery funds from the hero — a regression in the opposite direction. This wants an escrow bucket on `WalletBalance` upstream; `computeOffchainBalance` already has `isGenericallySpendable` in hand.

## The change

`computeOffchainBalance` already evaluated `isGenericallySpendable(vtxo)` to decide `available`. That single condition is split so gated funds land in their own accumulator:

```ts
// before
if (isGenericallySpendable(vtxo) && isUnlocked(vtxo)) {
    available += vtxo.value;
    addAssets(spendable, vtxo);
}

// after — gate first, so the buckets are disjoint
if (!isGenericallySpendable(vtxo)) {
    escrow += vtxo.value;
} else if (isUnlocked(vtxo)) {
    available += vtxo.value;
    addAssets(spendable, vtxo);
}
```

`available` is still reached on exactly `isGenericallySpendable && isUnlocked`, including short-circuit order — `isUnlocked` is evaluated only when `isGenericallySpendable` holds, in both forms. **No existing bucket changes value for any input.**

`escrow` is a *view* over coins already counted in `settled`/`preconfirmed`, not a fifth exclusive branch. It accumulates after the settled/preconfirmed split with no intervening `continue`, so `total` is untouched. That ordering is deliberate: `balance.ts` already carries a comment explaining why the gate is not applied earlier — a gated coin dropped from `recoverable` would fail `canSpendOffchain` too and land in no bucket at all.

The field is surfaced through both `WalletBalance` assembly sites — `Wallet.getBalance` and the service worker's `handleGetBalance`. The second is the one that matters in practice: apps render from `ServiceWorkerWallet`, so a field present only in the first would be permanently `0` in a wallet while typechecking fine. `expo/wallet.ts` and `serviceWorker/wallet.ts` forward rather than assemble, so there is no third site.

## Deriving the rest of the gap

The JSDoc states the whole identity, because the intent-locked portion is deliberately left derivable rather than bucketed:

```
total - available = boarding.total + recoverable + pendingRecovery + escrow + intentLocked
intentLocked      = settled + preconfirmed - available - escrow
```

`boarding.total` is in there because the public `total` is `totalBoarding + offchain.total` while `available` is `offchain.available` only. The twin JSDoc on the internal `OffchainBalance.escrow` correctly omits boarding, since that type's `total` has none.

## What `escrow` actually contains

Broader than swaps, and the JSDoc says so. `isContractGenericallySpendable` is default-closed, so a contract with no registered handler, a plugin type, or an `arkade` contract without `metadata.genericallySpendable === true` all land here. Swap escrow is the common case, not the only one.

The name follows existing vocabulary rather than introducing new: `index.ts` ("escrowed, intent-locked or awaiting recovery"), `balance.ts`, `wallet.ts` and `contracts/types.ts` already use "escrow" for precisely this set. This promotes documented prose into a machine-readable value.

## Compatibility

`escrow` is **required, not optional**. Optional would force `?? 0` at every call site and reintroduce the ambiguity the field exists to remove. Consumers that only *read* a balance are unaffected; code that *constructs* one — an alternate `IWallet` implementation, or a test mock building a full `WalletBalance` — needs the new field. This matches how `recoverable`, `pendingRecovery` and `availableAssets` were added.

## Tests

`test/wallet/balance.test.ts` covers the bucketing: gated coin counted as escrow and not available; owned buckets unchanged; intent-locked-but-ungated in neither; a coin both gated and intent-locked counted once as escrow; the closure invariant; and a recoverable coin staying out of escrow even when gated.

`test/wallet-message-handler.test.ts` pins the service-worker path through the real `handleGetBalance` with a gated `arkade` contract, using the existing fixture.

Full suite unchanged against `master`: ts-sdk 163 files / 2601 passed / 1 skipped / 0 failed, swap 22 / 368, boltz-swap 23 / 614 with no type errors.

## Deliberately not here

- **`escrowAssets`**, the asset-side twin. `assets - availableAssets` already expresses the held-not-selectable set, and wallet#913 handled the asset half. A symmetric bucket needs the same three-way split on the asset side — its own design question.
- **An `intentLocked` bucket.** Derivable from the four fields above, with the formula stated on the type.
- **Any wallet-side change.** The consumer stacks on wallet#913.

### DIFF
--- a/packages/ts-sdk/src/wallet/balance.ts
+++ b/packages/ts-sdk/src/wallet/balance.ts
@@ -11,6 +11,16 @@ export interface OffchainBalance {
     settled: number;
     preconfirmed: number;
     available: number;
+    /**
+     * The part of `settled + preconfirmed` held at contracts generic spending
+     * is gated from — swap escrow, chiefly. Counted in the owned buckets and in
+     * `total`, since it is still the user's money, but never in `available`.
+     *
+     * Disjoint from `available` by construction. The remainder,
+     * `settled + preconfirmed - available - escrow`, is what an in-flight
+     * settlement intent has locked.
+     */
+    escrow: number;
     recoverable: number;
     pendingRecovery: number;
     /** `settled + preconfirmed + recoverable + pendingRecovery` — the buckets are disjoint. */
@@ -57,6 +67,7 @@ export function computeOffchainBalance(
     let settled = 0;
     let preconfirmed = 0;
     let available = 0;
+    let escrow = 0;
     let recoverable = 0;
     let pendingRecovery = 0;
     const owned = new Map<string, bigint>();
@@ -94,7 +105,12 @@ export function computeOffchainBalance(
         } else {
             settled += vtxo.value;
         }
-        if (isGenericallySpendable(vtxo) && isUnlocked(vtxo)) {
+        // Gate first, so the two buckets are disjoint: a coin that is both gated
+        // and intent-locked is escrow, and the intent-locked remainder stays
+        // derivable as `settled + preconfirmed - available - escrow`.
+        if (!isGenericallySpendable(vtxo)) {
+            escrow += vtxo.value;
+        } else if (isUnlocked(vtxo)) {
             available += vtxo.value;
             addAssets(spendable, vtxo);
         }
@@ -107,6 +123,7 @@ export function computeOffchainBalance(
         settled,
         preconfirmed,
         available,
+        escrow,
         recoverable,
         pendingRecovery,
         total: settled + preconfirmed + recoverable + pendingRecovery,
--- a/packages/ts-sdk/src/wallet/index.ts
+++ b/packages/ts-sdk/src/wallet/index.ts
@@ -341,14 +341,18 @@ export interface WalletBalance {
      */
     available: number;
     /**
-     * Funds held at contracts that generic spending is gated from — swap escrow,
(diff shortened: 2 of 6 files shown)


Kukks:
Superseded by #751, which lands the same change and goes further:

- splits the unavailable remainder into both `gated` and `intentLocked`, pinning `settled + preconfirmed === available + gated + intentLocked` — this PR shipped one bucket and left the second derivable
- names the bucket for the mechanism rather than its most common occupant, so a type whose handler the runtime never registered isn't reported as something in flight
- fixes the `settled`/`preconfirmed` docs that described owned funds as spendable, which is the doc defect that misled the consumer to begin with

Closing in favour of it — nothing here needs porting.

PR #753, opened 2026-08-13
feat(swap): render a failed swap as one activity carrying its outcome

Kukks:
Makes a failed swap legible in wallet history.

Today `getTransactionHistory` classifies purely by VTXO created/spent and has no swap concept, so a failed swap's funding reads as an ordinary send and its refund as an unrelated receive — two rows, neither saying a swap happened or that it failed.

Two parts:

- `packages/ts-sdk` — an optional `outcome` on the activity-resolver types, so a resolver can declare something `Activity.settled` cannot express.
- `packages/swap` — `swapActivityResolver`, which groups a swap's transactions under one id carrying its real state.

## Why a new field rather than reusing `settled`

`Activity.settled` is `members.every(x => x.tx.settled)`, documented as "True once every member tx is settled". That is correct, and a failed swap satisfies it — the funding transaction confirmed. The problem is only that consumers read it as "succeeded".

Making `settled` overridable would silently change its meaning for every existing consumer to work around a naming collision. `outcome` is additive and leaves existing readers correct. It is carried at both places `buildActivities` folds membership data into a bucket's intent — `merge()`, and the bucket assembly a single-membership group takes.

## `outcome` is an opaque token, not display text

Values are lowercase machine tokens — `pending`, `settled`, `refunded`, `failed`, `lost` — for an app to map to its own copy or icon. Emitting display English would make the field unlocalizable and force apps to match on words to pick an icon.

## A refund on a receive leg is a loss, not money returned

The resolver keys the outcome on `(kind, state)`, not `state` alone. This package already names the case:

```ts
// swapManager.ts
const lostReceive = swap.kind === "lightning_receive" && swap.state === "refunded";
```

and `RfqSwapState`'s own docs say that on a receive leg this is a **loss** — the trader's incoming payment never arrived. Keying on `state` alone would render it as `refunded` and tell the user their money came back when it did not. That case emits `lost` instead.

The mapping keeps `Record<RfqSwapState, string>`, so adding a state upstream without mapping it is a compile error rather than a silent misrender. The receive-leg guard runs ahead of that table rather than replacing an entry, so the exhaustiveness check is unaffected.

## Shape

```ts
export interface SwapActivityInput {
  rfqId: string
  kind: "lightning_send" | "lightning_receive" | "onchain_send"
  state: RfqSwapState
  txids: readonly string[]   // funding, claim, refund — in any order
}

export function swapActivityResolver(deps: {
  listSwaps(): Promise<readonly SwapActivityInput[]>
}): ActivityResolver
```

`prepare()` loads once and indexes txid → swap; `resolve()` stays pure and synchronous, as the `ActivityResolver` contract requires. Grouping by `swap:${rfqId}` is what collapses a swap's funding and refund into one row.

Inputs are injected rather than read from a repository, so the resolver is testable with plain data and a consumer can serve it from whatever store it already has.

## Compatibility

Additive throughout. `outcome` is optional on both `GroupMembership` and `ActivityIntent`; nothing existing sets or reads it, and no existing activity changes. No change to the public `@arkade-os/sdk` surface — `packages/ts-sdk/src/index.ts` and `src/wallet/index.ts` are byte-identical to master across the branch.

## Tests

`packages/ts-sdk/test/wallet/activity.test.ts` — `outcome` reaches the intent; it is independent of `settled`, which stays `true` for a settled member tx; first-writer-wins across resolvers sharing a `groupId`; and it survives a single-membership group, which takes the bucket-assembly path rather than `merge()`.

`packages/swap/test/activity.test.ts` — group id, per-corridor labels, outcome per state, an unrelated transaction left plain, a rejecting `listSwaps` contributing nothing, `resolve()` before `prepare()` returning undefined, every txid of a swap indexing to the same group, and a receive-leg refund yielding a different outcome from a send-leg refund.

Full suite unchanged against master: ts-sdk 162 files / 2622 passed / 1 skipped, swap 25 / 400, boltz-swap 23 / 614, no type errors.

## Deliberately not here

**`rfqSwapActivityInputs`**, the helper that projects stored `RfqSwapRecord`s into `SwapActivityInput`s. It needs `getAllRfqSwaps()` and `RfqSwapRecord`, which live on #746 and are not on master. The resolver needs neither, so it ships now and the helper follows when #746 lands. At that point `SwapActivityInput["kind"]` should be bound to `RfqSwapRecord["kind"]` rather than repeating the union.

**A wallet consumer.** Separate change, in the wallet repo.

## Summary by CodeRabbit

* **New Features**
  * Added swap activity tracking with transaction matching, swap details, labels, and outcome states.
  * Added support for displaying activity outcomes, including refunded and lost transactions.
  * Activity outcomes now persist when related records are merged.

* **Bug Fixes**
  * Improved handling of transactionless, unrelated, and preparation-failure scenarios.
  * Ensured activity outcomes remain independent of transaction settlement status.

### DIFF
--- a/packages/swap/src/activity.ts
+++ b/packages/swap/src/activity.ts
@@ -17,7 +17,7 @@ export interface SwapActivityInput {
      * only on the unmerged rfq-persistence branch, not on master. Reconcile
      * with `RfqSwapRecord["kind"]` once that branch lands.
      */
-    kind: "lightning_send" | "lightning_receive";
+    kind: "lightning_send" | "lightning_receive" | "onchain_send";
     state: RfqSwapState;
     /** Funding, claim and refund txids, in whatever order. */
     txids: readonly string[];
@@ -27,21 +27,26 @@ export interface SwapActivityInput {
 const LABELS: Record<SwapActivityInput["kind"], string> = {
     lightning_send: "Lightning send",
     lightning_receive: "Lightning receive",
+    onchain_send: "Onchain send",
 };
 
 /**
- * How a swap state reads to a user. Keyed exhaustively off `RfqSwapState`, so a
- * state added upstream is a compile error here rather than silently rendering
- * as something misleading.
+ * How a swap state reads as an outcome token, keyed exhaustively off
+ * `RfqSwapState` so a state added upstream is a compile error here rather
+ * than silently rendering as something misleading. Opaque lowercase machine
+ * tokens, not display text — see `ActivityIntent.outcome`.
+ *
+ * `lightning_receive` + `refunded` is handled separately in `resolve`: it
+ * does not read off this table.
  */
-const STATUS: Record<RfqSwapState, string> = {
-    pending: "Pending",
-    claimable: "Pending",
-    claimed: "Pending",
-    needs_counterparty: "Pending",
-    settled: "Settled",
-    refunded: "Refunded",
-    failed: "Failed",
+const OUTCOME: Record<RfqSwapState, string> = {
+    pending: "pending",
+    claimable: "pending",
+    claimed: "pending",
+    needs_counterparty: "pending",
+    settled: "settled",
+    refunded: "refunded",
+    failed: "failed",
 };
 
 /**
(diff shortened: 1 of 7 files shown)


Kukks:
Thanks @arkana-ai-bot — all four addressed in `ee1bbdb0`:

- **`onchain_send` missing from the label matrix** — added as a third arm to the existing corridor-label test, asserting `"Onchain send"` so a misspelling would fail.
- **`needs_counterparty` → `pending` semantically compressed** and **`claimable`/`claimed` collapsing** — one comment above the three entries, covering both: the first two have no user-visible phase distinct from pending, while `needs_counterparty` is *blocked* rather than in flight. Noted that a future `"blocked"`/`"stuck"` token was considered, so the next reader doesn't have to re-derive it.
- **Line-number citations will drift** — `swapManager.ts:158-167` and `:1656-1661` replaced with symbol references (`RfqSwapState`'s `refunded` case, the `lostReceive` local in `outcomeOf`) in both the source comment and the test.

The mapping itself is unchanged, and `OUTCOME` keeps its `Record<RfqSwapState, string>` annotation so a new state stays a compile error.

pietro909 (approved):

PR #756, opened 2026-08-13
docs(balance): show the new buckets and warn against netting them out of total

Kukks:
Follow-up to #751, which split the unavailable remainder into `gated` and `intentLocked`. Documentation only — no behaviour, no API.

#751's motivation was a consumer who subtracted the wrong figure, and its second commit exists to fix "the docs that led the consumer astray". Three places were left saying the old thing, or inviting the same mistake.

## The README example was not updated

`packages/ts-sdk/README.md`'s "Checking Balance" still listed only the pre-#751 fields. That is the doc a consumer actually opens; the type's JSDoc is where they end up only if they go looking.

It now shows every bucket and states the identity that makes the arithmetic safe:

```
settled + preconfirmed === available + gated + intentLocked
```

with the point spelled out: subtract from `settled + preconfirmed`, **not** from `total`, which also carries `boarding.total`, `recoverable` and `pendingRecovery`. Those are still the user's funds, so netting a bucket out of `total` drops them from the figure with no signal.

## `gated`'s JSDoc invited the mistake it exists to prevent

It closed by describing what `total - gated` still carries — arithmetic against `total`, which is the operation that goes wrong. Reworded to point at `settled + preconfirmed` and to name what `total` additionally holds.

This is not hypothetical. Someone implementing the same feature independently documented the `total - available` gap and omitted `boarding.total`, for exactly this reason: the public `total` includes boarding while `available` does not.

## An `@see` block tag mid-sentence

`intentLocked`'s JSDoc had `(@see {@link BalanceCapabilities})` inside a sentence. TypeDoc hoists block tags out of the description, which would have detached the rest of the paragraph. Now a plain inline `{@link}`.

## Verification

`pnpm typecheck` and `pnpm lint` clean; full unit suite green through the pre-commit hook. No source behaviour is touched — the two `src` changes are both inside comments.

## Summary by CodeRabbit

* **Documentation**
  * Expanded balance documentation to explain gated, intent-locked, and pending-recovery funds.
  * Clarified how settled and preconfirmed balances relate to spendable funds.
  * Added guidance for calculating funds not committed to contracts or intents.
  * Improved wallet balance descriptions, including how gated and recoverable funds contribute to totals.
  * Corrected a balance documentation reference.

### DIFF
--- a/packages/ts-sdk/README.md
+++ b/packages/ts-sdk/README.md
@@ -341,7 +341,25 @@ console.log('Boarding Total:', balance.boarding.total)
 console.log('Offchain Available:', balance.available)
 console.log('Offchain Settled:', balance.settled)
 console.log('Offchain Preconfirmed:', balance.preconfirmed)
+console.log('Gated by a contract:', balance.gated) // swap escrow, chiefly
+console.log('Locked by an in-flight intent:', balance.intentLocked)
 console.log('Recoverable:', balance.recoverable)
+console.log('Awaiting recovery:', balance.pendingRecovery)
+```
+
+`settled` and `preconfirmed` are what the wallet OWNS; `available` is what generic
+spending will actually pick. The difference is accounted for exactly:
+
+```
+settled + preconfirmed === available + gated + intentLocked
+```
+
+To show "your money, minus what is tied up", subtract from `settled + preconfirmed`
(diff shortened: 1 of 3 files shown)


Kukks:
@coderabbitai review

PR #762, opened 2026-08-17
feat(swap): name the asset id in the arkade pair leg

pietro909:
Makes `@arkade-os/swap` interoperable with lightning-swap-service#113: an arkade asset leg is the asset id itself (`arkade:<68-hex>`, via `arkadeAssetLeg`) instead of the coarse `arkade:ASSET`, which no solver serves.

- the pair is what both sides route and subscribe on, so the asset id moves there; `profile` goes empty (the solver's schemas are `.strict()`)
- `ARKADE_ASSET` deprecated, not removed
- `MAX_PAIR_LENGTH` mirrors the wire's 158-char cap — dormant until asset↔asset lands
- second commit: `expectQuote` compares the quoted pair against the requested one, and the nostr transport's duplicate copy of it is gone. Independent of the asset work; revertable on its own.

Safe to ship before https://github.com/arkade-os/lightning-swap-service/pull/113: no solver serves arkade↔arkade in any spelling, so behaviour on live corridors is unchanged.

### DIFF
--- a/packages/swap/README.md
+++ b/packages/swap/README.md
@@ -290,7 +290,9 @@ message anywhere: **acceptance is funding**.
   offline. The solver observes the funding on-chain, pays the invoice, and claims with the
   preimage — which lands publicly in the claim witness as the receipt. A failed swap refunds by
   covenant to the trader's address, pushable by anyone, no trader keys or state.
-- **Arkade ↔ arkade** (BTC↔asset, asset↔asset): the trader accepts a quote by creating and funding
+- **Arkade ↔ arkade** (BTC↔asset, asset↔asset): an arkade asset leg names the asset id itself —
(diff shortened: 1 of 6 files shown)


pietro909:
Cross-link for tracking, now that this has shipped in `@arkade-os/swap@0.0.7`.

This closed the pair-naming half of the asset work — a leg can name an asset id,
matching lightning-swap-service#113. Two siblings remain, filed so they are
findable from here rather than rediscovered:

- **#768** — an asset↔asset offer can be *named* by a pair now, but still cannot be
  requested or filled: `offer.ts` has two covenant programs and a one-asset TLV, and
  four guards enforce "exactly one of wantAsset / offerAsset". lss#110 (merged)
  already decides such offers on the solver side.
- **#763** — `VHTLC.ScriptV2` denominated in an asset, the HTLC-side counterpart.

PR #763, opened 2026-08-17
feat(vhtlc): let a VHTLC be denominated in an Arkade asset (ScriptV2, and v1 by inheritance)

Kukks:
`VHTLC.ScriptV2` gains an optional `asset`, so a VHTLC can escrow an Arkade asset
rather than only sats.

```ts
asset?: {
  txid: Bytes      // the asset's genesis transaction id, 32 bytes
  groupIndex: number  // the asset group index within that genesis transaction
}
```

## What changes, and what deliberately does not

Only `nonInteractiveClaim` and `nonInteractiveRefund` — the two leaves carrying a
covenant the emulator enforces. The interactive signature leaves are
byte-identical, because a signature path asserts nothing about value, so an asset
makes no difference there. That is asserted, not assumed.

**Omitting `asset` leaves every script byte-identical**, so contracts already
funded still derive to the same address. There is a test for exactly that.

**The sat clause is RETAINED, not replaced.** An asset-carrying VTXO carries sats
too, so a covenant constraining only the asset would let a spend keep the asset
and strip the sats — the mirror of the bug the asset clause exists to prevent.
The sat covenant becomes the asset covenant's tail.

## The trap, in case it saves someone the afternoon

**The txid the script pushes is REVERSED relative to the serialized Asset ID.**
`assetId.slice(0, 64)` is the leading 32 bytes; the introspection opcode matches
only against those bytes reversed. Push them in wire order and
`OP_INSPECTOUTASSETLOOKUP` reports the asset **absent** — which then reads as
`amount = 0`, and `0 >= 0` passes, so the asset-stripping spend the covenant
exists to stop would succeed.

Nothing in the failure names it: the emulator says only `OP_VERIFY failed vin=0`,
and it fails identically whatever the amount comparison says, which is what
isolated it. `ScriptV2` does the flip internally, so callers pass the id in wire
order and never think about it.

Two other things found the same way, by elimination against the BTC-only control
in `non-interactive-htlc.test.ts`, which passes throughout — so the stack was
never the variable:

- **the asset input is LOCAL, not INTENT.** arkd refuses the alternative outright:
  `ASSET_INPUT_INVALID (35): unexpected asset input type: intent`
- **issuance is not immediately spendable** — the balance has to reflect it first.

## Verification

- **unit**: 163 files / 2630 tests pass, 1 skipped. `tsc --noEmit` clean.
- **e2e** (`test/e2e/asset-covenant.test.ts`): the emulator co-signs a spend that
  pays the asset through, **and rejects** one that pays the sats while keeping the
  asset. Both halves of the covenant's claim.
- Re-run on a **freshly created** regtest chain, not only the one it was developed
  against, so the reversed-txid behaviour is not an artefact of a single chain.

### Running the e2e

It needs arkd intent fees zeroed. The regtest stack applies real fees at the end
of `start()`, after which the probe's faucet dies with

```
INTENT_INSUFFICIENT_FEE (31): got 0 min expected 2000
```

which looks like a covenant rejection and is not. Zero them, run, restore:

```
curl -X POST localhost:7071/v1/admin/intentFees -H 'Content-Type: application/json' \
  -d '{"fees":{"offchainInputFee":"0.0","onchainInputFee":"0.0","offchainOutputFee":"0.0","onchainOutputFee":"0.0"}}'
```

## What this does NOT establish

Worth stating plainly, because it is the difference between "the covenant works"
and "a consumer can rely on it":

The e2e proves the **emulator enforces the asset covenant** — but it builds from a
hand-rolled artifact through `ark.contract()`, not through `ScriptV2`. `ScriptV2`'s
asset support is covered by unit tests only (script construction). **No test yet
spends a `ScriptV2`-built asset lockup against the emulator**, and no e2e in this
package builds through `ScriptV2` at all — `non-interactive-htlc.test.ts` defines
its own artifact inline too.

So the two halves each have evidence and the join does not. That join is what a
consumer locking funds to a `ScriptV2` asset lockup would depend on, and it is
the next thing to write rather than something this PR should be read as covering.

The refund leaf is also not e2e-covered here — note that is equally true of BTC in
this package, so it is a pre-existing coverage level rather than an asset-specific
hole.

## Summary by CodeRabbit

- **New Features**
  - Added optional asset support to VHTLC claim and refund covenants.
  - Asset-aware covenants verify asset identity, preserve satoshi requirements, and require complete asset forwarding.
  - Asset-backed VHTLC parameters now persist correctly through serialization and deserialization.
  - Satoshi-only behavior remains unchanged when no asset is configured.

- **Bug Fixes**
  - Added validation for malformed or incomplete asset configurations.

- **Tests**
  - Added unit and end-to-end coverage for asset covenant execution, compatibility, validation, and forwarding behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,7 @@ jobs:
             group: arkade-assets
             test_files: >-
               test/e2e/asset.test.ts
+              test/e2e/asset-covenant.test.ts
               test/e2e/arkadeCash.test.ts
               test/e2e/liquidation.test.ts
           - package: ts-sdk
--- a/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
+++ b/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
@@ -13,6 +13,35 @@ import { assertVhtlcSpendableNow, isCltvSatisfied, isCsvSpendable, resolveRole }
 import { sequenceToTimelock, timelockToSequence } from "../../utils/timelock";
 
 /**
+ * The stored `assetGroupIndex`, as a number, or a throw naming the row.
+ *
+ * `Number()` alone is not enough here, and it fails in exactly one direction
+ * that matters. `VHTLC.ScriptV2` already refuses a non-integer, a negative or
+ * anything past `0xffff`, so `"abc"`, `"1.5"` and `"-1"` die one frame down with
+ * a clear message. What survives is `Number("")`, `Number(" ")` and
+ * `Number("\t")` — all of which are **0**, a perfectly valid group index.
+ *
+ * A blank field therefore does not fail: it silently names group 0 of the same
+ * genesis transaction, which is a DIFFERENT asset. The contract then derives a
+ * different script from the one the row was written against, and registration
+ * dies at `upsertContractRow` with an opaque `Script mismatch` — the same
+ * silent-drop class the both-halves-or-neither check above exists to close, one
+ * value further in.
+ *
+ * Anchored and canonical: no sign, no point, no exponent, no leading zero, so a
+ * row that round-tripped through this handler's own `serializeParams` is the
+ * only shape accepted back.
+ */
+const parseGroupIndex = (raw: string): number => {
+    if (!/^(0|[1-9][0-9]*)$/.test(raw)) {
+        throw new Error(
+            `assetGroupIndex must be a canonical decimal integer, got ${JSON.stringify(raw)}`,
+        );
+    }
+    return Number(raw);
+};
+
+/**
  * Typed parameters for {@link VHTLC.ScriptV2} contracts.
  *
  * The eight mandatory fields are `VHTLCContractParams` verbatim — the two
@@ -233,7 +262,7 @@ export const VHTLCV2ContractHandler: ContractHandler<VHTLCV2ContractParams, VHTL
             params.assetTxid !== undefined && params.assetGroupIndex !== undefined
                 ? {
                       txid: hex.decode(params.assetTxid),
-                      groupIndex: Number(params.assetGroupIndex),
+                      groupIndex: parseGroupIndex(params.assetGroupIndex),
                   }
                 : undefined;
         return {
--- a/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
+++ b/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
@@ -31,6 +31,17 @@ export interface VHTLCV2ContractParams {
     unilateralClaimDelay: RelativeTimelock;
     unilateralRefundDelay: RelativeTimelock;
     unilateralRefundWithoutReceiverDelay: RelativeTimelock;
+    /**
+     * The Arkade asset the covenant leaves bind, if any.
+     *
+     * @see VHTLC.Options.asset
+     */
+    asset?: {
+        /** The asset's genesis txid, 32 bytes, canonical order — as the serialized Asset ID carries it. */
+        txid: Uint8Array;
+        /** The asset group index within that genesis transaction. */
+        groupIndex: number;
+    };
     /** @see VHTLC.Options.nonInteractiveClaim */
     nonInteractiveClaim?: {
         receiverPkScript: Uint8Array;
@@ -160,6 +171,19 @@ export const VHTLCV2ContractHandler: ContractHandler<VHTLCV2ContractParams, VHTL
                     params.nonInteractiveRefund.emulatorPubkey,
                 ),
             }),
+            // The asset must round-trip, and its absence here used to be silent
+            // in the worst way. `ContractManager` re-derives a contract from
+            // these params; dropping the asset derives the SAT-ONLY script, so
+            // registration died at `upsertContractRow` with a `Script mismatch`
+            // naming two hex strings and no cause. Same silent-drop class
+            // `validateOptions` refuses one layer up.
+            //
+            // Two keys rather than one blob, mirroring the script's own view of
+            // an Asset ID as a (txid, groupIndex) pair.
+            ...(params.asset && {
+                assetTxid: hex.encode(params.asset.txid),
+                assetGroupIndex: params.asset.groupIndex.toString(),
+            }),
         };
     },
 
@@ -176,12 +200,30 @@ export const VHTLCV2ContractHandler: ContractHandler<VHTLCV2ContractParams, VHTL
             "nonInteractiveRefundEmulatorPubkey",
             "nonInteractiveRefund",
         );
+        // Both halves or neither: a txid without its group index names no asset,
+        // and an index without a txid names nothing at all. Either alone is a
+        // corrupt row rather than a contract that happens to lack an asset, and
+        // reading it as the latter re-derives the sat-only script — the exact
+        // silent drop this round-trip exists to close.
+        if ((params.assetTxid === undefined) !== (params.assetGroupIndex === undefined)) {
+            throw new Error(
+                "asset params are incomplete: assetTxid and assetGroupIndex must both be present or both absent",
+            );
+        }
+        const asset =
+            params.assetTxid !== undefined && params.assetGroupIndex !== undefined
+                ? {
+                      txid: hex.decode(params.assetTxid),
+                      groupIndex: Number(params.assetGroupIndex),
+                  }
+                : undefined;
         return {
             sender: hex.decode(params.sender),
             receiver: hex.decode(params.receiver),
             server: hex.decode(params.server),
             preimageHash: hex.decode(params.hash),
             refundLocktime: BigInt(params.refundLocktime),
+            ...(asset && { asset }),
             unilateralClaimDelay: sequenceToTimelock(Number(params.claimDelay)),
             unilateralRefundDelay: sequenceToTimelock(Number(params.refundDelay)),
             unilateralRefundWithoutReceiverDelay: sequenceToTimelock(
--- a/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
+++ b/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
@@ -46,6 +46,8 @@ export interface VHTLCV2ContractParams {
     nonInteractiveClaim?: {
         receiverPkScript: Uint8Array;
         emulatorPubkey: Uint8Array;
+        /** @see VHTLC.Options.nonInteractiveClaim.strict */
+        strict?: { amount: bigint; assetAmount?: bigint };
     };
     /** @see VHTLC.Options.nonInteractiveRefund */
     nonInteractiveRefund?: {
@@ -162,6 +164,17 @@ export const VHTLCV2ContractHandler: ContractHandler<VHTLCV2ContractParams, VHTL
                 nonInteractiveClaimEmulatorPubkey: hex.encode(
                     params.nonInteractiveClaim.emulatorPubkey,
                 ),
+                // The opt-in quoted bound. Dropping it re-derives the DEFAULT
+                // claim covenant — a strictly weaker one — and registration dies
+                // at `upsertContractRow` with an opaque `Script mismatch`. Same
+                // silent-drop class as the asset keys above.
+                ...(params.nonInteractiveClaim.strict && {
+                    strictClaimAmount: params.nonInteractiveClaim.strict.amount.toString(),
+                    ...(params.nonInteractiveClaim.strict.assetAmount !== undefined && {
+                        strictClaimAssetAmount:
+                            params.nonInteractiveClaim.strict.assetAmount.toString(),
+                    }),
+                }),
             }),
             ...(params.nonInteractiveRefund && {
                 nonInteractiveRefundSenderPkScript: hex.encode(
@@ -210,6 +223,12 @@ export const VHTLCV2ContractHandler: ContractHandler<VHTLCV2ContractParams, VHTL
                 "asset params are incomplete: assetTxid and assetGroupIndex must both be present or both absent",
             );
         }
+        if (params.strictClaimAssetAmount !== undefined && params.strictClaimAmount === undefined) {
+            throw new Error(
+                "strictClaimAssetAmount without strictClaimAmount: reading this as 'not strict' " +
+                    "would re-derive the default claim covenant, which is weaker than the row asked for",
+            );
+        }
         const asset =
             params.assetTxid !== undefined && params.assetGroupIndex !== undefined
                 ? {
@@ -233,6 +252,14 @@ export const VHTLCV2ContractHandler: ContractHandler<VHTLCV2ContractParams, VHTL
                 nonInteractiveClaim: {
                     receiverPkScript: claim.destination,
                     emulatorPubkey: claim.emulatorPubkey,
+                    ...(params.strictClaimAmount !== undefined && {
+                        strict: {
+                            amount: BigInt(params.strictClaimAmount),
+                            ...(params.strictClaimAssetAmount !== undefined && {
+                                assetAmount: BigInt(params.strictClaimAssetAmount),
+                            }),
+                        },
+                    }),
                 },
             }),
             ...(refund && {
--- a/packages/ts-sdk/src/script/vhtlc.ts
+++ b/packages/ts-sdk/src/script/vhtlc.ts
@@ -546,18 +546,29 @@ function enforcePayToAsset(
             `asset group index must be an integer in [0, 65535], got ${asset.groupIndex}`,
         );
     }
+    // REVERSED, once, here. `asset.txid` is the id in WIRE order -- the leading
+    // 32 bytes of the serialized Asset ID -- but the introspection opcodes match
+    // against those bytes reversed. Push wire order and the lookup reports the
+    // asset ABSENT (`0 0`), so the covenant fails and the contract it guards is
+    // unspendable. Nothing in the failure says so: the emulator returns only
+    // `OP_VERIFY failed`, and returns it whatever the amount comparison says.
+    // Established on regtest against a real minted asset, by elimination against
+    // a passing BTC-only control.
+    //
+    // A copy rather than an in-place reverse: the caller's id is theirs.
+    const inspectionTxid = Uint8Array.from(asset.txid).reverse();
     return ArkadeScript.encode([
         // The output carries at least as much of the asset as the input did.
         // Output index is the input's -- the same index alignment the sat
         // covenant relies on, and the same liveness obligation on whoever
         // assembles the spend.
         "PUSHCURRENTINPUTINDEX",
-        asset.txid,
+        inspectionTxid,
         asset.groupIndex,
         "INSPECTOUTASSETLOOKUP",
         "VERIFY", // PRESENT on the output, not merely "zero of it"
         "PUSHCURRENTINPUTINDEX",
-        asset.txid,
+        inspectionTxid,
         asset.groupIndex,
         "INSPECTINASSETLOOKUP",
         "VERIFY", // ...and on the input, so the comparison means something
--- a/packages/ts-sdk/src/script/vhtlc.ts
+++ b/packages/ts-sdk/src/script/vhtlc.ts
@@ -68,6 +68,21 @@ export namespace VHTLC {
          * asset covenant while stripping the sats — exactly as the sat-only
          * covenant lets a spend strip the asset.
          *
+         * ONE ASSET IS BOUND, AND ONLY THAT ONE IS PROTECTED. If a VTXO funded to
+         * this contract carries ADDITIONAL assets alongside the bound one, those
+         * are not covered: whoever assembles a covenant spend chooses where they
+         * go, and can send them anywhere.
+         *
+         * `INSPECTOUTASSETCOUNT == 1` does not close that. It constrains the
+         * covenant's OUTPUT to exactly one asset — so the extras cannot ride
+         * along with the bound asset — but arkd's conservation rule is satisfied
+         * by routing them to a different output, which the covenant says nothing
+         * about. The bound asset arrives; the rest is the spender's to direct.
+         *
+         * So fund an asset contract with the asset it names and nothing else. A
+         * multi-asset VTXO behind this covenant is a loss waiting for whoever
+         * pushes the spend.
+         *
          * The id is the pair the introspection opcodes take. A canonical Asset
          * ID is `(genesis txid, group index)`, never a single blob.
          */
--- a/packages/ts-sdk/src/script/vhtlc.ts
+++ b/packages/ts-sdk/src/script/vhtlc.ts
@@ -517,6 +517,29 @@ function isP2trPkScript(pkScript: Bytes): boolean {
  * them; its own aggregate refund uses the interactive leaf precisely because
  * it lacks this per-index constraint, see `refund.ts`), never a safety
  * assumption.
+ *
+ * WHY `>= input` AND NOT `>= the quoted amount`. This bound is CONSERVATION,
+ * not agreement: it says the spend may not skim, and says nothing about what
+ * any quote promised. That is deliberate, on both leaves, and it is a question
+ * worth answering here because the covenant reads like the natural place to
+ * pin a quote and is not.
+ *
+ *  - A QUOTE IS NOT THE COVENANT'S TO KNOW. Pinning one compiles it into the
+ *    leaf, hence into the emulator key, hence into the ADDRESS. Re-quoting
+ *    would move the address of a contract that may already be funded.
+ *  - MISFUNDING IS THE SENDER'S EXPOSURE. They chose the amount. A lockup that
+ *    over- or underpays the quote is theirs to have created, and the
+ *    counterparty's protection is to decline it — refuse the swap and let it
+ *    refund — which is an application-layer decision, taken where a quote
+ *    actually lives, and revisable without moving anyone's address.
+ *  - NOTHING IS CLAIMED WITHOUT THE PREIMAGE. Both covenant leaves sit behind
+ *    the hash condition, so an underfunded lockup cannot be claimed out from
+ *    under the counterparty on the covenant's say-so. The covenant's job is to
+ *    stop a spend that satisfies the condition from redirecting the value; it
+ *    is not to adjudicate whether the trade was fair.
+ *
+ * So a funding gate that compares a lockup against its quote belongs in the
+ * consumer, not here. `lightning-swap-service`'s `lockupIsFunded` is that gate.
  */
 function enforcePayTo(destinationPkScript: Bytes): Bytes {
     // validateOptions already checked this for both current call sites — kept
--- a/packages/ts-sdk/src/script/vhtlc.ts
+++ b/packages/ts-sdk/src/script/vhtlc.ts
@@ -52,6 +52,31 @@ export namespace VHTLC {
             senderPkScript: Bytes;
             emulatorPubkey: Bytes;
         };
+        /**
+         * Optional: denominate this contract in an Arkade ASSET rather than in
+         * sats alone.
+         *
+         * Only the two NON-INTERACTIVE leaves change. Every other leaf is a
+         * signature path that asserts nothing about value, so an asset makes no
+         * difference to them — which is why this option reaches exactly the
+         * leaves whose covenant the emulator enforces.
+         *
+         * When set, those covenants additionally require the output to carry at
+         * least the input's amount of THIS asset, and to carry exactly one
+         * asset. The sat clause is RETAINED, not replaced: an asset-carrying
+         * VTXO carries sats too, so dropping it would let a spend satisfy the
+         * asset covenant while stripping the sats — exactly as the sat-only
+         * covenant lets a spend strip the asset.
+         *
+         * The id is the pair the introspection opcodes take. A canonical Asset
+         * ID is `(genesis txid, group index)`, never a single blob.
+         */
+        asset?: {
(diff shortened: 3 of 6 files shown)


Kukks:
Fixed in `de0309b6`. ① was real and I treated it as the money bug it is.

## ① asset with no non-interactive leaf — guarded

You are right, and right about why it is dangerous rather than untidy. Both
leaves are optional and they are the only ones carrying a covenant, so `asset`
with neither built a sat-only contract **and said nothing**. The caller funds it
believing the asset is bound, any spend satisfying the sat covenant walks off
with the asset, and the only outward difference is a `pkScript` matching a
non-asset address — not something a caller thinks to check.

`validateOptions` now refuses it. Established twice rather than once:

- wrote the test first and watched it **fail** against the old code
- then mutated: replacing the guard with `if (false)` fails both new tests and
  nothing else

Also asserted the inverse, since the guard should not be stricter than the
covenant: **either** leaf alone is enough to bind the asset, so neither is
individually required.

Worth noting your observation about why the existing malformed-id test missed
this was exactly right — it goes through a `withAsset` helper that always sets
both leaves, so the no-leaf path was unreachable from it.

## ② v1 inherits it — kept, and now tested

Deliberate rather than accidental, so I kept it in `BaseScript` instead of
narrowing the option to `ScriptV2`: the asset covenant is orthogonal to the
preimage-condition fragment, which is the *only* thing that differs between
versions. Restricting it would be arbitrary.

But your actual point stands — it was an untested money path. `VHTLC.Script`
now has a test asserting it binds the asset (a different `pkScript` than
without) and refuses the no-leaf case the same way.

## ③ unused binding — dropped

Replaced with a bare `await waitForVtxo(...)` and a line saying only the arrival
matters. You were right that `tsc --noEmit` will not catch it without
`noUnusedLocals`.

## ④ the join gap — agreed, and unchanged

Still the right next test, and still not this PR. For visibility: the missing
one is `ScriptV2` constructs the address → an asset VTXO funds it → the emulator
co-signs the spend. I have traced what a consumer needs for it — the asset rides
as an extension packet alongside the emulator packet, which the raw
`buildOffchainTx` path can already carry — and tracked it as the gating item in
arkade-os/lightning-swap-service#125.

## Separately: this PR's e2e was not running

Not from your review, found while chasing the CI failure — the e2e groups name
test files explicitly, so `asset-covenant.test.ts` was inert. Added to
`arkade-assets` in `e72d8ceb`, and that group passed on the following run, so the
probe is now actually exercised rather than just described.

## On the earlier `arkade-assets` red

It was `arkadeCash.test.ts`, which has zero vhtlc references, against six
consecutive master passes — no causal path from a vhtlc-only change, and the log
was full of `arkd not ready` / `dns: server misbehaving`. It passed on the next
run. `settlement-delegation` then failed instead, which I am treating the same
way pending this run.

pietro909:
Cross-link for tracking: the asset↔asset fill gap is now **#768**.

Adjacent to this PR rather than blocking it. #763 denominates a VHTLC in an asset
(one asset, one leg); #768 is the offer path's version of the same question — the
non-HTLC `arkade:<asset> <-> arkade:<asset>` case, which needs a third covenant
program and a two-asset TLV. Worth designing the two together, since
lightning-swap-service#110 (merged) already decides such offers on the solver side.

The three asset-covenant threads, so they are findable from each other:

- **#762** (merged) — the pair leg names the asset id
- **#763** — this PR, `VHTLC.ScriptV2` taking an asset
- **#768** — asset↔asset offers, unfillable today

pietro909 (changes requested):
Requesting changes on the e2e side — the script construction and validation design look right (I verified the reversal, stack order, flag semantics, and group-index encoding against the emulator source, both the pinned v0.0.3 and master, and against arkd's `AssetId` serialization; the flip is correct). What's missing is that the tests can't yet catch a wrong-but-plausible covenant.

## 1. The negative e2e assertion passes even if the covenant is decorative

The stripping case spends **without an asset packet** (`claim(...).to(...)` with no `.withAsset`). That attempt is refused no matter what the covenant says:

- at the emulator, the first asset opcode hits `vm.assetPacket == nil` → script error `"no asset packet"` — not the `0 0` + `VERIFY` path the comments describe as load-bearing;
- and even with a sat-only covenant the emulator would co-sign, arkd's `ValidateAssetTransaction` rejects it (`"asset packet not found in tx"`, plus `validateInputAssets` demanding every prevout asset declared and balanced). Stripping-by-omission is a protocol invariant for *any* asset VTXO, covenant or not.

Proof by construction: a covenant of just `… INSPECTOUTASSETLOOKUP VERIFY DROP` — asset present, any amount, no sat clause — still passes this negative test. The attack that **only this covenant** stops is a *valid* packet that redirects the asset while paying the sats, and it's never attempted. The test header calls assertion (2) the one that matters; as written it doesn't establish it.

Please add the discriminating case — attach the asset to a different output than the input index, so arkd conservation passes and only the covenant can refuse:

```ts
contract.functions.claim(PREIMAGE)
    .withAsset({ assetId,
        inputs: [{ vin: 0, amount: ASSET_LOCKED }],
        outputs: [{ vout: 1, amount: ASSET_LOCKED }] }) // asset NOT forwarded to output 0
    .to(receiverPkScript, CARRIER_SATS)
    .send()
```

## 2. No emulator execution of the ScriptV2 covenant — and the probed artifact differs from it

The PR states this honestly, but it's worth being precise about the gap: the e2e artifact uses a **fixed** `$assetAmount` and has **no** `INSPECTINASSETLOOKUP` and **no** `INSPECTOUTASSETCOUNT`. The `ScriptV2` covenant is input-relative and carries both — two opcodes nothing in the repo has ever executed. If either had a stack-order or counting quirk, a `ScriptV2` asset contract would be unspendable on its covenant leaves and no test would know (byte-pinning unit tests don't help — they passed for the pre-flip txid too).

The cheapest way to close both this and #1 with one test: run the improved negative case — and the positive one — through a contract **built by `VHTLC.ScriptV2`** rather than a hand-rolled artifact. That exercises the exact byte sequence consumers will lock funds to, including the input lookup and the count check.

## Medium (one-liners)

- **Handler round-trip drops `asset`**: `VHTLCV2ContractHandler.serialize/deserializeParams` (and v1's) have no `asset` field, so an asset-denominated contract registered via `ContractManager` derives a sat-only script and dies at `upsertContractRow` with a `Script mismatch` naming two hex strings and no cause — the same silent-drop class this PR fixes in `validateOptions`, one layer up. Round-trip it, refuse it explicitly, or document the gap.
- **Byte-order labels inverted in the comments**: the serialized Asset ID's leading 32 bytes are canonical/display order (arkd's `serializeTxHash` reverses chainhash internals "to match the canonical txid format"); wire order is what you get *after* the flip. The reversal is right — the words "WIRE order" say the opposite, in the comment block whose job is preventing exactly this mistake.
- **Reuse `AssetId`**: `src/extension/asset/assetId.ts` is the same `(txid, groupIndex)` pair with the same validation plus `fromString`; the new anonymous shape duplicates it, and the e2e hand-rolls `slice(0, 64)` + manual u16-LE parsing instead of `AssetId.fromString` (AGENTS.md: prefer reusing SDK primitives).
- **v1 inherits the option** (deliberately, per the test) but title/body say ScriptV2 only.
- **Undocumented caveat**: only the *bound* asset is protected — a VTXO carrying additional assets funded to this contract can have those redirected by whoever assembles the covenant spend.

Kukks:
Finding 1 confirmed and fixed in `a22e6047`. Finding 2 is now backed by a mutation rather than only by argument — details below, including one that still survives.

## 1. The negative assertion proved nothing — confirmed by construction

I ran your proof rather than reasoning about it. Replacing the amount comparison with a bare presence check:

```diff
-  "INSPECTOUTASSETLOOKUP", "VERIFY", "$assetAmount", "GREATERTHANOREQUAL", "VERIFY",
+  "INSPECTOUTASSETLOOKUP", "VERIFY", "DROP",
```

```
DECORATIVE covenant (amount check removed) -> *** SURVIVED: the negative test proves nothing ***
   Tests  1 passed (1)
```

`rejects.toThrow()` also carried no matcher, so any error passed it — a network failure included.

Replaced with two spends that **balance**, so arkd's conservation check is satisfied and the covenant is the only thing left to refuse them:

- **misdirection** — sats go exactly where the covenant demands, asset goes to another output entirely (your suggested case).
- **shortchanging** — the asset *is* on the covenant's output, so a presence-only covenant co-signs, but most of it has been skimmed to output 1. This is the discriminating one; misdirection alone still dies on the presence `VERIFY`.

Both need a second output, which the exact-amount sat clause leaves no room for, so the contract is funded with a dust output's surplus. The decorative covenant is now caught.

## 2. Your point about the artifact differing — a surviving mutation makes it concrete

After the fix I re-ran the mutation set. One still survives:

```
CAUGHT               decorative covenant: presence only, no amount
*** SURVIVED ***     drop the output-presence VERIFY
```

That is your finding stated as a test result. With a **fixed** `$assetAmount` the `0 >= 0` case the comment calls load-bearing cannot arise: an absent asset reports `0`, and `0 >= 500` fails on the amount comparison anyway, so the `VERIFY` is genuinely redundant *in the probed artifact*. It is not redundant in `ScriptV2`, where the comparison is input-relative — there an input that also carries none of the asset gives `0 >= 0`, which passes, and only the `VERIFY` stops it.

So the probed artifact cannot demonstrate the necessity of an opcode `ScriptV2` depends on. That is the gap, and it is now measurable rather than asserted. Recorded in the commit message so it does not get lost.

## Still to do

Closing that properly means running both cases through `ScriptV2`'s actual covenant, which is a larger change than it first looks: `ScriptV2`'s sat clause is `output0 >= INSPECTINPUTVALUE`, not `== $amount`, so a spend that pays the covenant leaves no sats for a second output — the misdirection cases need an extra funded input to be constructable at all. Doing that next, along with the medium items (handler `asset` round-trip, the inverted byte-order labels, reusing `AssetId`, the v1 scope in the title, and the "only the bound asset is protected" caveat).

Two of your medium points I want to flag as agreed rather than deferred, since they are the ones with a failure mode: the handler dropping `asset` (silent `Script mismatch` with no cause is exactly the class this PR fixes one layer up), and the byte-order comment saying the opposite of what it means in the block whose job is preventing that mistake.

CI is green on this push. Intent fees were zeroed for the run and restored afterwards.

Kukks:
Medium items done, plus a correction to one of my own claims. Four commits: `a22e6047`, `0396b6a1`, `0b72aeff`, `4334c3f4`.

### Handler `asset` round-trip — `0396b6a1`

Deeper than "serialize/deserialize have no field": **`VHTLCV2ContractParams` had no `asset` at all**, so neither direction could carry one. Round-trips it as two keys mirroring the script's own (txid, groupIndex) view, and refuses half an asset — a txid without its index names nothing, and reading either alone as "no asset" re-derives the sat-only script, the same silent drop reached by a corrupt row instead of a missing feature.

Mutation-checked: dropping either direction, or accepting half an asset, now fails.

**Not v1's handler, though.** `VHTLCContractParams` has no `nonInteractive*` fields, so a v1 contract routed through `ContractManager` can never carry a covenant leaf — and `validateOptions` already refuses an asset no leaf would bind. There is nothing to drop there. The v1 *script* class does inherit the option, which the existing test covers.

### Byte-order labels — `0396b6a1`

Fixed, and you were right about which way round. The serialized Asset ID's leading 32 bytes are canonical order (arkd's `serializeTxHash` already reversed the chainhash internals "to match the canonical txid format"); wire order is what the opcodes match, i.e. those bytes reversed back. `Options.asset.txid` now states the order it expects, since that is where a caller decides what to pass — and a caller who pre-reverses gets a contract unspendable on its covenant leaves with nothing naming why.

### Reuse `AssetId` — `0b72aeff`, with a trap

Done, and it surfaced something. The class is **not a bare export**: the entry point re-exports the module as `export * as asset`, so it is `asset.AssetId`. A bare `import { AssetId }` type-checks — the *type* comes through `export * from "./extension"` — and is `undefined` at run time. `tsc --noEmit` was clean and the e2e died with

```
TypeError: Cannot read properties of undefined (reading 'fromString')
```

I briefly read that red as evidence `AssetId` parsed differently from the hand-rolled code. It doesn't — namespaced correctly, the e2e spends against the emulator unchanged, which is the confirmation that the two parses agree. The import is namespaced with a note so it doesn't look like an oversight.

### The bound-asset caveat — `4334c3f4`

Documented, and I had to correct myself mid-write. My first version said a multi-asset VTXO "cannot be spent through the covenant leaves at all" because of `INSPECTOUTASSETCOUNT == 1`. That is the reassuring reading and it is wrong: the count constrains the covenant's **output**, so extras cannot ride along with the bound asset — but arkd's conservation is satisfied by routing them to a *different* output, which the covenant says nothing about. So they are redirectable, exactly as you said, not refused. Written as a funding rule rather than a footnote, since the safe-sounding reading is the wrong one.

### Title

Retitled to name v1's inheritance.

### Still open: finding 2

The ScriptV2 emulator execution is the one thing left, and I would rather do it properly than approximate it. Scoping note from reading `enforcePayToAsset`: its sat clause is `output0 >= INSPECTINPUTVALUE`, not `== $amount`, so a spend that satisfies the covenant leaves no sats for a second output — the misdirection and shortchanging cases both need an extra funded input (`.fund()`) to be constructable at all. That is the work, and it is also what makes the surviving presence-`VERIFY` mutation catchable.

Kukks:
`a84497ff` — the two never-executed opcodes now run against the emulator. And the mutation set that followed disagrees with one of this PR's own comments, so I want that on the record rather than buried.

## What landed

The asset clause is now **input-relative**, mirroring `enforcePayToAsset`: `INSPECTINASSETLOOKUP` for the comparison and `INSPECTOUTASSETCOUNT` for the bound. Both execute, both behave. That answers the substantive half of your point 2 — a stack-order or counting quirk in either would have made every ScriptV2 asset contract unspendable on its covenant leaves, and nothing was checking.

The sat clause stays `== $amount` rather than `>= INSPECTINPUTVALUE`. Deliberately, and it is the reason I did not go all the way to a ScriptV2-built contract in this pass: ScriptV2's form means a covenant-satisfying spend leaves **no sats for a second output**, so both misdirection cases need an extra funded input — and `.fund()` is not exercised by a single e2e in this package. That difference is BTC-side and `non-interactive-htlc.test.ts` covers it; the asset-side opcodes were the part with no coverage at all.

## The mutation set, including four survivors

```
CAUGHT       bound the output to TWO assets
SURVIVED     drop the INSPECTOUTASSETCOUNT bound entirely
SURVIVED     drop the output presence VERIFY
SURVIVED     drop the input presence VERIFY
SURVIVED     drop BOTH presence VERIFYs
```

**Gap 1 — nothing exercises what the count check is for.** No spend here injects a second asset alongside the bound one, so deleting the bound goes unnoticed. Only mis-setting it to `2` fails, and it fails on the *honest* spend, which is not the property the check exists for.

**Gap 2 — and this one contradicts the code.** The covenant comment calls the output presence `VERIFY` "load-bearing rather than defensive: without it an output carrying NONE of the asset reports `amount = 0`, and `0 >= 0` passes". Dropping **both** VERIFYs still refuses misdirection, still refuses shortchanging, and still allows the honest spend.

I do not think that disproves the claim — the `0 >= 0` path needs an **input** carrying none of the named asset, which a legitimately funded asset contract does not have. But it does mean the claim is currently unverified in the direction that matters, and the comment reads as though it were established. I have left it stated as unverified rather than either defending it or quietly deleting it, because working out whether that state is reachable — and whether anything is lost if it is — is a real question and not one I can answer by asserting.

If you already know the answer from the emulator source, that would settle it faster than I can from this side.

## Where that leaves your review

- **1** (vacuous negative) — done, `a22e6047`.
- **2** (ScriptV2 execution) — the asset opcodes are now executed, `a84497ff`. The remaining delta is the sat clause form and construction through `ScriptV2` itself, which needs `.fund()`.
- Handler round-trip `0396b6a1`, byte-order labels `0396b6a1`, `AssetId` reuse `0b72aeff`, bound-asset caveat `4334c3f4`, title retitled.

Two new items from the above (count-check coverage, and the VERIFY question) that I would rather track explicitly than leave implied. Happy to take them in this PR if you would prefer it not ship with them open.

Kukks:
`111968f7` — the count bound now has something to refuse, and one more previously-surviving mutation is closed.

Deleting `INSPECTOUTASSETCOUNT == 1` entirely used to survive: every spend carried a single asset, so the count was 1 whether or not anything checked it. Mints a second asset and funds the contract with both, which gets two things for one setup:

- **injection, refused** — the bound asset is forwarded in full, presence and amount both satisfied, and the tag-along is dumped onto the same output. Only the count bound refuses that.
- **the caveat, accepted** — the honest spend has to route the tag-along *somewhere*, and the covenant says nothing about where, so it goes to an address the contract never named and the spend succeeds. That turns "only the bound asset is protected" from a doc note into an executing assertion.

## Mutation ledger

```
CAUGHT     drop the INSPECTOUTASSETCOUNT bound                    (was surviving)
CAUGHT     compare the output against ITSELF rather than the input
SURVIVED   drop BOTH presence VERIFYs                             (still surviving)
```

The middle one is worth calling out: swapping `INSPECTINASSETLOOKUP` for a second output lookup now fails, so the **input-relative comparison is pinned** rather than merely present. That was the other half of what you flagged as never-executed — both opcodes are now not just running but load-bearing under mutation.

## The one still open

The presence `VERIFY`s. Unchanged: dropping both still refuses misdirection, shortchanging and injection, and still allows the honest spend. The `0 >= 0` path they guard needs an **input** carrying none of the named asset, and none of these spends can produce one — the contract is funded with the asset it binds.

I have deliberately not touched them. A surviving mutation says the tests cannot tell, not that the opcodes are idle, and removing a check from a money-path covenant on that basis is the wrong direction of error. The open question from my last comment stands: if the emulator's asset-lookup implementation settles whether that input state is reachable, that decides it faster than I can from here.

## Review status

- **1** vacuous negative — `a22e6047`
- **2** ScriptV2 execution — asset opcodes executed and mutation-pinned, `a84497ff` + `111968f7`. Residual: the sat clause form and construction through `ScriptV2` itself, which needs `.fund()` (unexercised anywhere in this package).
- handler round-trip + byte-order labels — `0396b6a1`
- `AssetId` reuse — `0b72aeff`
- bound-asset caveat — `4334c3f4`, now also tested
- title — retitled for v1 inheritance

Kukks:
`450a2a0f` — finding 2's core is closed. The emulator now executes the byte sequence `ScriptV2` builds.

## What it does

Adds a second artifact written to match `enforcePayToAsset` token for token — `PUSHCURRENTINPUTINDEX` for the index, `INSPECTINPUTVALUE` for the sats — and **proves** the match rather than asserting it by eye:

```ts
const fromSdk = new VHTLC.ScriptV2({ …, asset }).nonInteractiveClaimArkadeScript!
const fromArtifact = arkade.resolveAsm(scriptV2Shaped.functions.claim.arkadeScript.asm, binds)
expect(hex.encode(fromArtifact)).toBe(hex.encode(fromSdk))
```

`resolveAsm` binds the placeholders exactly as the compiler does, so this compares the bytes the emulator is about to run against the bytes `ScriptV2` emits for the same asset and destination. Equal means the spend that follows exercises **your** covenant, not a lookalike.

Then it funds that contract with the asset and claims it — accepted.

**Verified the binding is real**, not two things trivially equal:

```
CAUGHT     change ScriptV2's covenant (asset count 1 -> 2) — artifact stops matching
SURVIVED   CONTROL: pure rename inside the SDK — equality holds
```

## What it still does not do

The contract is not **built through** `ScriptV2`. The covenant bytes are proven and executed; the construction path is not. That needs a spend with an extra funded input, because `ScriptV2`'s sat clause is `out >= INSPECTINPUTVALUE` and leaves no sats for a second output — and `.fund()` is exercised by no e2e in this package. Stated in the commit rather than left implied.

If you consider the byte-equality sufficient for what you were after, this is done; if you specifically want construction covered too, that is the remaining piece and it is a bigger one.

## Three own-goals, since each looked like a covenant failure

Worth recording because the next person will hit them:

1. **`resolveAsm` is namespaced.** `arkade.resolveAsm`. A bare import type-checks and is `undefined` at run time — the same trap as `asset.AssetId` earlier on this branch. Twice in one day suggests the namespaced re-exports are a footgun in their own right.
2. **`Arkade.connect` defaults to mainnet.** The contract derived an `ark1` address on a `tark` network. `network: networks.regtest` is not optional in a test.
3. **The module-level receiver is already paid** by the test above, so `waitForVtxo` on it returned that spend's vtxo and the assertion compared two unrelated txids.

## Ledger

- 1 vacuous negative e2e — `a22e6047`
- 2 ScriptV2 execution — `a84497ff`, `111968f7`, `450a2a0f`
- handler `asset` round-trip + byte-order labels — `0396b6a1`
- `AssetId` reuse — `0b72aeff`
- bound-asset caveat — `4334c3f4`
- why conservation is the default — `b43534be`
- opt-in strict claim — `cac0ab7a`
- address-stability golden vectors — `0d8bdbd5`

The only thing I am still deliberately not doing is deleting the presence `VERIFY`s on the strength of a surviving mutation. That remains a coverage gap rather than a defect, and the case it guards needs an input carrying none of the bound asset.

Kukks:
I asked whether the presence `VERIFY`s are load-bearing, since a mutation dropping both survives the e2e, and said the emulator source would settle it faster than I could from this side. I went and read it rather than leaving that with you.

**Read from `arkade-os/emulator` `origin/master`** (the local checkout was 13 behind), `pkg/arkade/asset_opcodes.go`:

```go
// opcodeInspectOutAssetLookup pops a canonical AssetID (asset_txid asset_gidx)
// and an output index o, then looks up the matching asset output amount.
// Found:     pushes the amount as a minimally-encoded BigNum, then 1 (success flag).
// Not found: pushes 0 (BigNum), then 0 (failure flag).
```

and the not-found path confirms it:

```go
if err := vm.dstack.PushBigNum(BigNumFromUint64(0)); err != nil { return err }
vm.dstack.PushInt(0)
```

## So the comment is right, and here is the exact reachable case

With **both** `VERIFY`s removed, and neither input nor output carrying the bound asset:

```
out lookup, absent →  push 0, push 0     stack: [0, 0]
in  lookup, absent →  push 0, push 0     stack: [0, 0, 0, 0]
GREATERTHANOREQUAL →  pops two → 0 >= 0 → TRUE
```

The comparison succeeds and the asset clause is satisfied by a spend that moves none of the asset. That is exactly the `0 >= 0` the docstring describes, so **the claim was never wrong — the tests simply could not construct the state.**

## Why the mutation survives anyway

Reaching it needs an asset contract funded **without its asset**, plus a spend carrying an asset packet that declares the bound asset with zero amounts — a packet is mandatory, since `vm.assetPacket == nil` errors out at `"no asset packet"` before any lookup runs. None of the e2e cases produce that: every one funds the contract with the asset it binds, so the input lookup always succeeds and the comparison is against a real amount.

So the honest status is: **the VERIFYs guard a reachable state that no test constructs**, rather than a state that cannot exist. That is a coverage gap, and it is the correct kind — a check that is cheap, correct, and guards an edge, kept because the edge is real.

I am leaving the opcodes exactly where they are, and I would treat a future "these are redundant" simplification as needing this comment attached to it.

Constructing the case would mean funding an asset-bound contract with sats only and spending with a zero-amount packet. Worth having, not worth blocking this PR on — say the word if you would rather it land here.

Kukks:
Follow-up to my last comment: I tried to close the coverage gap I described, and **the state is not reachable through this SDK.** Worth recording, because it changes what the gap is.

## The attempt

Exactly the construction I named: a contract bound to an asset but funded with **sats only**, spent with a packet naming that asset with zero on both sides — balanced, so arkd's conservation has nothing to object to, and enough to get past `no asset packet` into the lookups.

```ts
.withAsset({ assetId, inputs: [{ vin: 0, amount: 0n }], outputs: [{ vout: 0, amount: 0n }] })
```

## What actually happens

```
asset output amount must be greater than 0
```

The SDK's **own client-side validation** refuses to build the packet. The spend never reaches the emulator, so nothing about the covenant is exercised.

## So the gap is narrower than I said

The `0 >= 0` path is real in the VM — the emulator's not-found branch pushes `0, 0` and `GREATERTHANOREQUAL` would compare `0 >= 0` — but **an SDK-built spend cannot express it.** Reaching it needs a client that bypasses this builder and hand-assembles the packet.

That does not make the `VERIFY`s pointless. A covenant guards against whoever assembles the spend, and "our own SDK will not build this shape" is a property of our tooling, not of the protocol. It does mean the check is defence against a non-SDK client rather than against a mistake this codebase can make — and that it cannot be covered by an e2e written through this API.

## One thing I got wrong on the way, since it is instructive

My first version of the test used a bare `rejects.toThrow()` and it **passed** — while proving nothing, because the SDK's validation error satisfied it just as well as a covenant refusal would have. It looked green, and dropping both `VERIFY`s still "passed" it.

That is precisely the vacuity you flagged in the original negative case, reproduced by me while trying to fix its sibling. It only surfaced because I mutated the covenant and saw the mutation survive, then pinned the matcher and read what was actually thrown. **A `rejects.toThrow()` with no matcher is not an assertion**, and I would treat any remaining one in this file as suspect.

The test is reverted — an unreachable test that reads as coverage is worse than the documented gap. Branch is unchanged at `450a2a0f`.

**Revised status for that finding:** the `VERIFY`s guard a VM-reachable state that this SDK cannot construct. Keep them; do not expect an e2e to cover them; and if the builder ever permits zero-amount packets, the test becomes writable and should be written then.

Kukks:
Closing the loop on the `assetGroupIndex` fix in `d87089c7`: `deserializeParams` does six other unguarded string→number conversions, and `Number("")`/`BigInt("")` are both **0**, so the obvious question is whether the fix should have been six wider. I checked each by construction rather than guessing.

| blank field | outcome |
|---|---|
| `strictClaimAmount` | caught — *strict claim amount must be positive, got 0* |
| `strictClaimAssetAmount` | caught — *strict claim assetAmount has no effect without asset* |
| `refundLocktime` | caught — *refund locktime must be greater than 0* |
| `claimDelay` | caught — *unilateral claim delay must greater than 0* |
| `refundDelay` | caught — *unilateral refund delay must greater than 0* |
| `refundNoReceiverDelay` | caught — *unilateral refund without receiver delay must greater than 0* |
| **`assetGroupIndex`** | **silent — derives a valid script for a different asset** |

All six die one frame down in `VHTLC.ScriptV2` with a message naming the field. `assetGroupIndex` was the sole exception for a reason worth stating: **0 is a legitimate group index.** A locktime of 0, a delay of 0 and an amount of 0 are each independently invalid, so the zero a blank string collapses to is caught for free everywhere else. Group 0 is a real asset, so nothing downstream has grounds to object — the contract simply derives a different script and surfaces as `Script mismatch` at `upsertContractRow`.

So the fix is correctly scoped at one field, and no follow-up is owed on the other six. Recording it here because "same pattern, six more instances" is the natural next review question and the answer is not obvious from reading.

Kukks:
@pietro909 — your review is the last thing blocking this and it predates every fix, so here is where your two findings landed. Arkana's latest pass calls the PR unblocked; I would rather you judge that than have me assume it.

## 1. The negative e2e passed even if the covenant was decorative — fixed

You were right, and the proof-by-construction was the useful part: a covenant of just `INSPECTOUTASSETLOOKUP VERIFY DROP` passed the old test, because spending with no asset packet dies at `vm.assetPacket == nil` before any covenant opcode runs.

Replaced with balancing spends — valid in every other respect, so only the claim under test can refuse them.

**One thing I got wrong while fixing it, worth recording.** My first replacement was the exact construction you suggested: a contract bound to an asset, funded with sats only, spent with a balanced zero-amount packet. It **passed while proving nothing** — the SDK's own client-side validation refuses to build that packet (`asset output amount must be greater than 0`), so the spend never reached the emulator. It looked green, and dropping both `VERIFY`s still "passed" it. I only caught it by mutating the covenant and seeing the mutation survive.

So the `0 >= 0` path is real in the VM but **unconstructable through this SDK**. The `VERIFY`s guard a state a non-SDK client could reach, not one this codebase can produce, and no e2e written through this API can cover it. That test is reverted rather than left in as false coverage.

## 2. No emulator execution of the ScriptV2 covenant — addressed, and here is exactly how far

`450a2a0f` adds a test that builds both:

- `fromSdk` — `new VHTLC.ScriptV2({…, asset}).nonInteractiveClaimArkadeScript`
- `fromArtifact` — `arkade.resolveAsm(scriptV2Shaped.functions.claim.arkadeScript.asm, …)`

asserts them **byte-identical**, then spends through it against the live emulator. Two independent sources, so the equality is not a value compared with itself.

**What that does establish:** the emulator executes the exact byte sequence `ScriptV2` emits for an asset covenant, including `INSPECTINASSETLOOKUP` and `INSPECTOUTASSETCOUNT` — the two opcodes nothing in the repo had ever run. A stack-order or counting quirk in either would fail here.

**What it does not:** the contract object spent is the artifact's, not one constructed by `ScriptV2` itself, so this covers the covenant **leaf bytes** rather than `ScriptV2`'s full taproot assembly — merkle root, control blocks, the other leaves. That narrower gap is what arkana tracks as its finding ④, routed to arkade-os/lightning-swap-service#125.

I think that is the honest boundary: your concern was that a stack-order quirk in the generated covenant would leave asset contracts unspendable with no test to say so, and that specific risk is now covered by execution. If you read the remaining gap as still blocking, say so and I will build the full `ScriptV2`-constructed lockup rather than argue the point.

Also in since your pass: `d87089c7` fixes a silent `assetGroupIndex` bug CodeRabbit found — `Number("")` is `0`, a valid group index, so a blank field deserialized as a *different asset* and surfaced only as an opaque `Script mismatch`. Ten refusal cases plus both range boundaries, mutation-checked.

CI is green including all four integration suites.

PR #777, opened 2026-08-20
feat(swap): open-RFQ client — broadcast bidding, sealed bids, pick the winner

Kukks:
Draft: the solver half of this is deployed and the wire is settled, but the
end-to-end path has only been exercised against unit tests here, not a live
relay. See **Not yet verified** below.

## What this is

Directed RFQ presumes the client already picked a solver from discovery data.
Open RFQ (`rfq-protocol` § 4.6) inverts that: publish trade intent to everyone
on a relay, collect competing **sealed** bids, and close with the winner over
the unchanged directed flow.

```
rfq_open  (kind 24860, PLAINTEXT broadcast, `t` = market key)   client → everyone
rfq_bid   (kind 24859, NIP-44 sealed to the open's author)      each solver → client
close     (the ordinary directed flow, untouched)               client → winner
```

Bids are sealed to the client rather than published, which makes this a
sealed-bid auction: competitors cannot read each other's prices, so there is no
undercutting race to run and no public price list to collude around.

It cannot be "the directed request minus the `p` tag". Directed content is
encrypted *to somebody* and an open RFQ has nobody to encrypt to, so whatever it
carries is public — which is why phase 1 carries no invoice and no address.

## Two modules

**`openRfq.ts` — pure, transport-free.** Parse a bid, price it, rank it. The
conformance arithmetic must round *identically* on both sides or a solver's
honest quote reads as a breach, so keeping it free of any relay is what makes it
testable against the spec's own worked example.

§ 4.6 fixes ceiling for exact-out and floor for exact-in precisely so a boundary
quote conforms from either direction. Both are pinned at the published numbers:
25 000 out costs at most **25 113**; 25 113 in yields at least **25 000** — the
same boundary from either side.

Every amount is `bigint`. § 2.1 encodes amounts as canonical decimal strings
because one whole token of an 18-decimal asset overruns a double, and a JSON
number is rounded inside `JSON.parse` before any validator could see it.

**`openRfqTransport.ts` — separate from `nostr.ts`, deliberately.** An open RFQ
is not the directed transport with a filter removed; it breaks in three places:

| directed transport | open RFQ needs |
| --- | --- |
| `authors: [solverPubkey]` | no author filter — bidders cannot be named in advance |
| one conversation key | one **per event**, derived from the sender |
| resolves one reply, closes | every reply inside a window |

The directed flow is untouched: once a winner is picked, the client builds an
ordinary `nostrRfqTransport` against that solver and runs § 4.1–4.5 as before.

## Written against what solvers publish, not only the spec

Two details that would each have shipped a client silently ignoring every real
bid:

- **bids arrive on the DIRECTED kind 24859**, not the broadcast kind. Only the
  `rfq_open` itself is 24860.
- **`min`, `max` and `fee_flat` go out as JSON numbers today**, while § 4.6
  describes canonical strings. So the parser accepts both forms, and the
  publisher always emits the canonical one — liberal in what it accepts, strict
  in what it sends.

## Two decisions worth review

- **A bid is attributed to the key that SIGNED its event, never to a field in
  the payload.** The signature is the whole reason a bid is a non-repudiable
  price commitment; reading a self-declared pubkey would discard that and let
  one solver post another's price.
- **Malformed input is dropped, not thrown.** On a shared bus anyone may reply,
  so one hostile payload must cost one bid rather than the whole auction.

## Verification

`pnpm typecheck`, `pnpm lint` and `pnpm test` all clean in `packages/swap` —
**28 files / 615 tests**, of which 32 are new here.

Mutation-checked rather than merely green: flipping the exact-out ceiling to a
floor — the single most consequential line, and the one a reimplementation is
most likely to get wrong — fails 3 tests including both worked examples.

Also covered: the canonical-decimal refusals (`1e18`, `1.5`, `" 42"`, leading
zeros), a number `JSON.parse` already rounded, an 18-decimal amount no double
could hold, exact-in bounds checked *after* pricing (§ 4.6 evaluates conformance
on the to leg), the unquotable case where `from_amount` does not exceed
`fee_flat`, and deterministic tie-breaking so relay arrival order cannot decide
a winner.

## Not yet verified — why this is a draft

- **No live-relay run.** `collectOpenRfqBids` is unit-shaped but has not
  published to a real relay and collected a real bid end to end. That is the
  claim I would most expect to be wrong.
- **No integration test against a running solver**, so the kind/tag/encoding
  agreement is verified by reading the solver's source, not by observing it.
- The `t` tag must carry the canonical corridor-qualified market key; a
  mis-spelled tag is *silently* unheard rather than refused, and nothing here
  can catch that without a relay.

### DIFF
--- a/packages/swap/src/index.ts
+++ b/packages/swap/src/index.ts
@@ -239,3 +239,23 @@ export {
     type RfqSwapActivityDeps,
     type SwapActivityInput,
 } from "./activity";
+export {
+    RFQ_BROADCAST_KIND,
+    exactInPayout,
+    exactOutCost,
+    formatWireAmount,
+    openRfqPayload,
+    parseBid,
+    parseWireAmount,
+    pickBestBid,
+    rankBids,
+    type OpenRfqSize,
+    type OpenRfqTerms,
+    type RankedBid,
+    type SolverBid,
+} from "./openRfq";
+export {
+    collectOpenRfqBids,
+    type CollectBidsOptions,
+    type CollectedBids,
+} from "./openRfqTransport";
--- a/packages/swap/src/openRfq.ts
+++ b/packages/swap/src/openRfq.ts
@@ -1,0 +1,287 @@
+/**
+ * Open RFQ — the client half of broadcast bidding (rfq-protocol § 4.6).
+ *
+ * Directed RFQ (§ 4.1–4.5) presumes the client already picked a solver. Open
+ * RFQ inverts that: publish trade intent to everyone on a relay, collect
+ * competing sealed bids, then close with the winner over the unchanged directed
+ * flow. This is the client half; the solver half is already deployed, which is
+ * why the wire details below are written against what solvers actually publish
+ * rather than only against the spec.
+ *
+ * Three phases, and only the first two are new:
+ *
+ *  1. `rfq_open` — PLAINTEXT broadcast on kind 24860. No recipient, so nothing
+ *     encryptable and nothing sensitive: no invoice, no address. Just the pair
+ *     and the size.
+ *  2. `rfq_bid` — each solver replies on the DIRECTED kind 24859, NIP-44
+ *     sealed to the broadcast's author key. Sealed to the client rather than
+ *     published, which makes this a sealed-bid auction: competitors cannot
+ *     read each other's prices, so there is no undercutting race to run and
+ *     no public price list to collude around.
+ *  3. close — the ordinary directed flow against the winner. Nothing here.
+ *
+ * This module is deliberately transport-free. Everything below is pure: parse
+ * a bid, price it, rank it. That keeps the part that must round IDENTICALLY to
+ * the solver testable against the spec's own worked examples, with no relay in
+ * the way.
+ */
+
+/** Broadcast kind for `rfq_open`. Directed traffic — including bids — is 24859. */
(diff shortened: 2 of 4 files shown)


Kukks:
`Integration (ts-sdk / settlement-delegation)` is red. It is not this diff, and
here is the reasoning rather than an assertion:

- the failing test is `packages/ts-sdk`'s
  `test/e2e/settlement.test.ts > Settlement - VtxoManager concurrent operations
  > should send offchain while VtxoManager background activity is running`,
  on `AssertionError: expected 195000 to be less than 100000`
- this branch touches **four files, all in `packages/swap`** — two new modules,
  their tests, and the index re-export
- `packages/ts-sdk` does not depend on any swap package; the dependency points
  the other way, so nothing added here is reachable from that suite
- the same run logs `ProviderUnavailableError` against
  `http://127.0.0.1:9/v1/indexer/script/subscribe` — port 9, i.e. an unset
  endpoint — which is environmental
- `master` is red on the same workflow right now, on a *different* integration
  job (`exit-providers-rotation`)

The two suites that would actually exercise this change — `Integration (swap /
all)` and `Integration (swap-rfq / all)` — both pass, as do lint, unit tests and
build.

Leaving it rather than re-running: a green re-run would prove flakiness, not
correctness, and the draft is not waiting on this check.

PR #785, opened 2026-08-20
feat(swap): fillOffer — the taker side an offer has no way to reach

### LINKED ISSUE #775: swap: no taker-side fill — a solver cannot fill an offer through this package
`@arkade-os/swap@0.0.7` ships the maker side of the atomic offer class and the observation side, but not the taker side. A solver can decide to fill an offer and cannot fill it.

Distinct from #768, which is about *expressing* an asset↔asset offer (two covenant programs, a one-asset TLV, the "exactly one of `wantAsset` or `offerAsset`" invariant). This is about a **supported** `arkade:BTC <-> arkade:<asset>` offer: it can be created, cancelled and classified, but not filled.

## What the package exports, grouped by role

Every `declare function` in `dist/index.d.ts`:

| role | functions |
|---|---|
| maker / funder | `createOffer`, `cancelOffer` |
| derivation | `offerVtxoScript`, `encodeOffer`, `decodeOffer`, `swapPrograms` |
| observation | `classifySpend`, `classifyDepositSpend`, `spendUpdate`, `watchOfferSwaps`, `restoreAssetSwaps`, `spendTxidsOf` |
| contract bookkeeping | `registerLockupContract`, `retireSettledOfferContracts` |
| **taker / filler** | **none** |

The only `fill`-named exports — `awaitOnchainFill`, `claimOnchainFill` — belong to the onchain HTLC class, not the offer class.

`fulfill` appears in the package's own JSDoc as the name of the covenant leaf:

> `fulfill` is signed by the **server alone**, but the covenant constrains it to pay output 0 to `makerWP` for at least `wantAmount` — a solver cannot take the deposit without delivering.

so the path exists in the contract. Nothing calls it.

## Why "constructible" is not the same as "provided"

`swapPrograms` is exported and documented as *"The contracts — pure data, shared verbatim with any other implementation."* So a solver CAN hand-build the spend: derive the script, construct a transaction paying output 0 to `makerWP` for at least `wantAmount`, and submit it for the Service to co-sign.

That is a poor place to leave the boundary. The fill is the money-path half — it is what takes the deposit — and the covenant's guarantee is only as good as the spend the solver builds against it. Every taker implements the same transaction shape independently, and gets it wrong differently: a wrong output index, an amount that satisfies the covenant but not the quote, or the asset-vs-sats duality on a `wantAsset` offer.

The asymmetry is also striking on its own: `createOffer` exists so a funder does not hand-build the funding transaction, and the funder is the side that risks less. `cancelOffer` exists so the funder does not hand-build the cancel. The one spend nobody is given is the one that moves the taker's own capital.

## What would close it

A counterpart to `createOffer` — given a decoded `Offer`, a wallet and the deposit outpoint, build and submit the fulfilling spend, returning the txid. Its shape can mirror `cancelOffer`'s, which already resolves the deposit by `fundingTxid` and throws when the address holds more than one spendable VTXO.

Two behaviours worth having in the package rather than in each caller, since both are already described in its JSDoc:

- **Racing a cancel.** `cancelOffer` documents that a fill in the same moment wins, and cancel then throws `"no spendable VTXO at the swap address"` — meaning the swap completed. A fill needs the mirror of that: losing to a cancel is a normal outcome, not an error.
- **Classifying its own result.** `classifySpend` already reads a spend as fill / cancel / `indeterminate`. A fill helper that returns the txid leaves every caller to re-derive whether its own submission is what landed.

## Consequence downstream

`lightning-swap-service` has the decision (`core/assetOffer.ts`, lss#110), the adapters (lss#131, lss#133) and the store (lss#152) for this corridor. The orchestrator is the remaining piece and cannot be written against this package — there is no call to make. Tracked at lss#130.

Kukks:
Closes #775.

`@arkade-os/swap` can build an offer, fund it, and cancel it. It cannot **fill**
one — so a solver holding the coins an offer asks for has no way through this
package to take it, and the atomic corridor has a maker half and no taker half.

## What it does

```ts
await fillOffer(wallet, arkServerUrl, offerHex, {
  fund,             // coins the taker supplies - inputs 1..n
  emulator,         // REQUIRED: co-signing service URL, or a provider of your own
  payoutScript,     // where the taker's proceeds land; defaults to its own address
  fundingTxid,      // selects the deposit when the swap address holds more than one
  swapAddress,      // pins the server key the covenant was built with
  assetCarrierSats, // sats at output 0 on an asset want
  emulatorPubkey,   // co-signer key override, for a network the SDK pins none for
})
```

Spends the offer's `fulfill` leaf: the deposit becomes input 0, the taker's coins
follow, output 0 pays the maker, and the change is the taker's payout. That
output-0 binding is the covenant's whole enforcement, so the fill either
satisfies it or the spend is invalid. There is no exposure window and nothing to
unwind on failure.

Both directions are supported: a **BTC want** pays `wantAmount` in sats at output
0, and an **asset want** pays through the asset packet with only a carrier on the
output itself. The wanted asset is emitted as group 0, which is the lookup index
`fulfill` reads.

## Decisions worth review

**`emulator` is required, with no default.** `fulfill` carries an
`arkadeScript`, so `ArkadeTransactionBuilder.send()` takes the covenant branch
and needs a co-signer; without one it builds the transaction and then refuses to
submit it. There is no discovery path - `ArkInfo` exposes no emulator endpoint -
so it is a pass-through parameter, mirroring `arkServerUrl`. This is the shape
@pietro909 confirmed against arkd and NArk master: arkd's `VerifyVtxoTapscriptSigs`
demands a signature from every non-server pubkey in the closure, and for a
covenant leaf that includes the emulator key tweaked by the arkade-script hash,
so a hand-assembled spend through plain `submitTx` dies as `INVALID_SIGNATURE`.
The taker of an offer is a solver, which already runs emulator-wired
infrastructure; the maker paths (`createOffer`, `cancelOffer`) need none.

**The deposit is checked against what the offer advertises.** `offerAsset` is a
TLV claim about a deposit the covenant never inspects - `fulfill` gates output 0,
what the *maker* is paid. So a fill against a deposit carrying a different asset,
or none, would go through: the taker pays `wantAmount` and receives whatever
happened to be there. Refused before the spend is built.

**`swapAddress` pins the server key.** The covenant is built against a specific
Arkade Service key; deriving it from anything but the funded address risks
reconstructing a different script than the one holding the money.

## Verification

Two test files, at two different seams, and the split is the point:

- `test/fill.test.ts` (23 tests) mocks `ArkadeContract` and records what the
  builder was asked for - the right level for shape, refusals and asset routing.
- `test/fillBuilder.test.ts` (4 tests) stubs **only** the three REST providers.
  `Arkade.connect`, the contract, the builder, the checkpoints and the packets
  are real, and the assertions read the ark tx the emulator was handed.

The second file exists because the first cannot catch a whole class of bug: with
`ArkadeContract` mocked, the real builder is never constructed, which is how a
`fillOffer` that could never submit anything passed a fully green suite. It pins
that the spend reaches the emulator, that the maker is at output 0 and the
taker's payout at output 1 (the ordering `withAsset({ vout: 1 })` depends on,
with an asset packet present), that the contract manager reaches the client, and
that `emulatorPubkey` is forwarded.

Rebased onto `master` (`45a5369`). `pnpm -r build` exit 0, then
`pnpm run test:unit`: **234 files / 4392 passed / 1 skipped / 0 failing**,
against a `45a5369` baseline of 232 / 4365 / 1 / 0.

## Not covered

**No live-relay or regtest run.** The emulator round trip is stubbed, not
exercised: this has never filled a real offer against a running Arkade Service
and emulator. That is the claim I would most expect to be wrong. A regtest e2e
against the existing `emulator` compose profile is the right next gate.

**Checkpoint completeness is not asserted on the emulator's reply.** NArk's
submitter checks that the response is fully co-signed;
`ArkadeTransactionBuilder.send()` does not. That belongs in `packages/ts-sdk`,
where it would change every covenant spend, and is deliberately not bundled
here.

The solver-side pieces that would exercise it — an offer-fill store, the fill
adapter, inventory, and a deposit observer — already exist in the service repo.
This is the SDK primitive they are all waiting on; the orchestrator that joins
them is the next piece and is deliberately not in this PR.

## Summary by CodeRabbit

* **New Features**
  * Added support for completing offers involving asset payments and asset-aware funding.
  * Added asset delivery to takers, maker payouts, surplus returns, and configurable satoshi carriers.
  * Exposed offer-filling functionality, funding types, and the asset-carrier amount for integrators.

* **Bug Fixes**
  * Improved validation for insufficient asset funding, unavailable or ambiguous deposits, unsupported assets, invalid keys, and deposits lacking the advertised asset.

### DIFF
--- a/packages/swap/src/offer.ts
+++ b/packages/swap/src/offer.ts
@@ -813,6 +813,25 @@ export async function cancelOffer(
     return txid;
 }
 
+/** Sats output 0 carries when the maker is paid in an ASSET rather than sats:
+ * the covenant checks the asset there, and the output still needs a carrier of
+ * its own. Overridable per fill via `assetCarrierSats`. */
+export const ASSET_CARRIER_SATS = BigInt(330);
+
+/**
+ * A coin the taker supplies, plus whatever assets it carries.
+ *
+ * `ArkTxInput` describes sats only, so the assets have to be declared — this is
+ * the one thing a fill cannot discover for itself, and omitting it is not
+ * cosmetic. **arkd refuses a spend whose asset packet omits an asset one of its
+ * inputs owns** (`ASSET_NOT_FOUND`), so a coin picked for its sats that happens
+ * to carry an asset takes the whole fill down unless it is named here. A
+ * wallet's own coins already carry `assets` in this shape.
+ */
+export type FillFunding = ArkTxInput & {
+    assets?: readonly { assetId: string; amount: bigint | number }[];
+};
+
 /**
  * Fill an offer — the TAKER's side, and the counterpart to {@link createOffer}.
  *
@@ -820,6 +839,17 @@ export async function cancelOffer(
  * spend to pay output 0 at least `wantAmount` to the maker's witness program, so
  * a taker cannot take the deposit without delivering. This composes that spend;
  * it does not weaken or reinterpret the covenant.
+ *
+ * **Both want sides.** For a BTC want output 0 pays `wantAmount` sats. For an
+ * ASSET want the covenant reads output 0 through `OP_INSPECTOUTASSETLOOKUP` with
+ * `lookup_index = 0`, so the wanted asset must be the FIRST group in the packet
+ * and output 0 carries only a dust sat carrier ({@link ASSET_CARRIER_SATS}).
+ * Group order follows the order the groups are added, which is why the wanted
+ * asset is added first — reordering it silently breaks the covenant.
+ *
+ * Every other asset in the spend — the deposit's own, and any the taker's
+ * funding coins carry — is routed to the taker's payout. Declaring those is not
+ * optional; see {@link FillFunding}.
  *
  * Unlike {@link cancelOffer} this writes NO local swap record. A taker filling
  * someone else's offer has no row to update — the repository dance there belongs
@@ -845,31 +875,47 @@ export async function fillOffer(
     arkServerUrl: string,
     offerHex: string,
     opts: {
-        /** Coins the taker supplies to pay `wantAmount`. Become inputs 1..n. */
-        fund: ArkTxInput[];
+        /** Coins the taker supplies to pay `wantAmount`, with any assets they
+         * carry. Become inputs 1..n. @see FillFunding */
+        fund: FillFunding[];
         /** Where the taker's proceeds land. Defaults to the wallet's own address. */
         payoutScript?: Uint8Array;
         /** Selects the deposit when the swap address holds more than one. */
         fundingTxid?: string;
         /** The funded address, to pin the server key the covenant was built with. */
         swapAddress?: string;
+        /** Sats at output 0 on an asset want. Defaults to {@link ASSET_CARRIER_SATS};
+         * raise it for a server whose dust threshold is higher. */
+        assetCarrierSats?: bigint;
     },
 ): Promise<string> {
-    const { fund, payoutScript, fundingTxid, swapAddress } = opts;
+    const {
+        fund,
+        payoutScript,
+        fundingTxid,
+        swapAddress,
+        assetCarrierSats = ASSET_CARRIER_SATS,
+    } = opts;
     const offer = decodeOffer(hex.decode(offerHex));
+    const wantedAssetId = offer.wantAsset?.toString();
 
-    // An asset want needs the taker to say which of its inputs carry the asset
-    // and how much, and needs asset change when they carry more than
-    // `wantAmount`. Guessing either produces a spend the covenant rejects for
-    // reasons the error does not explain, so this refuses rather than tries.
-    if (offer.wantAsset) {
-        throw new Error(
-            "fillOffer does not yet support an asset want: the covenant checks the asset at output 0 " +
-                "(INSPECTOUTASSETLOOKUP), which needs an explicit input->output asset binding from the caller",
-        );
-    }
     if (fund.length === 0) {
         throw new Error("fillOffer needs coins to pay wantAmount with — `fund` is empty");
+    }
+    // Checked before anything is read or spent: a fill that cannot deliver is
+    // refused here rather than by the emulator, which reports only that the
+    // covenant said no.
+    if (wantedAssetId !== undefined) {
+        const supplied = fund.reduce(
+            (sum, coin) => sum + amountOfAsset(coin.assets, wantedAssetId),
+            BigInt(0),
+        );
+        if (supplied < offer.wantAmount) {
+            throw new Error(
+                `fillOffer needs ${offer.wantAmount} of ${wantedAssetId} to pay the maker, but ` +
+                    `\`fund\` declares ${supplied} — pass coins carrying it, and declare their assets`,
+            );
+        }
     }
 
     const contractManager = await wallet.getContractManager();
@@ -908,30 +954,86 @@ export async function fillOffer(
     if (!vtxo) throw new Error("no spendable VTXO at the swap address");
 
     const payout = payoutScript ?? ArkAddress.decode(takerAddress).pkScript;
+    // A BTC want is paid in sats at output 0; an asset want is paid through the
+    // packet, so its sats leg is only the carrier the output needs to exist.
+    const makerSats = wantedAssetId === undefined ? offer.wantAmount : assetCarrierSats;
     const fill = contract.functions
         .fulfill()
         .from({ txid: vtxo.txid, vout: vtxo.vout, value: vtxo.value })
         .fund(fund)
         // Output 0, and the order is not cosmetic: the covenant inspects output
         // 0 specifically, so this must be the first `to`.
-        .to(offer.makerPkScript, offer.wantAmount)
+        .to(offer.makerPkScript, makerSats)
         // The taker's proceeds — the deposit it just took, plus any surplus of
-        // its own funding over `wantAmount`.
+        // its own funding. Appended after the outputs above, so it is vout 1.
         .change(payout);
 
-    // A deposit that IS an asset (`offerAsset`) carries it on the covenant
-    // input; move it to the taker rather than letting it vanish. Distinct from
-    // the `wantAsset` case refused above: this asset is not what the covenant
-    // checks, it is what the taker is being paid.
-    for (const a of vtxo.assets ?? []) {
-        fill.withAsset({
-            assetId: a.assetId,
-            inputs: [{ vin: 0, amount: BigInt(a.amount) }],
-            // vout 1 — output 0 is the maker's, which on a BTC want carries no asset.
-            outputs: [{ vout: 1, amount: BigInt(a.amount) }],
-        });
+    // Every asset in the spend, by the input holding it. The deposit is input 0
+    // and the funding coins are 1..n, matching the order the builder assembles.
+    const held = new Map<string, { vin: number; amount: bigint }[]>();
+    const hold = (vin: number, assets: FillFunding["assets"]) => {
+        for (const a of assets ?? []) {
+            const amount = BigInt(a.amount);
+            if (amount <= BigInt(0)) continue;
+            held.set(a.assetId, [...(held.get(a.assetId) ?? []), { vin, amount }]);
+        }
+    };
+    hold(0, vtxo.assets);
+    fund.forEach((coin, i) => hold(i + 1, coin.assets));
+
+    // The taker's asset proceeds land at vout 1 — but `change` only exists when
+    // there is a surplus, and an asset with nowhere to go is a spend arkd will
+    // refuse for a reason the error will not explain.
+    const outputsSum = makerSats;
+    const inputsSum = fund.reduce((s, c) => s + BigInt(c.value), BigInt(vtxo.value));
+    const hasPayoutOutput = inputsSum > outputsSum;
+
+    // THE WANTED ASSET FIRST — group index 0, which is the lookup index the
+    // fulfill script uses. Any other order makes the covenant read the wrong
+    // group and refuse.
+    const emitWanted = () => {
+        if (wantedAssetId === undefined) return;
+        const supplying = held.get(wantedAssetId) ?? [];
+        const supplied = supplying.reduce((s, i) => s + i.amount, BigInt(0));
+        const outputs = [{ vout: 0, amount: offer.wantAmount }];
+        const surplus = supplied - offer.wantAmount;
+        if (surplus > BigInt(0)) {
+            if (!hasPayoutOutput) {
+                throw new Error(
+                    `fillOffer has ${surplus} of ${wantedAssetId} to return but no payout output — ` +
+                        "fund with more sats than the maker's output takes",
+                );
+            }
+            outputs.push({ vout: 1, amount: surplus });
+        }
+        fill.withAsset({ assetId: wantedAssetId, inputs: supplying, outputs });
+        held.delete(wantedAssetId);
+    };
+    emitWanted();
+
+    // Everything else goes to the taker: the deposit's own asset when the maker
+    // wanted sats, and anything a funding coin happened to carry. Declaring the
+    // latter is what keeps arkd from answering ASSET_NOT_FOUND.
+    for (const [assetId, inputs] of held) {
+        const amount = inputs.reduce((s, i) => s + i.amount, BigInt(0));
+        if (!hasPayoutOutput) {
+            throw new Error(
+                `fillOffer has ${amount} of ${assetId} to return but no payout output — ` +
+                    "fund with more sats than the maker's output takes",
+            );
+        }
+        fill.withAsset({ assetId, inputs, outputs: [{ vout: 1, amount }] });
     }
 
     const { txid } = await fill.send();
     return txid;
 }
+
+/** How much of `assetId` a coin's declared assets add up to. */
+const amountOfAsset = (assets: FillFunding["assets"], assetId: string): bigint => {
+    let total = BigInt(0);
+    for (const a of assets ?? []) {
+        if (a.assetId === assetId) total += BigInt(a.amount);
+    }
+    return total;
+};
--- a/packages/swap/src/offer.ts
+++ b/packages/swap/src/offer.ts
@@ -27,6 +27,7 @@ import { concatBytes } from "@scure/btc-signer/utils.js";
 import {
     ArkAddress,
     RestArkProvider,
+    RestEmulatorProvider,
     RestIndexerProvider,
     arkade,
     asset,
@@ -36,6 +37,7 @@ import {
     type ArkTxInput,
     type IWallet,
     type NetworkName,
+    type EmulatorProvider,
     type RelativeTimelock,
 } from "@arkade-os/sdk";
 
@@ -887,6 +889,19 @@ export async function fillOffer(
         /** Sats at output 0 on an asset want. Defaults to {@link ASSET_CARRIER_SATS};
          * raise it for a server whose dust threshold is higher. */
         assetCarrierSats?: bigint;
+        /**
+         * The co-signing service — a base URL, or a provider of your own.
+         *
+         * Required, with no default: `fulfill` is a covenant path, so the
+         * emulator is what executes the arkade script and finalizes the spend
+         * with arkd. A client without one builds the transaction and then
+         * refuses to submit it.
+         */
+        emulator: EmulatorProvider | string;
+        /** Co-signer key override (33-byte compressed hex), as `createOffer`
+         * takes. Needed on a network the SDK pins no emulator key for, where
+         * connecting with an emulator otherwise throws. */
+        emulatorPubkey?: string;
     },
 ): Promise<string> {
     const {
@@ -895,6 +910,8 @@ export async function fillOffer(
         fundingTxid,
         swapAddress,
         assetCarrierSats = ASSET_CARRIER_SATS,
+        emulator,
+        emulatorPubkey,
     } = opts;
     const offer = decodeOffer(hex.decode(offerHex));
     const wantedAssetId = offer.wantAsset?.toString();
@@ -924,6 +941,14 @@ export async function fillOffer(
         indexer: new RestIndexerProvider(arkServerUrl),
         identity: wallet.identity,
         contractManager,
+        // `fulfill` is a covenant path, so the emulator executes the arkade
+        // script and finalizes with arkd. Without it the spend is built and then
+        // refused at submission — there is no default to fall back on.
+        emulator: typeof emulator === "string" ? new RestEmulatorProvider(emulator) : emulator,
+        // Only reached for the client's own emulatorKey, which this fill never
+        // derives against — the offer's own key is bound below. It still has to
+        // resolve, and on a network with no pinned key it throws without this.
+        ...(emulatorPubkey ? { emulatorPubkey } : {}),
     });
 
     // Rebuilt with the OFFER's keys, not the client's, for the same reason
--- a/packages/swap/src/offer.ts
+++ b/packages/swap/src/offer.ts
@@ -246,7 +246,7 @@ function u64(name: string, value: bigint): Uint8Array {
     if (value < BigInt(0) || value >> BigInt(64) > BigInt(0)) {
         throw new Error(`${name} does not fit the offer wire format (u64)`);
     }
-    const out = new Uint8Array(FIELDS.wantAmount.width);
+    const out = new Uint8Array(8);
     new DataView(out.buffer).setBigUint64(0, value, false);
     return out;
 }
@@ -977,6 +977,22 @@ export async function fillOffer(
     }
     const vtxo = fundingTxid ? vtxos.find((v) => v.txid === fundingTxid) : vtxos[0];
     if (!vtxo) throw new Error("no spendable VTXO at the swap address");
+
+    // The covenant checks OUTPUT 0 — what the maker is paid — and says nothing
+    // about what the deposit carries. `offerAsset` is a TLV claim, so a deposit
+    // holding a different asset or none at all still fills: the taker pays
+    // wantAmount and receives what was there, not what was advertised.
+    if (offer.offerAsset) {
+        const offered = offer.offerAsset.toString();
+        const deposited = amountOfAsset(vtxo.assets, offered);
+        if (deposited <= BigInt(0)) {
+            throw new Error(
+                `the deposit at the swap address carries no ${offered}, which this offer sells — ` +
+                    "filling it would pay wantAmount for nothing. The indexer may be behind, or " +
+                    "the offer may not be backed by what it advertises",
+            );
+        }
+    }
 
     const payout = payoutScript ?? ArkAddress.decode(takerAddress).pkScript;
     // A BTC want is paid in sats at output 0; an asset want is paid through the
--- a/packages/swap/src/offer.ts
+++ b/packages/swap/src/offer.ts
@@ -25,6 +25,7 @@
 import { hex } from "@scure/base";
 import { concatBytes } from "@scure/btc-signer/utils.js";
 import {
+    ASSET_CARRIER_SATS as SDK_ASSET_CARRIER_SATS,
     ArkAddress,
     RestArkProvider,
     RestEmulatorProvider,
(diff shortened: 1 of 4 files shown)


packages/swap/src/offer.ts
```diff
@@ -586,3 +587,126 @@ export async function cancelOffer(
...
+            "fillOffer does not yet support an asset want: the covenant checks the asset at output 0 " +
+                "(INSPECTOUTASSETLOOKUP), which needs an explicit input->output asset binding from the caller",
+        );
+    }
+    if (fund.length === 0) {
+        throw new Error("fillOffer needs coins to pay wantAmount with — `fund` is empty");
+    }
+
+    const contractManager = await wallet.getContractManager();
+    const client = await arkade.Arkade.connect({
```
pietro909, line 1034: `fulfill` is an `arkadeScript`-backed covenant path, and the real builder throws when `client.emulator` is absent. This `Arkade.connect` call only wires `arkade`/`indexer`/`identity`/`contractManager`, so every fill reaches `.send()` and fails with `covenant spends require an emulator on the Arkade client` before submit. Please pass a `RestEmulatorProvider` into the client here; the current tests mock `send()` and do not exercise this guard.

packages/swap/src/offer.ts
```diff
@@ -586,3 +587,126 @@ export async function cancelOffer(
...
+
+    const [vtxos, takerAddress] = await Promise.all([contract.getUtxos(), wallet.getAddress()]);
+    if (!fundingTxid && vtxos.length > 1) {
+        // Identical offers share one address, so guessing would fill an
+        // arbitrary deposit while the caller believes it filled a specific one.
+        throw new Error(
+            "multiple spendable deposits at the swap address — pass fundingTxid to select one",
+        );
+    }
+    const vtxo = fundingTxid ? vtxos.find((v) => v.txid === fundingTxid) : vtxos[0];
```
pietro909, line 1073: For want-BTC offers, the TLV declares the asset being sold in `offer.offerAsset`, but the selected VTXO is never checked for that asset. If the indexer is lagging or the selected deposit is malformed, this proceeds to pay the maker sats and later transfers only whatever assets happen to be present, which can leave the taker with no advertised asset. Please require the selected `vtxo.assets` to contain the declared `offer.offerAsset` with a positive amount before building the fill; `restoreAssetSwaps` already treats the missing declared deposit asset as unresolved.

packages/swap/src/offer.ts
```diff
@@ -586,3 +587,126 @@ export async function cancelOffer(
...
+    if (!vtxo) throw new Error("no spendable VTXO at the swap address");
+
+    const payout = payoutScript ?? ArkAddress.decode(takerAddress).pkScript;
+    const fill = contract.functions
+        .fulfill()
+        .from({ txid: vtxo.txid, vout: vtxo.vout, value: vtxo.value })
+        .fund(fund)
```
pietro909, line 1099: The core builder balances the spend using the contract input too (`vtxo.value + fund - outputs`), so this accepts taker funding that is short by the deposit carrier sats even though the API says `fund` pays `wantAmount`. In the exact-balance case (`fund + vtxo.value === wantAmount`) no change output is created, but the asset transfer below still targets vout 1. Please validate `sum(fund.value) >= offer.wantAmount` before sending.

pietro909 (commented):
Codex review findings.

pietro909 (changes requested):
**Blocking: `fillOffer` can never reach the network — it needs the emulator, which consuming apps don't (and can't) configure**

`fulfill` is an `arkadeScript`-backed covenant leaf, and the high-level builder routes every covenant spend through `client.emulator`:

```ts
// packages/ts-sdk/src/arkade/contract.ts (send, covenant path)
if (!client.emulator) {
    throw new Error("covenant spends require an `emulator` on the Arkade client");
}
```

But `fillOffer`'s `Arkade.connect` wires only `arkade` / `indexer` / `identity` / `contractManager` — no `emulator`. I reproduced this against the **real** `Arkade.connect` / `ArkadeContract` / `ArkadeTransactionBuilder` (mocking just the network providers): every fill reaches `.send()` and throws ``covenant spends require an `emulator` on the Arkade client`` before any submit. The current tests don't catch it because they stub out `ArkadeContract` entirely and replace `functions.fulfill()` with a recorder — so the builder that actually throws is never constructed.

**Why "just pass an emulator" is the wrong fix here.** The emulator is operator-level config, not something a wallet or this helper can discover:

- `ArkInfo` (the server's `/v1/info`) exposes **no emulator endpoint** — there's no discovery path from the Ark connection to the co-signing service.
- The only real emulator URLs in the repo are hardcoded in e2e tests (`localhost:7073`). Production code never wires one.
- NArk treats it the same way — `IEmulatorProvider` wired by DI (`AddArkadeEmulator`), i.e. explicit operator configuration, mirroring this SDK.

So adding an `emulator`/`emulatorUrl` option to `fillOffer` would force every consuming app to suddenly know about, locate, and configure a second service that nothing else in their stack needs. That contradicts how these apps are built.

**Option B — spend the covenant the way the rest of this package already does.** Every other covenant spend in `@arkade-os/swap` (the RFQ `nonInteractiveClaim` / `nonInteractiveRefund` paths) deliberately avoids `Arkade.connect`'s emulator entirely. They:

1. build the offchain tx directly with `buildOffchainTx(inputs, outputs, checkpointTapscript)`,
2. attach the covenant's arkade script as the emulator packet / `ConditionWitness` on the relevant input (the SDK already exposes `setArkPsbtField`, `EmulatorPacket`, `attachExtension`),
3. sign with the wallet identity, and
4. submit through the plain Ark provider — `ark.submitTx` → match server checkpoints → `ark.finalizeTx`.

No emulator client, no new config surface, and consistent with `refund.ts`'s existing pattern (it documents exactly this: *"the emulator is not involved… this transaction is submitted SIGNED, and the only counterparty is the Arkade server"*). `fillOffer` is currently the **only** code in the package driving a covenant through the fluent builder, which is the one place that hard-requires `client.emulator`.

Concretely, `fillOffer` would stop calling `contract.functions.fulfill()…send()` and instead assemble the `fulfill` spend by hand: deposit as input 0, `fund` coins as inputs 1..n, output 0 paying `wantAmount` to `makerPkScript`, the taker payout as change, the deposit asset moved to vout 1 via the asset packet, and the `fulfill` covenant's resolved arkade script attached to input 0 — then submit via the Ark provider it already constructs. All the validation it does today (covenant rebuild, server-key pinning, deposit selection, asset-want refusal) stays; only the spend mechanism changes.

**One protocol fact to confirm before committing to B:** whether arkd's plain `submitTx` path can validate/co-sign the offer covenant's `INSPECTOUTPUT*` introspection script itself, or whether that execution genuinely requires the separate emulator service. The RFQ covenants go through arkd directly, which suggests B is viable — but if the offer covenant's introspection opcodes are precisely what the emulator exists to execute, then the emulator isn't avoidable and we'd need Option A plus a discovery story (e.g. an emulator endpoint on `/v1/info`). That determination gates the choice, so it's worth confirming against the server/emulator before reworking the spend.

pietro909:
**Follow-up: retracting Option B — I checked arkd, and it can't work. Option A is the fix, and NArk confirms the shape.**

I owe this review a correction: the "one protocol fact to confirm" in my blocking comment is now confirmed, and it settles *against* the Option B I proposed. Verified against arkd master and NArk master:

**1. arkd has no covenant machinery at all — and it requires the emulator's signature at `submitTx`.** The covenant leaf's signer set is the tapscript signers *plus the emulator key tweaked by the arkade-script hash* (`packages/ts-sdk/src/arkade/program.ts` — `computeArkadeScriptPublicKey(emulatorKey, arkadeScript)`). For this program, `fulfill` is therefore an on-chain 2-of-2 over `{server, tweak(emulator, script)}`. At submit, arkd runs `VerifyVtxoTapscriptSigs(signedArkTx, false)` (`arkd/internal/core/application/service.go`), which mocks **only its own signer key** and demands a present, valid Schnorr signature from *every other* pubkey in the closure (`arkd/internal/infrastructure/tx-builder/covenantless/builder.go`). A hand-assembled fill submitted via plain `ark.submitTx` dies there with `missing 1 signature` → `INVALID_SIGNATURE`. The tx-builder is literally named `covenantless`; the only "covenant" hits in arkd are the unrelated `CovenantlessRedeemBranch`. An `EmulatorPacket` in the extension OP_RETURN would be inert data to arkd — it parses extensions for asset packets, never executes arkade scripts. The packet exists for the *emulator service*, which executes it, co-signs, and forwards to arkd itself (`EMULATOR_ARKD_URL`).

**2. My Option-B premise was wrong.** I claimed "the rest of this package already spends covenants this way." It doesn't: `claim.ts` and `refund.ts` spend **non-covenant** leaves (preimage-condition + receiver + server, and CLTV + sender + server) — `refund.ts` says so itself ("the emulator is not involved: it co-signs only the two covenant leaves"). The actual covenant leaves (`nonInteractiveClaim`/`nonInteractiveRefund`) are pushed by **covclaimd and the reference solver**, server-side daemons that wire the emulator. `EmulatorPacket` appears nowhere in `packages/swap/src`. There was no pattern to match.

**3. NArk master resolves this the way Option A should be shaped.** Upstream's design (which per our AGENTS.md is this repo's reference):

- Covenant detection is by inspection, not API: `ArkadePsbtExtensions.RequiresEmulatorCoSigning(coins)` — any arkade-bound input routes the whole spend to the emulator (`NArk.Arkade/Emulator/ArkadeEmulatorSpendSubmitter.cs`); everything else falls through to plain arkd.
- The emulator URL is app-level config wired once (`AddEmulatorClient(o => o.ServerUrl = …)`), per-deployment, operator-supplied. No discovery path exists there either.
- Its e2e covenant tests run the real two-service stack (arkd + emulator on `:7073` from the docker `emulator` profile) and go through the real submit handler — exactly the coverage gap this PR currently has.

**4. Consumer impact is narrower than my comment implied.** The taker of an offer is a *solver* — the same class of actor as covclaimd, which already runs emulator-wired infrastructure (`EMULATOR_ARKD_URL` in the regtest stack). The wallet (`arkade-os/arkade` web wallet) is maker-only: it calls `createOffer`/`cancelOffer`/`decodeOffer`, never `fillOffer`, and neither maker path needs a live emulator — `createOffer` bakes in only the emulator *pubkey* (network pin or deployment override), and `cancel` is a pure tapscript leaf through plain arkd. So: **`emulator`/`emulatorUrl` required on `fillOffer` (a pass-through to its internal `Arkade.connect`, mirroring `arkServerUrl`), absent everywhere else.** A consumer that ever grows a taker-side flow is by definition operating solver infrastructure and already knows its emulator URL. (A pinned per-network emulator *URL* table — next to the pinned keys in `networks.ts` — or an emulator endpoint on arkd's `/v1/info` are possible follow-ups; arkd master has no such field today.)

**So, to land this PR:**

1. Keep the spend exactly as built — the fluent builder and the emulator round-trip are correct.
2. Wire the emulator: required `emulator` (URL or `EmulatorProvider`) param on `fillOffer`, passed to `Arkade.connect`. No Option B.
3. Fix the test gap: the current tests stub `ArkadeContract` and never construct the real builder, which is why the throw was invisible. Minimum: a test that exercises the real `Arkade.connect`/`ArkadeTransactionBuilder` path with mocked network providers. Better: a regtest e2e against the existing `emulator` compose profile, matching `NArk.Tests.End2End/Arkade/ArkadeHtlcTests`.
4. Small parity nit: NArk's submitter asserts the emulator response is *fully* co-signed (signed ark tx present, checkpoint count matches) because the emulator forwards/finalizes only when it's the last required signer. `RestEmulatorProvider.submitTx` validates fields but `ArkadeTransactionBuilder.send()` doesn't check checkpoint completeness the same way — worth a line.

Kukks:
Updated: `fillOffer` now fills an **asset want** as well as a BTC want, and declares every asset it spends. Also merges `origin/master`, which this branch was 95 commits behind.

## The asset want was expressible all along

The refusal said the input→output binding had to come from the caller. That part is right, but the rest was solvable here: the covenant reads output 0 through `OP_INSPECTOUTASSETLOOKUP` with `lookup_index = 0`, and the builder's group order follows the order groups are added (`contract.ts:832-846` maps `assetSpecs` in call order). So adding the wanted asset **first** satisfies the covenant.

Output 0 then carries only a dust carrier — `ASSET_CARRIER_SATS` (330, overridable via `assetCarrierSats`) — not `wantAmount` sats. Paying `wantAmount` sats there, as the BTC path does, would hand the maker the taker's own money on top of the asset.

## A bug in the BTC path it already supported

arkd refuses a spend whose asset packet omits an asset one of its inputs owns (`ASSET_NOT_FOUND`), and only the deposit's assets were declared. A taker funding with sats coins that happen to carry an asset — routine for a solver holding a float — took the whole fill down with an error naming neither the coin nor the asset.

Every asset on every input is now declared: the deposit as input 0, the funding coins as 1..n, matching the order the builder assembles them (`contract.ts:670-680`).

`ArkTxInput` describes sats only, so what a funding coin carries is the one thing the fill cannot discover. `FillFunding` adds it, and a wallet's own coins already have that shape.

## Refusals moved ahead of the emulator

- A want the funding cannot pay for names the shortfall, instead of leaving the emulator to report only that the covenant said no.
- Assets return to the taker at the payout output, which the builder only creates when there is a sats surplus (`contract.ts:661`). An asset with nowhere to land is refused here rather than by arkd.

## Verification

- `pnpm -C packages/swap test` — 649 → **655** passing (7 added, 1 removed: the asset-want refusal test, replaced with a shortfall test). No change to the other 648.
- `pnpm run build`, `pnpm -C packages/swap run typecheck`, `smoke:dist`, `pnpm run test:unit` — all clean.
- Mutation-checked, since the two rules here are silent when wrong:
  - dropping funding-coin assets from the packet → 3 tests fail;
  - emitting the wanted asset after the other groups instead of first → the group-order test fails (`cc…` lands at group 0 instead of `bb…`).

The tests remain shape-only, as this suite is designed to be. The asset-want shape they pin is the one an implementation of this covenant settles for real on regtest — same group-0 rule, same 330-sat carrier, same every-input-declared packet.

Kukks:
Second push, `41d7b588` — and this one is the reason the answer to "can we drop our custom fill code for this" was not yet yes.

## `fillOffer` could never submit a fill

`fulfill` carries an `arkadeScript`, so `ArkadeTransactionBuilder.send()` takes the covenant branch and throws:

> covenant spends require an `emulator` on the Arkade client

`Arkade.connect` was called without one. There is no per-network default to fall back on — `emulator` is optional on `Arkade.connect` precisely because pure-tapscript contracts do not need it, which is why `cancelOffer` connecting the same way is correct and this was not.

So the fill built a transaction and then refused to submit it, every time. **Nothing caught it, and nothing could have**: the tests mock `ArkadeContract` wholesale, so `send()` returned a txid without ever reaching the real builder. It is a shape-only suite, and this was a wiring bug rather than a shape bug.

`emulator` is now a required option — a base URL, or a provider of your own — and `emulatorPubkey` passes through for a network the SDK pins no co-signer key for, where `Arkade.connect` otherwise throws while resolving one.

The mock now records what `connect` was configured with, so the gap has a test: removing the emulator from the connect call fails `connects WITH an emulator, without which the spend cannot be submitted`.

## Also worth knowing

While checking the emulator question I confirmed `getUtxos` needs no registration for a foreign script — with a `contractManager` set it is repository-backed for *registered* contracts and falls back to the indexer otherwise. So a taker filling someone else's offer does not register anything, which is the right behaviour and worth not regressing.

This branch now also carries `master` including #822, so `fillOffer` sits on top of the TLV/exit-closure work: an offer with an exit closure decodes, and `swapProgramBinding` rebuilds its three-leaf tree, so the `swapPkScript` check here passes for solverd-made offers rather than failing as a key rotation.

## Verification

- 649 → **656** passing on this branch (baseline taken after merging master), typecheck, build, `smoke:dist` and `test:unit` all clean.
- Mutation-checked: dropping `emulator` from the connect call fails the new test; dropping funding-coin assets from the packet fails 3; emitting the wanted asset after the other groups fails the group-order test.

Still shape-only, so the remaining gap is a settlement run — worth doing before this is relied on for real money.

Kukks:
Rebased onto `master` (`45a5369`, 18 commits ahead of the old base) and pushed `88dc95b3`. Six findings addressed, one disproved, two answered.

## @pietro909's blocking finding, and your own follow-up

Your retraction settled the fork, and your landing list is the one I worked from:

1. **Keep the spend as built** — unchanged.
2. **Wire the emulator** — done in `41d7b588`, before this push. `emulator: EmulatorProvider | string` is required with no default and is passed to `Arkade.connect`.
3. **Fix the test gap** — this push. Detail below; it is the substantive part.
4. **Checkpoint-completeness parity nit** — *not* taken, deliberately. It lives in `ArkadeTransactionBuilder.send()` in `packages/ts-sdk`, and it would change how **every** covenant spend treats an emulator response, not just this one. I have no way to confirm from here whether an emulator may legitimately return a different checkpoint count, and guessing wrong turns a working claim into a hard failure. Worth its own PR against the SDK with the emulator's contract in hand — flagging rather than smuggling it in.

### The test gap is closed by a second file, not by more of the first

You put your finger on exactly why the bug was invisible: `fill.test.ts` stubs `ArkadeContract` and replaces `functions.fulfill()` with a recorder, so the real `ArkadeTransactionBuilder` — the thing that throws — is never constructed. Adding assertions to that file cannot fix that; the seam is wrong.

New `packages/swap/test/fillBuilder.test.ts` stubs **only** `RestArkProvider`, `RestIndexerProvider` and `RestEmulatorProvider`. `Arkade.connect`, `ArkadeContract`, `ArkadeTransactionBuilder`, the checkpoints, the prev-ark-tx resolution and the asset/emulator packets are all real, and the assertions read the ark tx the emulator was actually handed. It pins four things:

- the spend **reaches** `emulator.submitTx` (the original bug, now a regression test against the real builder);
- output 0 is the maker's script for `wantAmount` and **output 1 is the taker's payout** — with an asset packet in the spend, so the extension output is present;
- the wallet's contract manager reaches the client, asserted behaviourally: the indexer returns nothing and the manager is the only source of the deposit, so a fill that finds it cannot have dropped it;
- `emulatorPubkey` is forwarded — a malformed one is rejected by `resolveEmulatorPubkey` inside `Arkade.connect`, which is the only thing on the path that looks at it.

## The critical one: a fill could pay for a deposit carrying nothing

@coderabbitai flagged this and @pietro909 raised the same thing on `offer.ts:978`. **Both are right, and it was still open.**

`offerAsset` is a TLV claim about a deposit the covenant never inspects. `fulfill` gates output 0 — what the *maker* is paid — and says nothing about what the deposit holds. So a fill against a deposit carrying a different asset, or none at all, went through: the taker paid `wantAmount` and received whatever happened to be there.

`fillOffer` now checks the selected deposit carries `offer.offerAsset` with a positive amount, before the spend is built.

**What made it invisible is worth stating plainly: the existing tests were filling an unbacked deposit throughout.** The default deposit fixture carried no assets, so every builder-shape test was asserting on a fill that should not have been buildable. Adding the guard turned seven of them red. The fixture now carries what the offer sells, and an unbacked deposit is its own case.

## @arkana-ai-bot's findings

| # | finding | outcome |
|---|---|---|
| 1 | `vout: 1` hardcoded, no builder guarantee cited | **Pinned**, against the real builder, with an asset packet present |
| 2 | no `contractManager` forwarding test | **Added**, behavioural |
| 3 | two assets on one deposit VTXO untested | **Added** |
| N1 | `decodeOffer` accepts a zero `exitDelay` | **Disputed** — below |
| N2 | `u64` buffer tied to `FIELDS.wantAmount.width` | **Fixed** — `new Uint8Array(8)` |
| N1 (2nd pass) | `emulatorPubkey` forwarding untested | **Added** |

On finding 1 specifically: the guarantee is real — `ArkadeTransactionBuilder.build()` copies `this.outputs`, appends the change output last, and inserts nothing between (`packages/ts-sdk/src/arkade/contract.ts`). But an implementation fact cited in a comment is not a test, so the assertion is on the built transaction instead. Mutating `vout: 1` to `vout: 0` is killed by three tests, including `expected [ { vout: +0, amount: 2000n } ] to deeply equal [ { vout: 1, amount: 2000n } ]`.

### Disputed: `decodeOffer` accepting a zero `exitDelay` is deliberate, and refusing it is a bug

Raised by both bots. I implemented it, and it broke an existing test that exists precisely to forbid this change:

```
FAIL test/offer.test.ts > swap offer > the optional records
     > DECODES a zero exit delay it would refuse to emit
Error: exitDelay must be a positive relative locktime
```

That test (`packages/swap/test/offer.test.ts`) carries the reasoning:

> deliberately asymmetric. The reference emits any ExitDelay it is given, so refusing one on the way in would make us unable to read an offer solverd can publish — and a useless exit closure is the maker's own protection to waive, not ours to reject on their behalf. Encode stays strict so WE never publish one.

The stated impact also does not follow. Arkana's premise is that "the sender, if it validated via `encodeOffer`/`assertExitDelay`, refused that value and compiled a two-leaf tree" — but a sender that refused the value never emitted the offer. A sender that *did* emit `exitDelay = 0` compiled the three-leaf tree, so decoding it faithfully is exactly what makes the two `swapPkScript` values agree. Refusing it is what would make us unable to read an offer that is well-formed on the wire.

And the mismatch it fears cannot pass silently in any case: both spend paths rebuild the covenant from the offer's own keys and compare it against `offer.swapPkScript`, throwing `rebuilt covenant does not match the offer's swapPkScript` before any network call — `offer.ts:754` in `cancelOffer` and `offer.ts:963` in `fillOffer`.

Encode stays strict, so this client never publishes one. No change made.

### Answered: the `fund` shortfall (`offer.ts:988`)

@pietro909, the half of this that could lose an asset is closed. `hasPayoutOutput` (added in `2f66c333`) refuses the exact-balance case you named — an asset routed to a vout the builder never creates — with `no payout output`, and there is a test for it.

The other half I have deliberately not added a check for. `sum(fund.value) >= wantAmount` would refuse fills that are correct: the deposit's carrier sats belong to the taker after the fill, so drawing on them leaves the taker's net position identical (`payout = vtxo.value + fund - wantAmount` either way). A gross shortfall already fails in the builder with `insufficient inputs - outputs N exceed inputs M`. Happy to add it as a strictness knob if you would rather the API mean literally "`fund` pays `wantAmount`", but as a hard refusal it costs real fills.

## Mutations on the new deposit guard

| # | mutation | result |
|---|---|---|
| 1 | guard deleted | **killed** — `promise resolved '...' instead of rejecting`, all 3 unbacked cases |
| 2 | `<= 0` weakened to `< 0` (a zero balance passes) | **killed** — same 3 |
| 3 | compares `wantAsset ?? offerAsset` instead of `offerAsset` | **survived — equivalent mutant.** The guard is inside `if (offer.offerAsset)`, and both codec paths enforce "exactly one of `wantAsset`/`offerAsset`", so `wantAsset` is necessarily `undefined` there. No test can distinguish it. |
| 4 | checks the taker's `fund` instead of the deposit | **killed** — same 3 |
| 5 | asset routed to `vout: 0` (the maker) instead of `vout: 1` | **killed** — `expected [ { vout: +0, amount: 2000n } ] to deeply equal [ { vout: 1, amount: 2000n } ]` |

## Gate

`pnpm -r build` exit 0, then `pnpm run test:unit`:

| | baseline `45a5369` | this head `88dc95b3` |
|---|---|---|
| test files | 232 | 234 (+2) |
| tests | 4365 passed, 1 skipped | 4392 passed, 1 skipped |
| failing | 0 | 0 |

`swap` goes 32 files / 780 tests to 34 / 807: `fill.test.ts` (23) and `fillBuilder.test.ts` (4) account for all 27, so nothing was overwritten.

Still no live-relay run — the emulator round trip is stubbed, not exercised. @pietro909, your suggestion of a regtest e2e against the existing `emulator` compose profile is the right next gate, and I have not done it here.

Kukks:
@coderabbitai's nitpick is real and taken in `1321bf12`. The SDK exports `ASSET_CARRIER_SATS = 330` from the ark asset rail and this file restated it as `BigInt(330)` — same protocol dust threshold, two spellings, free to drift. Now derived.

Verified the bare import resolves **at run time** (`330`, `number`) rather than trusting the type: several exports in this SDK are namespaced and type-check while being `undefined` at run time, so a compile-clean import is not evidence here.

Gate unchanged: 234 files / 4392 passed / 1 skipped / 0 failing, `pnpm -r build` exit 0.

Issue #734, opened 2026-08-12
Expose the Ark server info through IReadonlyWallet so plugins can drop the arkServerUrl parameter


tiero:
Every covenant-building entrypoint in `@arkade-os/swap` takes `arkServerUrl` next to the wallet — `createOffer(wallet, arkServerUrl, params)`, `request*(wallet, arkServerUrl, transport, params)` — even though the wallet was constructed from that same URL. The redundancy exists because neither `IWallet` nor `IReadonlyWallet` exposes the server connection: no URL, no provider, no `getInfo()`. The concrete `Wallet` holds it all internally (and already caches an info snapshot, `src/wallet/arkInfoSnapshot.ts`), but a plugin coding against the interface can't reach it without downcasting.

What the swap entrypoints actually need from `getInfo()`:
- `network` — per-network emulator pin (`resolveEmulatorPubkey`) and address hrp
- `signerPubkey` — the covenant's server key
- `unilateralExitDelay` — VHTLC claim-delay tiers

Proposal: add to `IReadonlyWallet` something like

```ts
/** Snapshot of the Ark server's info for the server this wallet is connected to. */
getArkInfo(): Promise<ArkInfo>;
```

backed by the wallet's existing cached snapshot (so plugins don't each re-fetch `/v1/info`). Exposing the ark provider itself would also work but leaks more surface than the use cases need.

Then `@arkade-os/swap` (and future plugins) can drop the URL positional entirely: `createOffer(wallet, params)`, `requestLightningSend(wallet, transport, params)` — the integration prerequisite becomes just a wallet. `registerOfferContract`'s `Arkade.connect` would still need provider construction, so either that also hangs off the wallet/contract manager, or the plugin keeps a URL only for that internal step — worth deciding together with this.

Context: #726, #732 (which reduced the prerequisites to wallet + URL), and #733 (companion API-surface issue). Repository-interface versioning note: this is an interface addition, so custom `IWallet` implementations need a compile-time bump.

PR #793, opened 2026-08-25
feat(swap): `createSwapClient` — one client over RFQ corridors and offer swaps

louisinger:
A facade, `createSwapClient`, unifying the package's two swap families behind one API:

- **RFQ corridors** (`arkade→lightning`, `lightning→arkade`, `arkade→onchain`) — quoted over a solver round trip, driven by `RfqSwapManager`.
- **Spot arkade↔arkade offers** — priced client-side from the market card's feed, funded as an offer covenant, filled non-interactively by any solver watching the arkd stream, driven by `watchOfferSwaps`.

Both families already worked as building blocks; every consumer had to write the same glue: the dispatch between them, the funding step each expects, the record/origin assembly `addSwap` needs, and the manager callback wiring the README documented as prose. That glue is now this module.

## The design rule: the market picks the backend

A market with a non-arkade leg carries an RFQ rendezvous (`discovery_pubkey` + `transports` on its card), so its quote is a solver round trip. A spot market carries a price feed and no rendezvous, so its quote resolves locally and the covenant is the whole protocol. **Swapping a backend is a card change, never a client change.**

Corridor fields are read off the market at runtime because `@arkade-os/solver-discovery` 0.1.x types predate corridor markets; a pre-corridor index dispatches spot, which is the correct reading of that data.

## Example

```ts
import { createSwapClient, findMarket, nostrRfqTransport } from "@arkade-os/swap";

const client = createSwapClient({
    wallet,
    arkServerUrl,
    repository,                        // one AssetSwapRepository backs both stores
    transportFor: (market) =>          // corridor markets only; never called for spot
        nostrRfqTransport({ relays: market.transports.nostr.relays,
                            solverPubkey: market.discovery_pubkey }),
    discovery: { network: "bitcoin", registryUrl: REGISTRY_URL },
});

await client.start();                  // RfqSwapManager + watchOfferSwaps, restored from the stores
client.onUpdate((u) =>                 // ONE stream, both families
    console.log(u.family, u.family === "offer" ? u.swap.status : u.swap.state));

const markets = await client.markets();

// Spot arkade↔arkade — no solver contact; ../solver fills from the stream
const spot = findMarket(markets, "btc", USDT_ID);
const q1 = await client.quote(spot.market, { kind: "spot", give: "base", giveAmount: "0.01" });
const swap1 = await client.accept(q1);           // createOffer + funding tx with the TLV
// later, if unfilled: await client.cancel(swap1.swap.id)

// Lightning send — RFQ, verified local derivation, then fund and go offline
const ln = markets.find((m) => m.quote_corridor === "lightning");
const q2 = await client.quote(ln, { kind: "ln_send", invoice: toInvoiceFacts(bolt11) });
await client.accept(q2);                          // funds the lockup; manager watches chain-only

// Lightning receive — nothing to fund; show the hold invoice, manager claims with P
const q3 = await client.quote(ln, {
    kind: "ln_receive", amount: 50_000, amountSide: "to",
    covclaimdPubkey, decodeInvoice,
});
showToPayer(q3.invoice);
await client.accept(q3);
```

## Summary by CodeRabbit

- **New Features**
  - Added a unified swap client supporting spot, Lightning, and on-chain swap flows.
  - Added market discovery, quote creation, swap acceptance, cancellation, monitoring, and lifecycle controls.
  - Added unified swap updates across supported swap types.
  - Added validation for unsupported market and quote combinations.

- **Bug Fixes**
  - Improved error handling for incomplete swap details and invalid corridors.

- **Tests**
  - Added coverage for market routing, quote validation, and swap status updates.


tiero:
- do we really need kinds? ie. ln_send and ln_receive, spot. I would not be that explicit imho, I don't see why it's important to branch out in kinds, when we have assets and corridors already. As mentioned assume that "spot market" are also RFQ and will soon go through Nostr for getting a quote

- covclaimdPubkey, why we need it? also not declared anywhere in the example

- decodeInvoice, not declared anywhere in the example

- arkServerUrl can come from wallets, this is pending work from @gringokiwi  https://github.com/arkade-os/ts-sdk/issues/734

PR #803, opened 2026-08-26
Remove arkServerUrl from swap methods 

### LINKED ISSUE #734: Expose the Ark server info through IReadonlyWallet so plugins can drop the arkServerUrl parameter

tiero:
It closes #734 

## Summary

Removes the `arkServerUrl` positional argument from the five covenant-deriving entrypoints in `@arkade-os/swap` (`createOffer`, `requestLightningSend`, `requestLightningReceive`, `requestOnchainSend`, `requestOnchainReceive`). These functions now source server information directly from `wallet.getArkadeInfo()` instead of building a provider from a URL, eliminating a redundant `/v1/info` round-trip and simplifying the API surface.

## Key Changes

- **API simplification**: Removed `arkServerUrl` positional from five entrypoints; callers now pass only `wallet` and other required arguments
- **New wallet method**: Added `IReadonlyWallet.getArkadeInfo()` that returns live server info with fallback to cached snapshot when offline
- **Type rename**: `ArkInfo` → `ArkadeInfo` (with deprecated alias for backward compatibility)
- **Provider flexibility**: `ArkadeConnectOptions.arkade` now requires only `getInfo()`; `submitTx`/`finalizeTx` are optional, allowing info-only clients
- **Wallet implementations**: Updated standard, service-worker, and Expo wallet implementations to provide `getArkadeInfo()`
- **Test refactoring**: Removed mocked `RestArkProvider` instances from tests; server info now comes from wallet stubs with `getArkadeInfo()` method

## Implementation Details

- `getArkadeInfo()` resolves live info first, falls back to boot snapshot on retryable failures, and propagates terminal errors
- The wallet's own derived state (server key, dust amount) remains pinned to boot epoch; only live fields like fees and checkpoint rotate
- `cancelOffer` and `watchOfferSwaps` retain their `arkServerUrl` argument since they need broadcast and indexer access beyond what `getArkadeInfo()` provides
- All test fixtures now define server info once (e.g., `ARK_INFO` constant) to ensure covenant derivation consistency between wallet and transport stubs

https://claude.ai/code/session_01XexX3GchteWXPZojgH1GgF

## Summary by CodeRabbit

- **Breaking Changes**
  - Swap offer, payment request, cancellation, and swap-watching APIs no longer require a server URL; details and capabilities come from the wallet.
  - `ArkInfo` is now `ArkadeInfo`; the former name remains available as a deprecated alias.

- **New Features**
  - Wallets provide live or cached Arkade information, chain-reading access, and transaction broadcasting.
  - Service-worker and Expo wallets support the new wallet-based capabilities.

- **Documentation**
  - Updated migration guides, READMEs, and examples for the revised APIs.


tiero:
Thanks — went through each point. One fixed and pushed, one correction with evidence, one out of scope, one I read the opposite way.

## Fixed: typed errors erased at the service-worker boundary

Valid, and pushed in `7b5e97a`. `ServiceWorkerReadonlyWallet.getArkadeInfo()` now attaches the original as `cause` instead of stringifying it, with a regression test.

Worth noting for anyone relying on this: `structuredClone` drops the prototype, so `instanceof ProviderUnavailableError` cannot survive the worker boundary regardless of what we do here. `name` does survive, and stringifying was losing even that. So the branch is `err.cause.name`, not `instanceof`.

## Correction: TypeScript does catch the old call sites

> TypeScript will not catch it cleanly — the extra argument would be silently ignored at runtime … the derivation would use a URL string as the `transport` argument and crash at the first RPC call

This isn't right, and it matters because it's the stated reason the cross-repo issue is blocking. Both old shapes are hard compile errors — excess arity, plus a `string` in a `RfqTransport` slot. Verified against this branch with `@ts-expect-error` probes compiled by `packages/swap/tsconfig.json`:

```ts
// @ts-expect-error old arity: URL positional is gone
createOffer(wallet, url, { wantAmount: 1000n, wantAsset });

// @ts-expect-error old arity: URL positional is gone
requestLightningSend(wallet, url, transport, { invoice: {...} });
```

Both directives fired (no `TS2578: Unused '@ts-expect-error'`), so both calls do error. Excess arguments to a fixed-arity function are `TS2554`. There is no silent-runtime-breakage path here — a downstream repo that typechecks will fail its build, loudly, at the call site.

## Out of scope: `arkade-wallet-stack` call sites

Correct that they need updating, and that's inherent to a breaking API change — it's recorded under **Breaking Changes** in `CHANGELOG.md` with before/after signatures, and in `packages/swap/MIGRATION.md`. This session is scoped to `arkade-os/ts-sdk`, so I can't read or edit that repo. Sequencing it against a `@arkade-os/swap` release is a maintainer call.

## Epoch skew: I think the reasoning is inverted

> the offer VTXO is built against a key that the wallet's own `arkProvider` will refuse to co-sign for

The `arkProvider` *is* the operator. If `getArkadeInfo()` returned epoch N+1, that is because the operator answered N+1 — so it will co-sign for N+1. Binding the covenant to the operator's *current* signer is the correct behaviour, and it's why this method resolves live rather than returning the boot snapshot.

The wallet's own pinned key going stale is real but separate: it affects the wallet's default VTXO address, not this covenant, which takes `server` explicitly from `info`.

The risk actually runs the other way — a cache fallback returning a *superseded* key when the operator has moved on. That is bounded by design: a rotated-out signer stays valid until its `cutoffDate`, which is what `ArkadeInfo.deprecatedSigners` exists to express, and the SDK keeps old-signer rows active and fans `checkpointUnrollCandidates` over them.

Happy to be shown wrong if you're modelling a case I'm missing.

---

tiero:
## CI: `Integration (ts-sdk / ark-core)` failed on `9ef74a1` — not this PR's

Standing down on this one, with the reasoning laid out so it can be checked.

**Why it isn't this change:**

- `7b5e97a` passed **all 8 checks**, including this exact suite, ~20 minutes earlier ([run #2907](https://github.com/arkade-os/ts-sdk/actions/runs/33002941099)).
- `9ef74a1` differs from `7b5e97a` by **one file**: `packages/ts-sdk/test/serviceWorker/wallet.test.ts` (+13/−1). No source change — `git diff 7b5e97a..9ef74a1 --stat` shows nothing else.
- This job runs exactly one file: `test_files: test/e2e/ark.test.ts` (`.github/workflows/ci.yml`). The file I changed is a unit test, excluded from the e2e run. **The failing suite cannot execute the only thing that changed.**
- In the same run, the other six completed integration jobs passed — including both swap suites, which are the ones this PR actually touches.

**Related prior occurrence:** `Integration (ts-sdk / exit-providers-rotation)` failed the same way earlier today on `b28b94d` ([run #2898](https://github.com/arkade-os/ts-sdk/actions/runs/32975766448)) and passed on the very next commit. These ts-sdk regtest suites appear to be flaky under load.

**No fix ported, because there is nothing identified to port.** I could not read the failing assertion: `get_job_logs` only tails, and the `Capture logs on failure` step dumps every container's full log, so the reachable window (19:28:08→19:28:33) is entirely that capture. The container logs it did surface show `VTXO_ALREADY_SPENT` on renewal and a fulmine `dial tcp [::1]:9005: connection refused` — environment/timing noise rather than a specific assertion.

**I cannot re-run.** Both `rerun-failed-jobs` and `workflow_dispatch` return `403 Resource not accessible by integration` for this token. A maintainer re-run would confirm or refute this in one click, and is the fastest way to settle it.

I'm deliberately not "hardening" `test/e2e/ark.test.ts` — it is outside this PR's scope and I would be guessing at a failure I could not read. Keeping the PR watched until it is green.

---

tiero:
Withdrawing the re-run request from my earlier comment — it settled itself.

`9233aff` is a CHANGELOG-only delta from `9ef74a1`, and [run 33007737689](https://github.com/arkade-os/ts-sdk/actions/runs/33007737689) came back green on all 8 checks including `Integration (ts-sdk / ark-core)`. Identical test code, opposite result, so the failure on `9ef74a1` was a flake in that e2e job rather than anything in this diff. No maintainer action needed.

PR is `mergeable_state: clean` and waiting on review only.

---

tiero:
### Reviewer decision: ship this breaking, or add a compat layer?

Three surfaces break. Only the first is likely to affect anyone:

1. The five swap entrypoints lose the `arkServerUrl` positional.
2. `IReadonlyWallet.getArkadeInfo()` is a new **required** member — breaks third-party wallet implementations only.
3. `Arkade.arkProvider` narrows to `ArkadeServerProvider` (`submitTx`/`finalizeTx` optional on the exposed type).

**(1) can be made fully non-breaking, cheaply.** The old and new second argument are structurally disjoint — `string` vs `RfqTransport`/`params` — and the arities don't collide (2 vs 3, 3 vs 4), so overloads are unambiguous with a `typeof a2 === "string"` demux. The compat body is nearly free, because the change is a one-expression swap at all five sites:

```ts
const resolveInfo = (wallet: IWallet, url?: string) =>
    url ? new RestArkProvider(url).getInfo() : wallet.getArkadeInfo();
```

New callers still skip the extra `/v1/info`; only legacy callers pay it. There's precedent in the SDK for keeping such a surface alive rather than deleting it — `BaseWalletConfig.arkServerUrl` carries `@deprecated Pass an explicit arkProvider instance instead. URL-based configuration will be removed in a future major version.` for exactly this URL→provider move.

**(2)** could instead follow the repo's own capability-probe convention (`src/wallet/hdWalletCapable.ts` + `isHDWalletCapable`), leaving `IReadonlyWallet` untouched — though that doesn't literally satisfy #734's "expose the server info *through IReadonlyWallet*".

**(3)** I'd leave as-is. Pinning the field to its old type would have it claim `submitTx` exists on an info-only client, reintroducing the assumption the runtime guard was added to catch.

Weighing against a compat layer: both packages are pre-1.0, there are no in-repo `src` callers of the five entrypoints (only docs and tests), and TypeScript hard-errors on stale call sites (`TS2554`) rather than failing silently — so downstream gets a compile error, not a runtime surprise.

Happy to add the overloads if you'd rather this land non-breaking.

---

tiero:
Decision made — disregard the open question in my previous comment.

These land in **v0.5**, which is taking breaking changes together with deprecations. No compat layer; this PR stays as-is.

---

bordalix:
@tiero  why not add the server url to `ArkadeInfo` so that we could also remove the param `arkServerUrl` from `cancelOffer()` and `watchOfferSwaps()`?

tiero:
@bordalix Right track — it *is* inconsistent that those two still take a URL when the other five no longer do. But I think `ArkadeInfo` is the wrong place for the cure.

### Why not `ArkadeInfo`

**It's the `/v1/info` response.** Every field on it is something the server advertises. A URL is how *you* reached the server: arkd binds to `0.0.0.0:7070` and has no idea it's fronted by `https://arkade.computer` — the hostname and TLS live in the proxy. It would need an operator-set config that can be wrong or unset, and one server often has several valid URLs at once (public domain, `.onion`, cluster DNS, localhost in dev). The alternative is the SDK stamping it on client-side, but the type is persisted — `arkInfoSnapshot.ts` serializes it into a versioned `StoredArkInfoSnapshot` — so we'd be storing a field the server never sent, next to the `digest` it did.

**It regresses the service-worker wallet.** `ServiceWorkerReadonlyWallet` holds no provider page-side by design; the worker owns the connection. Give the page a URL and it builds its own `RestIndexerProvider`, bypassing the worker's connection, rate gate and info cache.

**It doesn't address what those two actually need** — which isn't info, it's the **indexer** (plus broadcast, for cancel). The URL is only how they currently manufacture one.

### The actual gap

`ReadonlyWallet` already has a public `readonly indexerProvider` (`wallet.ts:739`), right next to `arkProvider`. A concrete wallet has everything those calls need *today*. What's missing is the seam: `IReadonlyWallet` exposes no provider at all, so functions taking the interface have nowhere to ask — hence `new RestIndexerProvider(url)` at `watch.ts:121` and `offer.ts:507-508`.

### Proposal

Not `getIndexer(): Promise<IndexerProvider>`. That needs 14 RPC message types for the SW wallet, leaks the transport into the interface, and hands plugins the raw `getVtxos` that `scripts/check-provider-boundary.mjs` **fails the build** over inside the SDK ("the only sanctioned way for SDK logic to read VTXOs from an `IndexerProvider`" is `getNormalizedVtxos`). It also hands out a call that `414`s on a wide script list and silently returns page one — `getAllNormalizedVtxos()` already solves both.

Instead, two narrow purpose-shaped seams, split on the readonly/full line:

```ts
/** Chain reads a plugin needs beyond its own wallet's VTXOs. */
export interface ArkadeReader {
    /** Normalized, chunked and paged to exhaustion — `getAllNormalizedVtxos` semantics. */
    getVtxos(scripts: string[], opts?: VtxoScriptQuery): Promise<NormalizedVirtualCoin[]>;
    getVirtualTxs(
        txids: string[],
        opts?: PaginationOptions,
    ): Promise<{ txs: string[]; page?: PageResponse }>;
}

/** Submitting a signed Arkade transaction. */
export interface ArkadeBroadcaster {
    submitTx(
        signedArkTx: string,
        checkpointTxs: string[],
    ): Promise<{ arkTxid: string; finalArkTx: string; signedCheckpointTxs: string[] }>;
    finalizeTx(arkTxid: string, finalCheckpointTxs: string[]): Promise<void>;
}

interface IReadonlyWallet {
    getArkadeInfo(): Promise<ArkadeInfo>;          // this PR
    getArkadeReader(): Promise<ArkadeReader>;
}

interface IWallet extends IReadonlyWallet {
    getArkadeBroadcaster(): Promise<ArkadeBroadcaster>;
}
```

**Why this shape:**

- The readonly/full split falls out of the semantics rather than being a special case — reads on `IReadonlyWallet`, broadcast on `IWallet`. Same line `IReadonlyAssetManager` vs `assetManager` already draws, and the invariant behind making `ReadonlyWallet.arkProvider` protected in this PR.
- Normalized at the boundary, so the provider-boundary rule extends to plugins for free. Today `swap` can bypass it: `RestoreIndexer` (`restore.ts:47`) is a raw `Pick<RestIndexerProvider, "getVirtualTxs" | "getVtxos">`.
- It matches how the SDK already hands out collaborators — `getContractManager() → IContractManager`, RPC-proxied on the SW wallet at `serviceWorker/wallet.ts:1217` — and how both packages already narrow providers everywhere (`RestoreIndexer`, `RefundIndexer`, `ArkadeConnectOptions.indexer?: Pick<IndexerProvider, "getVtxos">`, `ArkadeServerProvider`).

**Cost per implementation:**

- `ReadonlyWallet` / `Wallet` / `ExpoWallet` — return an object closing over `this.indexerProvider` / `this.arkProvider`, routing reads through the existing `getAllNormalizedVtxos`.
- `ServiceWorkerReadonlyWallet` — an RPC proxy exactly like `getContractManager()`: **4 new message types, not 27.**

One thing to get right: the SW wallet already has a `GET_VTXOS` message meaning *the wallet's own VTXOs, from repositories*. The indexer's is an arbitrary-script network query. Name the new ones `INDEXER_GET_VTXOS` / `INDEXER_GET_VIRTUAL_TXS`, or someone will eventually wire the wrong one.

**Payoff:**

```ts
cancelOffer(wallet, offerHex, opts)
watchOfferSwaps({ wallet, ... })
```

No URL left anywhere in `@arkade-os/swap`, and service-worker traffic stays in the worker.

Worth its own issue rather than growing this PR. One thing to settle first: NArk is our reference implementation, so we should check whether it already names this seam and borrow its naming rather than invent one.

---

packages/ts-sdk/src/wallet/wallet.ts
```diff
@@ -853,6 +867,65 @@ export class ReadonlyWallet implements IReadonlyWallet {
...
+     *
+     * And the answer may be the boot snapshot rather than live: on a retryable
+     * failure this falls back to the cache, so a caller deriving a covenant
+     * from `signerPubkey` can bind one to a key the operator has since
+     * rotated away from, instead of failing closed.
+     *
+     * @returns The Arkade server's info
+     * @see ArkadeInfo
+     */
+    async getArkadeInfo(): Promise<ArkadeInfo> {
```
pietro909: **Reading info suppresses the signer-rotation trigger** (unfixed; f7c0221 documents it above)

`RestArkProvider.getInfo()` ends with `this._digest = info.digest` (`providers/ark.ts:545`) and never calls `emitServerInfoChanged` — the only call site for that is the `DIGEST_MISMATCH` retry path at `ark.ts:472`. So every call here silently advances the digest that rides `X-Digest` on all other requests. Once advanced, arkd stops answering `DIGEST_MISMATCH`, `onServerInfoChanged` never fires, and `handleServerInfoChanged` → `rotateServerSigner` never runs. The wallet keeps deriving its offchain address, boarding script and checkpoint script under the deprecated signer for the rest of the session.

The docblock added in f7c0221 states this and concludes: *"Do not call this on a hot path expecting rotation to still work behind it."* But `@arkade-os/swap` calls it from six entry points, all on the covenant-deriving path:

- `offer.ts:390` (`createOffer`), `offer.ts:500` (`cancelOffer`)
- `rfq.ts:869`, `rfq.ts:1262`, `rfq.ts:1685`, `rfq.ts:1894`

The caveat and the callers currently contradict each other. Either `getInfo()` compares against the previous digest and emits on change, or these six read through a path that feeds `handleServerInfoChanged`.

Worth separating from the epoch-skew concern withdrawn earlier in this PR: that rebuttal ("if `resolveArkInfo` returns epoch N+1, the operator answered N+1 and will co-sign for N+1") is correct and doesn't apply here. This is not about which epoch gets reported — it's about the trigger that would close the window being suppressed by the act of reading.

packages/ts-sdk/src/wallet/wallet.ts
```diff
@@ -853,6 +867,65 @@ export class ReadonlyWallet implements IReadonlyWallet {
...
+        const { info } = await resolveArkInfo(this.arkProvider, this.walletRepository);
```
pietro909: **`source` is discarded, so a caller cannot refuse a cached snapshot**

`resolveArkInfo` returns `{ info, source, lastOnlineAt }`; this destructure keeps only `info`. The docblock added in f7c0221 now warns that the answer "may be the boot snapshot rather than live" and that a caller "can bind [a covenant] to a key the operator has since rotated away from, instead of failing closed" — but gives the caller no way to act on it, because the fact that it *was* a fallback never leaves the method.

Concretely: wallet booted a week ago (snapshot persisted), operator has since rotated `signerPubkey` and is unreachable now. `createOffer` (`offer.ts:390`) previously threw, because a caller-held `new RestArkProvider(url).getInfo()` rejected outright. Now the failure classifies as retryable, the snapshot hydrates, and `createOffer` derives `offerVtxoScript` against the superseded key, registers it, and hands back an address to fund — the exact state `cancelOffer` later diagnoses as "the server signing key has likely rotated since funding". The four `request*` VHTLC flows have the same problem, plus a stale `unilateralExitDelay` mis-sizing the claim/refund timelock ladder.

Returning `source` — or adding a live-only variant for covenant derivation — makes fail-closed expressible.

On the earlier suggestion that this is "bounded by `deprecatedSigners`/`cutoffDate` handling elsewhere": that classification (`signerRotation.ts:88-133`) reads the signer set out of `info`, and on this path the *same* stale snapshot supplies it — so a superseded key classifies as `CURRENT`. It bounds detection once live info returns; it doesn't stop the covenant being derived and funded.

pietro909 (commented):
Independent review of the full PR diff (`feat/v0.5`...`f7c0221`), posted from a Claude Code session at the author's request.

**Scope notes**
- Reviewed the branch as a whole rather than commit-by-commit. That's where most of these come from — several concern the package's export surface and cross-file consistency, which no single commit's diff makes visible.
- `packages/boltz-swap` findings dropped: deprecated and slated for removal.
- The missing `ArkadeReader` / `ArkadeBroadcaster` / `NormalizedVtxoPage` exports came up in this pass too, and are already fixed by f7c0221 — not re-reported.

**11 findings inline.** The three I'd resolve before merge:

1. **`wallet.ts:903`** — reading info suppresses the signer-rotation trigger. f7c0221 documents this and warns not to call it on a hot path; `@arkade-os/swap` calls it from six entry points. The caveat and the callers contradict each other.
2. **`wallet.ts:904`** — `source` is discarded, so "fail closed rather than derive a covenant against a possibly-stale signer key" is not expressible by any caller, despite the docblock naming that exact risk.
3. **`serviceWorker/wallet.ts:1082`** — `cause.name` does **not** survive the worker boundary. The test certifying that it does uses a hand-written clone stub that copies `name`; the real structured-clone algorithm normalizes it to `"Error"`. This is why it was closed as fixed earlier in the PR.

The remainder are smaller: a preemptable offline fallback, a shared mutable payload across deduped callers, two doc claims that don't match the code, a runtime-vs-compile-time capability claim, plus one efficiency and one conventions note.

Findings 1 and 2 are already known and deliberately deferred per f7c0221's commit message — flagged here because the deferral leaves six live callers on the path the new docs warn against, which reads like it wants an issue rather than silence.

packages/ts-sdk/src/wallet/wallet.ts
```diff
@@ -853,6 +867,65 @@ export class ReadonlyWallet implements IReadonlyWallet {
...
+        return info;
```
pietro909, line 919: **`getProviderConnectionState()` is never updated by this call**

`resolveArkInfo` has just learned whether the operator is reachable, but `_serverInfoSource` / `_serverInfoLastOnlineAt` are assigned only at construction (`wallet.ts:1179` and the `ReadonlyWallet` setup path). So the reported state is pinned to the boot-time verdict for the life of the object:

- Wallet booted while the operator was down → `_serverInfoSource = "cache"`. The network returns, this call succeeds live, and `getProviderConnectionState()` still reports `{ mode: "degraded", source: "cache" }` until the wallet is re-created. A UI banner never clears.
- The inverse is worse: wallet booted live, this call silently fell back to the snapshot, and the state still reports `mode: "online"`.

Both values needed are already in hand on the line above and discarded.

packages/ts-sdk/src/wallet/wallet.ts
```diff
@@ -3019,11 +3093,35 @@ export class Wallet extends ReadonlyWallet implements IWallet, HDWalletCapable {
...
+     */
+    declare readonly arkProvider: ArkProvider;
+
+    /**
+     * Broadcast access bound to this wallet's server, so a plugin needs only
+     * the wallet.
+     *
+     * Defined here and not on {@link ReadonlyWallet} for the same reason
+     * `arkProvider` is protected there: a readonly wallet, and every
+     * `toReadonly()` view, must not be able to submit.
```
pietro909, line 3147: **"must not be able to submit" is a compile-time claim only**

`protected` is erased at runtime. `Wallet.toReadonly()` passes `this.arkProvider` — the full `RestArkProvider` — into `ReadonlyWallet`'s protected slot, and `ReadonlyWallet.create()` likewise stores `setup.arkProvider`, which is always a complete provider (`config.arkProvider || new RestArkProvider(...)`). So `(readonlyView as any).arkProvider.submitTx(psbt, cps)` broadcasts.

That's perfectly fine as a type-level narrowing. The issue is that CHANGELOG.md states it as a capability boundary — "a readonly wallet — and every `toReadonly()` view — cannot submit" — which isn't true of the object at runtime. Either soften the CHANGELOG to say the *typed surface* cannot submit, or have `ReadonlyWallet` hold a narrowed provider so the guarantee is real.

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
```diff
@@ -203,6 +212,11 @@ export const DEFAULT_MESSAGE_TIMEOUTS: Readonly<Record<RequestType, number>> = {
     ADVANCE_SIGNING_DESCRIPTOR_WATERMARK: 10_000,
 
     // Medium reads — may involve indexer queries
+    // GET_ARKADE_INFO is a live `/v1/info` fetch queued behind the same
+    // per-origin rate gate as the indexer, not a repository read.
+    GET_ARKADE_INFO: 20_000,
```
pietro909: **This 20 s deadline can preempt the worker's own offline fallback**

The comment is right that this is a live fetch behind the rate gate — which is exactly why the deadline can fire first. In the worker, `resolveArkInfo` → `RestArkProvider.getInfo` → `rateGate.runHttp(url, () => fetch(url))` with no `AbortSignal`. On a black-holed connection (captive portal, dropped Wi-Fi that doesn't RST) that fetch hangs until the OS connect timeout, typically 30–75 s. At 20 s `sendMessageDirect` rejects with `ServiceWorkerTimeoutError` and the caller sees `Failed to get arkade info: ServiceWorkerTimeoutError…` — while the worker is still seconds away from hydrating the persisted snapshot that would have answered.

The direct `Wallet` path has no such deadline, so the two implementations of the same `IReadonlyWallet` method diverge precisely in the offline case this seam exists for. Threading an `AbortSignal` into the worker-side fetch with a budget below this deadline would keep the fallback reachable.

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
```diff
@@ -1038,6 +1059,76 @@ export class ServiceWorkerReadonlyWallet implements IReadonlyWallet {
...
+        try {
+            const response = await this.sendMessage(message);
+            return (response as ResponseGetArkadeInfo).payload.info;
+        } catch (error) {
+            // Keep the original reachable as `cause`. Unlike its siblings, this
+            // call reaches `resolveArkInfo` in the worker, which distinguishes
+            // `ProviderUnavailableError` (offline — retry) from
+            // `MalformedArkInfoSnapshotError` (corrupt cache — re-onboard).
+            // structuredClone drops the prototype, so `instanceof` cannot
+            // survive the boundary either way, but `cause.name` does — and
```
pietro909: **`cause.name` does not survive the boundary — the test's clone is a stub, not the real algorithm**

> structuredClone drops the prototype, so `instanceof` cannot survive the boundary either way, but `cause.name` does

The second half is wrong. The structured-clone algorithm serializes an Error's `name` only from the whitelist Error / EvalError / RangeError / ReferenceError / SyntaxError / TypeError / URIError; anything else normalizes to `"Error"`:

```js
const e = new Error('x'); e.name = 'ProviderUnavailableError';
structuredClone(e).name   // => "Error"
```

So the worker rejects with `ProviderUnavailableError`, `wallet-message-handler.ts` puts it on `response.error`, `postMessage` clones it, and the page reads `cause.name === "Error"`. The ProviderUnavailable-vs-MalformedSnapshot branch this comment documents is unreachable in a browser.

`test/serviceWorker/wallet.test.ts:335` passes because the harness never runs the real algorithm: `structuredCloneError` at line 29 is a hand-written simulation whose header says "Simulate the structured clone algorithm that postMessage uses", and which does `cloned.name = error.name` — precisely the step the real one omits. Note 9ef74a1 fixed this same class of gap for the *success* response by wrapping it in a real `structuredClone(...)`; the error response still goes through the stub, which is why this survived being closed as fixed earlier in the PR.

This file already contains the working answer: `serializeAggregateError` sends a plain object, whose `name` is an ordinary property and survives.

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
```diff
@@ -1038,6 +1059,76 @@ export class ServiceWorkerReadonlyWallet implements IReadonlyWallet {
...
+            getVirtualTxs: async (txids, opts) => {
+                const message: RequestIndexerGetVirtualTxs = {
+                    id: getRandomId(),
+                    tag: this.messageTag,
+                    type: "INDEXER_GET_VIRTUAL_TXS",
+                    payload: { txids, opts },
+                };
+                try {
+                    const response = await this.sendMessage(message);
+                    return (response as ResponseIndexerGetVirtualTxs).payload;
```
pietro909: **Deduped callers share one mutable payload object**

`INDEXER_GET_VIRTUAL_TXS` is in `DEDUPABLE_REQUEST_TYPES`, and `sendMessage` resolves every concurrent caller with the same dedup key to the *same* settled `WalletUpdaterResponse`. Returning `.payload` by reference means two callers doing `reader.getVirtualTxs(sameTxids)` receive the identical `{ txs, page }` object — one calling `txs.sort()` or `txs.pop()` silently mutates the other's result.

`getVtxos` above escapes this only incidentally, because `vtxos.map(normalizeVtxo)` allocates a new array; its `page` object is still shared.

A shallow copy (`{ txs: [...txs], page }`) closes it.

packages/ts-sdk/src/wallet/index.ts
```diff
@@ -977,6 +978,51 @@ export interface IAssetManager extends IReadonlyAssetManager {
...
+ * express.
+ */
+export type ArkadeReader = Pick<IndexerProvider, "getVirtualTxs"> & {
+    /**
+     * Required `opts`, unlike the `IndexerProvider` method it mirrors: this
+     * seam reads for *named* foreign scripts (or outpoints), and a bare
+     * `getVtxos()` would be an unscoped server-wide query.
+     *
+     * A single request, so a wide `scripts` list can exceed the query-string
+     * limit and only page one comes back — reach for {@link getAllNormalizedVtxos},
```
pietro909: **"only page one comes back" cannot happen with the built-in provider**

`RestIndexerProvider.getVtxos` (`providers/indexer.ts:730`) explicitly detects `pageIndex === undefined && pageSize === undefined` and follows the server cursor to exhaustion, dropping `page`. A caller reading this will believe a single `reader.getVtxos({ scripts })` silently truncates, and may layer defensive re-paging on top of a provider that already paged.

The other half of the sentence is the real hazard — query-string length for a wide `scripts` list — and it alone justifies reaching for `getAllNormalizedVtxos` and its `SCRIPT_QUERY_CHUNK_SIZE` chunking. Suggest dropping the paging clause.

Same docstring, line 1005: required `opts` isn't what prevents "an unscoped server-wide query". The provider's own `"Either scripts or outpoints must be provided"` throw at `indexer.ts:698` is, and the message boundary (`wallet-message-handler.ts:1098`) adds no validation of its own. Worth naming the real defence, since a reader may otherwise assume the seam validates.

packages/ts-sdk/src/providers/ark.ts
```diff
@@ -168,7 +168,7 @@ export interface DeprecatedSigner {
 
 export type ServiceStatus = Record<string, string>;
 
-export interface ArkInfo {
+export interface ArkadeInfo {
```
pietro909, line 184: **Divergence from the NArk reference is unnoted** (minor — conventions)

AGENTS.md: *"check how `NArk` does it and align with it unless there's a TypeScript- or platform-specific reason to diverge (and note the reason when you do)."*

NArk names this type `ArkServerInfo` and the call `GetServerInfoAsync`, and hangs it off the *transport* — with a dedicated `CachingClientTransport` — rather than off the wallet (`NArk.Core/Wallet/SingleKeyAddressProvider.cs:57`, `NArk.Swaps/Boltz/BoltzSwapsService.cs:35`, `NArk.Tests/CachingClientTransportTests.cs`). This PR picks `ArkadeInfo` / `getArkadeInfo()` on `IReadonlyWallet`, and the CHANGELOG entry ("The type name now matches the product and the accessor that returns it") gives the rationale for the *name* but not for diverging from the reference's naming or its placement.

Both choices may well be right for this platform — the ask is just the note AGENTS.md asks for, so the next person doesn't re-litigate it.

packages/swap/src/offer.ts
```diff
@@ -387,12 +387,12 @@ export async function createOffer(
         throw new Error("set exactly one of wantAsset (BTC->asset) or offerAsset (asset->BTC)");
     }
     const [info, makerAddress, makerPublicKey] = await Promise.all([
-        new RestArkProvider(arkServerUrl).getInfo(),
+        wallet.getArkadeInfo(),
```
pietro909: **Uncached `/v1/info` round-trip on every entry-point call**

`RestArkProvider.getInfo` always fetches; nothing memoizes. With `getArkadeInfo()` now called from six entry points (`offer.ts:390`, `offer.ts:500`, `rfq.ts:869/1262/1685/1894`), a maker creating N offers pays N `/v1/info` requests against the same per-origin rate gate — where the previous shape at least let a caller hold one provider across calls.

The swap README (line 20) says this design means "no second `/v1/info` round-trip", which holds per call site but not across a session.

NArk solves this with `CachingClientTransport`. A short-TTL memo would be safe here for the usual reason — the provider's `X-Digest` negotiation is what reports the cached info stale — but note it interacts with the digest issue on `wallet.ts:903`: fewer reads also means fewer opportunities to suppress the rotation trigger. Worth deciding both together.

packages/swap/README.md
```diff
@@ -12,6 +12,23 @@ React Native does not, so install `react-native-get-random-values` (or `expo-cry
...
+
+No offer entrypoint here takes a server URL. `cancelOffer` and `watchOfferSwaps` need more than
+server info — cancel broadcasts the refund and falls back to the indexer for a deposit made
+before contract registration existed, and the watcher reads spending transactions — so they
+ask the wallet for those too: `wallet.getArkadeReader()` for chain reads and
+`wallet.getArkadeBroadcaster()` for `submitTx`/`finalizeTx`. On a service-worker wallet both
+are proxied to the worker, so these reads stay on the wallet's own connection.
+The RFQ/refund/restore helpers (`restoreAssetSwaps`, `arkadeRefunder`, `claim`/`refund`)
+still take provider instances; an `ArkadeReader` satisfies those shapes structurally, so
+they can be fed `await wallet.getArkadeReader()` today.
```
pietro909: **`arkadeRefunder` and `claim`/`refund` cannot be fed an `ArkadeReader`**

`arkadeRefunder({ ark, indexer, wallet, repository })` takes `ark: RefundArkProvider = Pick<RestArkProvider, "getInfo" | "submitTx" | "finalizeTx">` (`refund.ts:126`). An `ArkadeReader` is `getVirtualTxs` + `getVtxos` — it satisfies the `indexer` field and nothing else, so it does not "satisfy those shapes structurally". No wallet seam returns the `getInfo`+`submitTx`+`finalizeTx` shape, so a caller has to hand-compose it:

```ts
{ getInfo: () => wallet.getArkadeInfo(), ...(await wallet.getArkadeBroadcaster()) }
```

A reader following this paragraph gets a compile error and no guidance. `restoreAssetSwaps` is fine (indexer-shaped); it's the other two that don't hold. The same claim is repeated in `packages/swap/MIGRATION.md:95` and `CHANGELOG.md:27`.

Either narrow the sentence to `restoreAssetSwaps`, or add the composition snippet.

packages/ts-sdk/src/wallet/wallet.ts
tiero: Fixed in 0f61472, with the mechanism you named: `getInfo()` now compares against the previous digest and emits `onServerInfoChanged` when it moved, so the six swap call sites *drive* rotation instead of suppressing it. Guarded on the previous digest being non-empty — the boot read stays silent, and the `DIGEST_MISMATCH` path (which clears the digest before its own refresh and emits explicitly) doesn't double-fire; the existing mismatch test pins the emit count at 1, and a new test covers boot-silent / unchanged-silent / moved-emits.

---

packages/ts-sdk/src/wallet/wallet.ts
tiero: Fixed in 0f61472 via your second option: `getArkadeInfo({ requireLive: true })` skips the fallback entirely and throws whatever the provider throws — the pre-#734 shape. All six covenant-deriving call sites (`createOffer`, `cancelOffer`, the four `request*`) now pass it, so the scenario you describe (week-old snapshot, rotated key, unreachable operator) fails the call instead of deriving a fundable address against the superseded key. The ordinary read keeps its fallback for non-covenant callers (fee display, status). Threaded through the SW wallet as an optional `GET_ARKADE_INFO` payload (an older worker ignores it and serves the fallback read — rolling-upgrade window only, noted on the wire type). Test: `getArkadeInfo({ requireLive: true }) fails closed instead of serving the cache`.

Your point about `deprecatedSigners`/`cutoffDate` reading from the same stale snapshot was the convincing detail — agreed that it bounds detection, not derivation.

---

packages/ts-sdk/src/wallet/wallet.ts
tiero: Fixed in 0f61472: every resolution now writes `_serverInfoSource`, and `_serverInfoLastOnlineAt` moves to `Date.now()` on a live answer (or the snapshot's `savedAt` on fallback). Both of your cases are covered by the new test — a wallet that booted live and falls back stops claiming `online`, and after a `requireLive` success the state leaves `degraded`.

---

packages/ts-sdk/src/wallet/wallet.ts
tiero: Softened in 0f61472 — the CHANGELOG now says the *typed surface* exposes no way to submit, states plainly that `protected` is erased at runtime (compile-time boundary, not a sandbox), and notes the readonly service-worker as the one place the line is also enforced at runtime (`SUBMIT_TX`/`FINALIZE_TX` refused with `ReadonlyWalletError`). Kept the type-level narrowing rather than a runtime-narrowed provider: `Wallet extends ReadonlyWallet` means the same object must carry the full provider for the subclass, and `toReadonly()` deliberately shares the instance.

---

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
tiero: Fixed in 0f61472: the live `/v1/info` fetch now carries `AbortSignal.timeout(12_000)` where the runtime supports it (feature-detected; older runtimes stay unbudgeted), leaving ~8s of the 20s page deadline for the rate-gate queue and the snapshot read. The abort rejects as a `TimeoutError`-named DOMException, which `isFetchTimeoutError` now recognizes, so it classifies retryable and `resolveArkInfo` reaches the fallback — the direct-`Wallet` and worker paths converge on the same behaviour instead of diverging exactly offline. Tested both halves: the fetch receives a signal, and the TimeoutError shape classifies retryable.

---

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
tiero: You're right, and the earlier fix was wrong — verified in Node: `structuredClone` of an Error named `ProviderUnavailableError` comes back named `"Error"`. The whitelist claim holds; arkana's spec rebuttal doesn't.

Fixed in 0f61472 with the mechanism this file already proved out for `AggregateError`: the worker stamps `errorName` beside the error as a plain string — once, in `tagged()`, the choke point every response exits through — and the page restores it at all three reject sites before the caller sees it. So `cause.name` now genuinely survives, by transport rather than by luck.

The test gap you identified is closed at the harness level too: `structuredCloneError` now normalizes names exactly like the real algorithm (whitelist or `"Error"`), so a stub can't certify this class of behaviour again; the cause test clones its envelope with the *real* `structuredClone`, and a companion test pins the rolling-upgrade window — an old worker sending no `errorName` yields the clone-normalized `"Error"` without crashing the page.

---

packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
tiero: Fixed in 0f61472: `getVirtualTxs` returns `{ txs: [...txs], page: page && { ...page } }`, and `getVtxos` — whose `map` already gave each caller its own array incidentally, as you noted — now copies `page` too. Both comments state the reason (dedup settles every concurrent caller with one response).

---

packages/ts-sdk/src/wallet/index.ts
tiero: Both halves addressed. The paging clause was already rewritten in an earlier pass on the same evidence (RestIndexerProvider pages `getVtxos` to exhaustion; `getVirtualTxs` is the call that actually truncates, which now carries its own warning naming the 50-txid chunking in `swap`'s restore scan). 0f61472 fixes the second half: the doc now says the required `opts` stops a bare `getVtxos()` at *compile time* and names the provider's own "Either scripts or outpoints must be provided" throw as the runtime defence — the message boundary validates nothing itself.

---

packages/ts-sdk/src/providers/ark.ts
tiero: Added in 0f61472 — `ArkadeInfo`'s docstring now carries the note: NArk's `ArkServerInfo`/`GetServerInfoAsync` lives on the transport with a `CachingClientTransport`; this SDK names for the product (per #734's ask) and hangs the read off the wallet because the wallet is the one object every consumer already holds, and in the service-worker model the page has no transport to hang it on — the provider lives worker-side. Names `CachingClientTransport` as the shape of the caching follow-up so the next person inherits the decision rather than re-litigating it.

---

packages/swap/src/offer.ts
tiero: Decided together with the digest issue, as you suggested — the resolution went the other way from a memo. With rotation fixed (an info read now *emits* on a moved digest) and `requireLive` on the six covenant sites, the per-entry-point read is deliberate: covenant derivation wants live info and fails closed offline, so a TTL memo on that path would reintroduce a window. The README claim is scoped to *per call* in 0f61472 and states the N-offers-N-reads cost plainly, naming `CachingClientTransport` (also noted on `ArkadeInfo` per the AGENTS.md divergence note) as the follow-up shape for the paths where caching is safe.

---

packages/swap/README.md
tiero: Fixed one commit before your review landed (b6a51e1, a `/simplify` pass caught the same over-claim): all three docs now scope the structural claim to the *indexer* parameter — `restoreAssetSwaps` takes a reader directly — and carry the composition snippet for the ark side: `{ getInfo: () => wallet.getArkadeInfo(), ...(await wallet.getArkadeBroadcaster()) }`. The activity resolver, which takes no provider at all, is out of the list.

---

pietro909 (approved):
All 11 review comments verified addressed against 313d4458; threads resolved. Unit suite and all integration jobs green.

Two non-blocking follow-ups: the stale `requestLightningReceive` sentence in MIGRATION.md:83-84 (contradicts :173-175), and the duplicate digest-emit implementation that will conflict with #676 when master flows into feat/v0.5 — master's `_hasServerInfo` version is the one to keep.

pietro909 (approved):

tiero:
merge? @pietro909

PR #811, opened 2026-08-28
Remove boltz-swap

pietro909:
## NOTE: CI is broken atm

Removes the deprecated `@arkade-os/boltz-swap` package and everything that pointed at it — workspace scripts, the release orchestrator, CI, docs, the lockfile, and the Boltz regtest tier no suite here uses. Two TODOs in `swapManager.ts` mark the one docblock whose rationale is framed as a diff against the removed package and needs restating on its own terms.


tiero:
resolve conflicts @pietro909 packages/swap/MIGRATION.md

PR #812, opened 2026-08-28
feat(vhtlc): add the non-interactive refund-without-receiver leaf

Kukks:
Adds a ninth, opt-in tapleaf to `VHTLC.ScriptV2`:
`nonInteractiveRefundWithoutReceiver` — the Arkade Service key plus the covenant-tweaked
emulator co-signer, spendable after `refundLocktime`.

## Why

Every existing route to the "refund the sender" tier needs a specific party online:

| leaf | needs | gate |
| --- | --- | --- |
| `refund` | sender + receiver + server | — |
| `refundWithoutReceiver` | sender + server | `refundLocktime` |
| `nonInteractiveRefund` | server + receiver + emulator | — |

A sender who funds a lockup and vanishes — or loses their key — therefore depends on the
**receiver's** cooperation. If the receiver is gone or uncooperative, the only path left is
`unilateralRefundWithoutReceiver`, a CSV leaf costing a real on-chain unilateral exit.

The script already implies the fix. Lining each interactive leaf up against its covenant twin
shows the rule the ladder follows, and the one cell it leaves empty:

| interactive leaf | beneficiary | covenant twin | present |
| --- | --- | --- | --- |
| `claim` | receiver | `nonInteractiveClaim` | yes |
| `refund` | sender | `nonInteractiveRefund` | yes |
| `refundWithoutReceiver` | sender | — | **no** |

The rule: *`nonInteractive X` is `X` with the beneficiary's signature replaced by an emulator key
tweaked by `enforcePayTo(their pkScript)`.* This fills the empty cell. The name is mechanical,
not chosen.

## What it is

```
CLTVMultisigTapscript.encode({
    absoluteTimelock: refundLocktime,
    pubkeys: [server, <the SAME cosigner nonInteractiveRefund already derives>],
})
```

Opt in with `nonInteractiveRefund.withoutReceiver: true`. It reuses the existing
`senderPkScript`/`emulatorPubkey` pair rather than taking its own, so the two leaves cannot be
given divergent destinations.

## Two properties worth checking in review

**The co-signer is shared structurally, not coincidentally.** Both refund covenant leaves pin the
same destination, so they must commit to the same key. There is exactly one call site to
`computeArkadeScriptPublicKey` on the refund path (`vhtlc.ts`, `nirCosigner`), consumed by both
`MultisigTapscript.encode` and `CLTVMultisigTapscript.encode`. Computing it twice would make
byte-identity a coincidence rather than a guarantee.

Consequently the emulator needs **no new key and no new evaluation path**. The two leaves cannot
cross-replay a signature despite sharing a key: BIP-341 tapscript sighashes commit to the tapleaf
hash.

**It is genuinely opt-in.** Leaf order fixes the merkle root, so the new leaf is pushed last and
every earlier leaf keeps its index. Without the flag, the derived pkScript is byte-identical to
before — asserted directly, comparing all eight prior leaves byte-for-byte rather than by count.

## Nothing else needs to change

- **arkd** — `TapscriptsVtxoScript.Decode` (`pkg/ark-lib/script/vtxo_script.go`) appends each
  decoded closure to a slice with no uniqueness constraint, so a tree carrying two
  `CLTVMultisigClosure` leaves parses. `ForfeitClosures()` already includes that type, and the new
  leaf carries `server`, satisfying the "forfeit closure must contain the signer pubkey" rule.
- **covclaimd** — decodes taptrees off the wire rather than deriving them.
- **the emulator** — same tweaked key, same covenant bytes.

## Cross-SDK conformance vectors

`test/fixtures/vhtlc-v2-nine-leaf.json` pins the pkScript, all nine leaf scripts, the serialized
taptree, and the covenant ArkadeScripts, from fixed non-random keys. It is intended to be consumed
verbatim by the C#, Go and Rust implementations, which must produce identical bytes or derive
addresses nobody can spend.

This is the failure it exists to prevent: when the SDK moved its preimage condition from
`HASH160 <h> EQUAL` to `SIZE 32 EQUALVERIFY HASH160 <h> EQUAL`, a consumer kept building the old
form, its endpoint answered **200**, and no claim was ever pushed — a silent no-op with nothing in
the logs.

The fixture carries both `leaves` (per-leaf, for byte comparison) and `tapTree` (the serialized
`encode()` blob), because the Go consumer's production path is `DecodeTapTree` over exactly those
bytes.

## Testing

- `test/vhtlc.test.ts` — 37 passed (34 pre-existing + 3 new)
- `test/contracts/vhtlcV2-handler.test.ts` — 65 passed (61 pre-existing + 4 new)
- `test/script/vhtlc-vectors.test.ts` — 3 new
- Full run: **2758 passed** vs **2746** at branch point. The 107 failures in 24 `test/e2e/*` files
  are pre-existing and require docker + regtest; the failing file list is byte-for-byte identical
  before and after.
- `tsc --noEmit` clean.

**Negative control on the vectors.** Swapping the last two leaf pushes made all three fixture
assertions fail — pkScript, leaf order, and `tapTree`. Restoring the order made them pass, with an
empty `git diff` confirming byte-identical restoration. The tests fail for the reason they claim.

## Two things reviewers should not take on trust

**No regtest spendability test.** Nothing here exercises the new leaf against arkd or the emulator
end-to-end. Byte-identity, leaf order, opt-in behaviour and key-sharing are proven; "a wallet can
actually spend this leaf" is not. The arkd analysis above says it should work, but that is
reasoning, not a spend.

**A green pre-commit hook is not evidence in this repo.** Its affected-package `lint` and
`test:unit` steps print `No projects matched the filters` and exit green **without running**. This
reproduces against unrelated diffs, so it is pre-existing and not introduced here — but it means
the test numbers above come from `npx vitest run` and `npx tsc --noEmit` invoked directly, not from
the hook.

## Summary by CodeRabbit

* **New Features**
  * Added an optional non-interactive refund path that does not require a receiver signature.
  * Added support for configuring and accessing the new refund script and covenant data.
  * Added VHTLC V2 support for the new nine-leaf contract structure.

* **Bug Fixes**
  * Added validation for invalid or incomplete refund configuration flags.

* **Tests**
  * Added coverage for serialization, script generation, leaf ordering, cross-SDK compatibility, and end-to-end refund spending.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,7 @@ jobs:
               test/e2e/exitDataCapture.test.ts
               test/e2e/vhtlc.test.ts
               test/e2e/non-interactive-htlc.test.ts
+              test/e2e/non-interactive-refund-without-receiver.test.ts
               test/e2e/indexer.test.ts
               test/e2e/electrum.test.ts
               test/e2e/onchain.test.ts
--- a/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
+++ b/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
@@ -131,21 +131,21 @@ function decodeCovenantLeaf<K extends string>(
  * the same ScriptV2 the two uses would fight over one row.
  *
  * **Which leaves this offers, and why the set is smaller than the ladder.**
- * ScriptV2 has eight leaves; a wallet holding ONE of the two participant keys
+ * ScriptV2 has nine leaves; a wallet holding ONE of the two participant keys
  * can satisfy four of them, and offering a leaf whose signature the caller
  * cannot produce is worse than offering fewer — it turns a refusal into a
  * transaction that gets built and then rejected.
  *
- * | leaf                              | needs                          | offered |
- * |-----------------------------------|--------------------------------|---------|
- * | `claim`                           | receiver + server, preimage    | yes, to the receiver |
- * | `refund`                          | sender + receiver + server     | no — needs the counterparty live |
- * | `refundWithoutReceiver`           | sender + server, CLTV          | yes, to the sender |
- * | `unilateralClaim`                 | receiver, CSV                  | yes, to the receiver |
- * | `unilateralRefund`                | sender + receiver, CSV         | no — needs the counterparty live |
- * | `unilateralRefundWithoutReceiver` | sender, CSV                    | yes, to the sender |
- * | `nonInteractiveClaim`             | server + emulator              | no — the wallet holds neither key |
- * | `nonInteractiveRefund`            | server + receiver + emulator   | no — the wallet holds neither key |
+ * | leaf                                  | needs                          | offered |
+ * |---------------------------------------|--------------------------------|---------|
+ * | `claim`                               | receiver + server, preimage    | yes, to the receiver |
+ * | `refund`                              | sender + receiver + server     | no — needs the counterparty live |
+ * | `refundWithoutReceiver`               | sender + server, CLTV          | yes, to the sender |
+ * | `unilateralClaim`                     | receiver, CSV                  | yes, to the receiver |
+ * | `unilateralRefund`                    | sender + receiver, CSV         | no — needs the counterparty live |
+ * | `unilateralRefundWithoutReceiver`     | sender, CSV                    | yes, to the sender |
+ * | `nonInteractiveClaim`                 | server + emulator              | no — the wallet holds neither key |
+ * | `nonInteractiveRefund`                | server + receiver + emulator   | no — the wallet holds neither key |
  * | `nonInteractiveRefundWithoutReceiver` | server + emulator, CLTV        | no — the wallet holds neither key |
  *
  * The two omitted interactive leaves are the same two the `vhtlc` handler
--- a/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
+++ b/packages/ts-sdk/src/contracts/handlers/vhtlcV2.ts
@@ -82,6 +82,8 @@ export interface VHTLCV2ContractParams {
     nonInteractiveRefund?: {
         senderPkScript: Uint8Array;
         emulatorPubkey: Uint8Array;
+        /** @see VHTLC.Options.nonInteractiveRefund.withoutReceiver */
+        withoutReceiver?: boolean;
     };
 }
 
@@ -144,6 +146,7 @@ function decodeCovenantLeaf<K extends string>(
  * | `unilateralRefundWithoutReceiver` | sender, CSV                    | yes, to the sender |
  * | `nonInteractiveClaim`             | server + emulator              | no — the wallet holds neither key |
  * | `nonInteractiveRefund`            | server + receiver + emulator   | no — the wallet holds neither key |
+ * | `nonInteractiveRefundWithoutReceiver` | server + emulator, CLTV        | no — the wallet holds neither key |
  *
  * The two omitted interactive leaves are the same two the `vhtlc` handler
  * omits, for the same reason: `refund` and `unilateralRefund` both need the
@@ -212,6 +215,13 @@ export const VHTLCV2ContractHandler: ContractHandler<VHTLCV2ContractParams, VHTL
                 nonInteractiveRefundEmulatorPubkey: hex.encode(
                     params.nonInteractiveRefund.emulatorPubkey,
                 ),
+                // Spread rather than a "false"/"undefined" string: a dropped
+                // flag re-derives the EIGHT-leaf script, a different pkScript,
+                // and registration dies at `upsertContractRow` with an opaque
+                // `Script mismatch`. Same silent-drop class as the asset keys.
+                ...(params.nonInteractiveRefund.withoutReceiver && {
+                    nonInteractiveRefundWithoutReceiver: "1",
+                }),
             }),
             // The asset must round-trip, and its absence here used to be silent
             // in the worst way. `ContractManager` re-derives a contract from
@@ -258,6 +268,24 @@ export const VHTLCV2ContractHandler: ContractHandler<VHTLCV2ContractParams, VHTL
                     "would re-derive the default claim covenant, which is weaker than the row asked for",
             );
         }
+        // Same silent-drop shape as the two checks above, one flag further in:
+        // a row naming this flag without the leaf it extends, or read as "not
+        // set", would re-derive a script missing this leaf.
+        if (params.nonInteractiveRefundWithoutReceiver !== undefined && !refund) {
+            throw new Error(
+                "nonInteractiveRefundWithoutReceiver without the nonInteractiveRefund keys it " +
+                    "extends: reading this as 'not set' would re-derive a script without the leaf",
+            );
+        }
+        if (
+            params.nonInteractiveRefundWithoutReceiver !== undefined &&
+            params.nonInteractiveRefundWithoutReceiver !== "1"
+        ) {
+            throw new Error(
+                `nonInteractiveRefundWithoutReceiver must be "1" when present, got ` +
+                    JSON.stringify(params.nonInteractiveRefundWithoutReceiver),
+            );
+        }
         const asset =
             params.assetTxid !== undefined && params.assetGroupIndex !== undefined
                 ? {
@@ -295,6 +323,9 @@ export const VHTLCV2ContractHandler: ContractHandler<VHTLCV2ContractParams, VHTL
                 nonInteractiveRefund: {
                     senderPkScript: refund.destination,
                     emulatorPubkey: refund.emulatorPubkey,
+                    ...(params.nonInteractiveRefundWithoutReceiver === "1" && {
+                        withoutReceiver: true,
+                    }),
                 },
             }),
         };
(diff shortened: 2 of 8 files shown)


Kukks:
Note on `Integration (ts-sdk / exit-providers-rotation)`: this looks like regtest infrastructure flake rather than a real failure, and the evidence is that **two runs of identical code failed different tests.**

- attempt 1 — `test/e2e/unilateralExit.test.ts > exits a vhtlc as receiver with the preimage (condition witness)`, a 300s timeout
- attempt 2 — `test/e2e/unilateralExitDeepChain.test.ts > prepare() emits a deep offchain send chain in physical-input topological order`

Both runs: 1 failed / 9 passed files, and both carry the same infrastructure symptoms underneath — `server not ready`, `connection refused` on `172.18.0.19:6060`, `failed to get wallet ready update stream`, `ark service not ready: wallet is locked or syncing`. A deterministic regression would fail the same test each time.

The other seven checks pass, including `Lint, Test & Build`, both swap integration suites, `ark-core`, `arkade-assets` and `settlement-delegation`.

Worth stating plainly rather than leaning on that alone: the failing tests are VHTLC tests, which is adjacent to this change. What argues against a real cause is that the new leaf is opt-in and nothing in the existing suite sets the flag, so every pre-existing script derives byte-identically — asserted directly, comparing all eight prior leaves byte-for-byte rather than by count. The unit-level golden-vector tests that do exercise the refactored shared-cosigner path all pass.

Re-running once more; happy to dig further if it fails a third time on yet another test.

Kukks:
Thanks — the emulator-spend gap is the right thing to lead with, and it is the same one the PR description flags as the claim not to take on trust. I am not going to argue around it: I am investigating whether an e2e spend test is feasible in the current harness, following whatever already exercises `nonInteractiveClaim` end-to-end. If it turns out infeasible I will say exactly what is missing rather than substitute a weaker test that looks like one.

Two corrections on cross-repo impact, where the review is working from stale clones:

**dotnet-sdk has a corresponding PR** — arkade-os/dotnet-sdk#193, open, targeting `feat/arkade-script` (deliberately, so the ninth leaf lands before #81 merges rather than becoming a second breaking change to a shipped `HTLCv2` type). It consumes `vhtlc-v2-nine-leaf.json` byte-identically — same SHA256 — and matched the fixture on the first run with zero hex mismatches. That is the first real evidence the fixture mechanism works rather than merely being asserted.

**go-sdk and rust-sdk are deliberately out of scope**, and it is worth being precise about why, because "consume the fixture" understates the work:

- go-sdk is at **seven** leaves — `vhtlc/opts.go` has `NonInteractiveClaim` but no `nonInteractiveRefund` at all (its own comment calls the claim leaf "a 7th covenant leaf").
- rust-sdk is at **six** — `ark-core/src/vhtlc.rs` has neither covenant leaf.

Both were already behind before this PR; it widens the gap by one. Reaching parity is a larger piece of work with its own review surface, and neither is on the critical path, because covclaimd decodes taptrees off the wire rather than deriving them. Each needs its own PR rather than being bolted onto this one.

On the **v1 inheritance** point — confirmed, and there is a detail that changes the fix. `Script` (v1) does inherit the leaf via `BaseScript`, but the v1 contract handler (`contracts/handlers/vhtlc.ts`) has no covenant support *at all*: no `nonInteractiveClaim`, no `nonInteractiveRefund`, no fields for either. So v1 could already build covenant leaves its own handler cannot round-trip. This PR widens a pre-existing gap rather than opening one. Adding only `withoutReceiver` to the v1 handler would leave the other two broken and imply a support level that does not exist, so I am documenting the limitation on the v1 class instead. Happy to be argued out of that if you would rather see v1 gain full covenant round-tripping — but that feels like its own PR.

On the merge gate: the flag is off by default and nothing in this repo sets it. The only consumer that does is arkade-os/intent-solver#216, which is a **draft** and blocked on this publishing — so the gate you are asking for exists de facto today. I would still rather close it with the e2e test than rely on that.

Kukks:
The blocking finding is closed. Pushed `c24cdaa4`, `2bd0283a`, `71984c86`, `da38643c`.

## The e2e spend test

`test/e2e/non-interactive-refund-without-receiver.test.ts` — six stages against a live regtest + emulator stack, run repeatedly, green:

1. fund an immature lockup
2. immature + correct destination → **rejected** (isolates arkd's CLTV gate)
3. advance the chain's median-time-past to maturity
4. mature + misdirected output → **rejected** (isolates the destination check)
5. mature + shortchanged value → **rejected** (isolates the conservation check)
6. mature + correct → **accepted**; `waitForVtxo` then independently queries the indexer and asserts the VTXO landed at `senderPkScript` with the expected txid

Stage 6 is the one that answers the finding directly: the emulator receives the signing request for `nirCosigner`, evaluates the covenant, returns a signature, and the spend confirms. Not "no exception was thrown."

## The three rejections are provably distinct — but not from the client

Worth reporting precisely, because it is also a finding. All three rejection stages surface a **byte-identical** client error:

```
Failed to submit tx to emulator: {"code":13, "message":"internal error", "details":[]}
```

The emulator's own logs for the same window show three genuinely different causes:

| stage | server-side cause |
| --- | --- |
| immature | `FORFEIT_CLOSURE_LOCKED (11): ... (blocktime)` |
| misdirected | `OP_EQUALVERIFY failed vin=0` |
| shortchanged | `false stack entry at end of script execution vin=0` — the trailing bare `GREATERTHANOREQUAL`, no `VERIFY` |

So each stage really does exercise a different covenant condition; the wire format flattens them. That sequence has been observed identically across multiple runs, and it is recorded in the test file's own comments rather than only in a transcript, so it survives.

The tests assert a deliberately identical pattern on all three, documented as intentional so nobody later tightens it into a distinction the client cannot actually make. **That the client cannot tell these three apart is arguably worth fixing on the emulator side**, independently of this PR.

## A correction to something in my earlier reply

I previously described the seconds-typed locktime as a workaround for arkd "refusing block-typed CLTV leaves". That was wrong, and I am glad to correct it before it propagated.

`blockTypeAllowed` is a parameter, threaded from `settings.AllowCSVBlockType()`:

```go
// pkg/ark-lib/script/vtxo_script.go:107
if !blockTypeAllowed && !c.Locktime.IsSeconds() { ... }

// internal/core/domain/settings.go:456
func (s Settings) AllowCSVBlockType() bool {
    return s.VtxoTreeExpiry.Type == arklib.LocktimeTypeBlock
}
```

One flag governs both the CLTV forfeit check and the CSV exit check, and it is true only when the operator's own `VtxoTreeExpiry` is block-typed. So this is operator configuration, not an arkd rule. The regtest operator here is seconds-typed, which is why block-typed leaves are refused *in this environment*.

The practical consequence, which is smaller than I first claimed: `vhtlc.test.ts`'s block-height fixtures are incompatible with a seconds-typed operator config — 3 of its 5 tests fail that way on unmodified master. Not a general fixture bug, and not a regression.

## The two documentation asks

- The emulator spend path is now documented on the `nonInteractiveRefundWithoutReceiver()` accessor, grounded in `providers/emulator.ts` and `arkade/contract.ts` rather than invented.
- The v1 note is on `VHTLC.Script`. One detail worth flagging: the v1 handler serialises **none** of the covenant options, not just this one — `VHTLCContractParams` carries no covenant fields at all. So v1 could already build covenant leaves it cannot register. Documented as the pre-existing limitation it is, rather than adding only `withoutReceiver` and implying a support level v1 does not have.

One process note: I understand a follow-up review from this bot lands as `COMMENTED`, which will not clear the existing `CHANGES_REQUESTED`. If the above satisfies the finding, that review will need dismissing by a maintainer.

Kukks:
@coderabbitai review

Kukks:
Thanks for re-reviewing — glad the e2e test lands the blocker.

One correction on **F3**, which repeats a point from the earlier pass: the fixture is no longer consumed by this repo only, and two of the three SDKs named do have corresponding PRs.

- **dotnet-sdk#193** — open. Commits `NArk.Tests/Arkade/vhtlc-v2-nine-leaf.json` byte-identical to the copy here (same SHA256, `2d3a2aab…f8dfb3`) and matched every leaf on the first run with zero hex mismatches. It targets `feat/arkade-script` rather than master so the ninth leaf lands before #81 merges, avoiding a second breaking change to a shipped `HTLCv2` contract type.
- **covclaimd#9** — open, all checks green. Commits the same fixture at `pkg/refund/testdata/`, also byte-identical, and decodes its serialized `tapTree` through the same `DecodeTapTree` + `TapscriptsVtxoScript.Decode` pair `pkg/preimage/claimer.go`'s `matchOutput` uses in production — rather than reassembling a tree from the leaf list, which would test our own reassembly instead of the SDK's bytes.

I re-verify that all three copies hash-match after every change; they still do.

**go-sdk and rust-sdk genuinely have no PRs, and that is deliberate.** It is worth being precise about why, because "consume the fixture" understates the work:

- go-sdk builds **seven** leaves — `vhtlc/opts.go` has `NonInteractiveClaim` but no `nonInteractiveRefund` at all (its own comment calls the claim leaf "a 7th covenant leaf").
- rust-sdk builds **six** — `ark-core/src/vhtlc.rs` has neither covenant leaf.

Both were already behind before this PR; it widens the gap by one. Reaching parity is a larger piece of work with its own review surface, and neither is on the critical path, because covclaimd decodes taptrees off the wire rather than deriving them. Each wants its own PR rather than being bolted onto this one.

Also noting for whoever merges: this bot's follow-up reviews land as `COMMENTED`, so the earlier `CHANGES_REQUESTED` is still the PR's `reviewDecision` even though the blocker it named is now closed. It will need dismissing rather than clearing itself.

pietro909 (approved):

PR #813, opened 2026-08-28
fix(sdk): push service-worker events instead of holding them for the tick

pietro909:
## Reason

On consumer's side, when using the Service Worker the events have to wait the ticker before they are notified. This explains the perception of slow swaps in receive.

## Description

Every event the worker pushes to the page — `CONTRACT_EVENT`, `VTXO_UPDATE`, `UTXO_UPDATE` and the settle/renew/recover/migrate progress events — was queued and released only when the tick timer fired, adding up to `tickIntervalMs` of latency (10s default, 5s in arkade) to a push the worker already had.

**Bus.** A single FIFO drain loop behind one `matchAll`, exposed to handlers as an optional third `start()` argument. The channel is generation-scoped, so one retained across a re-init or `stop()` goes inert on its own rather than relying on the implementor. `runTick` posts through the same queue, so ordering is preserved across both paths.

**Wallet handler.** Three generation counters — one per subscription field, one for requests — replace `scheduleForNextTick`. They cannot be merged: the incoming-funds re-subscribe would otherwise leave the live contract-event emitter permanently stale, and folding requests in would truncate an in-flight settle. They also guard the repository writes and the subscription-field assignments a stale initializer would otherwise perform.

`tick()` and `onNextTick` stay as the fallback when no channel is supplied.

No consumer change required — `tickIntervalMs` simply stops governing event latency.

## Summary by CodeRabbit

* **New Features**
  * Wallet and worker events can now be delivered immediately through a push channel, without waiting for the next update cycle.
  * Broadcasts are delivered to connected clients in order, with continued delivery after temporary failures.
  * Wallet restoration now preserves aggregate error details across worker communication.

* **Bug Fixes**
  * Prevented stale callbacks and events from being delivered after stopping or reinitializing the wallet.
  * Improved lifecycle handling for queued updates and contract events.

* **Documentation**
  * Updated worker messaging documentation to explain immediate broadcasts and scheduled responses.

### DIFF
--- a/packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts
+++ b/packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts
@@ -924,7 +924,9 @@ export class WalletMessageHandler
     }
 
     // lifecycle methods
-    async start(...params: Parameters<MessageHandler["start"]>): Promise<void> {
+    async start(
+        ...params: Parameters<MessageHandler<WalletUpdaterRequest, WalletUpdaterResponse>["start"]>
+    ): Promise<void> {
         const [services, repositories, channel] = params;
         this.readonlyWallet = services.readonlyWallet;
         this.wallet = services.wallet;
--- a/packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts
+++ b/packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts
@@ -874,6 +874,18 @@ export type WalletUpdaterResponse = ResponseEnvelope &
         | ResponseRestoreWallet
     );
 
+/** What a generation counter covers: a subscription field, or the request stream. */
+type Scope = "incomingFunds" | "contractEvents" | "request";
+
+/**
+ * Bound to the generation live when it was created. Re-check `stale` after
+ * every await that precedes a side effect — an emit, a repository write, or a
+ * subscription-field assignment.
+ */
+type Generation = { readonly stale: boolean };
+
+type Emitter = Generation & { emit(response: WalletUpdaterResponse | null): void };
+
 export class WalletMessageHandler
     implements MessageHandler<WalletUpdaterRequest, WalletUpdaterResponse>
 {
@@ -889,6 +901,18 @@ export class WalletMessageHandler
     private incomingFundsSubscription: (() => void) | undefined;
     private contractEventsSubscription: (() => void) | undefined;
     private onNextTick: (() => WalletUpdaterResponse | null)[] = [];
+    private channel: { broadcast(response: WalletUpdaterResponse): void } | undefined;
+
+    // One counter per subscription field, bumped exactly where that field is
+    // dropped — never where a neighbour's is. They cannot be merged:
+    // `onWalletInitialized` re-subscribes incoming funds while
+    // `ensureContractEventBroadcasting` deliberately keeps its subscription, so
+    // a shared bump would leave the live contract-event emitter permanently
+    // stale, and folding requests in would truncate an in-flight settle.
+    private incomingFundsEpoch = 0;
+    private contractEventsEpoch = 0;
+    /** Per-request progress callbacks: ends with the handler, not with a re-init. */
+    private requestEpoch = 0;
 
     /**
      * Instantiate a new WalletUpdater.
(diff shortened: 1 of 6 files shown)


pietro909:
Thanks — two of the three are addressed in 5d11cccb; the change-blocking one does not reproduce.

### 🔴 `epochOf` returning `undefined` as `number` — not reproducible

The premise is that `noImplicitReturns` is off, so TS won't flag a missing case. `noImplicitReturns` isn't the check that applies here: a function with a declared non-`undefined` return type and a reachable fall-through is **TS2366**, which fires under `strictNullChecks` (part of `strict: true`, which `config/tsconfig.base.json` does set).

Reduced to the exact shape, with `--strict` and nothing else:

```ts
type Scope = "a" | "b" | "c" | "d";   // 4th member added, epochOf not updated
private epochOf(scope: Scope): number {
    switch (scope) {
        case "a": return this.a;
        case "b": return this.b;
        case "c": return this.c;
    }
}
```
```
error TS2366: Function lacks ending return statement and return type does not
include 'undefined'.
```

So a fourth `Scope` member fails the build; it cannot reach the `born !== undefined` state described. The `default: { const _e: never = scope; ... }` guard would detect the same case at the same time (compile), only with a different message, so it buys no coverage — and the repo uses that pattern nowhere else. Left as is.

### 🟡 `start()` typed through the non-generic base — fixed

Now `Parameters<MessageHandler<WalletUpdaterRequest, WalletUpdaterResponse>["start"]>`, matching the generic the class actually implements, so the channel param no longer rests on method bivariance.

### 🟡 `drain()` silent drops — fixed

All three batch-level discard sites now log under `debug` via a `logDroppedBroadcasts` helper (batch size, reason, response tags), plus the per-client `postMessage` failure. Mirrors the existing `deliverResponse` precedent for a null `event.source`. No behaviour change when `debug` is off.

---

`pnpm lint`, `tsc --noEmit`, and the full unit suites are green (ts-sdk 2783, boltz-swap 616, swap).

The red `Integration (ts-sdk / exit-providers-rotation)` check is unrelated: it's `unilateralExit.test.ts > exits a vhtlc as receiver with the preimage` hitting its 300s timeout, with arkd logging `connection refused` on :6060 and `wallet is locked or syncing` throughout. This PR touches only service-worker code, which those node e2e tests never load. Integration is also intermittently red on `master` right now (`ts-sdk / ark-core`).

pietro909:
Round 2 — CodeRabbit's README finding is fixed in 5f259b6f. The `epochOf` blocker still does not reproduce; here is the disproof run against this repo's own config rather than a reduction.

### 🔴 `epochOf` exhaustiveness — disproved against `packages/ts-sdk/tsconfig.json` at head

The re-review's argument is *"`noImplicitReturns` is still absent from `config/tsconfig.base.json` (confirmed at head)"*. That is true and it is not the check that applies. TS2366 — *function lacks ending return statement and return type does not include `undefined`* — is a `strictNullChecks` diagnostic, and `strict: true` **is** set in that file.

I applied literally the change the review describes: added the `"reloadWallet"` member it names to `Scope`, left `epochOf` untouched, and ran the package's real typecheck.

```
$ node node_modules/typescript/bin/tsc -p packages/ts-sdk/tsconfig.json --noEmit
packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts(994,36):
error TS2366: Function lacks ending return statement and return type does not include 'undefined'.
exit=2
```

Line 994 is `epochOf` — the exact function and line the review flags as unguarded. With `noImplicitReturns` absent, exactly as stated. The build fails, so the described end state (`epochOf` returning `undefined`, `born !== undefined` always true, every emitter on that scope permanently stale) is unreachable: nobody can land that fourth scope without the compiler stopping them at this function.

The proposed `default: { const _exhaustive: never = scope; ... }` detects the same mistake at the same moment — compile time — differing only in the error text. It adds a runtime `throw` on a branch the type system proves dead, and the pattern appears nowhere else in this repo. I've left it out; happy to add it if a human reviewer wants the nicer diagnostic, but it is not a correctness blocker and I'd rather not carry the precedent on a false premise.

### 🟡 Unsigned head commit
Intentional — the author was away from the signing key. Squash-on-merge produces a signed commit, so this doesn't reach `master`.

### 📐 CodeRabbit — `tickIntervalMs` described as a delivery bound — fixed (5f259b6f)

Correct, and worse than the comment says. `scheduleNextTick()` is called from `runTick()`'s `finally`, so the interval is measured from the previous tick's **end**: the period is `tickIntervalMs` + tick duration. On top of that `runTick` awaits each handler's `tick()` in sequence, and returns early without ticking at all while `pendingInitCount > 0`. So the timer is a floor, never a bound. Reworded accordingly:

> Responses a handler holds for its `tick` wait for the `setTimeout` loop instead, which paces periodic work rather than bounding delivery: `tickIntervalMs` (default 10s) is a floor measured from the previous tick's end, and a slow handler or an in-flight init pushes it out further.

---

`pnpm lint` and the ts-sdk unit suite (2783) are green; the pre-commit hook ran the whole workspace. Checks unchanged otherwise — `Integration (ts-sdk / exit-providers-rotation)` remains the pre-existing `unilateralExit` e2e timeout under arkd instability, untouched by service-worker code.

PR #816, opened 2026-08-31
feat(sdk): treat isUnrolled as a location axis, not a spend fact

pietro909:
A unilateral exit is written `isUnrolled: true, isSpent: false`, and the terminal-spend predicate never read it — so an exited VTXO still counted as Arkade-spendable and stayed selectable for a send, a renewal batch or a recovery batch, all of which fail at the server.

Supersedes edd95df9 on `fix/unrolled-vtxo-balance`, which folded `isUnrolled` into `hasTerminalSpend`. NArk doesn't: `Unrolled` there means "this output lives onchain now" and is never part of the spend predicate. Same outcome for the bug, but it keeps `hasTerminalSpend` == `ArkVtxo.IsSpent()` and the copies of it around the codebase honest.

**Predicates.** `canSpendOffchain` / `canRecoverOnchain` both refuse an unrolled coin; new `canSweepOnchain` claims it. The three now partition the unspent set. `filterSnapshotVtxos` and the worker's `handleGetVtxos` branch on location first, so `withUnrolled` is authoritative.

**Balance.** New `unrolled` bucket, disjoint from the other four and inside `total` — the money is still the user's, onchain behind a CSV, with `completeUnroll` as the remedy. `WalletBalance` gains a field (semver minor). Note `Wallet.getBalance` has to ask for unrolled coins explicitly or the bucket reads zero on the main thread while the worker reports the truth; there's a parity test over one real wallet and one real repository.

**Gap A.** Neither exit path could reach a `ContractManager`, and delta sync can't see the change (`GetVtxosOptions.after` filters on creation time). `OnExitObserved` is a plain function type with no runtime imports, so `Unroll.Session` and the keyless `Executor` keep their independence; `Unroll.sessionFor` / `UnilateralExit.execute` wire it from a wallet so the default path is right without callers opting in. It's a prompt, not a guarantee — the hook fires on the explorer, the refresh reads the indexer, and a miss just leaves things where they already were. Documented at each surface.

**Also fixed, because reverting `hasTerminalSpend` would otherwise regress them:** `ArkadeContract.getUtxos`, the deprecated-signer migration rollup, `selectPendingRecoveryOutpoints`, `classifyCashVtxos` (new `exited` reason), `prepareUnrollTransaction`, and `intentReconciliation.isConsumed` — which also missed `spentBy` and read `settledBy: ""` as set.

**swap** (separate commit): `readLockupFate` stops keeping its own copy of the union and gains an `exited` fate; `runPass` reports it as `needs_counterparty` per half, so the independent L1 claim keeps running and a proven claim keeps its label until the refund window opens. `findLockupVtxos` drops exited outputs so no caller can push a doomed refund.

### Worth a reviewer's eye
- `boltz-swap` regresses deliberately (`arkade-swaps.ts:676,2501` now count an unrolled coin as unspent). Deprecated + unmaintained, stated rather than silently exempted.
- `ArkadeContract.getUtxos`'s no-manager fallback still trusts server-side `spendableOnly` — pre-existing, the two branches can disagree.
- `refundIfUnresolved` reports an exited lockup as `nothing_to_refund`, where the analogous swept case gets `needs_recovery` with outpoints.

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking funds that have been unilaterally exited onchain.
  * Added an `unrolled` balance category; these funds remain included in totals but are excluded from spending and recovery.
  * Added wallet helpers for executing exits and automatically refreshing exit status.
  * Added APIs to identify funds eligible for onchain sweeping.

* **Bug Fixes**
  * Prevented exited outputs from being treated as spendable, recoverable, or refundable.
  * Swap flows now correctly block or report exited lockups while preserving valid claims.

* **Documentation**
  * Updated unilateral-exit and balance guidance.

### DIFF
--- a/packages/swap/src/refund.ts
+++ b/packages/swap/src/refund.ts
@@ -52,6 +52,7 @@ import {
     assertSubmittedArkTxid,
     buildOffchainTx,
     getArkPsbtFields,
+    hasTerminalSpend,
     matchServerCheckpoints,
 } from "@arkade-os/sdk";
 
@@ -140,12 +141,15 @@ export interface LockupVtxo {
      * It is still the trader's money and it is still visible, which is why
      * {@link findLockupVtxos} returns it. What it is not is refundable by
      * {@link pushRefundWithoutReceiver}: that builds an offchain Ark
-     * transaction, and the SDK's own predicates make the two states mutually
-     * exclusive — `canSpendOffchain` is false exactly when `canRecoverOnchain`
-     * is true (`wallet/vtxo.ts`), and the latter is documented as "must be
-     * recovered into a fresh batch rather than spent offchain". Holding the
-     * trader's `sender` key does not change that; a sweep removes the leaf from
-     * the live tree, not the signature from the trader.
+     * transaction, and the SDK's own predicates put the two states on opposite
+     * sides — `canSpendOffchain` is false whenever `canRecoverOnchain` is true
+     * (`wallet/vtxo.ts`), and the latter is documented as "must be recovered
+     * into a fresh batch rather than spent offchain". Holding the trader's
+     * `sender` key does not change that; a sweep removes the leaf from the live
+     * tree, not the signature from the trader.
+     *
+     * Opposite, but not exhaustive: an unrolled output satisfies neither, and
+     * {@link findLockupVtxos} drops those before they reach this type at all.
      *
      * `packages/boltz-swap` splits on exactly this fact rather than working
      * around it: `settleRefundWithoutReceiver` sends a live VTXO through an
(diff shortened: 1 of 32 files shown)


pietro909:
Acted on the four actionable items in `5da847be`:

- **`withUnrolled` gotcha** — JSDoc on `GetVtxosFilter.withUnrolled` now says it is authoritative on the location axis and only that (spent exited outputs included, which is what `prepareUnrollTransaction` behind the worker needs), so unlike `withRecoverable` it doesn't narrow to a capability — test `canSweepOnchain` before acting on it.
- **`Session.do()` contract** — noted on `next()`: `do()` carries every effect, and at DONE it's the only place the observer fires.
- **`Unroll.sessionFor` / `UnilateralExit.execute` coverage** — both now tested directly, asserting the wiring by driving it rather than by inspecting it, plus `execute`'s override priority and provider forwarding. Each verified to fail when its src line is removed.

The rest I'm leaving as flagged in the description: the `boltz-swap` regression (deprecated, explicit non-goal), the `getUtxos` fallback branch, and `refundIfUnresolved`'s `nothing_to_refund` asymmetry.

2817 / 600 / 616 green.

PR #818, opened 2026-09-01
refactor(vhtlc)!: one nonInteractiveParameters opt-in replaces the per-leaf covenant options

Kukks:
Replaces the per-leaf opt-ins on `VHTLC.ScriptV2` (`nonInteractiveClaim?`,
`nonInteractiveRefund?` ╬ô├ç├╢ including the `withoutReceiver` flag shipped in 0.4.67) with a single
`nonInteractiveParameters` object. Present builds the full nine-leaf suite; absent builds the plain
six-leaf VHTLC; `legacy: "preTimelockedRefund"` rebuilds the pre-timelocked-refund eight-leaf
shape that shipped between 0.4.60 and 0.4.67, for talking to records written in that window.

## Why

The per-leaf shape described a freedom the covenant suite doesn't have. All three emulator
leaves derive from ONE emulator key ╬ô├ç├╢ one deployment decision, one key, one suite. The old
options admitted trees no deployment would ever produce and no row could faithfully describe:

- claim-only or refund-only trees (the suite is all-or-nothing);
- claim and refund leaves naming **different** emulator pubkeys (the constructor derives both
  cosigners from one `nirCosigner`; such a tree is unbuildable);
- `nonInteractiveRefund` with `withoutReceiver` omitted, i.e. an eight-leaf tree that is
  neither the legacy shape nor the current one ╬ô├ç├╢ a shape the deserializer then silently
  rebuilds *wrong*, because it can't distinguish "legacy" from "forgot the flag".

Collapsing to one object makes the type system say what the script already enforces: the
suite is one decision. The leaf bytes and their order are untouched ╬ô├ç├╢ claim, refund,
timelocked refund still append after the six signature leaves ╬ô├ç├╢ so the byte-frozen cross-SDK
fixture (`test/fixtures/vhtlc-v2-nine-leaf.json`, SHA256 pinned by the dotnet port) still
holds, and every address a 0.4.67 client derived stays identical.

## Handler round-trip

The contract handler has published `nonInteractiveClaim`/`nonInteractiveRefund`/
`nonInteractiveRefundWithoutReceiver` param keys since 0.4.60, so on-disk rows keep exactly
those keys; only the typed surface collapses. Deserialization now validates what the old code
accepted silently:

- a row naming only one of the two leaves throws (incoherent ╬ô├ç├╢ no matching tree);
- a row naming two different emulator pubkeys throws (unrepresentable under one key);
- a row without `nonInteractiveRefundWithoutReceiver:"1"` deserializes as
  `legacy: "preTimelockedRefund"` instead of building a tree that doesn't match what was
  written.

## `@arkade-os/swap`: derive both shapes, accept whichever the solver quoted

The builders (`lightningSendVtxoScript`, `receiveVtxoScript`) take a `legacy?` passthrough
and build through the group. On top, the four quote-verification paths stop deriving one
hardcoded shape: the new `matchQuotedLockup` builds both candidates ╬ô├ç├╢ full-suite first, then
legacy ╬ô├ç├╢ and `verifyLockupAddress` (now `string | string[]`, mirroring the reference client in
intent-solver's `examples/`) accepts whichever matches the quote's `lockup_address`, throwing
`AddressMismatch` carrying every candidate tried when neither does. The **matched** script is
what gets registered as the contract row and echoed in `treeParams`, so a record persisted
from the result rebuilds the lockup the solver actually funded rather than a shape the client
guessed.

Nothing on the wire says which shape a quote carries, so a single-shape client refuses every
quote from a solver on the other side of the rollout ╬ô├ç├╢ the failure intent-solver's
rfq-protocol Γö¼┬║7.1.1.1 exists to prevent. New pins in `rfqReceive.test.ts`: a nine-leaf-quoting
solver selects the full-suite candidate, an eight-leaf quote selects the legacy one, and
neither-match throws with both candidates on the error. The pre-existing goldens are
unchanged: their mock quotes pin the eight-leaf shape the deployed reference solver still
builds, now matched through the legacy candidate.

## Migration

```ts
// before
new VHTLC.ScriptV2({
    ...,
    nonInteractiveClaim: { emulatorPubkey, receiverPkScript },
    nonInteractiveRefund: { emulatorPubkey, senderPkScript, withoutReceiver: true },
});

// after
new VHTLC.ScriptV2({
    ...,
    nonInteractiveParameters: { emulatorPubkey, receiverPkScript, senderPkScript },
});

// rebuilding a pre-0.4.67 (eight-leaf) record
new VHTLC.ScriptV2({ ..., nonInteractiveParameters: { ..., legacy: "preTimelockedRefund" } });
```

`nonInteractiveRefund.senderPkScript` readers (e.g. `refund.ts`) now read
`nonInteractiveParameters?.senderPkScript`.

## Also removed: the `strict` claim bound

The opt-in quoted bound (`nonInteractiveParameters.strict`, and the
`strictClaimAmount`/`strictClaimAssetAmount` param keys the handler wrote for it) shipped in
0.4.67 and earlier and has no consumer, so it goes away here rather than being carried into
the regrouped option. The golden vectors ╬ô├ç├╢ captured before `strict` existed ╬ô├ç├╢ still pass:
the default path's bytes, and so every funded contract's address, are untouched. A stored row
written by a strict-capable release now fails by name at `deserializeParams` instead of
silently re-deriving the weaker default covenant at a different address.

BREAKING CHANGE: `VHTLC.Options.nonInteractiveClaim` / `.nonInteractiveRefund` are removed in
favour of `nonInteractiveParameters`; `VHTLCV2ContractParams` changes the same way. The
`nonInteractiveParameters.strict` option and its `strictClaimAmount`/`strictClaimAssetAmount` param
keys are removed. `AddressMismatch.derived` widens from `string` to `string | string[]` (it now carries every candidate tried). Param-key bytes on persisted contract rows are otherwise unchanged.

## Summary by CodeRabbit

- **New Features**
  - Added support for current and legacy lockup covenant formats, automatically selecting the format matching a quoted address.
  - Added emulator covenant configuration with support for legacy refund behavior.
  - Refund destinations now use the configured emulator covenant sender address when applicable.

- **Changes**
  - Simplified emulator covenant configuration into a single grouped option.
  - Removed strict quote-based amount and asset minimum enforcement.
  - Improved validation and error reporting for mismatched or incomplete covenant configurations.


Kukks:
@coderabbitai review

Kukks:
@coderabbitai review

Kukks:
On the three items:

**1 · Claim-only rows.** By construction none exist from the swap corridors: `lightningSendVtxoScript` / `receiveVtxoScript` have set claim and refund together since the V2 handler was introduced, so a single-sided row could only come from a direct-SDK consumer hand-building a claim-only tree and registering it. For that case the throw is deliberate — the alternative is the pre-PR behaviour of silently re-deriving a different tree against a funded lockup — and the error message names the recovery (restore the SDK version that wrote it). If anyone wants to scan a deployed store, the query is rows whose params carry exactly one of the two emulator-covenant keys; I don't have a deployed store to run it against.

**2 · `AddressMismatch.derived` widening.** Called out in the BREAKING CHANGE footer now. No read of `.derived` as a bare string exists in the visible consumers (all guard on `instanceof` only), and the class is thrown, not persisted, so the blast radius is compile-time only.

**3 · `strict` removal.** The exposure window is concrete: the `strictClaimAmount` / `strictClaimAssetAmount` keys shipped in `@arkade-os/sdk` 0.4.65–0.4.67 (tag `--contains cac0ab7a`), so a strict row could only have been written by a direct consumer who opted into `nonInteractiveClaim: { strict: ... }` on those three versions. No such consumer is known in any visible repo, and the option never appeared in a corridor, sample, or doc. The remaining risk is a private integration that discovered an undocumented option within that window — acknowledged; the throw at least fails loudly by name rather than silently spending with the wrong covenant set.

pietro909 (approved):

PR #819, opened 2026-09-02
fix(sdk): harden psbt and tree validation

Kukks:
Tightens a few validation gaps on data that arrives from the server.

- The bundled identities now sign under a narrower set of sighash types instead of accepting any of them, and reject a disallowed one on the bulk-signing path instead of silently skipping the input.
- Checkpoint PSBTs returned by the server are checked against that same policy before they reach a signer, on the send path and on both recovery paths.
- Tree graph validation now also covers node finality.

Behaviour is unchanged for well-formed inputs; the new checks reject shapes the SDK never produces itself.

## Test plan

- New unit coverage for each check, including the accept cases, the two silent-failure modes on bulk signing, and the cash recovery path. Disabling each check individually makes its own reject cases fail and leaves the accept cases passing.
- `pnpm -C packages/ts-sdk test:unit` — 174 files, 2870 passed, 1 skipped (baseline on master: 173 / 2846 / 1, 0 failing).
- `pnpm -C packages/boltz-swap test:unit` — 23 files, 616 passed, no type errors.
- `tsc --noEmit` clean; biome format clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Transaction signing now rejects unsupported sighash modes before processing, preventing incomplete or silently skipped signatures.
  * Pending checkpoint transactions with unsupported sighash settings are rejected during submission and wallet finalization.
  * Transaction tree validation now rejects non-final transactions with locktimes or non-default input sequences.

* **Improvements**
  * Signing consistently supports only the approved default and “all” sighash modes, including the permitted anyone-can-pay variant.


Kukks:
Thanks — first one was right and is fixed in `8d344caa`; second one I'm deliberately not taking.

**Bulk signing (accepted).** Confirmed against scure 2.0.1 with a direct probe rather than by reading:

- all inputs disallowed → `sign()` throws `No inputs signed` → the pre-existing catch swallowed it → returned a silently **unsigned** tx
- mixed → `sign()` returned `true`, signed input 0, silently **skipped** input 1

No signature was ever produced under a disallowed sighash, so the original hole stayed closed either way, but the failure was silent, which is its own problem. Both identities now preflight the whole PSBT before the bulk path. Three tests added covering all-disallowed, mixed, and the still-signs case; removing the preflight makes the two reject cases fail.

**Locktime vs. sequence (not taking).** The Bitcoin semantics are right — nLockTime is unenforced when every input is `SEQUENCE_FINAL`, so `lockTime != 0` with a final sequence is technically still final. I'm keeping the stricter check anyway:

- These trees are operator-supplied and the honest builder is a single function (`arkd` `pkg/ark-lib/tree/builder.go`, `getTx`) that hardcodes `psbt.New(..., version 3, nLockTime 0, [wire.MaxTxInSequenceNum])` for both the VTXO tree and the connector tree. A non-zero locktime is therefore never something we should be receiving, final sequence or not.
- Relaxing it would accept a shape no honest operator emits, in exchange for no compatibility benefit, on the exact data the finding is about.
- It matches the existing precedent in this validator family: the Go equivalent already hard-rejects `Version != 3` on the same "must look like what the builder produces" basis.

Happy to revisit if a legitimate producer of locktime-bearing tree nodes turns up.

Kukks:
@coderabbitai review

Kukks:
Both addressed in `0c2af2f4`. Thanks — the first was a real miss.

**1. Inconsistent guard on the arkadeCash recovery path — fixed.** You're right, and it's a genuine policy hole rather than just an inconsistency: that path signs server-returned checkpoints via `identity.sign(checkpoint)` in bulk mode, so it fell through to the identity-level `[DEFAULT, ALL, ALL_ANYONECANPAY]` instead of the `[DEFAULT]`-only policy the primary path enforces. A checkpoint declaring `SIGHASH_ALL` was rejected on one recovery path and accepted on the other. `assertAllowedSighashTypes(tx)` now runs at the decode in `wallet.ts` alongside `assertCheckpointsMatchInputs`, matching the primary site exactly.

Locked in with a test in `arkadeCashClaim.test.ts`: a pending sweep whose checkpoint declares `SIGHASH_ALL` is refused, `finalizeTx` is never called, and nothing is swept. Verified discriminating — disabling *only* that guard (leaving the primary one in place) makes it fail.

**2. Interleaved re-export — fixed.** `export { assertAllowedSighashTypes }` now sits after all import declarations rather than between two import groups.

Gate: 174 files, 2870 passed, 1 skipped, 0 failing (baseline on master 173 / 2846 / 1). `tsc --noEmit` clean, biome clean, and all 8 CI checks were green on the previous commit.

For the record on the earlier thread: I'm still keeping the strict `lockTime != 0` rejection rather than gating it on sequence finality, for the reason given above — arkd's `getTx` hardcodes locktime 0 for every tree node, so relaxing it would only widen what we accept.

Kukks:
All four closed in `93b3ad71`. Issue 4 was the interesting one — the concern was unfounded, but it pointed at a real gap in my *tests*.

**Issue 4 — PSBT round trip.** I probed scure directly rather than reasoning about it: a tx built with `lockTime: 0`, serialized with `toPSBT()` and re-parsed with `fromPSBT()`, comes back as `lockTime === 0` (`typeof number`), never `undefined`. Sequence likewise round-trips as `4294967295`. So `undefined !== 0` cannot occur and no guard is needed.

But you were right that it was untested, and more so than the review states: `TxTree.create()` parses every node via `Transaction.fromPSBT()`, so the round trip *is* the production path, while my tests constructed `Transaction` objects directly and never exercised it. Added a test that goes through `TxTree.create()` with base64-encoded PSBTs and asserts both the accept and the two reject cases survive.

**Issue 3 — indexed-path asymmetry.** Comment added at both sites noting that `signIdx` rejects a disallowed sighash itself, rather than skipping the input the way the bulk path does — which is precisely why only the bulk path needs the preflight.

**SeedIdentity coverage.** Added a `SeedIdentity sighash policy` block mirroring the SingleKey one: three reject cases on the bulk path, one on the indexed path, three accept cases. Verified discriminating — disabling only the SeedIdentity preflight fails all three bulk cases.

Gate: 174 files, 2878 passed, 1 skipped, 0 failing (was 2870; +8 new tests). `tsc --noEmit` clean, biome clean.

PR #821, opened 2026-09-02
feat(swap): v2 type foundation

pietro909:
Milestone 1 of the Swap SDK v2 track: the vocabulary every later milestone
imports. Pure types, one amount codec, one mapping table — no client, no
network, no persistence.

## What lands

`packages/swap/src/client/` — a directory rather than `src/v2/`, because the
package ends as two layers the spec already names (the client at the root, the
v1 building blocks demoted to `/protocol`), and a version-numbered directory
becomes a lie the day v2 is the only thing. `src/swapClient.ts` and the
`createSwapClient` name are occupied by the facade #793 landed; the exported
name stays an open question, the file question is settled.

| Module | Contents |
|---|---|
| `assetId.ts` | CAIP-19 grammar, `AssetId<R>`, parse/format, cross-rail sameness |
| `corridor.ts` | `Corridor` (discovery's, aliased), `CorridorId`, the rail bijection |
| `route.ts` | `Instrument`, `Endpoint<C>`, the closed `Route`, `Artifact` |
| `amount.ts` | the amount law and both decimal codecs |
| `rfqAmount.ts` | the wire adapter and the `amount_side` rename |
| `aliases.ts` | public ids down to discovery/RFQ vocabulary |
| `errors.ts` | the sixteen-member taxonomy |
| `primitives.ts` | `Hex`, `Pubkey` |

## Two invariants that are now compile errors

`Endpoint<C>` spells its asset `AssetId<RailOf<C>>`, so no value exists in which
an endpoint's corridor and its asset's rail disagree. And `onchain -> arkade` is
absent from `Route`, so the misroute is a type error once a route has been
resolved — before resolution it stays `UnsupportedRoute`, thrown ahead of RFQ
disclosure, artifact creation, persistence and funding.

Both are asserted in `test/client/route.types.ts`, and the branded
`AtomicDecimal` in `test/client/amount.types.ts`. These are enforced, not
decorative: `tsc --noEmit` reads a tsconfig whose `exclude` names `test`, and
vitest runs without `--typecheck` where `expectTypeOf` is `() => true`. The
first commit adds `tsconfig.test.json` scoped to `test/client/**` and extends
the `typecheck` script, which CI already runs. A `@ts-expect-error` that stops
erroring now fails the build (`TS2578`).

## Rulings the milestone owed

- **`CorridorId` survives** beside the rail namespaces. It is the vocabulary
  §3, §6 and §9 are written in, `bolt11` is a CAIP-2 artifact rather than a
  product word, and the correspondence is not a bijection once §9 is admitted —
  one `eip155` rail, many `eip155:<chain>` corridors. The redundancy objection
  is answered by the type, not a runtime check.
- **The arkade asset reference is the flat 68-hex identity**, not the spec's
  prose `<genesis_txid>.<idx>`. That identity is implemented in seven places
  across three repos and pinned by `ASSET_ID_VECTORS` precisely because every
  disagreement between sites fails silently; this layer adds no eighth spelling.
- **The network reference is core's `NetworkName`**, all five members.
  Discovery omits `testnet`; an asset on testnet exists whether or not anyone
  indexes it, so the partiality belongs at the discovery boundary.
- **Ids are lowercase throughout**, `erc20` included, and an uppercase one is
  refused rather than folded — folding destroys a caller's checksum silently,
  and loosening a validator later is the reversible direction.
- **Taxonomy members are bare condition nouns**, so `QuoteVerificationError`
  becomes `QuoteVerificationFailed`. The inverse is what earns it:
  `AssetIdError` and `AmountFormatError` are codec faults on caller input, not
  swap-boundary errors, and the suffix is what says so at the catch site.
- **The v2 modules stay internal.** `src/index.ts` is not in this diff, so
  nothing here reaches `dist` or npm; M8 slims the root export and creates
  `src/protocol.ts` (one exports key, one tsup entry, a barrel over the same
  declarations). Exporting now would make every M2–M7 refinement a change to a
  published surface.

## Existing code touched

One line, in `rfq.ts`: `SwapRefusal` moves the `name` it already set into a
literal-typed field. A `string` there collapses the discriminant for the whole
union. Same own property, same value, same runtime.

## Verification

`pnpm -C packages/swap run typecheck` (both projects), `test:unit` — 698 passing,
81 new — `lint`, `build`, `smoke:dist` and `npm pack --dry-run`. No regtest:
M1 has no network path, which is the point of the milestone.

### DIFF
--- a/packages/swap/src/client/corridor.ts
+++ b/packages/swap/src/client/corridor.ts
@@ -1,0 +1,69 @@
+/**
+ * The corridor axis, and its bijection with the rail namespaces.
+ *
+ * One axis, two vocabularies: discovery speaks `arkade | lightning | onchain`
+ * and an asset id's CAIP-2 namespace is `arkade | bolt11 | bitcoin`. They agree
+ * on one member of three. Collapsing them was the alternative and it loses
+ * either way — take the rail names and every market lookup translates on the
+ * way out to discovery; take the corridor names and `lightning:` overruns
+ * CAIP-2's eight-character namespace cap.
+ *
+ * So both stay, and the disagreement is spent once, here: `Corridor` is
+ * discovery's type verbatim, {@link railOfCorridor} is total, and `route.ts`
+ * ties an endpoint's corridor to its asset's rail in the type system so the two
+ * cannot disagree in a value.
+ */
+import type { Corridor as DiscoveryCorridor } from "@arkade-os/solver-discovery";
+import type { BitcoinRail, Rail } from "./assetId";
+
+/**
+ * The corridor a leg settles on.
+ *
+ * Aliased from discovery rather than re-declared: it is discovery's vocabulary,
+ * the alias layer has to speak it, and a re-declaration would drift silently
+ * the day discovery adds a corridor.
+ */
+export type Corridor = DiscoveryCorridor;
+
+/** Every corridor, in the order the route union uses them. */
+export const CORRIDORS = ["arkade", "lightning", "onchain"] as const satisfies readonly Corridor[];
+
+/**
+ * A corridor id: the three implemented corridors plus §9's EVM chains.
+ *
+ * The template arm stays open against the registry's closed enum on purpose —
+ * which chains a registry lists is a listing decision, not an id-grammar one.
+ */
+export type CorridorId = Corridor | `eip155:${number}`;
+
+/** The rail namespace a corridor's assets are spelled on. */
+export type RailOf<C extends CorridorId> = C extends "arkade"
+    ? "arkade"
+    : C extends "lightning"
+      ? "bolt11"
+      : C extends "onchain"
+        ? "bitcoin"
+        : "eip155";
+
+const RAIL_BY_CORRIDOR = {
+    arkade: "arkade",
+    lightning: "bolt11",
+    onchain: "bitcoin",
+} as const satisfies Record<Corridor, BitcoinRail>;
+
+const CORRIDOR_BY_RAIL = {
+    arkade: "arkade",
+    bolt11: "lightning",
+    bitcoin: "onchain",
+} as const satisfies Record<BitcoinRail, Corridor>;
+
+/** The rail a corridor's assets are spelled on. Total, and the type agrees. */
+export const railOfCorridor = <C extends CorridorId>(corridor: C): RailOf<C> =>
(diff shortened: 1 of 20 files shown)


packages/swap/src/rfq.ts
```diff
@@ -149,11 +149,14 @@ export const RFQ_TERMINAL_STATES = ["settled", "refused", "expired", "refunded",
 
 /** A refusal from the solver, carrying its closed-set reason. */
 export class SwapRefusal extends Error {
+    /** Literal-typed so the v2 error taxonomy's union discriminates on `name`
+     * — a `string` here collapses the discriminant for every member. Same value
+     * the constructor has always set, moved to a field initializer. */
```
pietro909, line 154: Redundant comment, remove it

packages/swap/src/client/corridor.ts
```diff
@@ -0,0 +1,69 @@
+/**
+ * The corridor axis, and its bijection with the rail namespaces.
+ *
+ * One axis, two vocabularies: discovery speaks `arkade | lightning | onchain`
+ * and an asset id's CAIP-2 namespace is `arkade | bolt11 | bitcoin`. They agree
+ * on one member of three. Collapsing them was the alternative and it loses
+ * either way — take the rail names and every market lookup translates on the
+ * way out to discovery; take the corridor names and `lightning:` overruns
+ * CAIP-2's eight-character namespace cap.
```
gringokiwi: Nope, just use caip-2 rail names 

No need for Corridor

packages/swap/src/client/corridor.ts
```diff
@@ -0,0 +1,69 @@
...
+ * The corridor a leg settles on.
+ *
+ * Aliased from discovery rather than re-declared: it is discovery's vocabulary,
+ * the alias layer has to speak it, and a re-declaration would drift silently
+ * the day discovery adds a corridor.
+ */
+export type Corridor = DiscoveryCorridor;
+
+/** Every corridor, in the order the route union uses them. */
+export const CORRIDORS = ["arkade", "lightning", "onchain"] as const satisfies readonly Corridor[];
```
gringokiwi: arkade, bolt11, bitcoin (as rail/network namespace)

No need to future proof for eip155 yet

gringokiwi: alternatively get rid of corridor entirely

PR #825, opened 2026-09-03
feat(swap): v2 corridor modules

pietro909:
Swap SDK v2, M2 — corridor modules. Plan: `meta/tracks/swap-sdk-v2/02-corridor-modules.md`. Stacked on M1 (#821).

One contract, three implementations under `packages/swap/src/client/corridors/`. Internal — nothing is added to either root barrel and `createSwapClient` is untouched.

- **Contract + registry.** Keyed on the corridor, not v1's `RfqSwap["kind"]`. `matches(raw)` answers `{claimed}` / `{refused}` / `undefined` — sync, non-throwing, amount-blind, since M7 wraps two of these in core's `PaymentRail.match`. The registry turns the counts into `AmbiguousDestination`; an LNURL is left unclaimed for M3.
- **Parsing.** Core's `arkTarget`/`btcTarget`/`invoiceTarget`; no second classifier. The network checks core omits are the modules': `assertRecipientArkAddress` (arkade), `L1_NETWORKS[l1NetworkFromArk]` (onchain), an authored HRP table (lightning).
- **Decoder.** `light-bolt11-decoder` becomes a dependency of `@arkade-os/swap`, behind the overridable `decode` seam. Absolute expiry, `bigint` millisats, a missing payment hash refused rather than defaulted to `""`.
- **Chain source.** First implementation of `ChainSource`, over an esplora URL. `getMtp` reads `mediantime` off `/blocks` and never forwards `getChainTip().time`.
- **Overrides.** `T | null`; `resolveCorridorDeps` throws `MissingCorridorDep` per corridor on first use.

Two core changes ride along: `BIP21.parse`/`create` stop lowercasing the address (base58 is case-sensitive) and the `ark=` filter becomes case-insensitive — both made a destination classify differently bare than as a param. Plus one root export, `assertRecipientArkAddress`.

Verified: build, both typechecks, lint, `test:unit`, `smoke:dist`, and the new `test/e2e/chainSource.test.ts` against a regtest stack.

## Summary by CodeRabbit

* **New Features**
  * Added swap destination support for Arkade addresses, Bitcoin on-chain addresses, and BOLT11 Lightning invoices.
  * Added corridor-based destination classification with network validation and clear handling for unsupported or ambiguous destinations.
  * Exposed address validation utilities and recipient address types through the SDK’s root API.
  * Added default chain and invoice processing for swap flows.

* **Bug Fixes**
  * Preserved case-sensitive addresses when creating and parsing BIP21 URIs.
  * Made `ark=` parameter matching case-insensitive while retaining valid address formatting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,14 @@ style and have not been backfilled.
 
 ### Features
 
+- **`assertRecipientArkAddress` and `RecipientAddressContext` are
+  root-exported.** The rotation-aware recipient check — hrp, then
+  `classifyAgainstSignerSet`, refusing `UNKNOWN_SIGNER` and `EXPIRED`
+  distinctly — was reachable only from inside core, while every
+  ingredient it needs already was. A plugin classifying a destination
+  address had to hand-roll a `serverPubKey ===` comparison, which
+  rejects valid addresses mid-rotation. Nothing about the function
+  changed.
 - **`IReadonlyWallet.getArkadeInfo()`.** The wallet is the single place
   that knows which Arkade server it speaks to, so a plugin needs only
   the wallet, never a server URL of its own. It answers with the live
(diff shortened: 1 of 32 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,14 @@ style and have not been backfilled.
 
 ### Features
 
+- **`assertRecipientArkAddress` and `RecipientAddressContext` are
```
tiero: Since is new let's call it ArkadeAddress please

tiero (commented):
Let's always chose Arkade instead of Ark for new variables and comments (add to your AGENTS md if needed) 

Please keep a README already so we monitor the dev UX as we go

PR #829, opened 2026-09-03
feat(swap): accept(), the durable record, and the storage defaults

pietro909:
Swap SDK v2, M4. Stacked on #821's successor — base is `feat/swap-v2-quote-path`, so the diff is M4 only.

`accept()` gets value committed and durable, then stops. One ordering on every route:

```
expiry guard -> idempotency read -> InsufficientFunds -> register covenant -> persist -> fund -> txid
```

Persist-first becomes a mechanism rather than a README instruction: v1 keyed a record on its funding txid, so the record could not exist before the money did. The key is now the client-minted `Quote.id`.

Three things the plan did not name, all settled in the track's Decisions table:

- **Covenant registration was unowned in v2.** The quote path reaches no contract manager by design, and a lockup's contract row is the only place its tree lives — so without registering inside `accept()`, every persisted record is unrebuildable. It is now a step in the ordering, before the persist.
- **The funding gate is the give _instrument_, not its asset.** A `lightning->arkade` receive gives BTC too, so an asset-branched pre-flight would refuse the canonical empty-wallet receive.
- **`version: 5` is a bump, not a fork.** The shared scan-cursor ruling makes a second interface impossible — it would force two repository objects over one database.

Also: `@arkade-os/swap/node` ships the Node storage default (file-backed SQLite, config-dir resolver, disposal that closes the connection it opened).

### Breaking

`AssetSwapRepository.version` is `5`, adding four record-store methods. Realm consumers gain a fifth schema. The IndexedDB migration is additive; v1 readers and rows are untouched.

### Verification

969 unit tests in the package (up from 870); four-backend conformance for the new store, an IndexedDB 2→3 migration case, a structural no-`bigint` type test, and `smoke:dist` over the new subpath under both `import` and `require`.

Not done: the regtest pass. The crash windows are covered at unit level by failing the next step and asserting what is durable, but the matrix wants two e2e profiles that cannot both be up at once.

## Summary by CodeRabbit

* **New Features**
  * Added `SwapClient.accept()` to durably accept quotes, move funds, recover from interruptions, and safely retry by quote ID.
  * Added durable swap-record support across repository backends.
  * Added a Node.js file-backed SQLite storage option with per-network database paths and automatic cleanup.
  * Added public types and conversion utilities for persisted swap records.
  * Added direct access to Realm schema exports.

* **Breaking Changes**
  * Repository version increased to 5; custom repository implementations must support swap-record operations.
  * Realm and IndexedDB schemas require migration updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,22 @@ style and have not been backfilled.
 
 ### Breaking Changes
 
+- **`@arkade-os/swap`: `AssetSwapRepository.version` is `5`, adding a v2
+  swap-record store.** Four methods land beside the existing ones —
+  `saveSwapRecord`, `getSwapRecord`, `getAllSwapRecords`,
+  `removeSwapRecord` — keyed by the client-minted quote id rather than a
+  funding txid, which is what lets a swap record exist *before* its
+  money does. An out-of-tree implementor of the interface fails to
+  compile until it adds them, which is what the version literal is for.
+  The four in-tree backends are updated; the IndexedDB one bumps
+  `DB_VERSION` to 3 and the migration is additive, so existing rows come
+  through untouched. Realm consumers gain a fifth schema,
+  `ArkadeSwapRecord` — spread `AssetSwapRealmSchemas` into your config
+  and bump your `schemaVersion`. The v1 readers are unchanged and keep
+  reading v1 rows; the two record families are deliberately invisible to
+  each other, while the restore-scan cursor and the markets cache stay
+  shared.
+
 - **`@arkade-os/swap`: the `arkServerUrl` positional is gone from the
   five covenant-deriving entrypoints.** `createOffer(wallet, params)`,
   and `requestLightningSend` / `requestLightningReceive` /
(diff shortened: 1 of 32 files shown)


pietro909:
**From M5 pre-entry validation: `failure`/`blockedReason` can't do the job their docblock claims.**

`client/record.ts:162-167` documents the two fields as "M5's ask: the durable claim-error strings its outcome table reads, so a restarted client ends a broken claim `failed` with a reason instead of `lapsed` with none."

They can't deliver that, because both are written only at terminal time:

- `failure` — `swapManager.ts:2004` (`fail()`) and `:1898` (the send-leg refund deadline)
- `blockedReason` — `:1948`, only from `block()`

The decision that needs a reason is the receive-leg deadline read at `:1593-1601`, and it runs while the swap is still non-terminal. At that point the record is by construction empty of both, so the round-trip can't separate `failed` from `lapsed` — the exact distinction the docblock says these fields recover.

What carries it today is `lastClaimError` (`:785`), written on every claim throw at `:1701` while the window is still open. It's process-local by design, and its own docblock at `:779-784` already accepts the cost: "after a restart the same swap ends `refunded` instead, which costs the caller a reason and nothing else — every throw was already reported through `onSwapFailed` as it happened."

So the ask is a new pre-terminal durable field written alongside `:1701`, not a read of the two that exist. Two ways to close it:

1. **Add it** — the record gains a claim-error string written on each throw at `:1701`, which makes the docblock's stated behaviour true.
2. **Drop the claim** — keep `failure`/`blockedReason` for what they do carry (terminal reasons) and reword the docblock to match, accepting the documented degradation.

M5 currently takes (2) and records the ask, on the grounds that adding to a milestone under review isn't M5's call. Raising it here while #829 is open, since (1) is cheap now and an M4 follow-up later.

Line pins: `swapManager.ts` against `feat/v0.5` `8769ad3b`; `client/record.ts` against this branch.

PR #832, opened 2026-09-03
feat(swap): a client-side ceiling on what a quote may charge

Kukks:
## Why

A solver's fee is `from_amount - to_amount` and the quote breaks it down no further. Nothing on the client could refuse a quote for being expensive.

That mattered less when a fee was a fixed spread. It matters now that solvers can price dynamically: the flat component tracks the chain fee market, so the same corridor quotes a different fee minute to minute. And the registry card cannot stand in for a ceiling — it publishes the corridor's *configured* flat fee, not the cap live pricing may reach, so a solver with a low fallback and a high cap advertises cheaper than it can charge.

The quote is the only honest number, and until now the only thing a client could do with it was pay.

## What

`assertFundable` takes an optional `maxFee`:

```ts
maxFee?: { bps?: number; sats?: number; referenceRate?: number }
```

**The ceiling is the GREATER of the proportional and absolute bounds.** That is the whole design, because a flat charge is a large proportion of a small swap — 420 sats is 8.4% of 5,420 and 0.084% of 500,000. A bare percentage refuses every small swap the moment fees rise; an absolute alone stops protecting a large one. `max()` lets the absolute cover the flat component without loosening the proportional bound where it does the work.

**Cross-asset is gated too, given a rate.** `from_amount - to_amount` is a fee only while both legs name the same asset; across assets it is the spread against a reference rate, and only the caller can supply one. With `referenceRate` (to-units per from-unit) the fee in FROM units is `(from * R - to) / R` and the same ceiling applies. Without it, a cross-asset pair is refused **loudly** rather than waved through — a caller who set a ceiling and silently got no gate would fund believing they were protected.

The rate must be the caller's **own**. Reading it off the solver's published market feed checks the solver against its own number and passes anything; the feed URL sitting on the card is what makes that the easy mistake rather than an unlikely one. The doc comment says so.

## Compatibility

Absent `maxFee` gates nothing, so every existing caller funds exactly what it funded before. Additive only.

## Testing

17 cases. Mutation-verified in both files:

| mutation | tests killed |
|---|---|
| `max` → `min` | 3 |
| `>` → `>=` | 2 |
| drop the cross-asset refusal | 1 |
| drop the no-rate guard | 1 |
| drop rate validation | 1 |
| `ceil` → `floor` | 1 |
| drop the `bps` range guard | 1 |
| drop the `sats` range guard | 1 |
| drop the `bps` integer check | 1 |
| drop the `sats` integer check | 1 |
| drop the `bps` upper bound | 1 |
| move the `bps` bound off by one | 1 |

The `ceil` → `floor` row is why the rounding is a pinned claim rather than a comment: it passed all ten earlier cases because every one of them divided evenly. The added case is the only one that can tell them apart — R = 3 with 2,999 to-units short is 999.67 sats, which `ceil` refuses against a 999-sat ceiling and `floor` funds. Erring toward refusing is the client's side to err on.

The last six rows come from [arkana's review](https://github.com/arkade-os/ts-sdk/pull/832#pullrequestreview-0): the guards on the *ceiling itself* were unexercised, so a mutation removing either went undetected. They are now pinned by reason code (`max_fee_out_of_range`), asserted on `reason` rather than the message — that code is the closed contract a caller branches on. The boundary values the guards must keep allowing (`bps: 10_000`, and a zeroed bound beside a set one) are pinned alongside, so the guard cannot become the bug it exists to prevent.

Unit suites green: ts-sdk 2886, swap 648, boltz-swap 616.

## Note

The reference client in `arkade-os/intent-solver` (`examples/lib/rfq-core.mjs`) already carries the same gate with identical reason codes, mirrored deliberately so the shipped client and the reference cannot disagree about a money gate.

## Summary by CodeRabbit

* **New Features**
  * Added optional fee ceilings when validating swap quotes.
  * Supports proportional, absolute, and reference-rate-based limits for cross-asset swaps.
  * Same-asset fees are assessed from the amount difference; cross-asset fees use the supplied reference rate with rounded-up calculations.
  * Quotes without configured limits continue to be accepted under existing behavior.

* **Bug Fixes**
  * Rejects quotes with unavailable, invalid, out-of-range, or excessive fee limits.
  * Added clearer refusal reasons for fee-limit validation failures.

### DIFF
--- a/packages/swap/src/rfq.ts
+++ b/packages/swap/src/rfq.ts
@@ -440,51 +440,21 @@ export const assertFundable = (input: {
         direction: "send" | "receive";
     };
     /**
-     * The most this client will pay, as the GREATER of a proportion of
-     * `from_amount` and an absolute number of sats. Absent means no ceiling —
-     * which is what every caller written before this got, and keeps.
-     *
-     * BOTH bounds, because a solver's fee is partly flat. With dynamic pricing
-     * that flat component tracks the chain fee market, and a flat charge is a
-     * large proportion of a small swap: 420 sats is 8.4% of 5_000 and 0.084% of
-     * 500_000. A bare percentage therefore refuses every small swap the moment
-     * fees rise, while an absolute alone stops protecting a large one. `max()`
-     * lets the absolute cover the flat component without loosening the
-     * proportional bound where that is the one doing the work.
-     *
-     * This is the ONLY ceiling a client has. The quote carries no fee field,
-     * and the registry card cannot stand in for one: it publishes the
-     * corridor's CONFIGURED flat fee, not the cap live pricing may reach, so a
-     * solver with a low fallback and a high cap advertises cheaper than it can
-     * charge. `from_amount - to_amount` off the quote is the real number.
-     *
-     * OMIT a bound rather than zeroing it. `{ bps: 0, sats: 500 }` is the right
-     * way to write an absolute-only ceiling and behaves as intended — but a lone
-     * `{ bps: 0 }` is a ceiling of ZERO, which refuses every quote carrying any
-     * fee at all. That cannot be rejected as a mistake, because it is also how a
-     * caller says "free or nothing"; the two are told apart only by what else is
-     * set. `{}`, which names no bound whatsoever, IS refused.
+     * The most this client will pay: the GREATER of the two bounds, because a
+     * flat fee is a large proportion of a small swap (420 sats is 8.4% of
+     * 5_000, 0.084% of 500_000). Absent means no ceiling. OMIT a bound rather
+     * than zeroing it: `{ bps: 0 }` alone is a ceiling of zero, indistinguishable
+     * from "free or nothing". `{}` IS refused.
      */
     maxFee?: {
+        /** Integer, 0..10_000 (10_000 = 100%). Out of range throws `max_fee_out_of_range`. */
         bps?: number;
+        /** Non-negative integer. Out of range throws `max_fee_out_of_range`. */
         sats?: number;
         /**
-         * To-units per from-unit, for a CROSS-ASSET pair. Without it such a pair
-         * is refused rather than waved through — the fee there is the spread
-         * against a rate, and this function is given no rate of its own.
-         *
-         * Must come from a source of YOUR OWN. Reading it off the solver's
-         * published market feed checks the solver against its own number and
-         * passes anything; the feed URL being right there on the card is what
-         * makes that the easy mistake.
-         *
-         * Ignored on a same-asset pair, where the exact fee is already known —
-         * so one ceiling can be set for every pair a wallet handles.
-         *
-         * Expect to need a LOOSER tolerance than same-asset: the spread you
-         * measure necessarily includes rate movement between your feed read and
-         * the solver's quote, so a tight bound refuses honest quotes on
-         * volatility alone.
+         * To-units per from-unit, required for a CROSS-ASSET pair. Must come
+         * from a source of YOUR OWN — the solver's own feed would check it
+         * against its own number. Needs a looser tolerance than same-asset.
          */
         referenceRate?: number;
     };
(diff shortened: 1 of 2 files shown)


Kukks:
Thanks — acted on both of the items you flagged as worth doing before merge, and pushed as `21db69a0`.

**Finding 2 (validation paths untested) — fixed.** Confirmed before acting: `max_fee_out_of_range` appeared in zero tests, and the `fundable` helper's own param type made the invalid cases unreachable without a cast. Six new cases now cover it — `bps: 10_001`, `bps: -1`, `bps: 12.5`, `sats: -1`, `sats: 0.5`, plus a boundary case pinning what the guards must keep *allowing* (`bps: 10_000`, and a zeroed bound beside a set one) so the guard cannot become the bug it exists to prevent.

They assert on `reason` rather than the message, since that code is the closed contract a caller branches on. Mutation-verified: each guard removed whole, each integer check dropped, the upper bound dropped, and the bound moved off by one — six mutations, all killed, baseline clean.

**Finding 3 (zero-bps footgun) — documented.** Added to the `maxFee` doc comment. Worth stating why it is a doc note rather than a guard: `{ bps: 0 }` alone cannot be *rejected* as a mistake, because it is also exactly how a caller expresses "free or nothing". The two are distinguishable only by what else is set, so the honest fix is to say "omit a bound rather than zeroing it" and leave the behaviour alone.

**Finding 1 (float precision) — no change, and I agree with your read.** The product exceeds 2^53 only when `from_amount * referenceRate` does, and at that point `ceil` biases toward refusal, so imprecision costs a false refusal rather than a false pass. Making this exact would mean threading bigint or a decimal type through a gate whose whole value is that a caller can drop it in — a real cost against a failure mode that already errs the protective way. Happy to revisit if a corridor lands where both operands are genuinely that large.

**Finding 4 (`split("->")` with no colon) — no change.** Your own note is the reason: the `legs.length === 2` guard is already there, and by the time a quote reaches `assertFundable` the pair has been through `expectQuote`, so a leg with no `:` is a bug in calling code rather than a solver-supplied value. Adding a second shape check here would put the error one frame further from where it was actually made. If it turns out to be worth catching, `expectQuote` is the place — it is where the pair is already being validated.

Kukks:
@coderabbitai review

packages/swap/src/rfq.ts
```diff
@@ -439,6 +439,55 @@ export const assertFundable = (input: {
...
+     *
+     * OMIT a bound rather than zeroing it. `{ bps: 0, sats: 500 }` is the right
+     * way to write an absolute-only ceiling and behaves as intended — but a lone
+     * `{ bps: 0 }` is a ceiling of ZERO, which refuses every quote carrying any
+     * fee at all. That cannot be rejected as a mistake, because it is also how a
+     * caller says "free or nothing"; the two are told apart only by what else is
+     * set. `{}`, which names no bound whatsoever, IS refused.
+     */
+    maxFee?: {
+        bps?: number;
```
pietro909, line 451: Worth documenting that ` 0 > bps < 10_000` ?

pietro909 (approved):

packages/swap/src/rfq.ts
Kukks: Good call — done in `55377316`. Both bounds now carry their range at the field:

```ts
/** Integer, 0..10_000 (10_000 = 100%). Out of range throws `max_fee_out_of_range`. */
bps?: number;
/** Non-negative integer. Out of range throws `max_fee_out_of_range`. */
sats?: number;
```

Named the reason code rather than just the range, since that is the closed contract a caller branches on — the message beside it is not.

Note the range is inclusive at both ends: `0` and `10_000` are both accepted. `{ bps: 0, sats: n }` is the intended spelling for an absolute-only ceiling, which is why zero cannot be rejected as a mistake — it is also how a caller says "free or nothing". Both boundaries are pinned by a test so the guard cannot quietly narrow.

Same commit also cuts the comments in this PR from 39% of added lines to 11.6%, deletions only — no logic moved, and the max/min, ceil/floor and bps-upper-bound mutations are all still caught.

PR #835, opened 2026-09-03
feat(swap): the EVM corridors' wire layer — pairs, amounts, requests, quotes

Kukks:
`@arkade-os/swap` can quote and fund the Lightning and onchain corridors but not the two EVM ones, which the reference solver has served for some time. This is the **negotiation half** of the client side: the pair form, the token identity, the amount encoding, the two `rfq_request` shapes, and the two quote shapes a solver answers with.

New module `packages/swap/src/evmRfq.ts`, exported from the package index.

## Which parts have production weight

Worth a reviewer's attention up front: the two EVM corridors are independently enabled per token and direction on the solver (`evmEnvStem(token, 'send')` vs `'receive'`), and LVGA — the first real deployment — serves **one** of them. A client receiving LVGA on EVM is, in this codebase's naming, the **`send`** corridor `arkade:BTC->ethereum:<token>`; the names are written from the solver's perspective, so the user-facing "receive" and the code's `send` are the same leg.

On that leg the solver locks, the **client claims**, and the solver refunds on timeout. The client claiming is structural: the solver learns the preimage by watching that claim and uses it to take the Arkade lockup.

So of the surface this PR establishes:

- **Load-bearing today** — the send pair and token identity, the send request builder, `readEvmSendQuote` and everything it validates, the token-amount codec, and `evm_refund_address` (the sixth field of the swap key *and* the explicit `claim` argument).
- **Ahead of demand** — the receive pair, `evmReceiveRequest`, `readEvmReceiveQuote`. Correct and tested, and the generic surface a second deployment or the opposite direction would use, but unexercised by any live corridor today.

Nothing branches on this. There is no direction flag, no LVGA mode, and no deployment-type check — direction is already a solver-config matter, and a client written against the shared surface works either way. The split above is about where to look hardest in review, not about how the code is shaped.

## What is deliberately NOT here

Deriving the Arkade covenant, comparing the two deadlines against each other, and funding.

`evm_timeout_block` is a **block height** and `refund_locktime` is **unix seconds**, and they sit adjacent in the same quote profile. Bridging them needs a per-chain block cadence, and the safe direction of that conversion differs by which side is reading it — reading someone else's timeout assumes the FASTEST cadence, sizing your own assumes the SLOWEST, and a single averaged constant is wrong for one of the two. That is its own decision and its own review, so this PR carries the field, documents the unit trap on it, and gates nothing on it.

## Amounts: bigint in the API, canonical decimal string on the wire

This is the one decision in the PR worth arguing about, so, explicitly: **a token quantity is a `bigint` at this package's API boundary and a canonical decimal string (`^(0|[1-9][0-9]*)$`) on the wire. Never a `number`, at either end.**

An ERC20 amount is 256-bit. A JSON number is an IEEE-754 double, exact only to 2^53 − 1, which at 18 decimals is 0.009 tokens — one whole DAI does not survive the round trip. The rounding happens inside `JSON.parse`, before any validator on either side can see it, and neither party can detect that it happened. This is why the solver's schema types `evm_amount` as a string, why its store declares the column TEXT, and what § 2.1 of the RFQ protocol specifies for every amount.

`bigint` at the boundary because it is the only exact integer JS has, and every comparison a client makes against these values — is this the amount I asked for, is the lock funded for what was quoted — has to be exact. `String(aBigint)` is already the canonical form for a non-negative value, so encoding needs no formatter and the real work is all on the way in: `evmAmountFromWire` **refuses a JSON number outright rather than coercing one**, at every magnitude, including values that happen to be exactly representable. Accepting those would make the check depend on the value rather than on the encoding.

The **sats** side of both corridors stays a JSON `number`, matching the four BTC corridors and the solver's own schemas. § 2.1 makes those strings too; that is a migration for corridors that already ship, in flight separately, and bundling it here would change five corridors to add one.

## Wire parity is pinned, not assumed

A client that encodes an EVM offer differently from the solver is not a bug that surfaces as an error — it is a swap that quietly never matches. So `test/evmRfq.test.ts` opens with a `describe("solver parity")` block that restates the solver's two `.strict()` request schemas **as data**, at a cited commit (`arkade-os/intent-solver@9751a1c`, `packages/solver-corridors-evm/src/wire/evmPayloads.ts` and `packages/solver-core/src/core/corridorPolicy.ts`), and runs both request builders through them. Restated rather than imported because they live in another repo behind a `zod` this package does not depend on — the same discipline `rfq.ts` already uses for the wire's pair cap, and the test is what turns a drift into a failure.

The **quote** direction is pinned separately, because the two fail differently. A wrong request field name produces a refusal from the solver; a wrong quote field name produces a reader that never finds the field and reports a broken quote against a solver that sent a correct one. So the exact key sets both quote builders emit are pinned too, checked both ways round, and every profile key the solver sends is asserted to be one the reader actually requires.

Three traps it pins, each of which silently produces an unmatched swap:

- **The receive envelope carries no `amount` at all.** Its absence is the rule, not an omission: what the client gives is a token quantity, the envelope's `amount` is a JSON number, so it rides the profile as `evm_amount` instead. The schema is `.strict()`, so an `amount` added for symmetry is an undeclared key and the whole request is refused as `unsupported_payload`.
- **The pair token must be LOWERCASE, the profile addresses need not be.** The solver's corridor regex is `[0-9a-f]` only while its profile address schema is `[0-9a-fA-F]` — so the same EIP-55 checksummed address is legal in one field of a request and refused in another. `evmTokenLeg` accepts either and emits lowercase; the profile keeps what it was given, because flattening a checksum throws away the typo protection it exists for.
- **Both corridors are exact-in.** `amount_side` is pinned to `"from"` on both, since the legs are different assets and exact-out would mean inverting a fetched, rounded, directional rate.

## The send quote's `evm_refund_address`

Added in `d65e6fa3`, paired with [arkade-os/intent-solver#3](https://github.com/arkade-os/intent-solver/pull/3), which publishes the field.

The send quote gave a client five of the six fields of the swap key. `ERC20Swap` stores no per-swap struct - only `mapping(bytes32 => bool)` over `keccak256(abi.encode(preimageHash, amount, tokenAddress, claimAddress, refundAddress, timelock))` - so without `refundAddress` a client cannot compute the key, cannot read `swaps(key)` to prove the solver locked before parting with its preimage, and cannot build the claim call, which takes that address explicitly.

`evm_refund_address` names whoever the CONTRACT refunds. On this leg the solver locks, so it is the SOLVER's - the mirror of the receive leg, where the client locks and sends its own in the request. Reading it as "the client's" on both legs is how it came to be missing.

Required rather than optional here: a send quote without it is one this client can neither verify nor claim, so the quote is the only point where that failure is cheap.

## `assertFundable`

Its `quote` parameter widens to `RfqQuote | EvmRfqQuote`. The gates an EVM quote reaches — `valid_until` and the 90-minute refund headroom — are the right ones for both corridors, and both read a plain `number`.

The widening is also what stops the amounts being read as sats. Across the union `from_amount` and `to_amount` are `number | string`, so no arithmetic compiles against them unsplit; read them through `evmQuoteSats` and `evmQuoteTokenAmount`, which pick the side off the pair rather than taking a direction argument.

### `maxFee` is refused on an EVM quote, not run on one

The max-fee gate merged to master (#838) while this branch was open, and the rebase put the two together: the widening turns its `from_amount - to_amount` arithmetic into five compile errors. That is the bug surfacing, not the bug.

The token leg is a canonical decimal string, and **it does not go NaN** — `*` and `-` coerce it, so the gate compares token base units against sats and answers with conviction. Measured, with the guard removed and the rest of the gate intact:

| quote | `maxFee` | what the gate does |
|---|---|---|
| send `arkade:BTC->ethereum:<token>` | `{ bps: 100, referenceRate: 0.5 }` | **passes** |
| send | `{ sats: 0, referenceRate: 0.5 }` | **passes** |
| receive `ethereum:<token>->arkade:BTC` | `{ bps: 100, referenceRate: 0.5 }` | refuses `fee_too_high` |
| receive | `{ sats: 0, referenceRate: 0.5 }` | refuses `fee_too_high` |

A ceiling of **zero sats** passing a live send quote is the one that matters: the caller has asked for the tightest bound expressible and been given no bound at all. The receive direction fails the other way, refusing a quote that is fine.

So `maxFee` on an EVM quote now throws `fee_gate_unavailable` — the reason code the gate already uses for "this ceiling cannot be computed here" — and says to omit it until the sum can be done in bigint against a rate in token base units per sat. Gating for real needs the token's decimals, which are not in the quote, and 256-bit arithmetic, which a double cannot hold; that belongs with the funding work, not here.

The BTC corridors are untouched: `assertFundable` still gates `{ sats: 500 }` against a 500-sat spread and refuses it at `{ sats: 499 }`.

**Note for #832 (`feat/max-fee-gate`), which is open at time of writing.** Its cross-asset branch already handles EVM pairs correctly at the pair level: `assetOf` on `arkade:BTC->ethereum:0x…` yields `BTC` vs `0x…`, so `sameAsset` is false and `referenceRate` is required, which is right. What it will need after this lands is to stop doing raw arithmetic on `from_amount`/`to_amount` — on an EVM quote one of them is a string, and `from_amount * rate - to_amount` coerces it silently and loses precision above 2^53. Measured, not assumed: applying #832's `rfq.ts` diff onto this head compiles with **five `TS2362`/`TS2363` errors at exactly the cross-asset fee lines**, so whichever of the two lands second cannot build until the amounts are read through `evmQuoteSats`/`evmQuoteTokenAmount`. (The coercion it prevents is not uniformly `NaN` — `Number("249750000000000000000")` is finite, so the wrong answer would be plausible below 2^53 and lossy above it, which is quieter and worse.) This PR does not duplicate that gate; the units it expects are documented on the quote types (`maxFee.referenceRate` is to-units per from-unit, so **token atomic units per sat** on the send leg and **sats per token atomic unit** on the receive leg — the easiest thing here to get wrong by a factor of the token's decimals).

## Reading a quote binds it to the negotiation

`readEvmSendQuote` / `readEvmReceiveQuote` take a required `rfqId` alongside the token. The pair check alone leaves the one case it cannot see: a second swap for the SAME token, whose quote arrives on the same relay subscription with a byte-identical pair. Only the id tells those apart, and taking the wrong one funds another swap's terms. `expectQuote` has always made this check for the four BTC corridors; these readers are reached by callers who may not have gone through it.

Required rather than optional, because nothing depends on these readers yet - so making it required costs nothing now and cannot be forgotten later, and an optional binding is one a caller skips exactly when several negotiations are in flight, which is the only time it matters. It is checked *before* the pair, since a reply to another negotiation may well be for the same market and the pair comparison would otherwise pass it and blame a field that was never wrong.

## A known type hole, documented rather than fixed

`RfqTransport.requestQuote` is typed `Promise<RfqQuote>`, and `RfqQuote` declares both amounts `number`. That is not accurate for an EVM quote. The transport never inspects the amounts, so the **value** is right and only the type is wrong — passing the result straight into `readEvmSendQuote` / `readEvmReceiveQuote` (both take `unknown`) is what makes the two agree again. Widening the transport's return type would make every existing caller narrow, which is a bigger change than this PR should carry; the reader's doc comment says so explicitly, including the concrete failure (`to_amount` types as a number, is a string, and `+` concatenates).

## Test plan

Baseline on `c6190783`, via the repo's own `pnpm run test:unit`: **224 files, 4139 passed, 1 skipped, 0 failing.**
After: **225 files, 4180 passed, 1 skipped, 0 failing** — +1 file, +41 tests, which reconciles exactly with the 41 in `test/evmRfq.test.ts`.

`pnpm -r build`, `pnpm -C packages/swap run typecheck`, `smoke:dist` and `npm pack --dry-run` all clean.

Every behaviour asserted here was mutation-tested — the code under test was broken and the suite confirmed to fail. **34 runtime mutations, all killed:**

| # | mutation | killed by |
|---|---|---|
| 1 | pair not lowercased | parity: lowercase-only corridor regexes |
| 2 | receive request carries an envelope `amount` | parity: strict schema / no envelope amount |
| 3 | `evmAmountFromWire` coerces a JSON number | amounts: refuses a JSON number at every magnitude |
| 4 | `TOKEN_AMOUNT` unanchored | amounts: refuses every non-canonical spelling |
| 5 | send profile key renamed | parity: strict schema |
| 6 | token amount read off the wrong leg | readers: opposite sides per direction |
| 7 | sats read off the wrong leg | readers: opposite sides per direction |
| 8 | negative amount encoded, not refused | amounts: refuses a negative |
| 9 | `min_age_seconds` check dropped | readers: refuses a missing profile field |
| 10 | `min_age_seconds` required positive | readers: accepts zero |
| 11 | send pair built in the receive direction | parity + pair tests |
| 12 | quote `pair` not compared | readers: refuses another token/pair |
| 13 | `refund_locktime` not required | readers: refuses a missing envelope field |
| 14 | `receiver_pk_script` unchecked | readers: refuses a missing profile field |
| 15 | `EVM_ADDRESS` accepts a wrong length | pair + request builder refusals |
| 16 | `solver_refund_pk_script` unchecked | readers: refuses a missing profile field |
| 17 | `evm_claim_address` not address-checked | readers: present-but-malformed address |
| 18 | `lockup_address` not required | readers: refuses a missing profile field |
| 19 | `evm_chain_id` not required | readers: refuses a missing profile field |
| 20 | `min_confirmations` not required | readers: refuses a missing profile field |
| 21 | `evm_contract_address` not address-checked | readers: present-but-malformed address |
| 22 | `assertHex` accepts uppercase and empty | readers: present-but-not-hex pkScript |
| 23 | `payment_hash` not required | readers: refuses a missing profile field |
| 24 | `evm_timeout_block` not required | readers: refuses a missing profile field |
| 25 | `valid_until` not required | readers: refuses a missing envelope field |
| 26 | `rfq_id` not bound to the negotiation | readers: refuses a quote for another negotiation |
| 27 | `solver_pubkey` not required | readers: refuses a missing envelope field |
| 28 | a quote with no `profile` at all tolerated | readers: refuses a missing envelope field |
| 29 | `assertHex` accepts an odd-length string | readers: present-but-not-hex pkScript |
| 30 | zero `evmAmount` accepted by the receive builder | request builders: refuses the inputs the solver would |
| 31 | `payment_hash` only string-checked | readers: refuses a non-32-byte hash |
| 32 | `assertHash32` made length-agnostic | readers: 62- and 66-character hashes |
| 33 | `evm_refund_address` not required on a send quote | readers: refuses a missing profile field |
| 34 | `evm_refund_address` present but not address-checked | readers: present-but-malformed address |
| 35 | `maxFee` guard deleted, EVM amounts coerced through | funding gate: refuses `maxFee` on an EVM quote |
| 36 | guard condition inverted | funding gate: the sats-only ceiling on a BTC quote |
| 37 | guard `\|\|` weakened to `&&` (refuse only if BOTH legs are strings) | funding gate: refuses `maxFee` on an EVM quote |
| 38 | refusal reason renamed to `max_fee_out_of_range` | funding gate: `reason` is `fee_gate_unavailable` |

Mutation 29 comes from review: `assertHex` checked the character class but not the length, so `"abc"` passed a validator whose whole job is to guarantee the value decodes - and `hex.decode` refuses an odd-length string, so the failure only moved to covenant derivation, where it reads as a broken covenant rather than a malformed quote field. Fixed in `91711eef`.

Three others (17, 21, 25) **survived the first round** and are why the second commit exists — 17 and 21 because two address checks were exercised only by absence, so the format check could have been deleted outright and nothing would have noticed on the two values a client later has to talk to a chain with; 25 because `valid_until` was required by the reader and by nothing else, and it fails in the shape that reads as success (`now >= undefined` is false, so a missing deadline does not fail `assertFundable`'s gate — it deletes it).

One further mutation is **type-level only**, so vitest cannot see it: reverting `assertFundable`'s parameter to `RfqQuote` alone. Verified with `tsc --noEmit` against a scratch caller — mutated, it fails with `TS2322: Type 'EvmSendQuote' is not assignable to type 'RfqQuote' … 'string' is not assignable to type 'number'`; unmutated, clean. Worth flagging that the package's `tsconfig.json` excludes `test/`, so nothing in CI typechecks the test tree, and this widening is therefore not gated by CI.

## Summary by CodeRabbit

* **New Features**
  * Added support for EVM RFQ corridors between Arkade BTC and Ethereum tokens.
  * Added creation and validation of send/receive RFQ requests and quotes.
  * Added conversion between token amounts and canonical wire-format values.
  * Added validation for token addresses, quote identifiers, pairs, deadlines, and funding requirements.
  * Exported the EVM RFQ functionality for use across the swap experience.

* **Bug Fixes**
  * Improved funding validation to correctly handle EVM quotes without interpreting token amounts as satoshis.

### DIFF
--- a/packages/swap/src/evmRfq.ts
+++ b/packages/swap/src/evmRfq.ts
@@ -534,9 +534,24 @@ const assertNonNegativeInteger = (value: unknown, field: string): void => {
     }
 };
 
+/**
+ * Lowercase hex of a whole number of bytes.
+ *
+ * The EVEN-LENGTH half is the part that is easy to leave out, and leaving it
+ * out makes this validator promise something it does not keep. Both fields
+ * checked with it are pkScripts a later step decodes to bytes, and
+ * `hex.decode` refuses an odd-length string — so without this, `"abc"` passes
+ * here and throws during covenant derivation instead, reporting a broken
+ * covenant for what is really a malformed quote field.
+ */
 const assertHex = (value: unknown, field: string): void => {
-    if (typeof value !== "string" || value.length === 0 || !/^[0-9a-f]+$/.test(value)) {
-        throw new Error(`${field} must be lowercase hex, got ${String(value)}`);
+    if (
+        typeof value !== "string" ||
+        value.length === 0 ||
+        value.length % 2 !== 0 ||
+        !/^[0-9a-f]+$/.test(value)
+    ) {
+        throw new Error(`${field} must be lowercase hex of whole bytes, got ${String(value)}`);
     }
 };
 
--- a/packages/swap/src/evmRfq.ts
+++ b/packages/swap/src/evmRfq.ts
@@ -1,0 +1,564 @@
+/**
+ * The EVM corridors' wire layer — `arkade:BTC->ethereum:<token>` and
+ * `ethereum:<token>->arkade:BTC`.
+ *
+ * The negotiation half only: the pair form, the token identity, the amount
+ * encoding, the two `rfq_request` shapes and the two quote shapes a solver
+ * answers with. Deriving the Arkade covenant, checking the two deadlines
+ * against each other and funding are NOT here — see the deadline note on
+ * {@link EvmSendQuoteProfile.evm_timeout_block} for why the ordering check
+ * needs a per-chain block cadence this module deliberately does not invent.
+ *
+ * Its own file rather than more of `rfq.ts` because nothing here is shared
+ * with the four BTC corridors: those dispatch on a pair CONSTANT, and an EVM
+ * pair cannot be one — it names its ERC20, and which tokens are served is a
+ * solver deployment's choice, known only at runtime.
+ *
+ * ## The reference solver, and where the parity is pinned
+ *
+ * Every shape below is the client half of a schema that already exists and is
+ * already strict. The source of truth is `arkade-os/intent-solver` at
+ * `9751a1c`:
+ *
+ * - `packages/solver-corridors-evm/src/wire/evmPayloads.ts` — the two
+ *   `.strict()` request schemas and the two quote payload builders.
+ * - `packages/solver-core/src/core/corridorPolicy.ts` — `evmCorridorFor`,
+ *   `evmDirectionOf`, and the LOWERCASE-only pair regex.
+ * - `docs/rfq-protocol.md` §§ 2.1, 7.1.5 — the amount encoding and the EVM
+ *   profiles.
+ *
+ * Restated here rather than imported, for the reason `rfq.ts` gives for
+ * restating the wire's pair cap: this repo does not own those numbers, and a
+ * copy that drifts is caught by a test that pins it (`test/evmRfq.test.ts`)
+ * rather than by a swap that silently never matches.
+ *
+ * ## Amounts
+ *
+ * **A token quantity is a `bigint` in this API and a canonical decimal string
+ * on the wire.** Not a `number`, at either end.
+ *
+ * An ERC20 amount is 256-bit. A JSON number is an IEEE-754 double, exact only
+ * to 2^53 − 1 — which at 18 decimals is 0.009 tokens, so one whole DAI does
+ * not survive the round trip. The rounding would happen inside `JSON.parse`,
+ * before any validator on either side could see it, and neither party could
+ * detect that it had happened. That is why the solver's schema types
+ * `evm_amount` as a string, why its store declares the column TEXT, and why
+ * § 2.1 of the protocol specifies a canonical decimal string for every amount.
+ *
+ * `bigint` at the API boundary because it is the only exact integer JS has,
+ * and every comparison a client makes against these values — is this the
+ * amount I asked for, is the lock funded for what was quoted — has to be
+ * exact. `String(aBigint)` is already the canonical form for a non-negative
+ * value: digits only, no separator, no exponent, no leading zero. So encoding
+ * needs no formatter, and the only real work is on the way IN — see {@link
+ * evmAmountFromWire}, which refuses a JSON number outright.
+ *
+ * The SATS side of both corridors stays a JSON `number`, matching the four BTC
+ * corridors and the solver's own schemas. § 2.1 makes those strings too; that
+ * is a migration for corridors that already ship, in flight separately, and
+ * bundling it here would change five corridors to add one.
+ */
+import { hex } from "@scure/base";
+
+import { ARKADE_BTC, assertPairLength, rfqPair } from "./rfq";
+
+// ── Pairs and token identity ────────────────────────────────────────────────
+
+/** The EVM leg's chain namespace. One value: the corridor is EVM-shaped, and
+ * WHICH chain it runs on is named by the quote's `evm_chain_id`, not by the
+ * pair. A second namespace here would be a second market key for one corridor. */
+export const EVM_CHAIN = "ethereum";
+
+/** `0x` then 40 hex, either case — what an EIP-55 checksummed address looks
+ * like, and what the solver's profile schema accepts for `evm_claim_address`
+ * and `evm_refund_address`. */
+const EVM_ADDRESS = /^0x[0-9a-fA-F]{40}$/;
+
+/**
+ * Atomic units as a canonical decimal string: digits only, no sign, no point,
+ * no exponent, no leading zero. The solver's `TOKEN_AMOUNT`, restated.
+ *
+ * Anchored, so `1e18` is refused rather than partially matched. Three
+ * spellings of exponent notation exist, and quietly reading one wrong
+ * misprices by eighteen orders of magnitude.
+ */
+const TOKEN_AMOUNT = /^(0|[1-9][0-9]*)$/;
+
+/** The two directional pair spellings, LOWERCASE token only — the solver's
+ * `evmDirectionOf`, restated. */
+const SEND_PAIR = /^arkade:BTC->ethereum:0x[0-9a-f]{40}$/;
+const RECEIVE_PAIR = /^ethereum:0x[0-9a-f]{40}->arkade:BTC$/;
+
+/**
+ * The `ethereum:<token>` leg for an ERC20, LOWERCASED.
+ *
+ * Accepts a checksummed address and emits a lowercase one, deliberately. The
+ * solver's pair regex is lowercase-only while its profile address schema is
+ * case-insensitive, so the same address is legal in one field of a request and
+ * refused in another — and a client that normalised in only one place would
+ * build a request whose profile parses and whose pair is `unsupported_pair`.
+ *
+ * The same rule, and the same reason, as `arkadeAssetLeg` for asset ids:
+ * solvers compare pair strings byte for byte, so one spelling normalised in
+ * one layer and not another derives the right market key and is then skipped
+ * as an unserved pair.
+ */
+export const evmTokenLeg = (tokenAddress: string): string =>
+    `${EVM_CHAIN}:${assertEvmAddress(tokenAddress, "token address").toLowerCase()}`;
+
+/** `arkade:BTC->ethereum:<token>` — the client locks sats, the solver pays
+ * tokens. */
+export const evmSendPair = (tokenAddress: string): string => {
+    const pair = rfqPair(ARKADE_BTC, evmTokenLeg(tokenAddress));
+    assertPairLength(pair);
+    return pair;
+};
+
+/** `ethereum:<token>->arkade:BTC` — the client locks tokens, the solver pays
+ * sats. */
+export const evmReceivePair = (tokenAddress: string): string => {
+    const pair = rfqPair(evmTokenLeg(tokenAddress), ARKADE_BTC);
+    assertPairLength(pair);
+    return pair;
+};
+
+/**
+ * Which EVM direction a pair names, or `null` when it names neither.
+ *
+ * Mirrors the solver's `evmDirectionOf`, and exists for the same reason: an
+ * EVM pair is not a constant, so nothing can dispatch on it by equality. A
+ * client holding a stored swap or an inbound status has only the pair string
+ * to decide what it is looking at.
+ */
+export const evmDirectionOf = (pair: string): "send" | "receive" | null => {
+    if (SEND_PAIR.test(pair)) return "send";
+    if (RECEIVE_PAIR.test(pair)) return "receive";
+    return null;
+};
+
+/** The lowercase ERC20 address an EVM pair names, or `null` when the pair is
+ * not an EVM one. */
+export const evmTokenOf = (pair: string): string | null => {
+    const direction = evmDirectionOf(pair);
+    if (direction === null) return null;
+    const leg =
+        direction === "send"
+            ? pair.slice(pair.indexOf("->") + 2)
+            : pair.slice(0, pair.indexOf("->"));
+    return leg.slice(EVM_CHAIN.length + 1);
+};
+
+// ── Amounts ─────────────────────────────────────────────────────────────────
+
+/**
+ * A token quantity in the wire's canonical decimal form.
+ *
+ * Refuses a negative value rather than emitting `-1`: the solver's schema is
+ * anchored, so it would answer `unsupported_payload`, which says nothing about
+ * which field was wrong.
+ */
+export const evmAmountToWire = (units: bigint): string => {
+    if (units < 0n) throw new Error(`token amount may not be negative, got ${units}`);
+    return units.toString();
+};
+
+/**
+ * A token quantity off the wire, as an exact `bigint`.
+ *
+ * **A JSON number is refused, never coerced.** This is the whole point of the
+ * string encoding: by the time a number is sitting in one of these fields,
+ * `JSON.parse` has already rounded it to the nearest double and the original
+ * value is gone. Accepting it — via `BigInt(Math.trunc(v))`, or the implicit
+ * coercion any arithmetic on it would do — turns a detectable protocol
+ * violation into a silently wrong amount, which is exactly the failure the
+ * encoding exists to make impossible.
+ *
+ * Non-canonical strings are refused for the same reason: `"0x10"`, `"1e18"`,
+ * `"01"` and `" 1"` all have a plausible reading and at least one wrong one.
+ */
+export const evmAmountFromWire = (value: unknown, field: string): bigint => {
+    if (typeof value !== "string") {
+        throw new Error(
+            `${field} must be a decimal string of atomic units, not a JSON ${typeof value} — ` +
+                `a number here has already been rounded by the parser`,
+        );
+    }
+    if (!TOKEN_AMOUNT.test(value)) {
+        throw new Error(`${field} is not a canonical decimal amount: ${JSON.stringify(value)}`);
+    }
+    return BigInt(value);
+};
+
+// ── Requests ────────────────────────────────────────────────────────────────
+
+/**
+ * The `rfq_request` for `arkade:BTC->ethereum:<token>`.
+ *
+ * **Exact-in only.** `amount_side` is pinned to `"from"` and `amount` is sats:
+ * the `to` leg is a different asset, so exact-out would mean inverting a
+ * fetched, rounded, directional rate. The solver's schema pins the literal, so
+ * an exact-out request is refused rather than served at a worse price.
+ *
+ * The client funds the Arkade covenant FIRST on this corridor, and funding is
+ * acceptance — so it holds that side's refund role, and `client_refund_pubkey`
+ * is its own x-only key for the covenant's client-side refund leaves. Same
+ * role and same field name as on the Lightning and onchain send legs.
+ *
+ * `evmClaimAddress` is the client's own EVM address, and the only party
+ * `ERC20Swap.claim` will pay. Sent as given: the solver's schema is
+ * case-insensitive here, and flattening an EIP-55 checksum would throw away
+ * the typo protection that spelling exists for.
+ */
+export const evmSendRequest = (input: {
+    rfqId: string;
+    /** The ERC20 being bought. Normalised into the pair — see {@link evmTokenLeg}. */
+    tokenAddress: string;
+    /** `sha256(P)`, hex — client-chosen; see `paymentHashOf`. */
+    paymentHash: string;
+    /** Where the CLIENT takes the tokens. */
+    evmClaimAddress: string;
+    /** The client's Arkade address — where the covenant refund must pay. */
+    refundAddress: string;
+    /** The client's own x-only key for the covenant's refund leaves. */
+    senderPubkey: Uint8Array;
+    /** Sats the client locks. Exact-in. */
+    amountSats: number;
+}): Record<string, unknown> => {
+    assertPositiveInteger(input.amountSats, "amountSats");
+    return {
+        v: 1,
+        type: "rfq_request",
+        rfq_id: input.rfqId,
+        pair: evmSendPair(input.tokenAddress),
+        amount_side: "from",
+        amount: input.amountSats,
+        profile: {
+            payment_hash: input.paymentHash,
+            evm_claim_address: assertEvmAddress(input.evmClaimAddress, "evmClaimAddress"),
+            refund_address: input.refundAddress,
+            client_refund_pubkey: hex.encode(input.senderPubkey),
+        },
+    };
+};
+
+/**
+ * The `rfq_request` for `ethereum:<token>->arkade:BTC`.
+ *
+ * **The envelope carries no `amount` at all**, and its absence is load-bearing
+ * rather than an omission. What the client gives is a token quantity; the
+ * envelope's `amount` is a JSON number, so it cannot hold one. `evm_amount`
+ * carries it in the profile as a decimal string instead. The solver's schema
+ * is `.strict()`, so an `amount` added "for symmetry" is not ignored — it is
+ * an undeclared key, and the whole request is refused as `unsupported_payload`.
+ *
+ * `amount_side` stays `"from"`: the client still names what it gives.
+ *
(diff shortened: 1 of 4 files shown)


Kukks:
@coderabbitai review

Kukks:
Thanks — the `assertHex` finding is a real bug and is fixed in `91711eef`. On the first one I think the review read a stale commit; evidence below.

## 1. `rfq_id` binding — already shipped, in `89f04f48`

This review landed at 22:35:44Z against `c5cdb773`, the **first of what were then three commits**. The binding it asks for shipped in `89f04f48`, pushed at 22:07:04Z — 28 minutes before the review was submitted.

The tell is in the review itself: it says "The test file (632 lines)". 632 is exactly the length of `test/evmRfq.test.ts` at `c5cdb773`. At `77e00faa` it was 815, and at `89f04f48` it was 853.

On `89f04f48`:

```
$ git show 89f04f48:packages/swap/src/evmRfq.ts | grep -n "expected: { tokenAddress"
448:    expected: { tokenAddress: string; rfqId: string },
470:    expected: { tokenAddress: string; rfqId: string },

$ git show 89f04f48:packages/swap/src/evmRfq.ts | grep -n "rfq_id !== expectedRfqId"
563:    if (quote.rfq_id !== expectedRfqId) {

$ git show 89f04f48:packages/swap/test/evmRfq.test.ts | grep -n "another negotiation"
625:    it("refuses a quote for another negotiation, even on the same market", () => {
```

The implementation matches what the review prescribes, including the ordering rationale — the id is compared *before* the pair, precisely because a reply to another negotiation may carry a byte-identical pair, so blaming the pair would name a field that was never wrong. Mutation 26 removes that comparison and the new test fails, which is what the table entry records.

The impact analysis was right, though, and it is why the check is there: two concurrent USDC buys on one relay subscription are indistinguishable by pair alone. The disagreement is only about whether it had already landed.

## 2. `assertHex` accepts odd-length strings — correct, fixed

Confirmed before fixing rather than taken on faith:

```
$ node -e "const {hex}=require('@scure/base');
           console.log(/^[0-9a-f]+$/.test('abc'));
           try { hex.decode('abc') } catch (e) { console.log(e.message) }"
true
hex.decode: expected string, got string with length 3
```

So the validator passed a value that does not survive the decode it exists to protect — the failure just moved one layer away, surfacing at covenant derivation as a broken covenant rather than here as a malformed quote field. Exactly as described.

Fixed in `91711eef` by adding `value.length % 2 !== 0`, with `"abc"` and a 65-nibble pkScript added to the negative cases. The message also changed from "must be lowercase hex" to "must be lowercase hex of whole bytes", since the old wording was true of the value it had just rejected.

Mutation 29 (remove the even-length term alone) now fails `refuses a pkScript that is present but not hex`; before this commit it did not.

## 3. `evm_timeout_block` unbounded above — leaving it, deliberately

Agreed on the facts, and I'd rather not add the bound here.

A ceiling on this value is not a shape check, it is a policy one, and the only defensible number comes from the ordering invariant this PR explicitly defers: `refund_locktime + margin <= evm_timeout_block`, across a block-height/unix-seconds boundary that needs a per-chain cadence. Any constant I picked now would be arbitrary, and — worse — it would look like the deadline gate while not being it. That is the failure mode the `maxFee` gate's own comment warns about: a caller who sees a bound and infers a protection that was never there.

The value is also the client's own choice on both directions where it appears, and over-setting it strands only the client's own tokens. So the bound belongs with the cadence work, and until then no bound is more honest than a decorative one.

---

Gate on `91711eef`: `pnpm run test:unit` → 225 files, 4178 passed, 1 skipped, 0 failing (baseline `c6190783` was 224 / 4139 / 1 / 0). 29 mutations, all killed.

Unrelated to this diff: `Integration (ts-sdk / ark-core)` failed on `89f04f48` and I do not believe it is attributable here — `packages/ts-sdk` declares no dependency on `@arkade-os/swap` in either `dependencies` or `devDependencies` and `test/e2e/ark.test.ts` never references it, the same job passed on `c5cdb773` and `77e00faa`, and the job log shows the stack never came up (`arkd not ready: dns: A record lookup error … server misbehaving`, `Boltz … shutting down with code: 1`) with the two assertions failing as `expected 0 to be greater than 0` downstream of a wallet that was never funded. Re-run requested to confirm rather than argue.

Kukks:
Thanks for the re-review, and for confirming the stale read on the first pass.

Took the non-blocking suggestion: `915cca48` asserts the missing-profile refusal on `readEvmReceiveQuote` too. You're right that both routes share `readEvmQuoteEnvelope` so the behaviour was already guaranteed — but which checks are shared is an implementation fact, and a test covering one direction stops covering the other the moment they stop sharing it. One assertion, so no reason not to.

Gate on `915cca48`: `pnpm run test:unit` → 225 files, 4178 passed, 1 skipped, 0 failing (baseline `c6190783`: 224 / 4139 / 1 / 0).

One process note for whoever merges: `reviewDecision` is still `CHANGES_REQUESTED` from the 22:35Z pass. A follow-up `COMMENTED` review does not supersede a changes-requested state, so despite "ready from my side" the PR stays formally blocked until that review is dismissed or a new approving one lands. Flagging rather than assuming it resolves itself.

Kukks:
@coderabbitai review

Kukks:
CI is fully green on `915cca48` — all 7 checks.

Two integration jobs each failed once and then passed on re-run with **no code change**, so the run history looks worse than it is:

| job | first attempt | re-run |
|---|---|---|
| `Integration (ts-sdk / ark-core)` | failed on `89f04f48` | SUCCESS |
| `Integration (ts-sdk / exit-providers-rotation)` | failed on `915cca48` | SUCCESS |

Both were the regtest stack failing to come up, not this diff:

- `ark-core`: `arkd not ready: dns: A record lookup error … server misbehaving`, `Boltz … shutting down with code: 1`; the two assertions failed as `expected 0 to be greater than 0`, downstream of a wallet that was never funded.
- `exit-providers-rotation`: `dial tcp 172.18.0.19:6060: connect: connection refused`, `ark service not ready: wallet is locked or syncing`.

Structurally neither can reach this PR: `packages/ts-sdk` declares no dependency on `@arkade-os/swap` in `dependencies` or `devDependencies`, and neither failing e2e file references it. The whole workflow also went green end-to-end on `91711eef`, which carries all the substantive code — `915cca48` only adds one assertion to a swap test.

Review state, for whoever merges: `reviewDecision` is still `CHANGES_REQUESTED` from the 22:35Z pass, whose blocking finding was a stale read of `c5cdb773` (confirmed by @arkana-ai-bot's follow-up). A `COMMENTED` re-review does not clear that state, so it needs an explicit dismissal or a fresh approving review. CodeRabbit reviewed `c6190783..915cca48` and generated no actionable comments.

Kukks:
All three advisories taken, in `2caeb417`. One correction on the #832 forward note — I tested it rather than reasoning about it, and it goes the other way.

## Correction: the union widening *does* block #832's compile

> The union type widening in `assertFundable`'s parameter won't block the compile there because #832's arithmetic is inside a conditional already typed `RfqQuote`.

It isn't. `sameAsset` is computed from `input.quote.pair` via a `split("->")`, which narrows nothing — `input.quote` stays `RfqQuote | EvmRfqQuote` through the whole block, so both amount fields are `number | string` at every arithmetic site.

Checked by applying #832's `rfq.ts` diff onto this head and compiling:

```
$ git apply -3 <832-rfq.patch>
Applied patch to 'packages/swap/src/rfq.ts' cleanly.

$ npx tsc --noEmit -p packages/swap/tsconfig.json
src/rfq.ts(565,15): error TS2362: The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.
src/rfq.ts(565,41): error TS2363: The right-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.
src/rfq.ts(567,20): error TS2362: ...
src/rfq.ts(567,74): error TS2363: ...
src/rfq.ts(572,25): error TS2362: ...
```

Five errors, at exactly the cross-asset fee lines. So the guard behaves as the PR description claims: whichever of the two lands second cannot compile until the amounts are read through `evmQuoteSats` / `evmQuoteTokenAmount`. That is the outcome I wanted — a merge conflict a compiler reports, rather than a silent coercion.

(Worth adding to your note: the coercion would not be uniformly `NaN`. `Number("249750000000000000000")` is finite, so `from_amount * rate - to_amount` yields a *plausible* wrong number for values inside double range and loses precision above 2^53 — quieter than `NaN`, and worse.)

## 1. `evmAmountToWire(0n)` accepted — fixed

Guard added to `evmReceiveRequest`, not to the codec. `"0"` is canonical per § 2.1 and `evmAmountToWire`/`evmAmountFromWire` have to keep emitting and accepting it, so the corridor's semantics belong in the builder. It covers negatives too — "not a positive amount" is one rule, not two — so a `-1n` now reports the builder's message; the codec keeps its own negative guard for direct callers and is tested separately.

## 2. `payment_hash` validated only as a string — fixed

Agreed, and the asymmetry you pointed at was the argument: pkScripts go through `assertHex`, both addresses through `assertEvmAddress`, and this one through `asString`.

Used a dedicated `assertHash32` (exactly 64 lowercase hex) rather than reusing `assertHex`, because a hash of the wrong *size* is a different value rather than a truncated spelling of the right one — `assertHex` would have accepted a 62-character hash. Tested against `""`, non-hex, 62 chars, 66 chars, uppercase, and a number, on both readers.

## 3. Discarded return values — commented

Both sites now say they are called for the throw, and that the value is read later off the returned quote via `evmQuoteTokenAmount`. Added the reason too: returning it there would hand callers a second, divergeable copy of a number the quote already carries.

## Mutation coverage

32 now, all killed. The three new ones are the checks above, and none was free:

| # | mutation | killed by |
|---|---|---|
| 30 | zero `evmAmount` accepted | request builders: refuses the inputs the solver would |
| 31 | `payment_hash` only string-checked | readers: refuses a non-32-byte hash |
| 32 | `assertHash32` made length-agnostic | readers: 62- and 66-character hashes |

Gate on `2caeb417`: `pnpm run test:unit` → 225 files, 4179 passed, 1 skipped, 0 failing (baseline `c6190783`: 224 / 4139 / 1 / 0). Build and typecheck clean.

Kukks:
Rebased onto `master` (`45a5369`, 14 commits ahead of the old base) and pushed `6d8296f7`.

The rebase was not clean in the way that matters. The max-fee gate landed on master in #838 while this branch was open, and it turns out the two collide — which is worth reading, because the collision **is** the union widening doing its job.

## The rebase turned five lines into compile errors

```
src/rfq.ts(524,15): error TS2362: The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.
src/rfq.ts(524,41): error TS2363: ...
src/rfq.ts(526,20): error TS2362: ...
src/rfq.ts(526,74): error TS2363: ...
src/rfq.ts(531,25): error TS2362: ...
```

All five are `assertFundable`'s fee arithmetic — `from_amount - to_amount`, `from_amount * referenceRate - to_amount`, `from_amount * bps / 10_000` — now reading `number | string` across the union.

## Correcting my own note above, and arkana's

I wrote in the #832 correction that the coercion "goes the other way" but described it as producing `NaN`. That was wrong, and I only found out by running it rather than reasoning about it. **The token leg does not go NaN.** It is a canonical decimal string, so `*` and `-` coerce it to a plain number, and the gate compares token base units against sats and answers confidently.

I removed the guard and measured what the rest of the gate then does:

| quote | `maxFee` | outcome |
|---|---|---|
| send `arkade:BTC->ethereum:<token>` | `{ bps: 100, referenceRate: 0.5 }` | **passed** |
| send | `{ sats: 0, referenceRate: 0.5 }` | **passed** |
| receive `ethereum:<token>->arkade:BTC` | `{ bps: 100, referenceRate: 0.5 }` | refused `fee_too_high` |
| receive | `{ sats: 0, referenceRate: 0.5 }` | refused `fee_too_high` |

A ceiling of **zero sats** passing a live send quote is the one that matters. The caller asked for the tightest bound the API can express and got no bound. The receive direction is wrong in the other direction — it refuses a quote that is fine.

Both are worse than NaN would have been, because NaN at least tends to surface. This one produces a plausible number.

## The fix

`maxFee` on an EVM quote now throws `fee_gate_unavailable` — the reason the gate already uses for "this ceiling cannot be computed here" — rather than running. Gating for real needs the token's decimals (not in the quote) and 256-bit arithmetic (a double cannot hold it); that belongs with the funding work.

The BTC corridors are unchanged, and there is a test pinning that: `{ sats: 500 }` still passes a 500-sat spread and `{ sats: 499 }` still refuses it.

## Mutations on the new guard

| # | mutation | killed by |
|---|---|---|
| 35 | guard deleted, amounts coerced through | `expected 'Error: maxFee cannot gate arkade:BTC-…' to match /EVM token leg/` |
| 36 | condition inverted | `expected [Function] to not throw an error but 'Error: maxFee cannot gate arkade:BTC-…' was thrown` — the BTC ceiling test |
| 37 | `\|\|` weakened to `&&` | `expected 'Error: …' to match /EVM token leg/`, both directions |
| 38 | reason renamed to `max_fee_out_of_range` | `expected 'max_fee_out_of_range' to be 'fee_gate_unavailable'` |

## Gate

`pnpm -r build` exit 0, then `pnpm run test:unit`:

| | baseline `45a5369` | this head `6d8296f7` |
|---|---|---|
| test files | 232 | 233 (+1, `evmRfq.test.ts`) |
| tests | 4365 passed, 1 skipped | 4409 passed, 1 skipped |
| failing | 0 | 0 |

`swap` goes 32 files / 780 tests → 33 / 824, so nothing was overwritten.

## On the review state

Dismissing the `CHANGES_REQUESTED` from the 22:35Z pass. Not because I disagree — the `assertHex` half of it was a real bug — but because both of its findings shipped before or shortly after it was submitted, and arkana has since confirmed that twice at two different heads, most recently at `d65e6fa3`: *"All prior blocking and advisory findings are closed… This is ready from my side."* A `COMMENTED` re-review cannot clear a changes-requested state, so the flag would otherwise sit there indefinitely against findings its own author has closed.

Kukks:
@coderabbitai review

Kukks:
@coderabbitai review

PR #827, opened 2026-09-03
Rename ts-sdk package to sdk

tiero:
## Summary

Rename the `packages/ts-sdk` directory to `packages/sdk` throughout the repository. This simplifies the package name while maintaining all functionality and API compatibility.

## Changes

- **Directory rename**: `packages/ts-sdk/` → `packages/sdk/`
- **Documentation updates**: Updated all references in README.md, CONTRIBUTING.md, and AGENTS.md to reflect the new package path
- **CI/CD pipeline**: Updated GitHub Actions workflows (.github/workflows/ci.yml, .github/workflows/tsdoc.yml) to reference the new package location
- **NPM scripts**: Updated package.json scripts and test commands to use the new `:sdk` suffix instead of `:ts-sdk`
- **Internal references**: Updated comments and documentation strings within source files to reference the new package location

## Implementation Details

- All file contents remain unchanged; this is purely a structural reorganization
- The package name in package.json remains `@arkade-os/sdk` (unchanged)
- All integration test references updated to use the new path
- Build, type-check, smoke-test, and publish verification steps all updated to reference `packages/sdk`

https://claude.ai/code/session_017TcYe3EsMM5q9dQrM7sU5x

PR #837, opened 2026-09-04
feat(swap): solver-routed send rails, and the ChainSource/MTP gaps they needed

Kukks:
`PaymentRouter` already does route selection, bounds matching, and
fallback-on-failure. The Arkade wallet's solver-routed onchain send
([wallet#941](https://github.com/arkade-os/wallet/pull/941)) reimplements all
three app-side, in 2,603 lines. Left there, the next client reimplements them
again and the two disagree about when to fall back.

This is the corridor as rails, plus the two things the rails needed first.

## The bug

`requestOnchainSend` never compared the quote it got back to the trade it
asked for. It returns `fundAmount: quote.from_amount` verbatim, so a solver
naming a different amount was funded at the solver's number — silently.

`assertQuotedAmount` has always made exactly this check. It was applied on
both receive legs and on neither send leg; `requestLightningSend` only escapes
because a BOLT11 profile carries no `amountSide` and it makes the same two
comparisons against the invoice instead. The helper moves above both
corridors, where it belongs, and is called.

Three fixtures were quoting `to_amount: 99_000` for an `amountSide: "to",
amount: 100_000` request. Eleven tests rode that and none of them noticed;
they now quote the trade. wallet#941 patches this client-side, and every other
client would have had to.

## `chainSourceFrom(provider, network)`

`ChainSource` shipped with no implementation, so the wallet hand-wrote 222
lines of Esplora client that `EsploraProvider` already covers. This adapts the
SDK's own `OnchainProvider` instead of introducing a second onchain interface
— it is the object a wallet already builds, it has two implementations, and
`ESPLORA_URL` / `ELECTRUM_WS_URL` already resolve a network to an endpoint.

Two mappings are not one-to-one, and both fail as a claim window that quietly
closed rather than as an error:

- **Outputs are looked up by address.** `getScriptUtxos` gets a `pkScript` and
  the provider is address-keyed. Esplora does expose `/scripthash/:hash/utxo`,
  but that hash is byte-reversed by convention and the wrong endianness
  returns an empty list rather than an error — indistinguishable from "not
  funded yet". Decoding costs one pure call and cannot fail silently.
- **Depth counts a tip-height confirmation as one**, and an unconfirmed output
  as zero rather than one: that number is what `minConfirmations` is compared
  against, and calling a mempool output "1 deep" lets a 1-confirmation policy
  claim against a transaction that can still be replaced.

I considered adding a script-keyed `getCoinsByScript` to `OnchainProvider`
instead. Both in-tree backends are natively script-keyed, so it would fit —
but as a required method it breaks every external implementer, and as an
optional one it adds a second code path with no user-visible benefit over a
decode that cannot fail quietly. Saying so here because it is a real fork and
the adapter is trivially swappable if the interface method is preferred.

## MTP is not the tip's timestamp

`OnchainProvider.getChainTip().time` is already read as median-time-past
everywhere it is consumed — the wallet's `ContractManagerConfig.chainTip`
("what seconds-typed timelocks are judged against"), and the exit executor's
maturity test. That is also what BIP-113 evaluates a seconds-typed
`OP_CHECKLOCKTIMEVERIFY` against.

`EsploraProvider` honoured it, returning esplora's `mediantime`.
`ElectrumOnchainProvider` returned the tip header's own `nTime` — a different
number, which can sit up to two hours ahead of MTP and matures a timelock
before the network does. It does not surface as a visible bug; it surfaces as
a broadcast the node rejects.

Electrum serves raw headers and no MTP, so it now computes the median of the
eleven-block window itself, cached per tip hash (`getChainTip()` is on the
balance-read path, and ten header fetches per call would be a round trip per
coin listing) and falling back to the tip's own time rather than throwing when
the window cannot be read. The interface's `@remarks` now say what `time` is.

## The rails

**`solverOnchainRail`** (`arkade:BTC -> onchain:BTC`) and
**`solverLightningRail`** (`arkade:BTC -> lightning:BTC`).

wallet#941 grew a six-case refusal enum and an `OnchainRouteUnavailable` error
to express falling back to the collaborative exit. The router's `available()`
contract already *is* that: a rail returning false is dropped from `options()`
and never takes the router with it. So every refusal here is a `false` — no
card, a card that serves the pair but not the SIZE, a destination the claim
could not pay to, an invoice that is amountless or expired. Discovery failing
is left to propagate, because the router catches it, warns, and drops the
rail, which reaches the same place more loudly.

What stays in the app is the user-facing half: the copy explaining why an exit
was collaborative. `options()` names the rails that survived, which is the
fact that copy is about.

They live in `@arkade-os/swap` and are registered by the app, not shipped in
`ts-sdk/payment/rails`: the dependency runs one way, `PaymentRail` is
structural, and putting them in core would make a cycle. Dependencies are
taken at construction rather than read off `RouterContext.swaps`, which
boltz-swap's client already occupies.

Two orderings are load-bearing:

- **`quote()` negotiates.** An RFQ quote is a real quote with a `valid_until`,
  and `PaymentOption.quote()` is lazy for exactly this reason — listing
  options costs nothing, asking for a price costs a negotiation.
- **`send()` persists before it funds**, and refuses to fund if the write
  throws. A funded lockup with no record cannot be refunded.

Amounts stay receiver-exact. The onchain rail asks `amountSide: "to"`, so the
user's number is the L1 payout and the corridor's fee sits on top — which the
`assertQuotedAmount` fix then pins. The claim transaction's own fee still
comes out of that output hours later at a rate nobody knows now;
`meta.claimFeeDeductedFromPayout` says so rather than dressing a guess as a
total. On Lightning the invoice fixes the amount, so a `req.amount` that
contradicts it is refused rather than ignored.

Driving a swap after funding — watching the fill, claiming it, refunding a
stranded lockup — stays `RfqSwapManager`'s job. Wire `awaitSettlement` for a
handle that reaches `"settled"`; without one it stops at `"sent"`, because the
recipient has nothing until the corridor completes and reporting `"settled"`
on a funded lockup would be a lie.

## Not in this PR

The asset dimension. `PaymentRequest.amount` and `RouteQuote.amount`/`fee`/
`total` are bare `number`s with no unit, so the router can say "500" but never
"500 USDX". Asset payments and cross-asset-via-corridor both need that, and it
changes a merged public API — it belongs in its own PR with its own argument.

`@arkade-os/boltz-swap`'s `lightning` rail is untouched and stays registered:
both match the same invoices, `options()` surfaces both, and the priority is a
preference rather than a restriction.

## Interaction with open PRs

- **#637** (`/v1/blocks` fallback) edits the `baseFetch` line in
  `getChainTip()`; this edits the `return` a few lines below and does not
  change which field is read. Compatible either order.
- **#830** (leaked watchers) reshapes a wide span of `EsploraProvider`
  including around `getChainTip()`. A textual conflict is likely; it does not
  change what `getChainTip` returns, so resolution is mechanical.
- **#834** (`/utxo` probe) touches `watchAddresses` only — no overlap.
- **#833** (baseline anchor) reads `getChainTip().height`, and its fixture
  already supplies `mediantime`. Agrees with the clarified contract.
- **#827** (`ts-sdk` → `sdk` rename) will conflict as a path move on every
  file here; nothing semantic.
- **#835** and **#832** touch `packages/swap/src/rfq.ts` at lines 85/428 and
  439/454 respectively; this touches ~1370 and ~1424. No overlap. #835 also
  edits `packages/swap/src/index.ts` at line 130 against this PR's ~154.

## Review round 1 (`2af2efef`)

Two findings taken, one part of one declined — see [the reply](https://github.com/arkade-os/ts-sdk/pull/837#issuecomment-5536514421).

- **Re-gate before funding.** `assertFundable` ran only while quoting, and
  `RouteQuote.send()` is a separate user action minutes later. Both rails now
  re-run it at the top of `send()`, before `persist` and before the wallet
  spends: a quote that lapsed, an invoice that expired, or an L1 claim window
  that stopped being safe all buy a swap that can only refund.
- **Declined: gating on both sides of `persist` with rollback.** A persisted
  record for an unfunded swap is the safe failure — the manager retires it. A
  funded lockup with no record is the unrecoverable one. Moving the gate after
  the write would create the case the ordering exists to prevent.
- **Emulator pin held to shape.** A 33-byte compressed key encodes to 66 hex
  and was adopted verbatim for a card advertising none, reaching `connect` as
  an `emulatorPubkey` the type documents as x-only. `solverRendezvous` now
  applies the same `XONLY_HEX` check to the pin.
- **`SolverOnchainRendezvous` deprecated on arrival** (arkana): the alias is
  gone rather than un-deprecated — the shape turned out corridor-independent
  once the Lightning rail landed, and nothing has shipped that could depend on
  it.
- **`payoutPkScript` and JSON** (arkana): a `Uint8Array` that `JSON.stringify`
  turns into `{"0":81,…}`, coming back as an object with no `.length` any
  script builder accepts. `persist` is injected, so the type now says a JSON
  store must `hex.encode`/`hex.decode` around it.
- **The MTP e2e proved less than it claimed** (CodeRabbit) — see the test plan.

## Test plan

- `pnpm -r build && pnpm run test:unit`
  - baseline (`c6190783`): ts-sdk 174 files / 2886 passed + 1 skipped, swap 27
    / 637, boltz-swap 23 / 616 — 4139 passing, zero failures.
  - after: ts-sdk 174 / 2893 (+7), swap 31 / 719 (+82), boltz-swap 23 / 616
    (unchanged) — 4228 passing, zero failures. The +89 reconciles exactly:
    +7 in the Electrum suite, and +82 across four new swap files
    (`onchainSendQuoteGate` 5, `chainSource` 15, `payment/solverOnchain` 37,
    `payment/solverLightning` 25).
  - all six integration jobs passed on `cb5da060`; `11b5c757` adds one e2e
    case to `exit-providers-rotation`, which already enumerates
    `test/e2e/electrum.test.ts`.
- Every added behaviour was mutation-tested; each mutation was killed:
  - removing `assertQuotedAmount` from `requestOnchainSend` — 4 of 5 fail
  - `getChainTip` returning the tip header's `nTime` instead of MTP — 5 fail
  - dropping the `+1` from the depth arithmetic — 2 fail
  - dropping the `Math.max(0, …)` depth clamp — 1 fail
  - `getMtp` reading the host clock — 2 fail
  - `getSpendingTx` reading outspend 0 instead of the requested vout — 2 fail
  - encoding the lookup address for mainnet regardless of network — 1 fail
  - the rendezvous ignoring the card's size bounds — 3 fail
  - `send()` funding before persisting — 2 fail
  - the onchain rail quoting `amountSide: "from"` — 1 fail
  - `available()` swallowing a discovery failure and claiming to fit — 1 fail
  - `available()` skipping the destination-script pre-check — 1 fail
  - routing an expired invoice — 1 fail
  - routing an amountless invoice — 1 fail
  - ignoring a request amount the invoice contradicts — 1 fail
  - the rendezvous matching on the base corridor rather than the payout side —
    2 fail
  - the rendezvous reading the base side's bounds — 11 fail
  - the emulator pin adopted without an x-only shape check — 1 fail
  - `send()` skipping the re-gate, onchain — 3 fail; lightning — 2 fail
  - the lightning re-gate forgetting the invoice expiry — 1 fail
- The Electrum MTP path is covered against a real Fulcrum (`11b5c757`,
  corrected by `3b717063`). The pre-existing e2e asserted only `tip.time > 0`,
  which the fallback also satisfies, so the new case asserts the fallback's
  warning did NOT fire on a fresh provider.
  **Correction:** as first written that was still not enough — below height 10
  the short-window branch fetches fewer than ten headers (none at height 0),
  warns about nothing, and passes both assertions without exercising the batch
  at all. CodeRabbit caught it. The test now mines up to a full window first
  (with two blocks of headroom for a subscription that observes the mine a
  block late) and asserts the height it read, so the ten-header batch is what
  answered. The earlier green run demonstrated the path runs; it did not
  demonstrate the full window, and that only holds from `3b717063` onward.
- The claim I would most expect to be wrong is now the solver rails' `connect`
  seam: no integration test drives a real RFQ negotiation through a rail, so
  whether `deps.connect` is the right shape for a Nostr transport's lifetime
  is argued rather than demonstrated.


Kukks:
Fixed in `2af2efef`, with one part declined.

**Both deadline findings: taken.** The window is real and I had not seen it — `assertFundable` runs while quoting, and `RouteQuote.send()` is a separate user action that can be minutes later. Both rails now re-run the gate at the top of `send()`. On the onchain leg `htlcParams.refundLocktime` is the same number `deriveOnchainSend` gave the quote-time gate, so they cannot disagree. Six tests added across the two rails (quote expired, invoice expired between quote and send, claim window too short, inverted timelock order), and removing either gate fails 2–3 of them.

**Declined: gating on both sides of `persist` with rollback.** The meaningful gap is between `quote()` and `send()`, not the microseconds between persisting and funding — nothing observable changes in between, so the second check cannot fail when the first passed.

More importantly it would invert the one ordering the rail is built around. A persisted record for a swap that was never funded is the **safe** failure: `RfqSwapManager` finds an unfunded lockup and retires it. A funded lockup with no record is the unrecoverable one — `cancelOffer`/the refund path rebuild from the record, and without it nobody can refund. So `persist` deliberately has no rollback contract, and moving the gate after the write to gain one would create the case the ordering exists to prevent. The gate runs once, before both.

**Emulator pin: taken.** Correct that a 33-byte compressed key encodes to 66 hex and was adopted verbatim for a card advertising none. `solverRendezvous` now holds the pin to the same `XONLY_HEX` shape a card's own key is held to, so both rails fail closed at the shared boundary rather than each re-deriving it.

Kukks:
Thanks — both non-blocking items taken in `14edab7c`.

- **`SolverOnchainRendezvous` deprecated on arrival.** Agreed, and the alias is gone rather than un-deprecated. The shape turned out corridor-independent once the Lightning rail landed, so `SolverRendezvous` is the only name now; nothing has shipped that could depend on the alias.
- **`payoutPkScript` and JSON.** Good catch. The docstring now says it explicitly: `JSON.stringify` turns the `Uint8Array` into `{"0":81,"1":32,…}`, which comes back as an object with no `.length` any script builder accepts, and a `persist` writing to a JSON store must `hex.encode`/`hex.decode` around it. Exactly the failure worth naming — silent at write time, surfacing as a claim that cannot be built long after funding.
- **MTP fallback observability.** Agreed and left as-is: degrading beats throwing on a balance-read path, and the `console.warn` is the signal. Worth noting the fallback is deliberately not cached, so a systematically-refusing Fulcrum re-attempts the batch on every call rather than pinning the early-maturing value for the life of a block.

**One correction — the review is a push behind.** Its body says *"Reviewed diff at `cb5da060`"*, and two commits have landed since:

- `2af2efef` re-runs `assertFundable` at the top of `send()` on both solver rails (a CodeRabbit finding: `RouteQuote.send()` is a separate user action minutes after `quote()`, long enough for a quote or invoice to lapse), and holds `fallbackEmulatorPubkey` to `XONLY_HEX`.
- `11b5c757` **is** the real-Fulcrum MTP integration this review asks to be run before release. It is already merged into this branch and green: `Integration (ts-sdk / exit-providers-rotation)` passed, and the case appears by name in the log —

  ```
  ✓ test/e2e/electrum.test.ts > ElectrumOnchainProvider integration tests
    > computes median-time-past from real headers, not the tip's own time  11ms
  ```

  It asserts the fallback's warn did *not* fire on a fresh provider, which is what separates "the median was computed" from "the window read failed and degraded" — `tip.time > 0` alone passes either way. So that tracking item is closed, not pending.

All eight checks are green on the head commit.

Kukks:
Taken in `3b717063` — and this one materially corrects a claim I made in the description, so thank you.

You are right that the test proved less than it said. Below height 10 `medianTimePast` takes the short-window branch: fewer than ten header fetches, none at all at height 0. It warns about nothing and returns a median of what exists, so **both** existing assertions pass while the batch of ten `blockchain.block.header` requests the test exists to exercise never happens. The regtest stack does not guarantee a depth, so the green run I cited above does not by itself prove the batch worked.

Rather than assert and risk a red on a shallow stack, it now mines up to a full window first — with two blocks of headroom so a headers subscription that observes the mine a block late still sees one — and then asserts the height it actually read:

```ts
const start = await provider.getChainTip();
mine(Math.max(0, MTP_WINDOW + 1 - start.height));
…
expect(tip.height).toBeGreaterThanOrEqual(MTP_WINDOW - 1);
```

I have corrected the test-plan section of the description accordingly: the previous integration pass demonstrated the code path runs against a real Fulcrum, not that the full ten-header batch did. That is only established from this commit's run onward.

Kukks:
Comment volume cut in `25d62775` — style only, no logic, no test assertions changed.

**Before:** 864 comment-or-blank of 2414 added = 36% (comments alone 650 = 27%), with a 48-line block on one function.
**After:** 416 of 1966 = 21.2%. **Comments alone 202 = 10.3%.**

Both numbers, because the combined one cannot reach 10% honestly: blank lines are 214 of the 1966 (10.9%) on their own, so even with zero comments the combined figure floors at ~11%. I have not deleted blank lines to make the number look better.

Kept what is not recoverable from the code — the MTP-vs-`nTime` distinction (the field is called `time`, two providers returned different quantities for it, and the wrong one matures a timelock early), the byte-reversed scripthash that returns `[]` rather than erroring, persist-before-fund, the `Uint8Array` that JSON silently destroys, and why a re-gate is needed between `quote()` and `send()`.

Cut restatement of the line below, narration of how each gap was found (that is already in the commit messages and this description), and rationale repeated between a definition and its test or between the two sibling rails — kept once at the definition, with the sibling pointing at it. No block is over six lines except three module headers.

Gate unchanged: ts-sdk 2893, swap 719, boltz-swap 616 — 4228 passing, 0 failing on this branch alone. Staged blobs verified biome-clean by extracting them with `git show :<file>` rather than reading the working tree.

pietro909 (commented):
Read-only pass on `25d62775` — no tests or linters run. Ten findings: four blockers, six warnings. Nits omitted.

Nine are inline below. The tenth can't be anchored because the file isn't in this diff:

### `packages/ts-sdk/src/wallet/exit/executor.ts:348` — warning

```ts
: tip.time >= (dep.blockTime ?? 0) + step.delay.value;
```

This PR redefines `tip.time` as median-time-past, but `dep.blockTime` is still the confirming header's own `nTime` — `getTxStatus` returns `block_time` on both Esplora (`providers/onchain.ts:277`) and Electrum. Before this PR both sides were header time; now the two sides of one inequality use different bases.

It errs conservative — seconds-typed CSV sweeps wait an extra MTP-lag interval rather than firing early — which is why it's a warning. But it's a regression this PR introduces into the path it set out to fix. Either read MTP on both sides, or keep header time on both. Worth noting the earlier review pass listed this exact line as a beneficiary of the MTP fix and verified the median arithmetic against Bitcoin Core, but didn't check the right-hand operand.

---

Two findings were dropped as already covered here: the MTP fallback returning tip `nTime` (raised, and consciously declined) and the `fallbackEmulatorPubkey` encoding (fixed in `2af2efef`).

The blockers are four independent fixes, none large — the biggest is reusing the existing `vin` fallback in `chainSourceFrom.getSpendingTx`.

packages/swap/src/chainSource.ts
```diff
@@ -0,0 +1,57 @@
...
+                    : 0,
+        }));
+    },
+
+    async getSpendingTx(txid: string, vout: number): Promise<{ txHex: string } | null> {
+        const outspends = await provider.getTxOutspends(txid);
+        const outspend = outspends[vout];
+        // Some deployments omit the spender txid even when `spent` is true; the
+        // only caller reads P out of the spend and retries.
+        if (!outspend?.spent || !outspend.txid) return null;
```
pietro909: **Blocker.** `null` here collapses back into `phase: "unfunded"` at `onchainHtlc.ts:492` — so a funded-and-claimed HTLC reports as never funded.

That defeats the distinction the classifier's own docstring says it maintains (`onchainHtlc.ts:472-474`: `funding` is what separates "never funded" from "funded and already spent"). And the comment above doesn't hold — the caller doesn't retry: `nextOnchainAction` maps `unfunded` to `wait` (`swapManager.ts:340-342`), so the preimage receipt is never recorded and past `refundLocktime` the manager tries an Arkade refund of a lockup the solver already claimed with P.

`wallet.ts:1563-1590` already has the fallback for providers that omit the spender txid — index the address history's `vin`. Worth reusing rather than returning `null`.

packages/swap/src/payment/solverOnchain.ts
```diff
@@ -0,0 +1,162 @@
...
+
+export const SOLVER_ONCHAIN_RAIL = "solver-onchain";
+
+/** What a record needs; `onchainSendProfile()` maps it to an `RfqSwapOrigin`. */
+export type SolverOnchainSend = Awaited<ReturnType<typeof requestOnchainSend>> & {
+    rendezvous: SolverRendezvous;
+    /** Where the claim pays. Named nowhere else — the claim's output is the
+     *  spender's choice. NOT JSON-safe: a JSON store must `hex.encode` it, or
+     *  it returns as an object with no `.length` and the claim cannot be built. */
+    payoutPkScript: Uint8Array;
```
pietro909, line 24: **Blocker**, separate from the JSON hazard the docstring now names.

`payoutPkScript` has no slot in `OnchainSendProfile`: `onchainSendProfile()` (`rfqCorridors.ts:167-181`) returns `{claimKey, refundKey, htlcLocktime, network, htlcAddress, minConfirmations}` and drops it. So an app persisting the way line 17 points it to loses the claim destination — correct `hex.encode`/`decode` round-tripping doesn't help, the value never reaches the record.

After a restart `driveOnchain` reaches `claimable` and calls `claimOnchain`, which has nothing to build `buildHtlcClaim`'s `payoutPkScript` from (`onchainHtlc.ts:243`). Add the field to `OnchainSendProfile` and to the mapper.

packages/swap/src/payment/solverOnchain.ts
```diff
@@ -0,0 +1,162 @@
...
+                        await ctx.wallet.send({
+                            address: swap.address,
+                            amount: swap.fundAmount,
+                        });
+                        // Not "settled": the recipient has nothing until the
+                        // solver fills the HTLC and this wallet claims it.
+                        emit({ status: "sent" });
+                        const result = { railId: SOLVER_ONCHAIN_RAIL, swapId: swap.rfqId };
+                        if (!deps.awaitSettlement) return result;
+                        const { txid } = await deps.awaitSettlement(swap);
```
pietro909: **Blocker.** A rejection here is terminal: `handle.ts:38-39` catches it and emits `{status: "failed"}`, and `settled()` rejects.

But funding already succeeded two lines up. A poll timeout or a transient esplora error in the watcher reports an in-flight payment as failed — inviting a retry that funds a second lockup.

A watcher failure isn't a payment failure. Suggest resolving to `result` (the payment is sent; settlement is unobserved) or a distinct non-terminal status. Same at `solverLightning.ts:147`.

packages/swap/src/payment/rendezvous.ts
```diff
@@ -0,0 +1,78 @@
...
+): SolverRendezvous | undefined => {
+    // A 33-byte compressed key encodes to 66 hex and would otherwise pass as
+    // x-only. Neither rail re-derives the pin, so this is the only place both
+    // fail closed.
+    const encoded = fallbackEmulatorPubkey ? hex.encode(fallbackEmulatorPubkey) : undefined;
+    if (encoded !== undefined && !XONLY_HEX.test(encoded)) return undefined;
+    const pinned = encoded;
+
+    for (const market of markets) {
+        if (marketCorridor(market, "base") !== "arkade") continue;
```
pietro909: **Blocker.** Corridor is checked, asset is not. A market `{base_asset: "USDT", base_corridor: "arkade", quote_asset: "BTC", quote_corridor: "onchain"}` passes both lines, `sideLimits` then bounds-checks a sats amount against USDT limits, and the loop returns it — never reaching the real arkade:BTC → onchain:BTC card.

Both rails go on to negotiate their hard-coded pair against a solver that doesn't serve it: a burned negotiation that also leaks the intent, which is the cost the bounds check exists to avoid.

`markets.ts:211` already does corridor + asset + limits in one call — `bestMarket(markets, {baseId, quoteId, wantSide})`.

packages/swap/src/payment/solverOnchain.ts
```diff
@@ -0,0 +1,162 @@
...
+            }
+
+            // `"to"` keeps the quote receiver-exact: the user's number is the
+            // L1 payout and the fee sits on top.
+            const negotiated = await deps.connect(rendezvous, (transport) =>
+                requestOnchainSend(ctx.wallet, deps.arkServerUrl, transport, {
+                    amount,
+                    amountSide: "to",
+                    payoutPubkey: deps.payoutPubkey,
+                    ...(deps.emulatorPubkey ? { emulatorPubkey: deps.emulatorPubkey } : {}),
```
pietro909, line 110: **Warning.** `rendezvous.emulatorPubkey` is validated by `rendezvousOf` and then never used: only `deps.emulatorPubkey` reaches `requestOnchainSend`, which otherwise falls back to `resolveEmulatorPubkey(network)`. Line 112 keeps the rendezvous on the swap, but the covenant is already built.

Safe today precisely *because* the validated value is ignored — the pinned default is the right key. The hazard is that it reads as a trust boundary: with no `fallbackEmulatorPubkey` a card advertising an arbitrary key is accepted anyway, and with one set the check gates a value that has no effect on the covenant.

Either bind the card's key into the covenant, or drop `emulatorPubkey` from `SolverRendezvous`. Same at `solverLightning.ts:108`.

packages/swap/src/payment/solverOnchain.ts
```diff
@@ -0,0 +1,162 @@
...
+            // A discovery failure propagates: the router catches it, warns, and
+            // drops this rail — which is the intended fallback, and louder than
+            // swallowing it here would be.
+            return (await rendezvousFor(amount)) !== undefined;
+        },
+
+        quote: async (req, ctx: RouterContext): Promise<RouteQuote> => {
+            const address = btcTarget(req.raw)!;
+            const amount = resolveSendAmount(SOLVER_ONCHAIN_RAIL, req.raw, req.amount);
+            const payoutPkScript = l1ScriptForAddress(address, deps.l1Network);
```
pietro909, line 95: **Warning.** `deps.l1Network` is never reconciled with `negotiated.l1Network` (available at line 112), which `requestOnchainSend` returns so callers needn't guess.

Construct the rail with `l1Network: "bitcoin"` against a regtest Ark server: `available()`'s `l1ScriptForAddress` throws for every `bcrt1…`, the catch at line 77 swallows it, and the rail answers false for every payment with no diagnostic anywhere. Where HRPs coincide (testnet/signet/mutinynet all map to `"testnet"`) it succeeds and persists a payout script derived on a network the swap wasn't negotiated for.

Compare the two after line 112 and throw.

packages/swap/src/payment/solverOnchain.ts
```diff
@@ -0,0 +1,162 @@
...
+                    payoutPubkey: deps.payoutPubkey,
+                    ...(deps.emulatorPubkey ? { emulatorPubkey: deps.emulatorPubkey } : {}),
+                }),
+            );
+            const swap: SolverOnchainSend = { ...negotiated, rendezvous, payoutPkScript };
+
+            return {
+                railId: SOLVER_ONCHAIN_RAIL,
+                amount,
+                fee: swap.fundAmount - amount,
```
pietro909, line 126: **Warning.** `amount` is specified as "Sats delivered to the recipient" (`ts-sdk/src/payment/types.ts:48`) with no estimate escape hatch — the `fee` docstring just above grants one, `amount` doesn't. `onchainRail` goes out of its way to honour that, grossing up so the recipient receives exactly `quote.amount` (`ts-sdk/src/payment/rails/onchain.ts:50-52, 76`).

Here `amount` is the pre-claim-fee HTLC value; the recipient actually gets `amount - claimTxFee`. `claimFeeDeductedFromPayout: true` in `meta` records it, but this is the only rail whose `amount` isn't what the type says it is.

Not a router-selection bug — `router.ts:69-74` ranks on the configured `priority` list, not on cost. It costs any consumer that displays `amount` or compares quotes across rails. Folding a claim-fee estimate into `fee` and leaving `amount` receiver-exact would restore the contract.

packages/ts-sdk/src/providers/electrum.ts
```diff
@@ -889,21 +892,61 @@ export class ElectrumOnchainProvider implements OnchainProvider {
...
+    /**
+     * Consensus's `GetMedianTimePast`, ending at and including `height`.
+     *
+     * Cached per tip HASH, not height: this runs on every balance read, and
+     * hashing means a reorg at the same height recomputes rather than serving
+     * the orphaned chain's median. Degrades to the tip's `nTime` rather than
+     * throwing — a caller that cannot read the tip at all is worse off.
+     */
+    private async medianTimePast(height: number, tipHex: string, hash: string): Promise<number> {
+        if (this.mtpCache?.hash === hash) return this.mtpCache.mtp;
```
pietro909, line 921: **Warning.** No in-flight dedupe: `mtpCache` is only written after the fetch resolves, so N concurrent `getChainTip()` calls on a fresh tip each fire the full 10-header batch. This runs on every balance read (`wallet.ts:2123`) and every `chainSourceFrom.getScriptUtxos` poll — five readers on a new block make 50 requests where 10 would do.

`contractManager.ts:1893-1902` already models the shape with `chainTipInflight`.

Minor, same path: line 903 parses `tip.hex` for the hash and line 922 parses the identical bytes again for the timestamp. Passing the parsed header in would drop one parse.

packages/swap/test/payment/solverLightning.test.ts
```diff
@@ -0,0 +1,364 @@
...
+    expiresAt: NOW() + 3600,
+    ...over,
+});
+
+const negotiated = (fundAmount: number) =>
+    ({
+        rfqId: "rfq-ln-1",
+        address: "tark1lockup",
+        fundAmount,
+        quote: { from_amount: fundAmount, to_amount: 100_000, valid_until: 1_800_000_000 },
```
pietro909: **Warning.** `valid_until: 1_800_000_000` is a fixed instant — 2027-01-15. After that the send-time re-gate added in `2af2efef` throws `quote expired` and every `solverLightningRail.send` case fails for a reason unrelated to the code under test.

`solverOnchain.test.ts:54` has it right: `NOW() + 3600`.

packages/swap/src/chainSource.ts
Kukks: Fixed in `5ade00fe`. `getSpendingTx` now falls through to the address history's `vin` when the outspend names no spender (`chainSource.ts:24-31`, `:63`), reusing the shape you pointed at. `||` not `??`, for the same reason `wallet.ts:1590` gives — the electrum provider's `txid: ""` sentinel.

`ChainSource.getSpendingTx` takes the spent output's `pkScript` to index that history (`onchainHtlc.ts:364-369`), passed by `classifyOnchainHtlc` from `input.htlc.pkScript` (`onchainHtlc.ts:496`). Required rather than optional, so there is no silent no-fallback path; a required parameter on the interface does not break existing implementations, which may still implement with two.

Reproduced as the loss path end to end rather than just the adapter: `chainSource.test.ts` builds a real claim spend with `buildHtlcClaim`, serves it through `chainSourceFrom` over a provider that omits the spender txid, and asserts `classifyOnchainHtlc` reports `{phase: "claimed", preimage}`. Against the old code that assertion got `{phase: "unfunded"}`. 3 tests failed before the fix, 0 after.

The old case asserting the null — "reports no spend when the deployment omits the spender txid" — encoded the bug, so it is now "reports no spend when neither the outspend nor the history names one".

packages/swap/src/payment/solverOnchain.ts
Kukks: Fixed in `5ade00fe`, both rails. The watcher call is wrapped and a rejection resolves to `result` — the handle stays at `"sent"` and the failure is logged, not emitted (`solverOnchain.ts:168-174`, `solverLightning.ts:151-157`).

Your read of the mechanism is exact: `makeHandle`'s `done.catch` turns any rejection from the run body into `{status: "failed"}`, and funding has already succeeded by then, so a poll timeout reported an in-flight payment as failed. Resolving to `result` rather than adding a status, since "sent" already means what is true here — the money has left, settlement is unobserved — and `RfqSwapManager` drives the swap either way.

Two tests failed first, one per rail, each pointing at exactly the lines you named (`solverOnchain.ts:154`, `solverLightning.ts:147`); both assert the handle reaches `"sent"`, that `settled()` resolves to the sent result, and that no subscriber ever sees `"failed"`.

packages/swap/src/payment/solverOnchain.ts
Kukks: Fixed in `5ade00fe` — with one deviation you did not ask for, flagged up front.

`payoutPkScript: string` is now a slot on `OnchainSendProfile` (`rfqCorridors.ts:137`), written by `onchainSendProfile()` (`:175`, `:184`), and decoded back on the swap by `hydrate` (`:261-263`). `SolverOnchainSend` is what you hand the mapper, since only it carries the field (`solverOnchain.ts:16-24`).

**The deviation:** `hydrate` tolerates a missing `payoutPkScript` instead of throwing, and `OnchainSendSwap.payoutPkScript` is optional. Required at the write, so no new record can omit it — but every record written before this commit lacks it, and those swaps are still owed their refund. Throwing at restore would strand them, which is worse than the claim they already could not make. If you would rather it threw like `minConfirmations` does, say so and I will change it.

`project()` still does not emit the field, which is correct: the merge is a spread, `profile: { ...origin.profile, ...handler.project(swap) }` (`rfqRecord.ts:226`, `:262`), so the value written at creation survives every later update.

2 tests failed first: the mapper assertion, and a create-persist-rebuild round-trip that now asserts the script comes back as bytes. A third, new, pins the legacy path — a record with no `payoutPkScript` restores with the HTLC intact and the field `undefined`.

packages/swap/src/payment/rendezvous.ts
Kukks: Fixed in `5ade00fe`. `solverRendezvous` now filters through `selectMarkets` on corridor **and** asset before the loop (`rendezvous.ts:72-77`), which is `bestMarket`'s own predicate.

I kept the loop over the filtered set rather than the single `bestMarket` call, so the skip-to-next-card behaviour for a malformed or pin-disagreeing `emulator_pubkey` is provably unchanged, and left the size check exactly where it was for the same reason. `selectMarkets` therefore gets corridor + asset only, not `wantSide`/`wantAmount`.

3 tests failed first: a card with a non-BTC base and one with a non-BTC quote must both be skipped, and a later BTC card must be taken over an earlier non-BTC one.

**The check immediately caught two wrong fixtures**, which is the more interesting half of this. Both rail test cards carried `base_asset: { id: "BTC" }` — the *ticker*, not the id. Confirmed against the emitter rather than the consumer, in `arkade-os/intent-solver@origin/main`: `packages/solver-core/src/core/marketKey.ts:41` defines `BTC: { id: 'btc', name: 'Bitcoin', decimals: 8 }`, `packages/solver-core/src/core/registryCard.ts:33` publishes `{ id: ASSETS.BTC.id, ticker: 'BTC', … }`, and `packages/solver-core/src/core/assetMarketConfig.ts:97` documents the distinction. So `id` is lowercase `btc` and `ticker` is `BTC` — separate fields, and the strict `!==` in `selectionPredicate` is right without case-folding. The fixtures are corrected to `id: "btc"`, matching `test/fixtures.ts:10`.

That those fixtures were wrong and nothing noticed is the evidence the asset was never checked.

packages/swap/src/payment/solverOnchain.ts
Kukks: Fixed in `5ade00fe`, taking your second option: `emulatorPubkey` is off `SolverRendezvous` (`rendezvous.ts:11-20`). `rendezvousOf` still resolves the card's key and still fails closed on a malformed one and skips a card that disagrees with the pin (`:29-36`) — it is a filter, and now only a filter.

I did not take the first option. Binding the card's key into the covenant would mean that with no `fallbackEmulatorPubkey` configured, any card could dictate the co-signer — which is the hazard you describe rather than a fix for it. Dropping the field keeps today's behaviour exactly and removes the reading that it is a trust boundary.

A test asserts the rendezvous carries no `emulatorPubkey` at all, and the existing pin cases now assert on `solverPubkey`. Mutation check: re-adding the field to the returned object is caught.

packages/swap/src/payment/solverOnchain.ts
Kukks: Fixed in `5ade00fe`. `negotiated.l1Network` is compared against `deps.l1Network` right after the negotiation returns and throws on a mismatch (`solverOnchain.ts:113-120`) — before the swap is built, so nothing is persisted or funded.

Your HRP point is what makes it worth throwing rather than logging: the mismatch that fails loudly is the harmless one. The failing-first test used `l1Network: "bitcoin"` against a `regtest` rail; adding the check then failed 13 tests, because the shared `negotiated()` fixture never modelled `l1Network` at all even though `requestOnchainSend` returns it. Fixture corrected to `regtest`.

1 test failed before the fix, 0 after. Mutations: deleting the comparison, and inverting it, are both caught.

packages/swap/src/payment/solverOnchain.ts
Kukks: Declined, and happy to be overruled — the contract violation is real, I disagree only about the remedy.

The collaborative exit can be receiver-exact because it *performs* the deducting spend: `onchainRail` grosses up and then hands the same `FeeInfo` to `offboard` at settlement, which is what `rails/onchain.ts:47-49` means by "the two cannot drift". This rail never claims. The claim is the app's `RfqSwapManagerCallbacks.claimOnchain`, which picks its own feerate at its own time (`onchainHtlc.ts:426-438`), and the rail has no feerate at `quote()`. So a gross-up here would negotiate `amount + estimate` against the solver and the recipient would still receive something else — trading a flagged, visible shortfall for an invisible variable one, and charging the user for the guess.

What I did instead: amended the contract so it is no longer silently violated. `RouteQuote`'s docblock now names the exception (`payment/types.ts:37-41`) alongside the gross-up rule, pointing at `claimFeeDeductedFromPayout`.

If you would rather have the estimate — a feerate on `SolverOnchainRailDeps`, gross up the `to` side, fold it into `fee` — say so and I will do it; the claim tx's vsize is deterministic from the HTLC and payout scripts, so only the rate is a guess. Your call.

packages/ts-sdk/src/providers/electrum.ts
Kukks: Fixed in `5ade00fe`, on `chainTipInflight`'s model: `mtpInflight` holds the pending read keyed by tip hash, adopted only for a matching hash, cleared in a `finally` (`electrum.ts:561`, `:920-931`). The window fetch moved to `readMedianTimePast` (`:936`).

The second parse is gone too — `getChainTip` now destructures `{ hash, timestamp }` from its one `parseBlockHeader` and passes the timestamp down (`:903`, `:908`), so `medianTimePast` takes `tipTime: number` instead of re-parsing the hex.

It reads as more than an efficiency bug once tested. Five concurrent `getChainTip()` calls on a fresh tip returned `[2000, 1005, 2000, 2000, 2000]` — four of the five got the tip's own `nTime` via the fallback rather than the median, because the redundant fetches had nothing left to serve them. A rate-limiting server does the same thing to a real caller, and that fallback matures early. So the burst is not just wasteful, it is a route to the wrong answer.

Two tests: the concurrent-readers case, and one where a new tip arrives while another's window is still in flight, asserting it never gets served the older tip's median. That second one exists because the first mutation run left "adopt a pending read for ANY hash" surviving.

packages/swap/test/payment/solverLightning.test.ts
Kukks: Fixed in `5ade00fe` — `valid_until: NOW() + 3600`, matching `solverOnchain.test.ts:54`.

Worth recording where the blast radius stops, since the constant appears elsewhere in the suite: the send-time re-gate reads the wall clock at `solverLightning.ts:134` and `solverOnchain.ts:148`, and those are the only two places a test fixture is compared against real `Date.now()` without an injected `now`. The other `1_800_000_000`s in `packages/swap/test` are script parameters or golden-vector inputs — `rfq.ts`'s pinned locktime among them — which must stay fixed. So this was the one that would have started failing, and `solverOnchain` already had it right.

Kukks:
## `executor.ts:348` — the MTP base mismatch, and the review round generally

Posting here since the file is not in the diff. Fixes for all ten findings are in `5ade00fe`; the nine inline ones are answered in thread.

### The decision: MTP on the left, header time on the right, deliberately

Kept as this PR left it, with the reasoning written at the site (`wallet/exit/executor.ts:345-352`). Two legs:

1. **The left operand got *more* correct, not less.** Consensus evaluates a spend for inclusion in block *N+1* against the median-time-past of block *N* — the tip. `tip.time` is now exactly that. Before this PR it was the tip header's own `nTime`, which *overstates* MTP, so the old symmetry could fire early and eat a `non-BIP68-final` rejection.
2. **The right operand is a provable upper bound.** BIP-68 measures a coin from the MTP of the block *before* the one containing it. A block's `nTime` must be greater than the previous block's MTP, so `dep.blockTime` — the confirming header's `nTime` — is always at least the value consensus uses. Requiring more elapsed time than consensus does can only wait, never fire early.

Both deviations point the same way, so the mixed bases are safe in the direction that matters. Reading MTP on both sides is the fully correct form, but the coin-side value is `MTP(depHeight - 1)`, which `getTxStatus` cannot give — it would need a new `OnchainProvider` capability, and every custom provider implementation would have to grow it. Not worth it for an operand that is already conservative.

**Marked as inferred:** this is reasoning from Bitcoin Core's rules (`CalculateSequenceLocks` / `EvaluateSequenceLocks`, and the `time-too-old` header check), not something executed here. If you want it demonstrated rather than argued, the honest test is a regtest one, and I have not written it. You were right that the earlier pass checked the median arithmetic and not the right-hand operand.

### Evidence for the round

- **Each blocker reproduced as a failing test first**, before any fix: 3, 2, 2, 3 failures respectively, all 0 after. The chainSource one asserts the loss path end to end — a real `buildHtlcClaim` spend served through `chainSourceFrom` over a provider that omits the spender txid, classified as `claimed` rather than `unfunded`.
- **23 mutations run, 23 killed, 0 survived.** Each blocker has a mutation that restores the pre-fix code exactly, plus `||`→`??`, inverted guards, wrong-network address indexing, and writing the claim key as the payout script. Two survived on the first pass; I dropped the one that was an untestable optimisation and wrote a test for the one that was load-bearing rather than leave either unproven.
- **Gate: 4228 → 4241 passing**, 1 skipped, 0 failing either side, `pnpm run test:unit` exit 0. The +13 reconciles exactly against the tests added, and the file count is unchanged at 228.
- **Comments: 9.81%** of added non-blank lines.

### One thing to flag

**#838 is stacked on this branch and now needs a rebase** — it touches `solverLightning.ts`, `solverOnchain.ts` and `types.ts`, all changed here. I have not touched it.

PR #838, opened 2026-09-04
feat(sdk): an asset dimension for PaymentRouter, and the asset + cross-asset rails

Kukks:
> **Rebased onto `master`.** #837 landed as a squash (`fa3a2d1f`), so this
> branch was replayed with `git rebase --onto origin/master 5ade00fe`, taking
> only its own five commits. Base retargeted from `feat/payment-rails-router`
> to `master`; the diff is now this PR's work alone (16 files).
> No conflicts: #837's review fixes landed in `send()` and in `rendezvous.ts`,
> while this PR's rail changes are asset guards in `available()`/`quote()`.

`PaymentRequest.amount` and `RouteQuote.amount`/`fee`/`total` are bare
`number`s with no unit or asset field. The router can say "500" but never "500
USDX". An Arkade asset payment therefore has nowhere to go — and an Arkade
address is the **same string** whether it is paid in BTC or in an asset, so
the target cannot tell them apart either. The asset is named by the AMOUNT.

## The public-API decision

Three options, and one of them the codebase had already taken.

**Widen the existing fields** (`number | { atomic: bigint; asset: string }`).
Rejected: it breaks every consumer that does arithmetic on `quote.total`, and
the asset half would still be lossy anywhere a `number` survives.

**A separate request type / parallel router.** Rejected: it duplicates
matching, ranking and preferences, and it re-creates exactly the "two things
that disagree about when to fall back" problem the router exists to avoid.

**Add an optional field alongside** — taken, because `Wallet.send`'s
`Recipient` already solves this problem and solves it this way:

```ts
export interface Recipient {
    amount?: number;      // sats, defaults to dust
    assets?: Asset[];     // { assetId: string; amount: bigint }
}
```

and `Asset.amount` is documented as `bigint` for the reason that settles the
whole question: *"asset supplies routinely exceed `Number.MAX_SAFE_INTEGER`
and silently truncating in arithmetic would corrupt balances."*

So `PaymentRequest` grows the same optional `assets`, using the same `Asset`
type — no new vocabulary, and a request is now structurally close to the
`Recipient` a rail hands to `wallet.send`. Every existing sats field keeps its
meaning, and nothing that compiled before stops compiling: boltz-swap's 616
tests are untouched and still pass, which is the check that the addition is
genuinely additive.

`Recipient` also settles a subtler point. An Arkade asset does not move on its
own — it rides a sats-carrying output — so a 500 USDX payment really does move
sats as well. The sats fields are therefore **not zeroed** on an asset route:
`amount` is the carrier and `total` goes on meaning what leaves the wallet, so
ranking on `total` still means the same thing on every rail.

On `RouteQuote` the asset view is a **pair, not a triple**:

```ts
assets?: { delivered: Asset; spent: Asset };
```

because a cross-asset route spends BTC and delivers USDX, and `total = amount
+ fee` cannot hold across two units. On a same-asset route the two name the
same asset and the difference between them is the fee.

`RouterContext.swaps` is left as `unknown`. It did not need typing: the rails
here take their dependencies at construction, which is stricter than a typed
escape hatch and does not make core name a type it cannot import.

## Did cross-asset composition fit in a rail?

**Yes, and the router needed no composition primitive.** That was the thing to
test rather than assume, so `crossAssetRail` is the demonstration.

Both legs are quotable at quote time — `quoteOffer` prices the swap from the
market's own feed, and delivery is an off-chain transfer with no counterparty
fee — so the rail prices them into one `RouteQuote` and reports the offer on
`RouteResult.swapId`, which already had a slot. `send()` funds the offer,
waits for a filler, and only then pays the recipient.

The argument for keeping this out of the router: a hop is only routable if it
is quotable, and a rail that owns both hops can quote both. Pushing it up
would mean inventing a quote-composition rule every future rail has to
satisfy, for one caller. The one thing that genuinely could not stay inside a
rail is the two-unit quote — hence `assets`.

What it does not do is asset → asset in one offer. `createOffer` refuses
anything but BTC↔asset ("set exactly one of wantAsset or offerAsset"), so
paying USDX out of EURX is two offers. That is still rail-internal — more
legs, not a different shape — and it is not implemented here.

Its BTC-only restriction is `findMarket` being keyed on BTC, **not** a
separate guard. `selectionPredicate` matches asset ids exactly in either
orientation, so a EURX/USDX market is never returned for `(btc, USDX)`. I
wrote a give-side check first; mutation testing showed removing it broke
nothing, I traced why, found it unreachable, and deleted it rather than keep a
test that passed for the wrong reason.

## The rails

**`ark-asset`** — pay an Arkade address in an asset. Splits from `ark` on
`available()` rather than `match()`: both legitimately match the address, and
which one can pay is a question about the amount. `options()` returns exactly
one, so their relative priority is immaterial.

**`cross-asset`** — pay an asset the wallet does not hold. Lives in
`@arkade-os/swap` (it needs `createOffer`); `ark-asset` is core, because an
asset transfer out of a balance you already have needs no swap at all.

`ark-asset` is registered in **both** default factories. A Lightning-capable
app picks boltz-swap's `createDefaultPaymentRouter(wallet, swaps)` over the
core one, and without it there that choice silently costs it asset payments —
`route()` would throw "no rail for" with nothing saying why. It never competes
with `ark`, so every existing BTC ranking in that factory's tests is unchanged
and still asserted exactly as it was.

## One thing worth arguing about

On a cross-asset route the sats triple comes out as `amount: 330` (the
carrier), `fee: 100_000` (the swap deposit), `total: 100_330`. Calling the
purchase price a "fee" reads oddly — it is not a charge, it is what the asset
cost.

It is forced rather than chosen. `RouteQuote` defines `amount` as sats
delivered to the recipient and `total` as sats that leave the wallet; the
recipient receives USDX and 330 sats of carrier, and 100,330 sats leave. With
`total = amount + fee` holding, `fee` can only be 100,000. `assets.spent`
names the same number properly, as BTC.

The alternative is to let an asset route report `total` as only the carrier
and leave the deposit off the sats triple entirely — which would make
`options()` rank a cross-asset route as costing 330 sats, and that is worse.
Flagging it because a reviewer may prefer a third field over overloading
`fee`, and that is a reasonable position.

## The silent-drop this closes

Every BTC-only rail now refuses a request carrying an asset instead of
ignoring it: `ark`, `onchain`, `lightning`, `onchain-swap`, `solver-onchain`,
`solver-lightning`. Without that, a request for 500 USDX handed to a BTC rail
would pay its carrier sats, report success, and deliver none of the asset the
user asked for. Refusing at `available()` is what leaves the router free to
rank a rail that can, and `assertNoAssets` at `quote()` is the backstop for a
caller reaching a rail directly.

## Self-review found a bug (`c040b8a2`)

`Asset.assetId` is a hex string — that is what `Wallet.send` takes — while
`createOffer`'s `wantAsset` is an `asset.AssetId`, a class it reads `.txid`
and `.groupIndex` off to bind the covenant. I had written `wantAsset:
asset.assetId as never`, and the cast is what hid it. It does not throw: both
fields come back `undefined`, so the offer publishes wanting nothing in
particular — the user funds it and nobody can fill it.

The tests missed it because they mock `createOffer`, and a mock accepts any
shape. The regression test therefore asserts on the value passed rather than
the call: `txid` is a `Uint8Array`, `groupIndex` is a number, `toString()`
round-trips. Re-introducing the cast fails it.

Asset ids are also 34 bytes, not 32, and both fixtures were 32 — long enough
to look right. `available()` now parses the id up front, so one `createOffer`
could not bind drops the rail instead of failing at send time with an offer
already funded.

## Review round 1 (`1fa1a647`)

All five findings from arkana-ai-bot taken — [reply](https://github.com/arkade-os/ts-sdk/pull/838#issuecomment-5537079792).

- **The window between the fill and the payment had no state**, and the record
  actively lied: frozen at `"quoted"` after a fill landed, it points recovery
  at `cancelOffer` when the asset is in fact in the wallet and the recipient is
  still owed. `CrossAssetSwap` now carries a `phase` and `persist` runs at both
  transitions — `"quoted"` before funding, `"filled"` after delivery and before
  the recipient send. Both sit before an irreversible step, the same rule the
  send rails follow.
- **The deposit's `bigint → number` narrowing is asserted, not assumed.** BTC's
  supply fits a safe integer; the fragile assumption is the give side being
  BTC, which a comment would not catch.
- **Doc-only** on `RouteQuote.assets` (on a cross-asset route `fee` is the
  purchase price, and `assets.spent` is what to display), on `awaitFill` (the
  caller owns the deadline, and resolving on a timeout makes the rail pay from
  whatever the wallet then holds), and on `discover` (called twice per route,
  and what caching makes that one round trip rather than four).

## Test plan

- `pnpm -r build && pnpm run test:unit`
  - base (#837 head, `14edab7c`): ts-sdk 174 files / 2893 + 1 skipped, swap 31
    / 719, boltz-swap 23 / 616 — 4228 passing, zero failures.
  - after: ts-sdk 175 / 2914 (+21), swap 32 / 748 (+29), boltz-swap 23 / 617
    (+1, the new factory case) — 4279 passing, zero failures. The 51 added
    tests reconcile exactly: +50 across the two new files, +1 in the boltz
    factory suite. **No existing boltz-swap test changed**, which is the check
    that the type addition is genuinely additive.
- Mutations run, each killed:
  - `cross-asset` routing when `findMarket` finds no corridor — 3 fail
  - `findMarket` keyed on the asset instead of BTC — 1 fail
  - `cross-asset` ignoring `validatePlan`'s verdict — 2 fail
  - paying the recipient before waiting for the fill — 1 fail
  - dropping the swap deposit from the sats total — 1 fail
  - reporting the delivered asset as the spent one — 1 fail
  - dropping the BTC short-circuit (a registry round trip on every BTC send) —
    1 fail
  - dropping the well-formedness short-circuit — 1 fail
  - accepting a multi-asset request — 1 fail
  - `ark-asset.available()` ignoring whether an asset was named — 4 fail
  - `ark-asset` sending the carrier but dropping the asset — 1 fail
  - `ark-asset` quoting zero sats instead of the carrier — 3 fail
  - `ark` accepting an asset request (the silent drop) — 4 fail
  - `resolveAssetAmount` accepting a `Number` quantity — 2 fail
  - re-introducing the `as never` cast (the bug above) — 1 fail
  - `available()` skipping the asset-id parse — 1 fail
  - not recording the fill before the recipient send — 3 fail
  - recording it after the send instead of before — 2 fail
  - leaving the phase at `"quoted"` — 2 fail
  - removing the deposit narrowing guard — 1 fail
  - One mutation **survived** and was acted on rather than explained away: the
    `givesBtc` give-side guard in `cross-asset`. It was unreachable; it is
    gone, and the test that appeared to cover it now pins the real invariant.
- Not run: regtest integration for the asset rails. `ark-asset` bottoms out in
  `Wallet.send({ assets })`, which the arkade-assets integration job already
  covers; `cross-asset`'s fill wait is injected and has never been driven
  against a real filler. **That is the claim I would most expect to need
  adjusting** — specifically whether `deps.awaitFill` is the right seam or
  whether the rail should take a `watchOfferSwaps` handle directly.

## The fields are tagged `@experimental` (`b268a9b8`)

This PR targets `master`, which ships `@arkade-os/sdk` 0.4.x every few days
(0.4.65 -> 0.4.68 in the last fortnight). So merging it publishes
`PaymentRequest.assets` and `RouteQuote.assets` as stable API within days.

Meanwhile #843 — the v2 swap client, on `feat/v0.5` — models assets natively
as `give`/`take`/`amountOn` over `AssetRef`, and `AssetId` the class has lived
in this package all along at `extension/asset/assetId.ts` while
`Recipient.assets` uses hex-string ids. **That split predates both PRs**, and
following `Recipient` here is still the right call for the reason above. But
converging the two vocabularies after a 0.4.x release turns a refactor into a
breaking change, so both fields carry `@experimental` until v0.5 settles it.

Doc-only, no behaviour change; typecheck and lint clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for sending Arkade assets through the default payment router.
  * Added cross-asset payments, allowing BTC to be exchanged for an asset before delivery.
  * Payment quotes and requests now support asset details.
  * Cross-asset payments now track payout progress through completion and support recovery after interruptions.

* **Bug Fixes**
  * BTC-only payment methods now reject asset-bearing requests instead of attempting incompatible routes.
  * Improved validation for missing, malformed, or unsupported asset amounts.
  * Asset routing preserves existing BTC payment priorities.

### DIFF
--- a/packages/swap/src/payment/crossAsset.ts
+++ b/packages/swap/src/payment/crossAsset.ts
@@ -21,9 +21,13 @@ import { BTC_ASSET_ID } from "../store";
 
 export const CROSS_ASSET_RAIL = "cross-asset";
 
-/** The two states need different recovery: a `"quoted"` record whose offer
- *  never filled is cancelled; a `"filled"` one owes the recipient a send. */
-export type CrossAssetPhase = "quoted" | "filled";
+/**
+ * Recovery branches on this. `filled` means the payout is NOT yet broadcast;
+ * `paying` means it WAS and the outcome is unknown — resolve that against
+ * chain state, never by resending. Persisting only afterwards would leave a
+ * crash between send and persist looking exactly like `filled`.
+ */
+export type CrossAssetPhase = "quoted" | "filled" | "paying" | "settled";
 
 export interface CrossAssetSwap {
     phase: CrossAssetPhase;
@@ -34,6 +38,8 @@ export interface CrossAssetSwap {
     asset: Asset;
     payTo: string;
     plan: OfferPlan;
+    /** The payout, once `settled` — the receipt a `paying` record looks for. */
+    txid?: string;
 }
 
 export interface CrossAssetRailDeps {
@@ -44,9 +50,8 @@ export interface CrossAssetRailDeps {
     quote(market: DiscoveredMarket, give: Side, wantAmount: bigint): Promise<OfferPlan>;
     btcBalance(): Promise<bigint>;
     dust(): Promise<bigint>;
-    /** Called twice, both before an irreversible step: `"quoted"` before the
-     *  offer is funded, `"filled"` after delivery and before the recipient is
-     *  paid. Keyed by `offerHex`, stable across both. */
+    /** Before every irreversible step and once after the last. See
+     *  {@link CrossAssetPhase}. Keyed by `offerHex`. */
     persist(swap: CrossAssetSwap): Promise<void>;
     /** Resolve once a filler has delivered the asset to this wallet.
      *
@@ -174,11 +179,14 @@ export function crossAssetRail(deps: CrossAssetRailDeps): PaymentRail {
                         await deps.awaitFill(swap);
                         // Before paying: "quoted" would say it never filled.
                         await deps.persist({ ...swap, phase: "filled" });
+                        // Before the send, not after: a crash must not read as unpaid.
+                        await deps.persist({ ...swap, phase: "paying" });
                         const txid = await ctx.wallet.send({
                             address: payTo,
                             amount: carrier,
                             assets: [asset],
                         });
+                        await deps.persist({ ...swap, phase: "settled", txid });
                         const result = {
                             railId: CROSS_ASSET_RAIL,
                             txid,
--- a/packages/swap/src/payment/crossAsset.ts
+++ b/packages/swap/src/payment/crossAsset.ts
@@ -1,44 +1,10 @@
 /**
  * `cross-asset` — pay an asset the wallet does not hold, by swapping into it
- * first.
+ * first. Both legs are quotable up front, so they price into one `RouteQuote`
+ * and the router needs no multi-hop primitive.
  *
- * ## Composition fits inside a rail
- *
- * The obvious reading of "swap, then pay" is that the router needs a
- * multi-hop primitive: two `RouteQuote`s chained, with a way to price the pair.
- * It does not, and this rail is the demonstration. Both legs are known at
- * quote time — `quoteOffer` prices the swap from the market's own feed, and
- * the delivery leg is an off-chain transfer with no counterparty fee — so the
- * rail prices them together into one `RouteQuote` and reports the swap on
- * `RouteResult.swapId`, which {@link RouteResult} already had a slot for.
- *
- * Nothing about the router had to change. That is the argument for keeping
- * composition out of it: a hop is only routable if it is quotable, and a rail
- * that owns both hops can quote both. Pushing this into the router would mean
- * inventing a quote-composition rule that every future rail has to satisfy,
- * for one caller.
- *
- * What DID have to change is the request and quote shape, and only because a
- * cross-asset route spends one asset and delivers another: `total = amount +
- * fee` cannot hold across two units, which is why `RouteQuote.assets` carries
- * `spent` and `delivered` separately rather than a third triple.
- *
- * ## What this rail does not do
- *
- * Asset → asset in one offer. `createOffer` refuses anything but BTC↔asset
- * ("set exactly one of wantAsset or offerAsset"), so paying USDX out of EURX
- * is two offers. That is still a rail-internal concern — it is more legs, not
- * a different shape — but it is not implemented here, and `available()` says
- * so by requiring the wallet's BTC balance to cover the deposit.
- *
- * ## Order
- *
- * `send()` funds the offer, waits for a filler to deliver, and only then pays
- * the recipient. The wait is injected (`awaitFill`) because observing a fill is
- * `watchOfferSwaps`/`restoreAssetSwaps`' job and this package does not start a
- * watcher of its own. `persist` runs before the offer is funded, for the same
- * reason it does on the send rails: an offer funded without a record is one
- * `cancelOffer` cannot rebuild.
+ * BTC -> asset only: `createOffer` refuses anything but BTC<->asset, and
+ * keying `findMarket` on BTC is what enforces that.
  */
 import type { DiscoveredMarket, OfferPlan, Side } from "@arkade-os/solver-discovery";
 import type { Asset, PaymentRail, RouteQuote, RouterContext } from "@arkade-os/sdk";
@@ -53,84 +19,48 @@ import { createOffer } from "../offer";
 import { findMarket, validatePlan, type PlanError } from "../markets";
 import { BTC_ASSET_ID } from "../store";
 
-/** This rail's id, and what `RouterPreferences.priority` ranks it by. */
 export const CROSS_ASSET_RAIL = "cross-asset";
 
-/**
- * How far the route has got. Persisted, because the two states need different
- * recovery: a `"quoted"` record whose offer never filled is cancelled, while a
- * `"filled"` one has the asset in the wallet and owes the recipient a send.
- */
+/** The two states need different recovery: a `"quoted"` record whose offer
+ *  never filled is cancelled; a `"filled"` one owes the recipient a send. */
 export type CrossAssetPhase = "quoted" | "filled";
 
-/** The offer this rail funded, and the payment it was funded for. */
 export interface CrossAssetSwap {
-    /** Where the route has got to; see {@link CrossAssetPhase}. */
     phase: CrossAssetPhase;
-    /** `createOffer`'s encoded offer — the only input `cancelOffer` needs. */
     offerHex: string;
-    /** The swap address the deposit went to. */
     address: string;
     swapPkScript: Uint8Array;
-    /** Sats deposited into the offer. */
     depositSats: number;
-    /** What the offer buys, and what the recipient is then paid. */
     asset: Asset;
-    /** Where the asset goes once the offer fills. */
     payTo: string;
     plan: OfferPlan;
 }
 
 export interface CrossAssetRailDeps {
-    /** Arkade Service REST URL — `createOffer` reads `getInfo()` from it. */
     arkServerUrl: string;
-    /**
-     * Markets to price and route over.
-     *
-     * Called once by `available()` and again by `quote()` — the router
-     * deliberately does not carry a result between them, so a rail that
-     * decided it could route must decide again when asked for a price. With
-     * `discoverMarkets` (1h cache) and `makeCachedFeedFetch` (30s, with
-     * in-flight collapsing) behind these, an `options()`-then-`route()` pass
-     * costs one round trip rather than two. Passing a bare registry fetch
-     * makes it four.
-     */
+    /** Called by `available()` and again by `quote()`; pass the caching
+     *  `discoverMarkets`, not a bare registry fetch. */
     discover(): Promise<DiscoveredMarket[]>;
-    /**
-     * Price the swap leg. Pass `quoteOffer` from `@arkade-os/solver-discovery`,
-     * partially applied with whatever feed fetch the app uses — `markets.ts`
-     * ships `makeCachedFeedFetch` for exactly this.
-     */
     quote(market: DiscoveredMarket, give: Side, wantAmount: bigint): Promise<OfferPlan>;
-    /** Spendable sats, for the deposit the plan asks for. */
     btcBalance(): Promise<bigint>;
-    /** The server's dust limit — `validatePlan` protects the BTC leg with it. */
     dust(): Promise<bigint>;
-    /** Write the offer down BEFORE it is funded: an offer funded without a
-     *  record is one `cancelOffer` cannot rebuild. */
+    /** Called twice, both before an irreversible step: `"quoted"` before the
+     *  offer is funded, `"filled"` after delivery and before the recipient is
+     *  paid. Keyed by `offerHex`, stable across both. */
     persist(swap: CrossAssetSwap): Promise<void>;
-    /** Resolve once a filler has delivered the asset to this wallet. */
+    /** Resolve once a filler has delivered the asset to this wallet.
+     *
+     *  OPEN QUESTION for review: is an injected promise the right seam, or
+     *  should the rail take a `watchOfferSwaps` handle? No deadline is applied
+     *  here, and resolving on a timeout rather than rejecting would make the
+     *  rail pay from whatever the wallet then holds. */
     awaitFill(swap: CrossAssetSwap): Promise<void>;
-    /** Sats the delivery leg carries. Defaults to the SDK's dust carrier. */
     carrierSats?: number;
-    /** Co-signer key override for the offer covenant. */
     emulatorPubkey?: string;
 }
 
-/** Sats the delivery output carries when nothing else says. Mirrors
- *  `Recipient.amount`'s own default. */
 const DEFAULT_CARRIER_SATS = 330;
 
-/**
- * The asset id as `createOffer` needs it, or undefined when it is not one.
- *
- * `Asset.assetId` is a hex STRING (that is what `Wallet.send` takes), and
- * `createOffer` wants an `AssetId` — it reads `.txid` and `.groupIndex` off it
- * to bind the covenant, so handing it the string would build an offer with an
- * undefined want-asset rather than fail. Ids are 34 bytes; `fromString` throws
- * on anything else, and returning undefined here lets `available()` drop the
- * rail instead of the router.
- */
 const parseAssetId = (assetId: string): assetExt.AssetId | undefined => {
     try {
         return assetExt.AssetId.fromString(assetId);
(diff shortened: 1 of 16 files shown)


Kukks:
@coderabbitai review

(Auto-review skipped because this PR is stacked on `feat/payment-rails-router` rather than the default branch — the check reports pass, but no review ran.)

Kukks:
Note for whoever reviews this: **CodeRabbit will not review this PR while it is stacked.** Its check reports `pass`, but the detail line reads `Review skipped: reviews are disabled for this base branch` — the org config only auto-reviews `master` and `next-version`, and an explicit `@coderabbitai review` does not override it. It should get a real review once #837 merges and this retargets `master`; until then a green CodeRabbit check here means nothing.

Kukks:
All five taken in `1fa1a647`. Thank you — finding 1 is a genuine hole and I had not seen it.

**1. No recovery state between fill and payment.** Correct, and the record actively lied: frozen at `"quoted"` after the fill landed, it points recovery at `cancelOffer` when what is actually true is that the asset is in the wallet and the recipient is still owed.

`CrossAssetSwap` now carries a `phase`, and `persist` is called at both transitions — `"quoted"` before the offer is funded, `"filled"` after delivery and before the recipient send. Both sit before an irreversible step, which is the same rule the send rails' single persist follows, so there is one principle rather than two.

Four tests, and the mutations confirm each is load-bearing: dropping the second persist fails 3, moving it after the send fails 2, leaving the phase at `"quoted"` fails 2.

**2. `bigint → number` narrowing.** Taken, and asserted rather than commented — a comment states the assumption, `Number.isSafeInteger` enforces it. You are right that BTC's supply fits; the fragile part is the *give side being BTC*, and a non-BTC deposit corridor is exactly the change that would slip past a comment. It now throws with the amount named. (Writing the test for it also surfaced that `validatePlan`'s `above-max` fires first on a normal market, so the test has to widen the bounds to reach the guard.)

**3. `planFor()` twice per route.** Agreed, non-blocking, and documented on `discover` rather than changed: the router deliberately does not carry a result from `available()` into `quote()`, and caching a plan inside the rail would mean serving a price the market may have moved off. The JSDoc now says it is called twice and that `discoverMarkets` (1h) plus `makeCachedFeedFetch` (30s, in-flight collapsing) make an `options()`-then-`route()` pass one round trip rather than four with a bare registry fetch.

**4. `fee` on a cross-asset route.** Taken where you pointed: the `RouteQuote.assets` JSDoc in `types.ts` now says that `fee` absorbs the purchase price because `total = amount + fee` must hold, and that `assets.spent` is the same number correctly labelled and is what a cost display or fee-threshold check should read. That is the right home — the explanation was in `crossAsset.ts`, where no consumer reads it.

**5. `awaitFill` deadline.** Documented on the dep, including the failure mode you named: resolving on a timeout rather than rejecting makes this rail attempt the recipient send against whatever the wallet holds at that moment, so a caller enforcing a deadline should reject.

---

One note on base: the review is against `b1a70d86`, and the branch has since been rebased onto #837's head — `1fa1a647`'s parents are the same three commits, so every finding applied cleanly and none was stale.

Also worth flagging for whoever merges: **CodeRabbit has not reviewed this PR at all.** Its check reports `pass`, but the detail reads `Review skipped: reviews are disabled for this base branch` — the org config only auto-reviews `master` and `next-version`, and an explicit `@coderabbitai review` does not override it. It should get a real pass once #837 merges and this retargets `master`.

Kukks:
Comment volume cut in `68cc5b3e` — style only, no logic, no test assertions changed.

**Before:** 481 comment-or-blank of 1339 added = 35.9% (comments alone 383 = 28.6%), with a 41-line block on `crossAsset.ts`.
**After:** 210 of 1068 = 19.7%. **Comments alone 112 = 10.5%.**

Both numbers for the same reason as #837: blanks are 98 of 1068 (9.2%) on their own, so the combined figure cannot approach 10% without deleting separators.

The two items flagged for your decision are still in place but tightened to the question rather than the argument — `awaitFill`'s seam and `fee` carrying the swap deposit each now read as an `OPEN QUESTION for review` of three or four lines, with the full reasoning left on this PR where it can be acted on.

Gate: ts-sdk 2914, swap 748, boltz-swap 617 — **4279 passing, 0 failing**, unchanged. Staged blobs verified biome-clean via `git show :<file>`, not the working tree.

packages/ts-sdk/src/payment/rails/arkAsset.ts
```diff
@@ -0,0 +1,47 @@
...
+ * They split on `available()` rather than `match()` because both legitimately
+ * match the address; which one can pay is a question about the amount, so
+ * `options()` returns exactly one and their relative priority is immaterial.
+ */
+export function arkAssetRail(): PaymentRail {
+    return {
+        id: "ark-asset",
+        match: (req) => arkTarget(req.raw) !== undefined,
+        // A malformed asset is `quote()`'s to name, not this rail's to drop.
+        available: (req) => assetsOf(req).length > 0,
```
pietro909, line 26: This rail returns true for any asset request even if the wallet has zero balance of that asset. While quote() or send() will catch it, returning true here makes the router include it in options(). Should it perform a quick balance check to match the 'ark' rail behavior?

pietro909 (commented):
Reviewing the latest head (68cc5b3e). Technical findings on recovery and narrowing are confirmed fixed. Adding three new observations for consideration:

packages/swap/src/payment/crossAsset.ts
```diff
@@ -0,0 +1,193 @@
...
+                    `${CROSS_ASSET_RAIL}: deposit of ${plan.deposit.atomic} does not fit a JS number`,
+                );
+            }
+            const depositSats = Number(plan.deposit.atomic);
+            const carrier = deps.carrierSats ?? DEFAULT_CARRIER_SATS;
+
+            return {
+                railId: CROSS_ASSET_RAIL,
+                // OPEN QUESTION for review: `total = amount + fee` forces the
+                // purchase price into `fee`, which reads as a service charge.
```
pietro909, line 140: This assumes the carrier is paid from the wallet's remaining balance. If the wallet was swept to fund the swap, the second send() will fail. Does the solver provide the carrier in the fill, or must the wallet maintain this buffer?

packages/swap/src/payment/crossAsset.ts
```diff
@@ -0,0 +1,193 @@
...
+                        await ctx.wallet.send({
+                            address: offer.address,
+                            amount: depositSats,
+                            extensions: [offer.extension],
+                        });
+                        emit({ status: "sent" });
+
+                        await deps.awaitFill(swap);
+                        // Before paying: "quoted" would say it never filled.
+                        await deps.persist({ ...swap, phase: "filled" });
```
pietro909, line 181: Floating emit. While sync today, it should be awaited to support future async-iterator handles or cross-process event sinks.

Kukks:
Heads-up for whoever picks this up: **this branch needs a rebase onto its base.**

`feat/payment-rails-router` moved from `25d62775` to `5ade00fe`, which addresses pietro909's review on #837 (four blockers, six warnings). That commit touches three files this PR also edits:

- `packages/swap/src/payment/solverOnchain.ts`
- `packages/swap/src/payment/solverLightning.ts`
- `packages/ts-sdk/src/payment/types.ts`

The overlaps to expect when rebasing:

- Both send rails changed shape inside `send()` — `awaitSettlement` is now wrapped, and a rejection resolves at `"sent"` instead of propagating. Any edit in that block will conflict.
- `SolverRendezvous` lost its `emulatorPubkey` field (`rendezvous.ts`), so anything reading it off the rendezvous no longer compiles.
- `ChainSource.getSpendingTx` takes a third argument, the spent output's `pkScript`.
- `OnchainSendProfile` gained a required `payoutPkScript`, and `onchainSendProfile()` now needs it on its input.

Deliberately not rebased here — flagging rather than touching this branch.

packages/swap/src/payment/crossAsset.ts
```diff
@@ -0,0 +1,193 @@
...
+                        const txid = await ctx.wallet.send({
```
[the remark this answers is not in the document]
Kukks: Fixed in `33c2a10f` — confirmed as a real double-send before fixing, and the fix is deliberately not the obvious one.

Persisting a terminal state *after* the payout would have moved the bug rather than closed it: a crash between `wallet.send` and that persist still leaves `"filled"`, whose documented meaning is "owes the recipient a send". So the attempt is recorded **before** it is made.

`CrossAssetPhase` is now `"quoted" | "filled" | "paying" | "settled"` (`crossAsset.ts:24-37`):

- `filled` — asset arrived, payout **not** broadcast. Safe to send.
- `paying` — payout **was** broadcast, outcome unknown. Resolve against chain state, never by resending on the record alone.
- `settled` — landed; `CrossAssetSwap.txid` carries the receipt.

`paying` is persisted at `:194` before the send, `settled` at `:200` after it.

The residual window is `paying`, and it is the one I chose: it costs recovery a chain lookup, where the old shape silently told a compliant consumer to pay twice. Closing it entirely needs an idempotent send, which this path does not have.

Reproduced first, then fixed: a test runs the documented recovery rule (resend iff `phase === "filled"`) over the record a *completed* payment leaves, and asserted `send` was never called — it failed with "expected spy to not be called at all, but actually been called 1 times". A companion test runs the same rule over the mid-flight `filled` record and asserts it *does* send, so the guard is not blanket.

8 mutations, all killed — including reverting to no-persist-after-payout, and moving the `paying` persist to after the send (the "same bug one step later" case).

`CrossAssetPhase` is now exported from `payment/index.ts`, so consumers can branch on it without string literals.

packages/ts-sdk/src/payment/rails/arkAsset.ts
```diff
@@ -0,0 +1,47 @@
...
+        quote: async (req, ctx: RouterContext): Promise<RouteQuote> => {
+            const address = arkTarget(req.raw)!;
+            const asset = resolveAssetAmount("ark-asset", req);
+            // Defaulted here so the quote states the sats that will leave.
+            const carrier = tryResolveSendAmount(req.raw, req.amount) ?? ASSET_CARRIER_SATS;
```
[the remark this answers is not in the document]
Kukks: Fixed in `33c2a10f`, and you were right that the default was the wrong answer for a *stated* value.

`arkAsset.ts:31` now validates an explicit amount and defaults only an absent one:

```ts
if (req.amount !== undefined) assertSendableAmount("ark-asset", req.amount);
const carrier = tryResolveSendAmount(req.raw, req.amount) ?? ASSET_CARRIER_SATS;
```

I kept `tryResolveSendAmount` for the BIP21 path rather than restructuring, so an `amount=` in the URI still resolves the same way; the new line only distinguishes "stated and invalid" from "not stated".

A test covers `0`, `-1` and `1.5` — it failed before with "promise resolved instead of rejecting". Two mutations (dropping the check, inverting the guard) are both killed.

Kukks:
## Findings A and G — both fixed in `33c2a10f`

### G — no terminal persist after payout (double-send)

Confirmed in the code before fixing, at the same place you and CodeRabbit both landed: exactly two `deps.persist` calls, and nothing after the recipient send.

The fix is **not** just adding a terminal phase. That would have moved the window rather than closed it — a crash between `wallet.send` and a post-send persist still leaves `"filled"`, which the docs define as "owes the recipient a send". So the attempt is recorded *before* it is made:

`CrossAssetPhase = "quoted" | "filled" | "paying" | "settled"` (`crossAsset.ts:24-37`), with `paying` persisted at `:194` before the send and `settled` at `:200` after, carrying `CrossAssetSwap.txid`.

The phases are now true of the moments they name:

- `filled` — payout **not** broadcast; safe to send.
- `paying` — payout **was** broadcast, outcome unknown; resolve against chain state, never by resending on the record.
- `settled` — landed, `txid` recorded.

The window I am left with is `paying`, chosen rather than overlooked: it costs recovery a chain lookup, where the previous shape silently instructed a compliant consumer to pay twice. Eliminating it needs an idempotent send primitive, which this path does not have — so the honest move is to name the ambiguity instead of hiding it.

Your read on the test was right, so I did not just extend the array. `crossAsset.test.ts` now has a test that runs the documented recovery rule (resend iff `phase === "filled"`) over the record a **completed** payment leaves and asserts nothing is sent — it failed against the old code with *"expected spy to not be called at all, but actually been called 1 times"*. A companion runs the same rule over the mid-flight `filled` record and asserts it **does** send, so the guard is not blanket. The ordering test now pins `persist:filled → persist:paying → send → persist:settled`.

### A — `CrossAssetPhase` not exported

Fixed: added to the export block in `packages/swap/src/payment/index.ts`, which `src/index.ts` re-exports with `export * from "./payment"`. Recovery handlers can narrow on it without string literals — which, as you noted, matters more now that a phase carries a "do not resend" meaning.

### Evidence

- **3 tests failed first** on the cross-asset change, 0 after; 1 on the `arkAsset` carrier fix.
- **8 mutations, 8 killed, 0 survived** — including reverting to no-persist-after-payout (the original bug), moving the `paying` persist to *after* the send (the "same bug one step later" case), a terminal record still labelled `filled`, and a terminal phase written without the txid.
- **Gate 4354 → 4359 passing**, 1 skipped, 0 failing, reconciling exactly to the 5 tests added.
- Comments 9.48% of added non-blank lines.

Also in this push, from CodeRabbit's second finding: `arkAssetRail` treated an explicit invalid `req.amount` as absent, so `{ assets: [a], amount: 0 }` silently used the 330-sat default carrier. A stated amount is now validated; only an absent one defaults.

PR #843, opened 2026-09-04
feat(swap): the three verbs, and the rails that route them

pietro909:
Implements **M7 — Swap verbs** of the swap-sdk-v2 track. Stacked on #842 (M6); retarget to `feat/v0.5` when that merges.

`pay`, `receive` and `exchange` on the v2 client, plus the two rails that put lightning and `arkade -> onchain` back into a payment app's reach. M7 adds no capability — everything the verbs call was built in M3–M6. What it adds is the fee ceiling; what it subtracts is vocabulary.

## The verbs

Each compiles a `QuoteInput`, calls `quote`, checks `quote.fee` against a ceiling, and calls `accept`.

```ts
await client.pay(bolt11)                                  // -> { kind: "swap", swap }
await client.pay(bc1Address, { amount, maxFee })          // -> { kind: "swap", swap }
await client.pay(arkAddress, { amount })                  // -> { kind: "payment", txid }
const r = await client.receive({ amount, via: "lightning" })   // r.artifact is non-optional
await client.exchange({ give: "BTC", take: "USD", amount, amountOn: "give" })
```

`pay` to a plain Arkade address **delegates** rather than rejecting: it classifies with core's own `arkTarget`, resolves the amount under core's own law, and settles through `wallet.send`. Rejecting it as `UnsupportedRoute` is cheaper here and dearer everywhere else — every product with one pay box would carry the branch. That is what widens the return to a union whose non-swap arm carries a txid and no swap id; manufacturing a `Swap` for same-asset, same-rail, rate-1 would be a swap record for something nothing swapped.

`receive` returns only after `accept()` has persisted, and its `ReceiveRequest = Swap & { artifact: Artifact }` is where that guarantee gets stated: you cannot show a payer an invoice whose claim secret is still in memory.

## The ceiling

`maxFee` is `{ amount: bigint; asset: AssetId }` on every verb and on the new `SwapPolicy.maxFee`, and the effective ceiling is the **minimum** of the two — a policy ceiling a call can raise is decorative. A bare `bigint` is refused as under-specified: the fee sits on the give leg on corridors and the take leg on asset swaps, so one number names no asset. `MaxFeeExceeded` is thrown between `quote` and `accept`, carrying the terms (`quoteId`, `asset`, `fee`, `maxFee`) and nothing funded.

## The rails

`lightningRail(client)` and `onchainSwapRail(client, { claimFeeRateSatVb })`, plus `createSwapPaymentRouter(wallet, client, config)` — the shape #811 deleted with `packages/boltz-swap`, rebuilt on the v2 client. The rails close over their own deps rather than re-opening `RouterContext`, so **nothing under `packages/ts-sdk/src/payment/` is edited**; they write M6's tagged `AssetSwapId` into a field core still declares `string`.

`available()` never quotes — it resolves, which is network-free — because a swap rail's `quote()` is an RFQ round trip that discloses an invoice and an amount, and an app that ranks then quotes several options would disclose to every solver.

`onchainSwapRail` **grosses the take leg up** by the claim fee. On `arkade -> onchain` the trader claims the solver's L1 HTLC itself and that claim's fee comes out of the payout, so a rail quoting the solver's spread alone would beat the collaborative exit on a fee it does not charge — and would under-deliver, which core's receiver-exact contract forbids. It asks the solver for `amount + claimFee` and reports the estimate inside `fee`, keeping `total === amount + fee`.

## The projection

Fourteen `Outcome` members onto four `PaymentStatus` literals, declared total — a partial map is where the collapse hides. It is lossy by design and the loss is recoverable: `refunded` and `lapsed` both land on `failed`, and the difference survives on the update's `error` as `SwapPaymentFailedError.outcome`. The rail goes terminal at `refunding`, not `refunded`, because holding terminality back would hang every `settled({ timeoutMs })` caller for a whole refund window. A handle observes a payment up to its first terminal outcome; every recovery past that — the `needs_recovery -> funded` backslide included — is observed on `client.onUpdate`, keyed by the tagged swap id that round-trips to `swaps()`.

## One stale premise, corrected

The plan claimed a grep for `PaymentRail|PaymentRouter|RouterContext` over `packages/swap` returns nothing. It was pinned at `a16d93b2`, which predates #837 (`fa3a2d1f`): `solver-lightning` and `solver-onchain` already ship from the root barrel — the same two corridors on the **v1 RFQ path**. M7's rails land beside them rather than replacing them. They are one generation apart, not two designs: the v1 rails take `discover`/`connect`/`persist`/`awaitSettlement`/`payoutPubkey`/`l1Network` from the app and own no persistence, drive, recovery or outcome vocabulary, where the v2 rails take one `SwapClient`. Retiring the `solver-*` pair is M8's — removing published root exports inside the milestone that ships no deprecations would be a breaking edit in the wrong place.

## Verification

- 49 new unit tests; swap suite **1280 passing** (from 1231).
- `pnpm --filter @arkade-os/sdk exec vitest run test/payment` green **unedited** — nothing under `packages/ts-sdk/` is touched.
- Full workspace `test:unit` green (ts-sdk 2977, swap 1280); build, lint and both typechecks clean.
- M6's `pending: "M7"` coverage marker is retired, leaving `InconsistentRoute` the one declared-inert member.
- `ONCHAIN_CLAIM_VSIZE` (152) is pinned against a real sizing pass in `onchainHtlc.test.ts`.

**Not run:** `test/e2e/verbs.test.ts` ships unrun. It needs the seconds-typed `swap-rfq` regtest profile; the checked-in `.env.regtest` is block-denominated, so `unilateralClaimDelay` refuses the stack's 20-block exit delay.

https://claude.ai/code/session_01NNWu98iCogpRogwMgrkyS1

### DIFF
--- a/packages/swap/package.json
+++ b/packages/swap/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@arkade-os/swap",
-    "version": "0.1.0-rc.0",
+    "version": "0.1.0-rc.1",
     "type": "module",
     "description": "Client-side Arkade Intents asset swaps: discover markets, quote, create/track/cancel offers, restore from chain.",
     "repository": {
--- a/packages/swap/src/client/client.ts
+++ b/packages/swap/src/client/client.ts
@@ -21,6 +21,11 @@
  * thing — it registers the swap it just persisted — and everything else the
  * lifecycle needs lives behind {@link createSwapDrive}.
  *
+ * From M7 the three verbs hang off the same object: `pay`, `receive` and
+ * `exchange` are `quote` -> fee ceiling -> `accept` and nothing else, and they
+ * add no capability — everything they call was already here. What they add is
+ * the ceiling; what they subtract is vocabulary.
+ *
  * From M6 the surface closes: `cancel`, `swaps`, `markets`, and `ClientDisposed`
  * enforced across every member after disposal. Disposal is terminal; what the
  * terminal gate refuses is a NEW act, and an `Unsubscribe` already handed out
@@ -60,6 +65,17 @@ import { feedFetch, quoteFromFeed, type FeedFetch } from "./quoteOffer";
 import { quoteViaRfq } from "./quoteRfq";
 import type { Quote, QuoteId, QuoteInput, RouteResolution } from "./quote";
 import { resolveRoute, type ResolvedRoute } from "./resolve";
+import {
+    exchange,
+    pay,
+    receive,
+    type ExchangeOptions,
+    type PayOptions,
+    type PayResult,
+    type ReceiveOptions,
+    type ReceiveRequest,
+    type VerbDeps,
+} from "./verbs";
 import { cardMarketOf, marketBackendOf, marketKeyOf, usableMarkets, type Market } from "./market";
 import { nostrTransportFactory, type RfqTransportFactory } from "./transport";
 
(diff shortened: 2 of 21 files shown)


Kukks:
Heads-up on an interaction with #838 (the asset dimension for `PaymentRouter`, targeting `master`). Not a blocker for this PR as it stands — `feat/v0.5` hasn't taken #838 yet — but worth pinning before the merge rather than after.

#838 adds `assets?: Asset[]` to `PaymentRequest`, and closes the silent drop that opens by making every BTC-only rail refuse an asset-bearing request: `ark`, `onchain`, `lightning`, `onchain-swap`, `solver-onchain` and `solver-lightning` all get `assetsOf(req).length > 0 -> false` at `available()`, with `assertNoAssets` at `quote()` as the backstop for a caller reaching a rail directly.

M7's two rails gate through `railAvailable` in `swapRail.ts`:

```ts
export const railAvailable = async (client, input) => {
    try { return (await client.resolve(input)).eligible > 0; } catch ...
};
```

Asset-blind, and correctly so on this base — `assetsOf` doesn't exist on it. But once v0.5 merges master after #838 lands, `lightning` and `onchain-swap` become the only rails in that barrel that *accept* an asset-bearing request: they'd pay the carrier sats, report success, and deliver none of the asset the caller asked for. That's the exact failure #838 exists to close, reintroduced by the two newest rails.

Cheap to close at the rebase — the same two lines each rail already has a slot for. Flagging now so it doesn't get lost in a merge that otherwise looks clean.

Two smaller notes:

- **Expect a conflict in `packages/swap/src/payment/index.ts`.** Both PRs append at the same anchor after the `solverLightning` export block. Export names are disjoint (`cross-asset` vs `lightning`/`onchain-swap`/`router`/`status`/`swapRail`), so it resolves by keeping both.
- **No objection to the corridor overlap you already call out.** M7's rails landing beside `solver-lightning`/`solver-onchain` rather than replacing them reads right to me, and retiring the pair in M8 is the correct place for it. Only noting that #838 edits both of those files, so the two generations will be visibly side by side for a release or so.

On the asset vocabulary itself: I've tagged `PaymentRequest.assets` and `RouteQuote.assets` `@experimental` on #838 (`b268a9b8`), so core's `assets?: Asset[]` and this client's `give`/`take`/`amountOn` can converge on v0.5 without that being a breaking change on 0.4.x. Happy to go either direction on which one wins — that decision reads better with both landed than blocked on it.

PR #847, opened 2026-09-07
test: e2e for the activity resolver and the payment rails, and the market bounds they depend on

Kukks:
## What this is

A coverage pass on three subsystems, which started by measuring what was
actually covered rather than assuming. One of the three turned out to be well
covered, one had a happy path only, and one had nothing. Measuring also turned
up a production defect, which is the one non-test change here.

## The inventory I found

Every swap e2e file that exists **does** run: the `swap` and `swap-rfq` CI jobs
pass an empty `test_files` and the package scripts glob
(`vitest run test/e2e --exclude 'test/e2e/rfq*.test.ts'` and
`vitest run test/e2e/rfq`). Read off the run logs for #845, not the job
conclusions: `swap/all` ran `swap`, `offerCancel`, `chainSource` (3 files, 13
tests) and `swap-rfq/all` ran `rfqVerbs`, `rfqDrive`, `rfqRegister` (3 files, 12
tests). The ts-sdk matrix enumerates instead of globbing, and it was complete —
25 files listed, 25 on disk, no orphans either way.

| Subsystem | Unit | e2e that runs | Verdict |
|---|---|---|---|
| Activity resolver | `ts-sdk/test/wallet/activity.test.ts`, `swap/test/activity.test.ts` | **none** — no e2e in either package referenced `buildActivities`, `getActivityHistory` or `ActivityRegistry` | gap |
| Payment rails | `swap/test/payment/*` (5 files) | one case: `rfqVerbs.test.ts` "the lightning rail through a real router", happy path | partial |
| Swap v2 | `swap/test/client/*` (~26 files) | `rfqVerbs`, `rfqDrive`, `offerCancel`, `chainSource` — including unhappy paths (fee ceiling, repeat cancel, refused ids, rebuild mismatch) | **already covered; nothing added** |

## The defect

`selectMarkets` takes a `wantAmount` ("trade size, checked against that side's
bounds") and `eligibleMarkets` never passed it, so every card's
`min_base_amount` / `max_base_amount` / `min_quote_amount` / `max_quote_amount`
was dead data — none of the four is read anywhere under `packages/swap/src`.
`resolve()` answered `eligible: 1` for a size 100x a card's ceiling.

That is what breaks the swap rails' documented self-healing. `payment/router.ts`
and `payment/onchainSwap.ts` both promise an out-of-range amount drops
`onchain-swap` at `available()` and the collaborative exit wins with no error,
and `available()` decides on `resolve().eligible > 0`. The existing unit test
asserts the fallback against a client hand-stubbed to `eligible: 0` — a value
the real client could not produce for an in-pair route — so it pinned a number
rather than the behaviour. In practice an out-of-range payment reached `quote()`
and failed at the solver instead of falling back to a working exit.

Only a take-side pin is forwarded; a give-side pin would need a price this read
deliberately does not have, so it stays unbounded. That asymmetry is documented
on the parameter.

**This is the one change here that is not a test.** It is separable: revert
`be29c318` and the three unit cases in `market.test.ts` / `resolve.test.ts` go
red, everything else stands.

## What was added

- `packages/ts-sdk/test/e2e/activity.test.ts` — real boarding, settle and
  offchain send, then the built-in resolvers against the shapes the wallet
  actually emits, plus the unhappy half: a resolver that throws, one whose
  `prepare()` rejects, one returning an empty `groupId` and a NaN amount, two
  claiming one group, and registry removal/override. **Added to the
  `settlement-delegation` group in `ci.yml`** — without that entry it would run
  nowhere while every job stayed green.
- `packages/swap/test/e2e/rfqRails.test.ts` — the two ways a rail declines: an
  out-of-range amount dropping `onchain-swap` so the real exit wins (priced by
  arkd's own fee schedule), a size under the card's floor doing the same, and a
  held `RouteQuote` outliving its `Quote` so `send()` fails `QuoteExpired`
  having funded nothing and written no record. The expiry wait polls the quote's
  own `expiresAt` rather than sleeping, so it does not race the clock it asserts
  on.
- Three unit cases pinning the bounds fix.
- Two Windows-only test-harness fixes, which took 2 of the swap package's 60
  suites out of any local run.

## Baseline vs post

Captured on `f6804f04` before any edit.

| | baseline | post |
|---|---|---|
| ts-sdk unit | 177 files, 3000 tests (2999 pass, 1 skip) | 177 files, 3000 tests (2999 pass, 1 skip) |
| swap unit | 60 files, **2 failing**, 1301 tests | 60 files, **0 failing**, 1318 tests |
| swap-rfq e2e | 3 files, 12 tests | 4 files, 16 tests |
| settlement-delegation e2e | 9 files, 25 tests | 10 files, 35 tests |

The two baseline unit failures were Windows-only and green on Linux CI:
`test/client/errors.test.ts` (the suite would not collect — `ENOENT scandir`
on a `/C:/…` path) and `test/exports.test.ts > the M8 disposition record >
binds each R name to the v2 declaration…` (a backslashed module path asserted
against a posix pattern).

+17 unit tests reconciles exactly: +14 from `errors.test.ts` now collecting at
all, +3 new.

## Mutations run

Each applied, pushed to a scratch branch, run in CI, then reverted. The e2e ones
needed the unit gate dropped on the scratch branch so the integration jobs would
run at all — the unit suite catches several of these first.

| Mutation | Test that caught it | Failure |
|---|---|---|
| Drop the `wantAmount` pass-through in `eligibleMarkets` | `market.test.ts > bounds the take leg against the card…` | `expected [ { card: {…(15)}, …(3) } ] to deeply equal []` |
| (same) | `resolve.test.ts > lands a size the cards do not serve there…` | `expected 1 to be +0` |
| Forward `amount?.value` regardless of side | `resolve.test.ts > leaves a give-side pin unbounded…` | `expected +0 to be 1` |
| `boardingResolver` match inverted | `activity.test.ts` — 5 cases | `expected undefined to be defined`, `expected false to be true`, `expected undefined to be -2000` |
| `railAvailable` returns true unconditionally | `rfqRails.test.ts` — 2 cases | `expected [ 'onchain-swap', 'onchain' ] to deeply equal [ 'onchain' ]` |
| `prepare()` rejection no longer isolated | `activity.test.ts > isolates one whose prepare rejects…` | `Error: could not load` propagated out of `getActivityHistory()` |
| Membership merge becomes last-writer-wins | `activity.test.ts > keeps the first writer's intent…` | `expected 'Second' to be 'First'` |
| `accept()` stops refusing an expired quote | `rfqRails.test.ts > fails the send with QuoteExpired…` | `expected Error: settle timeout to match object { name: 'QuoteExpired' }` |
| `prepare()` called but not awaited | `activity.test.ts > prepares before it resolves…` | `expected undefined to be 'Prepared'` |

Two of these changed the tests rather than confirming them, which is the point
of running them:

- The expiry case originally called `handle.settled()` unbounded. With
  `accept()`'s expiry check removed the swap funds, the handle never goes
  terminal, and the case burned its full 180s budget instead of reporting.
  It now waits with `settled({ timeoutMs: 30_000 })`.
- The prepare-ordering case originally pushed its marker on the synchronous
  first line of `prepare()`, so it passed even when `buildActivities` called
  `prepare()` without awaiting it (CodeRabbit spotted this). The first fix used
  a microtask boundary, which the mutation also survived — `await Promise.all`
  flushes those anyway. It now uses `setTimeout`.

## What I could not verify locally

The regtest stack on this machine belongs to another project and has been up for
36 hours; `container_name: arkd` is hardcoded in the compose file, so a second
stack cannot coexist, and its arkd has intent fees enabled (min 2000) while the
swap profile zeroes them — the existing e2e suites fail on it in `beforeAll`.
I did not tear it down. **The two new e2e files have therefore only ever run in
CI, not on my machine**; the unit, lint and typecheck results above are what I
verified locally.

One flake surfaced and was fixed rather than retried: the boarding-fallback case
predicted the ungrouped bucket would be keyed by the boarding txid, but the
natural key is `arkTxid || commitmentTxid || boardingTxid`, so once the settle's
commitment reached the row the lookup found nothing. It passed three runs before
failing one. It now finds the activity by its member transaction.

A separate, unrelated flake was observed once and cleared on re-run:
`offerCancel.test.ts > names the rebuild mismatch…` returned `needs_recovery`
instead of `cancelled` in a job whose log also showed `arkd not ready` and a DNS
lookup failure during bring-up. That file is untouched by this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,7 @@ jobs:
               test/e2e/arkade-script.test.ts
               test/e2e/contractWatchState.test.ts
               test/e2e/lookAhead.test.ts
+              test/e2e/activity.test.ts
           - package: ts-sdk
             group: exit-providers-rotation
             test_files: >-
--- a/packages/swap/src/client/market.ts
+++ b/packages/swap/src/client/market.ts
@@ -115,9 +115,14 @@ export const usableMarkets = (
  * is what makes a card's own `min/max` bounds mean something: without it a
  * snapshot serving the pair answers "eligible" for any size, and the swap rails'
  * documented self-healing — an out-of-range amount drops the rail at
- * `available()` and the collaborative exit wins — never fires. Only a take-side
- * pin is passed, because that is the side `wantSide` names; converting a
- * give-side pin would need the price this read deliberately does not have.
+ * `available()` and the collaborative exit wins — never fires.
+ *
+ * @param takeAmount The pin on the TAKE leg only, which is the side `wantSide`
+ *   names. Omitting it skips the bounds entirely — an unpinned read is not a
+ *   zero-sized trade. A GIVE-side pin is deliberately not converted and passed:
+ *   that would need a price this read does not have, so a give-side amount past
+ *   a card's ceiling still answers `eligible: 1` here and is refused by the
+ *   solver instead. That asymmetry is the known sharp edge of this parameter.
  */
 export const eligibleMarkets = (
     snapshot: DiscoverySnapshot,
--- a/packages/swap/src/client/market.ts
+++ b/packages/swap/src/client/market.ts
@@ -110,11 +110,20 @@ export const usableMarkets = (
  * `source` and not `discovery_pubkey` because the latter is the field a cached
  * card carries unvalidated: filtering trust on untrusted content would give the
  * allowlist the shape of a check and none of the effect.
+ *
+ * `takeAmount` is the pinned trade size on the leg the trader receives, and it
+ * is what makes a card's own `min/max` bounds mean something: without it a
+ * snapshot serving the pair answers "eligible" for any size, and the swap rails'
+ * documented self-healing — an out-of-range amount drops the rail at
+ * `available()` and the collaborative exit wins — never fires. Only a take-side
+ * pin is passed, because that is the side `wantSide` names; converting a
+ * give-side pin would need the price this read deliberately does not have.
  */
 export const eligibleMarkets = (
     snapshot: DiscoverySnapshot,
     legs: { give: DiscoveryLeg; take: DiscoveryLeg },
     policy?: SwapPolicy,
+    takeAmount?: bigint,
 ): MarketCandidate[] => {
     const markets = usableMarkets(snapshot, policy);
 
@@ -135,6 +144,7 @@ export const eligibleMarkets = (
             // The trader receives the take leg, so that side must be one the
             // solver can pay out; a direction nobody solves yields no market.
             wantSide: side === "base" ? "quote" : "base",
+            ...(takeAmount === undefined ? {} : { wantAmount: takeAmount }),
         }).map((card) => ({
             card,
             give: side,
--- a/packages/swap/src/client/resolve.ts
+++ b/packages/swap/src/client/resolve.ts
@@ -227,7 +227,15 @@ export const resolveRoute = async (
     const amount = pinAmount(input, take.instrument);
 
     // 7. The market, after the policy filters that must run before disclosure.
-    const candidates = eligibleMarkets(snapshot, { give: giveLeg, take: takeLeg }, deps.policy);
+    //    A take-side pin also bounds-checks the card, so a size no solver on
+    //    this snapshot serves is `eligible: 0` here rather than an RFQ round
+    //    trip that discloses the amount only to be refused.
+    const candidates = eligibleMarkets(
+        snapshot,
+        { give: giveLeg, take: takeLeg },
+        deps.policy,
+        amount?.on === "take" ? amount.value : undefined,
+    );
     const market = chooseMarket(candidates, deps.policy);
 
     const resolution: RouteResolution = {
--- a/packages/swap/test/client/errors.test.ts
+++ b/packages/swap/test/client/errors.test.ts
@@ -1,5 +1,6 @@
 import { readFileSync, readdirSync } from "node:fs";
 import { join } from "node:path";
+import { fileURLToPath } from "node:url";
 import { describe, expect, it } from "vitest";
 import { AddressMismatch, SwapRefusal } from "../../src/rfq";
 import { SwapDriveRefusedError } from "../../src/client/drive";
@@ -167,7 +168,8 @@ const COVERAGE = {
 /** Every `.ts` under `src/`, concatenated — what "has a throwing site" is read
  * from, so the map cannot claim coverage the code does not have. */
 const SOURCES = ((): string => {
-    const root = new URL("../../src", import.meta.url).pathname;
+    // `fileURLToPath`, not `.pathname`: the latter yields `/C:/…` on Windows.
+    const root = fileURLToPath(new URL("../../src", import.meta.url));
     const walk = (dir: string): string[] =>
         readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
             const path = join(dir, entry.name);
--- a/packages/swap/test/client/market.test.ts
+++ b/packages/swap/test/client/market.test.ts
@@ -102,6 +102,17 @@ describe("the eligible set", () => {
         );
     });
 
+    it("bounds the take leg against the card, and only when a pin names it", () => {
+        const snapshot = snapshotOf([lightningCard]);
+        const legs = { give: ARKADE_BTC, take: LIGHTNING_BTC } as const;
+        // The card serves 1000..50_000_000 on either side.
+        expect(eligibleMarkets(snapshot, legs, undefined, 100_000n)).toHaveLength(1);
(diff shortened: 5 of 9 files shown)


Kukks:
Thanks -- both nits read correctly, and one of them is now taken.

**JSDoc on `takeAmount`** (`d67b2a67`): taken. The prose said a give-side pin is
not converted but never said what that costs a caller, so the parameter now
names the asymmetry directly: a give-side amount past a card's ceiling still
answers `eligible: 1` here and is refused by the solver rather than by this
read. That is the sharp edge, stated where someone will hit it.

**`as unknown as DiscoveredMarket`** in the e2e card builder: left as-is,
deliberately. `rfqVerbs.test.ts` and `rfqDrive.test.ts` already build their
discovery cards through the same cast, so removing it in one file would make
the odd one out rather than raise the bar. If it is worth tightening it is
worth tightening across the three together, which is a separate change.

Two corrections to the record, since the review was taken at `bdb473d` and the
head has moved twice since:

- The expiry case no longer calls `handle.settled()` unbounded. Mutation-testing
  it showed the failure mode was a stall, not a report: with `accept()`'s expiry
  check removed the swap funds, the handle never goes terminal, and the case
  burned its full 180s budget. It now waits with `settled({ timeoutMs: 30_000 })`,
  and the same mutation fails in 32s with
  `expected Error: settle timeout to match object { name: 'QuoteExpired' }`.
- `UNDER_FLOOR_SATS` was 400 in the reviewed commit and the case failed in CI on
  a correct rail -- 400 plus the 304 sat claim-fee gross-up is inside a 100 sat
  floor, so `onchain-swap` rightly stayed available. The onchain card now serves
  5_000..50_000 and the size under test is 1_000, which is clear of the rail's
  own dust check, so it is the card refusing. The lightning card keeps its low
  floor because the expiry case quotes a 990 sat invoice against it.

On the shared-registry fragility note: agreed, and it is why every mutation site
restores in `finally`. Vitest runs a file's suites sequentially by default and
this file sets no `concurrent`, so the ordering assumption holds today; the
`finally` blocks are what keep it from mattering if that ever changes.

Kukks:
@coderabbitai review

Kukks:
@coderabbitai review

Kukks:
@coderabbitai good catch, and it went further than the diff you saw.

You were right that `seen.push("prepare")` on the synchronous first line meant
the case asserted call order rather than the await. I took the fix, then
mutation-tested it by changing `buildActivities` to call `prepare()` without
awaiting it — and the case still passed. The microtask boundary is not enough
here: the map callback is `async`, so the surrounding `await Promise.all(...)`
flushes a microtask anyway and the correlation data was in place by the time
`resolve` ran.

It now uses `await new Promise((r) => setTimeout(r, 0))`, which outlives that
flush. Re-running the same unawaited-`prepare` mutation, the case fails with:

```
× activity.test.ts > prepares before it resolves, and can correlate on what prepare loaded
  AssertionError: expected undefined to be 'Prepared'
```

So the finding was real and is now actually closed rather than nominally closed.
`7e7dcba1` and `3edaeb13`.

Kukks:
One correction to the follow-up's ledger: the JSDoc item is **not still open** —
it was taken in `d67b2a67`, which landed after the `0fb5b78` you reviewed at, so
it fell outside the increment. On the current head, `eligibleMarkets` carries:

```
 * @param takeAmount The pin on the TAKE leg only, which is the side `wantSide`
 *   names. Omitting it skips the bounds entirely — an unpinned read is not a
 *   zero-sized trade. A GIVE-side pin is deliberately not converted and passed:
 *   that would need a price this read does not have, so a give-side amount past
 *   a card's ceiling still answers `eligible: 1` here and is refused by the
 *   solver instead. That asymmetry is the known sharp edge of this parameter.
```

Three commits also landed after that one, all test-only:

- `5a4ac643` / `3edaeb13` — CodeRabbit found that the prepare-ordering case
  asserted call order rather than the await, because its marker was pushed on
  the synchronous first line of `prepare()`. The first fix used a microtask
  boundary and the unawaited-`prepare` mutation survived it (`await Promise.all`
  flushes microtasks anyway); it now uses `setTimeout`, and the same mutation
  fails with `expected undefined to be 'Prepared'`.
- `7e7dcba1` — the boarding-fallback case flaked. It predicted the ungrouped
  bucket would be keyed by the boarding txid, but the natural key is
  `arkTxid || commitmentTxid || boardingTxid`, so once the settle's commitment
  reached the history row the lookup found nothing. Whether it had by the time
  the case ran was a race with the indexer, which is why it passed three runs
  before failing one. It now finds the activity by its member transaction and
  asserts it carries no intent, which is the claim being made.

Current head `7e7dcba1`: all seven checks green, `activity.test.ts` 10/10 and
`rfqRails.test.ts` 4/4, both confirmed by name in the run logs.

PR #851, opened 2026-09-07
fix(sdk): inject the fee source into onchainRail, not a wallet type parameter

Kukks:
## What changed

`RouterContext.wallet` is now `IWallet`, with no type parameter anywhere. The one rail member that needed more than `IWallet` — `onchainRail`'s single `arkProvider.getInfo()` read, used to price the collaborative exit — becomes a **required constructor dependency**:

```ts
export function onchainRail(deps: { feeInfo: () => Promise<FeeInfo> }): PaymentRail
```

Both `createDefaultPaymentRouter` factories already hold a concrete `Wallet`, so they wire it and their own signatures are unchanged:

```ts
.use(onchainRail({ feeInfo: walletFeeSource(wallet) }))
```

`Ramps` already takes `IWallet`, so `new Ramps(ctx.wallet)` inside the rail is unaffected.

## This supersedes the generic design, on purpose

Earlier commits on this branch introduced `RouterContext<W extends IWallet = Wallet>` and parameterized `PaymentRail<W>` and `PaymentRouter<W>`. **Both review bots reviewed that shape.** It has been replaced deliberately — please read this as a new design rather than diffing it against the earlier review.

The measurement that decided it. Across all rails in all three packages, `ctx.wallet` is read 12 times:

- 7 are `ctx.wallet.send(...)`, which is on `IWallet`
- 4 pass the whole wallet to functions that **already take `IWallet`**: `createOffer`, `requestLightningSend`, `requestOnchainSend`, `new Ramps`
- **exactly 1** reads `.arkProvider`

A type parameter threaded through three exported types, propagating into every rail signature in two downstream packages, existed to carry that single member. One constructor argument carries it instead, and the type parameter disappears.

## Why the dependency is required, not optional

The boltz rails use the *optional context* pattern — `packages/boltz-swap/src/payment/lightning.ts:34` does `if (ctx.swaps == null) return false`, silently making the rail unavailable. This PR deliberately does **not** follow that precedent for `feeInfo`.

The two cases are not alike. A swap client being absent is a genuine deployment state: an Arkade-only app legitimately has no Boltz, and the right behaviour is for the Lightning rail to be unavailable. A fee source the router cannot obtain is a *wiring bug* — and a fee silently going missing on an offboard is a money-path failure. An optional `deps?:` with a fallback would reintroduce exactly the silent rail-drop this design exists to prevent, so the rail fails to build instead.

## `PaymentRail` goes back to method syntax

Its members were changed to function properties earlier on this branch purely so `strictFunctionTypes` would make `PaymentRail<Wallet>` non-assignable to `PaymentRail<IWallet>`. With no type parameter that variance concern no longer exists, so they revert to method syntax — the original public shape, and the friendlier one for an external implementer using class methods. Keeping the properties would be a gratuitous change to an exported interface with no remaining justification. The comment documenting the variance rule is removed with it.

## Test plan

The `.test-d.ts` typecheck harness added by this PR stays — `packages/ts-sdk/tsconfig.typecheck.json` plus the `typecheck` block in `vitest.config.ts`, without which a `.test-d.ts` assertion is a silent no-op. Its assertions now pin the injection design:

- `RouterContext["wallet"]` is exactly `IWallet`
- a `ServiceWorkerWallet` (IWallet-only) is accepted as the router's wallet
- every rail registers on a router built from an IWallet-only wallet
- `onchainRail()` and `onchainRail({})` both fail to compile (`@ts-expect-error`)
- `IWallet` has no `arkProvider`

The harness was verified to bite rather than assumed: a deliberately false assertion was added, the suite reddened (`Type Errors 1 failed`, exit 1), and it was removed again.

Three mutations, each killed:

| mutation | result |
| --- | --- |
| `deps?:` optional, falling back to `ctx.wallet.arkProvider` | `TypeCheckError: Unused '@ts-expect-error' directive.` — `tsc --noEmit` on src still passed, so only the test catches this |
| drop `feeInfo` from the core factory call site | `src/payment/index.ts(36,14): error TS2554: Expected 1 arguments, but got 0.` |
| revert `RouterContext.wallet` to concrete `Wallet` | 3 assertions fail, including `Type 'ServiceWorkerWallet' is missing the following properties from type 'Wallet'` |

Gate, unchanged from the pre-change baseline measured on the same base: `pnpm -r build`, `pnpm -r typecheck` and `pnpm -r lint` all exit 0; `pnpm -r test:unit` is 233 files / 4377 passed / 1 skipped / 0 failed.

## Follow-ups from review (975c3420)

- **`onchainRail` throws if `feeInfo` is missing or not a function.** Types do not bind JavaScript callers, and `options()` only try/catches `available()` — `quote()` is called unwrapped, so the router does not re-rank around a throw. Without the guard a JS caller who omits the dep fails mid-payment instead of at wiring time, which is the failure this design exists to prevent.
- **`walletFeeSource(wallet)` is exported next to the rail** and used by both factories, so the `arkProvider.getInfo().fees` reach lives in one place and third-party callers have canonical wiring.
- **`typecheck.include` is pinned to the tsconfig's scope, plus `exclude: ["test/e2e/**"]`.** Vitest's default glob was wider than `tsconfig.typecheck.json`: a false assertion planted under `test/e2e/` was collected, reported `Type Errors no errors`, and passed with exit 0. Pinning the include alone does not fix it — that path still matches `test/**/*.test-d.ts` — so the exclude is what makes collection a subset of the tsc program.
- **`typecheck:types` added and wired into CI.** The config that makes the assertions bite had no direct gate; `typecheck` builds from `tsconfig.json`, which excludes `test/`.
- **Stubs use `{} as FeeInfo`** rather than `as never`.

Gate: build / typecheck / typecheck:types / lint all exit 0; unit 233 files, 4377 passed, 1 skipped, 0 failed.

### Note on diff size

The design itself is smaller than the generic it replaces: 4 source files, +13/−6, against 5 files and +26/−19. The total is larger — 12 files, +123/−30, against 8 files and +79/−19 — for two reasons worth naming rather than hiding. Making the dependency required forces all ten existing `onchainRail()` call sites across three test files to be rewired; an optional dependency would have left them untouched and produced a smaller total, but that is precisely the design being rejected here. The review follow-ups then added the runtime guard, the promoted fee source, and the harness gate. Both are deliberate trades: the churn buys a rail that cannot be built unpriced, and a type harness that cannot report a green it did not earn.

### DIFF
--- a/packages/swap/test/payment/solverOnchain.test.ts
+++ b/packages/swap/test/payment/solverOnchain.test.ts
@@ -6,7 +6,7 @@
  */
 import { describe, expect, it, vi } from "vitest";
 import type { DiscoveredMarket } from "@arkade-os/solver-discovery";
-import { PaymentRouter, onchainRail, type RouterContext } from "@arkade-os/sdk";
+import { PaymentRouter, onchainRail, type FeeInfo, type RouterContext } from "@arkade-os/sdk";
 import {
     SOLVER_ONCHAIN_RAIL,
     solverOnchainRail,
@@ -253,7 +253,7 @@ describe("solverOnchainRail.available", () => {
 
 describe("the router drops this rail rather than failing the payment", () => {
     /** A request the collaborative-exit rail can price without a live wallet. */
-    const feeInfo = async () => ({}) as never;
+    const feeInfo = async () => ({}) as FeeInfo;
     const routerCtx = (rail: ReturnType<typeof solverOnchainRail>) =>
         new PaymentRouter({
             wallet: {} as never,
--- a/packages/swap/test/payment/solverOnchain.test.ts
+++ b/packages/swap/test/payment/solverOnchain.test.ts
@@ -253,13 +253,13 @@ describe("solverOnchainRail.available", () => {
 
 describe("the router drops this rail rather than failing the payment", () => {
     /** A request the collaborative-exit rail can price without a live wallet. */
-    const arkProvider = { getInfo: async () => ({ fees: {} }) };
+    const feeInfo = async () => ({}) as never;
     const routerCtx = (rail: ReturnType<typeof solverOnchainRail>) =>
         new PaymentRouter({
-            wallet: { arkProvider } as never,
+            wallet: {} as never,
             prefs: { priority: [SOLVER_ONCHAIN_RAIL, "onchain"] },
         })
-            .use(onchainRail())
+            .use(onchainRail({ feeInfo }))
             .use(rail);
 
     it("ranks the solver route first when a card takes the send", async () => {
--- a/packages/ts-sdk/src/payment/index.ts
+++ b/packages/ts-sdk/src/payment/index.ts
@@ -6,12 +6,12 @@ export * from "./handle";
 export { PaymentRouter, AmbiguousRouteError } from "./router";
 export { arkRail } from "./rails/ark";
 export { arkAssetRail, ASSET_CARRIER_SATS } from "./rails/arkAsset";
-export { onchainRail } from "./rails/onchain";
+export { onchainRail, walletFeeSource } from "./rails/onchain";
 
 import { PaymentRouter } from "./router";
 import { arkRail } from "./rails/ark";
 import { arkAssetRail } from "./rails/arkAsset";
-import { onchainRail } from "./rails/onchain";
+import { onchainRail, walletFeeSource } from "./rails/onchain";
 import type { Wallet } from "../index";
 
 /**
@@ -33,5 +33,5 @@ export function createDefaultPaymentRouter(wallet: Wallet): PaymentRouter {
     })
         .use(arkRail())
         .use(arkAssetRail())
-        .use(onchainRail({ feeInfo: async () => (await wallet.arkProvider.getInfo()).fees }));
+        .use(onchainRail({ feeInfo: walletFeeSource(wallet) }));
 }
--- a/packages/ts-sdk/src/payment/index.ts
+++ b/packages/ts-sdk/src/payment/index.ts
@@ -33,5 +33,5 @@ export function createDefaultPaymentRouter(wallet: Wallet): PaymentRouter {
     })
         .use(arkRail())
         .use(arkAssetRail())
-        .use(onchainRail());
+        .use(onchainRail({ feeInfo: async () => (await wallet.arkProvider.getInfo()).fees }));
 }
--- a/packages/ts-sdk/src/payment/rails/onchain.ts
+++ b/packages/ts-sdk/src/payment/rails/onchain.ts
@@ -1,5 +1,4 @@
 import type { PaymentRail, RouterContext } from "../types";
-import type { Wallet } from "../../index";
 import { btcTarget } from "../targets";
 import { assertNoAssets, assetsOf, resolveSendAmount } from "../amount";
 import { makeHandle } from "../handle";
@@ -57,14 +56,18 @@ function grossUpOffboard(
  *
  * An explicit amount is mandatory. To sweep the full balance, call
  * `Ramps.offboard(address, feeInfo)` directly — the router has no amountless path.
+ *
+ * @param deps.feeInfo - Source of the operator's fee schedule. Required: a rail
+ * that cannot price an offboard must not be constructible — a silently dropped
+ * rail on the money path is worse than a wiring error at startup.
  */
-export function onchainRail(): PaymentRail<Wallet> {
+export function onchainRail(deps: { feeInfo: () => Promise<FeeInfo> }): PaymentRail {
     return {
         id: "onchain",
         match: (req) => btcTarget(req.raw) !== undefined,
         // BTC only: an Arkade asset has no L1 representation to offboard to.
         available: (req) => assetsOf(req).length === 0,
-        quote: async (req, ctx: RouterContext<Wallet>) => {
+        quote: async (req, ctx: RouterContext) => {
             assertNoAssets("onchain", req);
             const address = btcTarget(req.raw)!;
             // Reject missing/zero/fractional amounts up front: 0 sats would
@@ -72,7 +75,7 @@ export function onchainRail(): PaymentRail<Wallet> {
             const amt = resolveSendAmount("onchain", req.raw, req.amount);
             // Priced here rather than in send() so the quote carries a real fee;
             // the same FeeInfo is reused at settlement, so the two cannot drift.
-            const { fees } = await ctx.wallet.arkProvider.getInfo();
+            const fees = await deps.feeInfo();
             const script = hex.encode(offboardDestinationScript(address));
             const { gross, fee } = grossUpOffboard(amt, fees, script);
             return {
--- a/packages/ts-sdk/src/payment/rails/onchain.ts
+++ b/packages/ts-sdk/src/payment/rails/onchain.ts
@@ -1,4 +1,5 @@
 import type { PaymentRail, RouterContext } from "../types";
+import type { Wallet } from "../../index";
 import { btcTarget } from "../targets";
 import { assertNoAssets, assetsOf, resolveSendAmount } from "../amount";
 import { makeHandle } from "../handle";
@@ -57,13 +58,13 @@ function grossUpOffboard(
  * An explicit amount is mandatory. To sweep the full balance, call
  * `Ramps.offboard(address, feeInfo)` directly — the router has no amountless path.
  */
-export function onchainRail(): PaymentRail {
+export function onchainRail(): PaymentRail<Wallet> {
     return {
         id: "onchain",
         match: (req) => btcTarget(req.raw) !== undefined,
         // BTC only: an Arkade asset has no L1 representation to offboard to.
         available: (req) => assetsOf(req).length === 0,
-        quote: async (req, ctx: RouterContext) => {
+        quote: async (req, ctx: RouterContext<Wallet>) => {
             assertNoAssets("onchain", req);
             const address = btcTarget(req.raw)!;
             // Reject missing/zero/fractional amounts up front: 0 sats would
--- a/packages/ts-sdk/src/payment/rails/onchain.ts
+++ b/packages/ts-sdk/src/payment/rails/onchain.ts
@@ -5,7 +5,14 @@ import { makeHandle } from "../handle";
 import { Ramps, offboardDestinationScript } from "../../wallet/ramps";
 import { Estimator } from "../../arkfee";
 import type { FeeInfo } from "../../providers/ark";
+import type { Wallet } from "../../index";
 import { hex } from "@scure/base";
+
+/** The canonical fee source for {@link onchainRail}: the operator's schedule, as
+ *  the wallet already reads it. Keeps the one `arkProvider` reach in one place,
+ *  so an app registering the rail itself need not know where fees live. */
+export const walletFeeSource = (wallet: Wallet) => async (): Promise<FeeInfo> =>
+    (await wallet.arkProvider.getInfo()).fees;
 
 /** Iterations allowed when solving the gross-up fixpoint. A fee schedule charging
  *  less than one sat per extra sat converges in a handful of rounds; the cap only
@@ -62,6 +69,11 @@ function grossUpOffboard(
  * rail on the money path is worse than a wiring error at startup.
  */
 export function onchainRail(deps: { feeInfo: () => Promise<FeeInfo> }): PaymentRail {
+    // Types do not bind JavaScript callers, and the router does not re-rank after
+    // a quote() throws — so refuse at wiring time rather than mid-payment.
+    if (typeof deps?.feeInfo !== "function") {
+        throw new Error("onchain rail: a feeInfo source is required");
+    }
     return {
         id: "onchain",
         match: (req) => btcTarget(req.raw) !== undefined,
--- a/packages/ts-sdk/src/payment/types.ts
+++ b/packages/ts-sdk/src/payment/types.ts
@@ -1,4 +1,4 @@
-import type { Asset, IWallet, Recipient, Wallet } from "../index";
+import type { Asset, IWallet, Recipient } from "../index";
 
 export type PaymentStatus = "pending" | "sent" | "settled" | "failed";
 
@@ -86,8 +86,10 @@ export interface RouterPreferences {
     tieBreak?: "first" | "require-choice";
 }
 
-export interface RouterContext<W extends IWallet = Wallet> {
-    wallet: W;
+export interface RouterContext {
+    /** A rail needing more than this takes it as a constructor dep, rather than
+     *  narrowing the context every other rail shares. */
+    wallet: IWallet;
     /** Loosely typed in core to avoid a dependency on boltz-swap; swap rails cast it. */
     swaps?: unknown;
     prefs: RouterPreferences;
@@ -108,16 +110,13 @@ export interface PaymentRequest {
 }
 
 /** A payment rail — registered by id, mirrors the ActivityRegistry resolver shape. */
-// Properties, not methods: property parameters are contravariant under
-// `strictFunctionTypes`, so a `PaymentRail<Wallet>` cannot register on a
-// `PaymentRouter<IWallet>`. Method syntax is bivariant and would allow it.
-export interface PaymentRail<W extends IWallet = Wallet> {
+export interface PaymentRail {
     id: string;
     /** Classification only — amount-blind; takes the request for uniformity. */
-    match: (req: PaymentRequest, ctx: RouterContext<W>) => boolean;
+    match(req: PaymentRequest, ctx: RouterContext): boolean;
     /** Availability gate — where a rail drops itself for an out-of-limits amount. */
-    available?: (req: PaymentRequest, ctx: RouterContext<W>) => boolean | Promise<boolean>;
-    quote: (req: PaymentRequest, ctx: RouterContext<W>) => Promise<RouteQuote>;
+    available?(req: PaymentRequest, ctx: RouterContext): boolean | Promise<boolean>;
+    quote(req: PaymentRequest, ctx: RouterContext): Promise<RouteQuote>;
 }
 
 export interface PaymentOption {
--- a/packages/ts-sdk/src/payment/types.ts
+++ b/packages/ts-sdk/src/payment/types.ts
@@ -108,8 +108,9 @@ export interface PaymentRequest {
 }
 
 /** A payment rail — registered by id, mirrors the ActivityRegistry resolver shape. */
-// Properties, not methods: method parameters stay bivariant even under
-// `strictFunctionTypes`, letting a `PaymentRail<Wallet>` onto a `Router<IWallet>`.
+// Properties, not methods: property parameters are contravariant under
+// `strictFunctionTypes`, so a `PaymentRail<Wallet>` cannot register on a
+// `PaymentRouter<IWallet>`. Method syntax is bivariant and would allow it.
 export interface PaymentRail<W extends IWallet = Wallet> {
     id: string;
     /** Classification only — amount-blind; takes the request for uniformity. */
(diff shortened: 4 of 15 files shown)


packages/ts-sdk/src/payment/types.ts
```diff
@@ -108,13 +108,15 @@ export interface PaymentRequest {
 }
 
 /** A payment rail — registered by id, mirrors the ActivityRegistry resolver shape. */
-export interface PaymentRail {
+// Properties, not methods: method parameters stay bivariant even under
+// `strictFunctionTypes`, letting a `PaymentRail<Wallet>` onto a `Router<IWallet>`.
```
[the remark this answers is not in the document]
Kukks: Valid finding, applied in 12bd749a.

Every clause in the original was true, but "letting a `PaymentRail<Wallet>` onto a `Router<IWallet>`" had no explicit subject, so it could attach to the property syntax it was defending rather than the method syntax it was warning against. For a comment whose only job is to stop someone reverting to method syntax, that ambiguity is the whole failure mode.

Took the substance of your suggestion in three lines rather than four, to stay inside this repo's comment budget:

```ts
// Properties, not methods: property parameters are contravariant under
// `strictFunctionTypes`, so a `PaymentRail<Wallet>` cannot register on a
// `PaymentRouter<IWallet>`. Method syntax is bivariant and would allow it.
```

Both directions are stated explicitly now. Both are also asserted in `test/payment/routerContext.test-d.ts` rather than left to the comment: widening `onchainRail` to `PaymentRail<IWallet>` fails that file with `Unused '@ts-expect-error' directive`.

Kukks:
Confirmed the substantive finding by symbol search, and it is correct. All five sibling rails return a bare `PaymentRail`, which under the new default resolves to `PaymentRail<Wallet>`:

- `packages/boltz-swap/src/payment/lightning.ts:27` — `lightningRail(): PaymentRail`
- `packages/boltz-swap/src/payment/onchain-swap.ts:58` — `onchainSwapRail(): PaymentRail`
- `packages/swap/src/payment/crossAsset.ts:79` — `crossAssetRail(deps): PaymentRail`
- `packages/swap/src/payment/solverLightning.ts:67` — `solverLightningRail(deps): PaymentRail`
- `packages/swap/src/payment/solverOnchain.ts:65` — `solverOnchainRail(deps): PaymentRail`

Your reading of what they need is right too — I enumerated every `ctx.wallet` access across all three packages before writing this PR, and none of the five reaches past `IWallet`: the three `@arkade-os/swap` rails hand the wallet to `createOffer`/`requestLightningSend`/`requestOnchainSend`, whose parameters are already declared `IWallet`; `onchainSwapRail` only calls `send`; `lightningRail` never touches `ctx.wallet` at all. `onchain` is the sole rail in the monorepo that needs the concrete class.

Taking this as the tracked follow-up rather than folding it in here, which your review allows. The reason for keeping it separate: this PR was scoped to the core type change, and the annotation touches two further published packages. It is a safe change in the direction that matters — `PaymentRail<IWallet>` is assignable to `PaymentRail<Wallet>` by contravariance, so a default-router consumer is unaffected, and that direction is already asserted in `routerContext.test-d.ts` ("keeps an IWallet-only rail usable on a concrete-Wallet router").

The same gap exists one layer further out and is worth stating together: the rails wallet#944 defines in its own `sendRouter.ts` are also annotated bare `PaymentRail` and will need `PaymentRail<IWallet>` for the same reason. Recorded on arkade-os/wallet#950 so that lands with the cast removal.

Two small corrections for the record, since both matter if someone follows your citations. The line numbers are off: `ark.ts:18` is the `arkRail` signature rather than the `quote` arrow, `arkAsset.ts:27` is `available`, the `arkProvider` read is at `onchain.ts:75` not `:74`, and the `new PaymentRouter(...)` in the type test is at `:21` not `:22`. The findings themselves all hold — I located each by symbol. And on the "redundant annotation" nit: the explicit `ctx: RouterContext<IWallet>` on the `quote` arrows is indeed inferable, but I have kept it deliberately, because it is what makes the rail's wallet requirement legible at the call site rather than only in the return type.

Kukks:
@coderabbitai review

Heads-up for reviewers: this branch was force-pushed and the design was **replaced**, not amended. Your earlier reviews were against `12bd749a`, which used a generic (`RouterContext<W extends IWallet = Wallet>`, `PaymentRail<W>`, `PaymentRouter<W>`). That approach was rejected in favour of construction-time dependency injection.

New head is `a6c77a25`. In short:

- `RouterContext.wallet` is plainly `IWallet`; the type parameter is gone from all three types.
- The single member that forced the generic — `onchainRail`'s one `arkProvider.getInfo()` read — is now a **required** constructor dependency, `onchainRail(deps: { feeInfo: () => Promise<FeeInfo> })`.
- `PaymentRail`'s members revert to method syntax; they were only made function properties to get a variance behaviour that no longer has anything to constrain.

The required-vs-optional choice is deliberate and is the main thing worth challenging: it departs from the `ctx.swaps == null` optional-context precedent in the boltz rails. Rationale is in the PR description under "Why the dependency is required, not optional".

packages/ts-sdk/vitest.config.ts
```diff
@@ -24,6 +24,10 @@ export default mergeConfig(
             sequence: {
                 sequencer: CustomSequencer,
             },
+            typecheck: {
+                enabled: true,
+                tsconfig: "./tsconfig.typecheck.json",
```
pietro909, line 29: **`typecheck.include` is left at vitest's default, which is wider than the tsconfig — so a type test can still be a silent no-op.**

`typecheck.include` defaults to the package-wide `**/*.{test,spec}-d.?(c|m)[jt]s?(x)`, while `tsconfig.typecheck.json` includes only `test/**/*.test-d.ts` and excludes `test/e2e` and `examples`. A `.test-d.ts` outside `test/` (or under `test/e2e`) is therefore collected by vitest, **reported as passing, and never compiled by tsc** — exactly the silent no-op the new config's own comment says it prevents.

Second, related effect: typecheck only runs when a file matching that glob survives the CLI filter, but the `Type Errors  no errors` summary line prints regardless. Verified on the identical config:

- `vitest run test/payment/handle.test.ts` with a type error planted in `src/` -> `Type Errors  no errors`, **exit 0** (missed).
- same error under `--dir test/payment` -> exit 1 (caught, surfacing as `Unhandled Source Error`, though the `Type Errors` line still reads `no errors`).

So `test:integration` (`vitest run test/e2e/**`) matches no `.test-d.ts` and does no typechecking at all, while printing a clean type result it did not earn. Same for the single-file run documented in `AGENTS.md`.

Pinning the include to the tsconfig's own scope fixes both:

```ts
typecheck: {
    enabled: true,
    tsconfig: "./tsconfig.typecheck.json",
    include: ["test/**/*.test-d.ts"],
},
```

Cost is not the concern — typecheck was 2.98s of a 120s `test:unit`.

packages/ts-sdk/tsconfig.typecheck.json
```diff
@@ -0,0 +1,7 @@
+// Mirrors packages/boltz-swap/tsconfig.typecheck.json, which documents why a
```
pietro909, line 1: **Nothing compiles this config except the vitest run.**

`ci.yml:37` runs `pnpm -C packages/ts-sdk run typecheck` = `tsc --noEmit` on `tsconfig.json`, which excludes `test/` — so CI's dedicated type-check step builds a program of 224 `src` files and **zero** test files. No npm script references `tsconfig.typecheck.json` either. Its only gate is `pnpm run test:unit` at `ci.yml:67`, which reaches it indirectly through vitest.

That is thin for the file that makes the new assertions bite, and it interacts with the comment above: if the vitest include and this config drift apart, nothing independently notices. A small script would give it a direct gate:

```json
"typecheck:types": "tsc -p tsconfig.typecheck.json --noEmit"
```

Worth adding to the CI type-check job next to the existing three.

packages/ts-sdk/src/payment/index.ts
```diff
@@ -33,5 +33,5 @@ export function createDefaultPaymentRouter(wallet: Wallet): PaymentRouter {
     })
         .use(arkRail())
         .use(arkAssetRail())
-        .use(onchainRail());
+        .use(onchainRail({ feeInfo: async () => (await wallet.arkProvider.getInfo()).fees }));
```
pietro909: **The fee-source thunk is duplicated verbatim across two packages — worth promoting it.**

The identical expression appears here and at `packages/boltz-swap/src/payment/index.ts:47`:

```ts
async () => (await wallet.arkProvider.getInfo()).fees
```

Two copies today, and every third-party app that registers `onchainRail` itself becomes a third — each having to know that the fee schedule lives behind `wallet.arkProvider.getInfo().fees`. That is the knowledge the new dep was meant to let callers *not* care about, and `AGENTS.md` asks for shared wallet/protocol behaviour to be promoted into `ts-sdk` rather than copied outward.

A one-liner exported next to the rail would carry it:

```ts
export const walletFeeSource = (wallet: Wallet) => async () =>
    (await wallet.arkProvider.getInfo()).fees;
```

then `.use(onchainRail({ feeInfo: walletFeeSource(wallet) }))` in both factories. It also gives app authors the canonical wiring to copy, and keeps the one `arkProvider` reach in a single place — which is most of what made the original premise error possible.

packages/swap/test/payment/solverOnchain.test.ts
```diff
@@ -253,13 +253,13 @@ describe("solverOnchainRail.available", () => {
 
 describe("the router drops this rail rather than failing the payment", () => {
     /** A request the collaborative-exit rail can price without a live wallet. */
-    const arkProvider = { getInfo: async () => ({ fees: {} }) };
+    const feeInfo = async () => ({}) as never;
```
pietro909: `({}) as never` says less than it could, and absorbs more than you want:

```ts
const feeInfo = async () => ({}) as never;
```

`never` is assignable to anything, so this stub keeps compiling through any future change to `FeeInfo` or to the `feeInfo` signature — including one that should have failed here. `{} as FeeInfo` states the intent (an empty fee schedule, which is what `grossUpOffboard` tolerates via `feeInfo?.intentFee ?? {}`) and keeps the stub honest against the real type.

Minor, and the `wallet: {} as never` on line 259 is fine as-is — that one genuinely stands in for a wallet the rail never touches.

packages/ts-sdk/src/payment/rails/onchain.ts
```diff
@@ -56,8 +56,12 @@ function grossUpOffboard(
  *
  * An explicit amount is mandatory. To sweep the full balance, call
  * `Ramps.offboard(address, feeInfo)` directly — the router has no amountless path.
+ *
+ * @param deps.feeInfo - Source of the operator's fee schedule. Required: a rail
+ * that cannot price an offboard must not be constructible — a silently dropped
+ * rail on the money path is worse than a wiring error at startup.
  */
-export function onchainRail(): PaymentRail {
+export function onchainRail(deps: { feeInfo: () => Promise<FeeInfo> }): PaymentRail {
```
pietro909, line 71: This signature is the fix, and it is worth saying why it is better than the alternative it replaced.

`ctx.wallet` is read 12 times across all rails in all three packages; exactly one of those reads needed more than `IWallet`. The generic design paid for that one member with a type parameter on three public types — and the payment did not even clear: the invariance that made it sound existed only in `src`, because `PaymentRouter`'s private member types are elided in declaration emit, so consumers compiling against `dist/index.d.ts` could launder `PaymentRouter<IWallet>` into `PaymentRouter<Wallet>` with no cast and register this rail anyway.

One argument removes the member gap at its source instead of encoding it in the type system, and `FeeInfo` is already public (`src/index.ts:239`), so third-party callers can name the dep. The doc addition explaining *why* it is required rather than optional is the right thing to have written down — the contrast with the boltz rails' `if (ctx.swaps == null) return false` is the part a future reader would otherwise get wrong.

pietro909 (commented):
## Review — re-run against `a6c77a25`

I reviewed this at `12bd749a` (the generic design) and the review was largely superseded while it ran. Recording both halves, since the first half is what motivated the second.

### What `a6c77a25` resolves

The generic design had one blocking defect and two premise errors, all now moot:

1. **The guarantee did not survive declaration emit.** `PaymentRouter`'s invariance in `W` — the property that blocked every laundering route to `ctx.wallet.arkProvider` — came *only* from its two private members, and declaration emit elides private member types. Against `src` the laundering was rejected; against the published `dist/index.d.ts` it compiled clean:

```ts
const rIW = new PaymentRouter<IWallet>({ wallet: sw, prefs: {} });
const rW: PaymentRouter<Wallet> = rIW;   // no cast, no `any`
rW.use(onchainRail());                    // accepted -> ctx.wallet.arkProvider undefined
```

   `src`: `error TS2322: Type 'PaymentRouter<IWallet>' is not assignable to type 'PaymentRouter<Wallet>'`. Published `.d.ts`: exit 0, with negative controls in the same file firing correctly. With both private types erased, the only remaining `W` uses were the constructor and `use` — and `use` was method syntax, so bivariance made the two instantiations mutually assignable. Every consumer, including `packages/swap` and `packages/boltz-swap`, compiles against that version, and a `src`-based type test could never catch it.

2. **`arkProvider` was not "only on the concrete class."** `ExpoWallet` (`src/wallet/expo/wallet.ts:123-127`, published via `./wallet/expo`) does `implements IWallet` while declaring `readonly arkProvider`. It had the member the rail needed and was still refused, so it would have had to reintroduce the `as unknown as` cast the PR set out to delete.

3. **The method -> function-property change was source-breaking**, contradicting "every existing consumer sees the type it saw before". Compiled against a real `origin/master` worktree, five third-party rail shapes went from clean to broken — all of them rails annotating a *narrower* `ctx`/`req` (`TS2322`, `TS2416`, `TS2430`), and it reached the artifact (`dist/index.d.ts:6173`, same in `.d.cts`).

Injecting the fee source removes the cause of all three rather than patching them: no type parameter, so no variance to preserve or lose in emit; `RouterContext.wallet: IWallet`, so `ExpoWallet` and `ServiceWorkerWallet` both work; `PaymentRail` back to method syntax, so the published extension point is unchanged. The commit message's trade — one argument against threading a parameter through three public types to carry one member — is the right call, and the 12-read census backs it. Making the dep required rather than optional is also right, and for the reason given: a silently dropped rail on the money path is worse than a wiring error at startup.

One thing I had verified on the old head that still holds and is worth recording: the change is provably type-level only. All five changed files transpiled **byte-identical** through the repo's own esbuild, and `IWallet` appears zero times in `dist/index.js`/`.cjs`.

### What still stands

Four comments below, all on the test harness or on wiring — nothing blocking. The harness findings were mutation-verified on the old head and both config files are byte-identical here, so they carry over unchanged.

**Not re-run on `a6c77a25`:** CI was still pending when I posted, and I did not re-run the gate locally, so I am not asserting build/typecheck/test status on this head. On `12bd749a` the gate passed at exactly the claimed counts (233 files / 4376 passing / +1 skipped).

<sub>Reviewed with 71 agents: 21 findings raised, 15 survived three-lens adversarial verification, 6 refuted. Four of the refuted ones were things I had initially reported and withdrew — the `createDefaultPaymentRouter` "Wallet-only rails" TSDoc (that phrase means "needs no swap provider"), the variance comment's wording, the `tsconfig.typecheck.json` mirror claim, and a claim that the interleaved `//` comment detached the TSDoc (typedoc and `dist/index.d.ts:6172` both show it attached).</sub>

Kukks:
Thanks both — all five findings taken, in `975c3420`. Verified each against the code before acting rather than applying on trust; notes where my result differed from the report:

**@coderabbitai — validate `feeInfo` at construction (Major).** Accepted. Confirmed the failure mode: `options()` only try/catches `available()`, while `quote()` is called unwrapped through `PaymentOption.quote`, so a JS caller who omits the dep gets a `TypeError` mid-payment and the router does *not* re-rank around it. That is the same money-path argument that made the dep required, so it now throws at build time. Runtime coverage added for missing, empty-bag and non-function.

**@pietro909 — `typecheck.include` wider than the tsconfig.** Confirmed, and it is worse than a style point: I planted `expectTypeOf<string>().toEqualTypeOf<number>()` under `test/e2e/` and the suite reported **`Test Files 1 passed`, `Type Errors no errors`, exit 0**.

One correction to the proposed fix: `include: ["test/**/*.test-d.ts"]` alone does not close it, because `test/e2e/foo.test-d.ts` still matches that glob while `tsconfig.typecheck.json` excludes `test/e2e`. It needs the exclude too, so vitest's collection is a subset of the tsc program:

```ts
include: ["test/**/*.test-d.ts"],
exclude: ["test/e2e/**"],
```

Same probe now exits 1 with "No test files found" — no invented green.

**@pietro909 — nothing compiles that config but vitest.** Accepted: added `typecheck:types` (`tsc -p tsconfig.typecheck.json --noEmit`, verified exit 0) and a CI step beside the existing three.

**@pietro909 — duplicated fee thunk.** Accepted; `walletFeeSource(wallet)` is exported next to the rail and used by both factories.

**@pietro909 — `as never` stub.** Accepted, now `{} as FeeInfo`.

Gate on `975c3420`: build / typecheck / typecheck:types / lint all exit 0; unit 233 files, 4377 passed, 1 skipped, 0 failed (+1 vs the previous head is the new guard test).

I did not action the CodeRabbit autofix checkboxes.

pietro909 (approved):
All five taken, and the two corrections to my own reports are right. I re-verified against `975c3420` locally rather than reading the summary — including re-running the mutations rather than trusting the ones you reported.

**`typecheck.include` — your correction is the important part.** You are right that `include` alone does not close it: `test/e2e/foo.test-d.ts` matches `test/**/*.test-d.ts`, so my proposed fix would have left the hole open. The `exclude` is what makes vitest's collection a subset of the tsc program. Re-ran the probe on this head: a false `expectTypeOf<string>().toEqualTypeOf<number>()` planted under `test/e2e/` is now declined — `No test files found, exiting with code 1`, with the run printing `typecheck include: test/**/*.test-d.ts` / `typecheck exclude: test/e2e/**`. The same probe under `test/payment/` still fails the suite (exit 1). Collects what it compiles, refuses what it does not, no invented green either way. `test/e2e` is a real directory driven by `test:integration`, so the exclude is guarding something live, not a hypothetical.

**`typecheck:types`.** `tsc -p tsconfig.typecheck.json --noEmit`, exit 0 in 2.7s, and wired at `ci.yml` beside the existing type-check step. The config now has a gate that does not run through vitest.

**`walletFeeSource`.** Exported next to the rail, re-exported through `payment/index.ts` and so through the package root — `boltz-swap` imports it from `@arkade-os/sdk` rather than keeping a second copy. The `arkProvider.getInfo().fees` reach now appears exactly once in the repo, which is what makes it a canonical wiring for third-party callers rather than a private helper.

**`{} as FeeInfo`.** Applied. The remaining `as never` at line 259 is the wallet the rail never touches, which is the one I said was fine.

**CodeRabbit's runtime guard.** Agreed on the reasoning, and it holds here: `options()` wraps only `available()`, so a `quote()` throw is not re-ranked around. Guard covers missing, empty bag, and non-function, with runtime tests for all three.

Gate reproduced locally on this head: `pnpm run build`, `pnpm -r typecheck`, `pnpm -C packages/ts-sdk run typecheck:types` and `pnpm run lint` all exit 0; `pnpm run test:unit` is 233 files, 4377 passed, 1 skipped, 0 failed — matching your numbers exactly.

One note for the changelog rather than for this diff: `onchainRail()` -> `onchainRail({ feeInfo })` is a breaking signature change for anyone registering the rail themselves. It is the right trade and the PR argues it explicitly — I only mean it should be called out at release, since the failure is loud but happens at their build, not ours.

The design reads better than the generic it replaces, and the harness earns its green now. Approving.

Kukks:
@coderabbitai review

The previous run reported "Review limit reached", so this is a re-request rather than a duplicate. Please review `975c3420`, which adds the construction-time `feeInfo` guard (your Major finding), the promoted `walletFeeSource`, and the `typecheck.include`/`exclude` pinning plus the new `typecheck:types` CI gate.

PR #854, opened 2026-09-07
test(swap): the swaps arkade.money makes, end to end against solverd

tiero:
## Goal

End-to-end coverage for the swap flows arkade.money runs and this SDK never exercised against a solver: swap all in both directions, asset -> BTC, a wrong-priced offer that stays pending and is cancelled, and the record status after a wallet restore or reload. One file changes: `packages/swap/test/e2e/swap.test.ts`, which gains a second block that runs through the regtest stack's own solverd (BTC/RGT market, 1 sat = 1 RGT, fee 0).

Update (c321a429): the fix is now on this branch, so the suite is green end to end. Two product defects, two fixes:

- **ts-sdk** gains `WalletBalance.maxSendable` — `available` less one dust carrier whenever a spendable coin carries an asset (floored at zero). `send` names that ceiling when asset change would be stranded instead of reporting a misleading bare "Insufficient funds". This is the "swap all" hazard arkade.money's Max button hits.
- **swap** re-covers offer covenants after a restore: `ensureOfferContracts` registers a landed deposit's covenant, `restoreAssetSwaps` takes a `cover` option (handing the registration its funding-time server key, correct across rotations), and `watchOfferSwaps` sweeps live records at start-up — re-registering covenants for non-terminal records and reconciling each deposit against the chain, so a fill that landed while nothing watched is picked up. The same reconcile re-runs on `connection_reset`.

The two places the tests disagreed with the fix's semantics were reconciled to the fix: the swap-all-BTC case plans off `maxSendable` (not `available`), and `restoreFromKey` passes `cover`, as arkade.money's restore does. The fix was developed as commits c484d68c + a513ffeb on the author's follow-up branch and is squashed into c321a429 here.

Original report below, for the record:

---

Tests only, on purpose. The cases that fail on master are the report.

## Cases

| # | Asked for | Test |
|---|---|---|
| 1 | Swap ALL BTC -> asset | `swap ALL BTC -> RGT: the balance the wallet reports available funds the offer, and it fills` |
| 2 | Swap ALL asset -> BTC | `swap ALL RGT -> BTC: the whole asset balance goes, and the solver pays sats` |
| 3 | Asset -> BTC at the wrong price, stays pending, cancel | `RGT -> BTC at the WRONG price stays pending, and cancels` |
| 4 | Asset -> BTC leaving change | `RGT -> BTC leaving asset change: part of the balance goes, the solver pays sats, the rest stays` |
| 5 | BTC -> asset leaving change | `BTC -> RGT leaving change: quoted off the card, validated, funded, and filled` |
| 6 | Wrong price, restore, cancel from the restored wallet | `RGT -> BTC at the WRONG price, restored from the key alone, then cancelled from the restored wallet` |
| 7 | Asset -> BTC, restore, then filled | `RGT -> BTC restored before the solver takes it: the restored wallet sees the fill` |
| + | Statuses stay right across a reload | `reload on the same storage: a fill that landed while nothing watched is picked up, a pending offer still cancels` |
| + | Statuses stay right across a restore | `restored from the key alone, every record reads the status the live wallet holds` |

Every case goes through the path arkade.money takes: `planOffer` off the solver's card, `validatePlan`, `createOffer`, `wallet.send`, the record, then `watchOfferSwaps`, `restoreAssetSwaps` or `cancelOffer`. "Restore" is the same key on fresh repositories. "Reload" is the same key on the same repositories, with a new `Wallet` instance and watcher. Case 7 pauses the `solver` container across the funding, so the solver only sees the offer once the restored wallet is watching.

## Results on master (CI run 3150, head 0f3f6db): 9 pass, 5 fail

The five failures are the report. Each is a product defect, with the actual CI message:

- **Swap all BTC -> RGT** — `funding the offer with the wallet's own available balance (10330 sats, 10000 RGT held) was refused: Insufficient funds`. A wallet holding an asset offers its whole `available`; every asset sits on a dust carrier that `available` counts, so the send leaves the asset change nothing to ride on and `send` refuses it. This is the "swap all" arkade.money's Max button hits.
- **Wrong price, restore, cancel from the restored wallet** — `expected 6700n to be 8000n`. After a restore the escrowed asset should still read as owned. The live maker, which kept the covenant registration, counts it (8000); a wallet restored from the key alone does not (6700), because nothing re-covers the covenant, so the 1300-unit escrow is invisible until the cancel brings it back.
- **Restore before the fill** — `record … reads "pending" 90000 ms on, expected "fulfilled" — its deposit is spent on chain`. `restoreAssetSwaps` rebuilds the record, but nothing registers the covenant with the restored wallet, so its watcher never receives the spend.
- **Reload on the same storage** — same message. A fill that lands while no watcher runs produces no event afterward, so the record stays pending.
- **Restore, every record's status** — `expected [ Array(1) ] to deeply equal []`. Follows from the reload gap: a live record still pending whose deposit is spent.

Swap all RGT -> BTC passes, which is correct: sending the whole asset frees its carriers, so that direction has no swap-all hazard.

The candidate fix for all five failures (`maxSendable`, re-registration on restore, a watcher start-up reconcile) was developed at commits c484d68c + a513ffeb and now lands here as c321a429, with its unit tests.

https://claude.ai/code/session_014vNcqAg1BGU3KJgU5jVJsL

## Summary by CodeRabbit

* **New Features**
  * Added `maxSendable` to wallet balances to show the maximum transferable amount while preserving asset-backed funds.
  * Improved send validation with detailed errors when asset dust reserves limit transfers.
  * Restored swaps can automatically re-register and watch live deposits.
  * Swap monitoring now reconciles missed, swept, and filled deposits at startup.
  * Added support for watching already-landed deposits without recording a new issuance.

* **Documentation**
  * Updated balance, swap funding, and restore guidance for sendable amounts and deposit recovery.

### DIFF
--- a/packages/swap/src/watch.ts
+++ b/packages/swap/src/watch.ts
@@ -15,7 +15,7 @@
  * txid is the one `cancelOffer` recorded is a cancel, and anything else that
  * spends an offer deposit is a fill. The fallback, for a cancel this device did
  * not make (another device, or a wiped store), reads the covenant leaf off the
- * spending transaction ({@link classifySpend}).
+ * spending transaction ({@link classifyDepositSpend}).
  *
  * What this deliberately does not do:
  *
@@ -30,6 +30,12 @@
  * - **Keep watching a script it has finished with.** A persisted terminal
  *   status retires the contract to `retained`, once nothing at that script
  *   still needs coverage ({@link retireOfferContract}).
+ *
+ * Layout: the public surface comes first — {@link spendUpdate}, the pure
+ * transition, then {@link watchOfferSwaps}, which reads as the sequence it
+ * performs. Everything below it is module-private machinery, taking the
+ * watcher's dependencies as an explicit {@link WatchContext} rather than
+ * closing over them.
  */
 import { base64, hex } from "@scure/base";
 import {
@@ -38,9 +44,10 @@ import {
     RestIndexerProvider,
     Transaction,
     type ContractVtxoEvent,
+    type IContractManager,
     type IWallet,
 } from "@arkade-os/sdk";
-import { RETIRABLE, retireOfferContract } from "./coverage";
+import { RETIRABLE, retireOfferContract, type OfferContractRetirer } from "./coverage";
 import { decodeOffer, ensureOfferContracts, OFFER_CONTRACT_KIND } from "./offer";
 import type { AssetSwapRepository } from "./repository";
 import { classifyDepositSpend, spendTxidsOf, type RestoreIndexer, type SpendKind } from "./restore";
@@ -64,6 +71,23 @@ const TERMINAL: readonly AssetSwapStatus[] = ["fulfilled", "cancelled", "recover
  */
 const NEEDS_COVERAGE: readonly AssetSwapStatus[] = ["pending", "cancelling", "recoverable"];
 
+/** A running watcher. `idle()` exists because the writes are async: shutdown
+ * and tests both need to know when in-flight updates have settled. */
+export interface OfferSwapWatcher {
+    stop(): void;
+    idle(): Promise<void>;
+}
+
+export interface WatchOfferSwapsParams {
+    wallet: IWallet;
+    /** Same URL `createOffer`/`cancelOffer` take; used to read a spending tx
+     * when the exact classifier cannot answer. */
+    arkServerUrl: string;
+    repository: AssetSwapRepository;
+    /** Called after a change is persisted. A notification, not a store. */
+    onUpdate?: (swap: AssetSwap) => void;
+}
+
 /**
  * The record change a classified spend implies, or `undefined` when it implies
  * none — an already-resolved swap, or a spend nobody could classify.
@@ -86,23 +110,6 @@ export function spendUpdate(
         // mirrors restore.ts: a completion time is a fill's, not a cancel's
         ...(status === "fulfilled" && spend.at ? { completedAt: spend.at } : {}),
     };
-}
-
-/** A running watcher. `idle()` exists because the writes are async: shutdown
- * and tests both need to know when in-flight updates have settled. */
-export interface OfferSwapWatcher {
-    stop(): void;
-    idle(): Promise<void>;
-}
-
-export interface WatchOfferSwapsParams {
-    wallet: IWallet;
-    /** Same URL `createOffer`/`cancelOffer` take; used to read a spending tx
-     * when the exact classifier cannot answer. */
-    arkServerUrl: string;
-    repository: AssetSwapRepository;
-    /** Called after a change is persisted. A notification, not a store. */
-    onUpdate?: (swap: AssetSwap) => void;
 }
 
 /**
@@ -137,11 +144,16 @@ export async function watchOfferSwaps({
     onUpdate,
 }: WatchOfferSwapsParams): Promise<OfferSwapWatcher> {
     const manager = await wallet.getContractManager();
-    // Current server key at watcher start. TODO: persist the funding-time key
-    // with swap records; a signer rotation during a long session makes leaf
-    // classification return indeterminate rather than guessing.
-    const serverPubkey = ArkAddress.decode(await wallet.getAddress()).serverPubKey;
-    const indexer: RestoreIndexer = new RestIndexerProvider(arkServerUrl);
+    const context: WatchContext = {
+        manager,
+        indexer: new RestIndexerProvider(arkServerUrl),
+        repository,
+        // Current server key at watcher start. TODO: persist the funding-time
+        // key with swap records; a signer rotation during a long session makes
+        // leaf classification return indeterminate rather than guessing.
+        serverPubkey: ArkAddress.decode(await wallet.getAddress()).serverPubKey,
+        onUpdate,
+    };
 
     // events arrive independently but the update is read-modify-write over
     // the whole list, so two concurrent handlers would lose one of the writes
@@ -153,119 +165,69 @@ export async function watchOfferSwaps({
         });
     };
 
-    /** The two halves of a spend, as an event or the indexer reports them. */
-    type SpentDeposit = { txid: string; vout: number; arkTxId?: string; spentBy?: string };
-
-    const classify = async (swap: AssetSwap, vtxo: SpentDeposit, spentTxid: string) => {
-        // the exact answer: only the user can cancel, and cancelOffer records
-        // the txid it submitted
-        if (swap.spentTxid === spentTxid && swap.status === "cancelling") return "cancelled";
-        try {
-            // both halves of the spend: the checkpoint carries the deposit
-            // outpoint, the ark tx is the id the record and history name
-            const candidates = spendTxidsOf(vtxo);
-            if (candidates.length === 0) return "indeterminate";
-            const { txs } = await indexer.getVirtualTxs(candidates);
-            return classifyDepositSpend(
-                decodeOffer(hex.decode(swap.offerHex)),
-                serverPubkey,
-                txs.map((psbt) => Transaction.fromPSBT(base64.decode(psbt))),
-                { txid: vtxo.txid, vout: vtxo.vout },
-            );
-        } catch {
-            return "indeterminate" as const;
-        }
-    };
-
-    /**
-     * What a classified spend does to its record: written, announced, and the
-     * script retired once nothing live is left at it. One path for a spend an
-     * event delivers and for one the start-up reconcile finds on the indexer.
-     */
-    const resolveSpend = async (
-        swap: AssetSwap,
-        vtxo: SpentDeposit,
-        spentTxid: string,
-        at?: number,
-    ) => {
-        const kind: SpendKind = await classify(swap, vtxo, spentTxid);
-        const changes = spendUpdate(swap, { txid: spentTxid, kind, at });
-        if (!changes) return;
-
-        // notify only on a write that landed: `onUpdate` is documented as
-        // firing after the change is persisted, and a consumer that caches
-        // from it would otherwise run ahead of the store
-        const { persisted, swaps } = await updateAssetSwapBestEffort(repository, swap.id, changes);
-        // a lost write must not retire: the next restore scan still
-        // believes this deposit is live
-        if (!persisted) return;
-        onUpdate?.({ ...swap, ...changes });
-        // `swaps` is the post-update view, so the liveness check sees this
-        // record's new status without a third read
-        if (changes.status && RETIRABLE.includes(changes.status)) {
-            await retireOfferContract(manager, swaps, swap.swapPkScript);
-        }
-    };
-
-    const handleSpend = async (event: Extract<ContractVtxoEvent, { type: "vtxo_spent" }>) => {
-        if (event.contract.metadata?.kind !== OFFER_CONTRACT_KIND) return;
-
-        for (const vtxo of event.vtxos) {
-            const spentTxid = vtxo.arkTxId || vtxo.spentBy;
-            if (!spentTxid) continue;
-            // Identical offers share one script and are told apart by the
-            // deposit that funded them. Repository v1 has no indexed lookup, so
-            // this is O(history) per spend event; add a query API if offer
-            // event volume makes this hot.
-            const swap = (await getAssetSwaps(repository)).find(
-                (s) => s.fundingTxid === vtxo.txid && s.swapPkScript === event.contractScript,
-            );
-            if (!swap) continue;
-            await resolveSpend(swap, vtxo, spentTxid, event.timestamp);
-        }
-    };
-
-    /**
-     * The fate of a deposit the sweep just covered, read off the indexer. A
-     * spend that happened while the script was uncovered never becomes an
-     * event: this is the one read that catches it. A deposit not indexed yet
-     * is left alone — its event will come, or the next start will look again.
-     */
-    const reconcile = async (swap: AssetSwap) => {
-        const { vtxos } = await indexer.getVtxos({ scripts: [swap.swapPkScript] });
-        const vtxo = vtxos.find((v) => v.txid === swap.fundingTxid);
-        if (!vtxo) return;
-        if (vtxo.virtualStatus.state === "swept") {
-            // mirrors restore.ts: a swept deposit is still the user's money at
-            // that script, so it keeps its script watched and is never retired
-            if (swap.status === "recoverable") return;
-            const changes = { status: "recoverable" as const };
-            const { persisted } = await updateAssetSwapBestEffort(repository, swap.id, changes);
-            if (persisted) onUpdate?.({ ...swap, ...changes });
-            return;
-        }
-        if (vtxo.virtualStatus.state !== "spent") return;
-        const spentTxid = vtxo.arkTxId || vtxo.spentBy;
-        if (!spentTxid) return;
-        // No timestamp available on this path: the indexer VTXO only carries
-        // `createdAt` (its own creation), not the spend's time. A reload caught
-        // here is persisted without `completedAt`; the consumer re-queries the
-        // spending transaction if it needs the fill time.
-        await resolveSpend(swap, vtxo, spentTxid);
-    };
-
     const unsubscribe = manager.onContractEvent((event) => {
         // An offer is an owned contract; a watched script has no `metadata`.
         if (!isContractVtxoEvent(event) || event.type !== "vtxo_spent") return;
-        enqueue(() => handleSpend(event));
+        enqueue(() => handleSpendEvent(context, event));
     });
 
     // After subscribing, never before: registration is what starts the
     // manager's own watch on a script, and an event it delivers during the
     // sweep must find a handler already attached.
-    const uncovered = (await getAssetSwaps(repository)).filter(
+    const uncovered = await liveOfferRecords(repository);
+    await coverLiveOffers(wallet, arkServerUrl, uncovered);
+    // Through the queue, like an event: a spend delivered mid-sweep and the
+    // reconcile of the same deposit must not write over each other. Not
+    // awaited — `idle()` is how a caller waits for it — so a slow indexer
+    // delays the reconcile, never the watcher.
+    for (const swap of uncovered) enqueue(() => reconcileQuietly(context, swap));
+    enqueue(() => retireUntouchedScripts(context, uncovered));
+
+    return {
+        stop: unsubscribe,
+        idle: () => queue,
+    };
+}
+
+// ── internals ────────────────────────────────────────────────────────────────
+
+/** The two halves of a spend, as an event or the indexer reports them. */
+type SpentDeposit = { txid: string; vout: number; arkTxId?: string; spentBy?: string };
+
+/** The contract-manager surface the helpers below use, and no more. */
+type OfferWatchManager = OfferContractRetirer & Pick<IContractManager, "getContracts">;
+
+/**
+ * What a running watcher hands its helpers: the dependencies they read, passed
+ * rather than closed over, so each step below is callable — and testable — on
+ * its own.
+ */
+interface WatchContext {
+    manager: OfferWatchManager;
+    indexer: RestoreIndexer;
+    repository: AssetSwapRepository;
+    /** Server key as of watcher start; the leaf classifier rebuilds against it. */
+    serverPubkey: Uint8Array;
+    onUpdate?: (swap: AssetSwap) => void;
+}
+
+/** The records whose covenant may still hold a deposit, so still need coverage. */
+async function liveOfferRecords(repository: AssetSwapRepository): Promise<AssetSwap[]> {
+    return (await getAssetSwaps(repository)).filter(
         (swap) => typeof swap.offerHex === "string" && NEEDS_COVERAGE.includes(swap.status),
     );
+}
+
+/**
+ * Register every live record's covenant. Best effort: a record left uncovered
+ * is logged and tried again at the next start, rather than failing the watcher
+ * that the other records still need.
+ */
+async function coverLiveOffers(
+    wallet: IWallet,
+    arkServerUrl: string,
+    uncovered: readonly AssetSwap[],
+): Promise<void> {
     try {
         await ensureOfferContracts(
             wallet,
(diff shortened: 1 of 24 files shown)


tiero:
@gringokiwi just cosmetic changes:

1. helpers should be at the end of the file as much as possible
2. add a comment that the test are AI generated, and need to be redone
3. there is a lot on coupling and nesting. ask calud to do /simplify

gringokiwi:
## What changed since the last review pass

Three commits on top of `f829a9ad`. The `CHANGES_REQUESTED` above predates all of them — re-review would be welcome.

### `7b06d538` — the three findings from the pass on `f829a9ad`

**1. `completedAt` absent from reconcile-path fulfillments.** Resolved as documentation rather than code. The VTXO that `indexer.getVtxos` returns carries `createdAt` — its own creation — and no spend timestamp, so there is nothing on that path to pass as `at`. The call site now says so, and says what a consumer that needs the fill time must do instead (re-query the spending transaction). If the preference is a timestamp field on `SpentDeposit` sourced from a second indexer read, say so and I will do that instead — it was a deliberate choice not to add a network round trip to the reload path for a display field.

**2. Post-sweep retire called for never-registered scripts.** Fixed. `row?.watch === "retained"` does not fire when `row` is `undefined`, which is exactly the partial-`ensureOfferContracts`-failure case. Now:

```ts
const [row] = await manager.getContracts({ script });
if (!row) continue; // never registered (partial ensureOfferContracts failure)
if (row.watch === "retained") continue;
await retireOfferContract(manager, swaps, script);
```

**3. `LIVE_DEPOSIT` omits `"cancelling"` without explanation.** Fixed — the constant now carries the cross-reference to `NEEDS_COVERAGE` and the reason the two sets differ.

### `001e49fa` — merged `master`

The branch was seven commits behind and conflicting. The conflict was in `watch.ts` and was `#857`'s doing: `master` narrowed the event type to `ContractVtxoEvent` and added the `isContractVtxoEvent` guard. Resolved by keeping this branch's feature set (`resolveSpend`, `reconcile`, `ensureOfferContracts`, `NEEDS_COVERAGE`) and taking `master`'s narrower type and guard. `MERGEABLE` / `CLEAN` since.

### `b11291f6` — the cosmetic pass

**Helpers at the end.** `watch.ts` now reads: docstring, constants, public types, `spendUpdate`, `watchOfferSwaps`, an `internals` divider, then every private helper.

**Coupling and nesting.** `watchOfferSwaps` was one ~180-line function with four closures nested inside it, each capturing `manager`, `indexer`, `repository`, `serverPubkey` and `onUpdate` implicitly. They are now module-level functions taking those as an explicit `WatchContext`:

| was | now |
|---|---|
| `classify` | `spendKindOf(context, …)` |
| `resolveSpend` | `applySpend(context, …)` |
| `handleSpend` | `handleSpendEvent(context, …)` |
| `reconcile` | `reconcileDeposit` + `reconcileQuietly` |
| inline start-up sweep | `liveOfferRecords`, `coverLiveOffers`, `retireUntouchedScripts` |

The entry point is left as the sequence it performs:

```ts
const unsubscribe = manager.onContractEvent(…);
const uncovered = await liveOfferRecords(repository);
await coverLiveOffers(wallet, arkServerUrl, uncovered);
for (const swap of uncovered) enqueue(() => reconcileQuietly(context, swap));
enqueue(() => retireUntouchedScripts(context, uncovered));
```

`OfferWatchManager` also narrows the manager to the two methods these helpers actually call, instead of the whole `IContractManager`.

**The AI-generated tests.** Marked, not yet rewritten — see the question below. The e2e suite carries an `AI-generated, and to be redone` section in its file docstring saying it is scaffolding rather than the specification and is not to be extended with more AI-written cases. Each new block in `watch` / `coverage` / `register` / `restore` and the two `ts-sdk` suites carries a three-line marker to the same effect.

### Verification

No behaviour change from the cosmetic pass: the same 74 watch/restore/coverage/register cases pass before and after it. Full unit run on the merge result — ts-sdk 3007 passed, swap 830 passed, boltz-swap 617 passed, no type errors; lint and build clean.

### One question before merge

On *"the tests are AI generated, and need to be redone"* — I have done the marking half only. If **redone** means rewritten by hand before this merges, I would rather do it as a follow-up PR than grow this one: it is already past 2,000 changed lines and Danger has flagged the size, and the rewrite is mechanical enough to review on its own. If **redone** means flagged so the next person knows not to trust their shape, it is done. Which did you mean?

gringokiwi:
One more commit since the summary above — `421b51cd`, from the review of the "swap all" half.

The ceiling was reachable only by matching a regex against an `Error` message. Since the point of `maxSendable` is that a "send max" control prefills from it, the caller most likely to need it back is the one whose control raced a receive and got refused — so it should be a field.

```
AssetChangeCarrierError                          (abstract base: changeAmount, assetChangeCount, dustAmount)
├── SendAboveMaxSendableError                    + maxSendable
└── SelectedVtxosCannotCarryAssetChangeError     (no ceiling, deliberately)
```

Two types rather than one because the remedies differ — the same split `ForeignDescriptorError` / `WalletCannotSignError` already makes. Generic selection ran out of coins: lower the amount to `maxSendable`, or send the assets too and free the reserve. Caller pinned the inputs: `send` may not reach for a coin it was not given, so there is no ceiling to name and "lower the amount" would be the wrong advice. The base is there for consumers that only care that the assets had no carrier.

**New public exports:** `AssetChangeCarrierError`, `SendAboveMaxSendableError`, `SelectedVtxosCannotCarryAssetChangeError`, `AssetChangeCarrierFacts`.

**Messages are byte-identical**, so the existing assertions still hold and anything already parsing them keeps working. Purely additive.

Two cases added: the generic path pinned by type and all four fields, and the caller-pinned path pinned as a *different* type with no `maxSendable` to read.

ts-sdk 3009 passed (+2), swap 830, boltz-swap 617, no type errors; lint and build clean.

Also worth flagging from that review, not changed here: `WalletBalance.maxSendable` is a required field, so it is non-breaking to read but breaking to *produce* — any mock or alternative implementation building a `WalletBalance` literal needs updating. Belongs in the release notes.

gringokiwi:
@tiero — one thing left from your review, and it needs a yes/no from you.

> add a comment that the test are AI generated, and need to be redone

I read that two ways and did the first:

- **A — label them.** Mark the suites so nobody treats them as the specification. Done: a banner in the e2e suite's docstring, and a marker on each new block in `watch` / `coverage` / `register` / `restore` and the two `ts-sdk` suites.
- **B — label them *and* rewrite them before this merges.** Not done.

If you meant B, that is ~1,600 lines across seven files, and the e2e suite has to be iterated against the live regtest stack including the paused-solver window. Folded in here it takes an already 2,000-line PR — Danger has flagged the size — and makes both halves harder to read.

**I would rather split it**, and I have filed #890 so the split is a commitment rather than an intention: full inventory, what "redone" has to mean (shape that follows behaviour rather than implementation, fixture sprawl, waits standing in for invariants, an unverified case matrix), and acceptance criteria including "the cases that currently fail against a reverted fix must still fail".

The argument for splitting: the fixes are reviewed, green across all seven CI suites, and now waiting on process rather than substance — and this branch has been broken by a moving `master` twice while waiting. The argument against: markers are how "temporary" becomes permanent. #890 is my answer to that, and I am happy to be held to it.

For the record, the two typed-error cases I added in `421b51cd` are AI-written by the same standard. They carry the same marker and are in #890's inventory.

**So: split, or rewrite here first?** Either is fine — I just need the answer to unblock it.

Also, since the last comment: `master` moved again (#859, #874, #880, #881, #876 all landed) and I have merged it in at `960a10d5`. One real conflict, an import collision in `recipient-address-binding.test.ts`; #881 had also converted that file's fetch mocks to `jsonResponse` and missed the block this PR adds, so I aligned it. ts-sdk 3035, swap 831, boltz-swap 617, no type errors, lint and build clean.

packages/ts-sdk/test/recipient-address-binding.test.ts
```diff
@@ -444,6 +452,155 @@ describe("send with caller-selected vtxos", () => {
...
+    });
+
+    it("refuses a send of the whole balance and names the ceiling", async () => {
+        const wallet = await makeWallet();
+        // 40_700 is `available`; every sat is an input, so the asset change
+        // has nothing left to ride on. The ceiling is `available - dust`.
+        await expect(wallet.send({ address: ADDR, amount: 40_700 })).rejects.toThrow(
+            /0 sats of change cannot carry 1 asset change\(s\), needs 1000 — send at most 39700 sats \(WalletBalance\.maxSendable\)/,
+        );
+    });
```
tiero, line 526: These three regexes (526, 535, 556) pin a whole human-readable sentence, including the `(s)` pluralisation artifact, the `(WalletBalance.maxSendable)` parenthetical, the clause order and a literal em dash. The one thing you most want freedom to reword, a user-facing error, is now frozen by three tests in another package.

The tests are ugly because the API underneath is. `_sendImpl` computes the ceiling and then embeds it in prose, so a consumer wanting the number has to regex-parse an error message or recompute the rule.

**The SDK already had the right pattern, in this exact domain.** `packages/ts-sdk/src/wallet/ramps.ts`:

```ts
export class DustChangeError extends Error {
    readonly change: bigint;
    readonly dustAmount: bigint;
```

exported from the root index, with the doc comment *"Lets callers (e.g. wallet UI) react with appropriate UX, for instance offering to exit the full balance, instead of forwarding a server-side dust rejection to the user."* That is precisely the arkade.money Max button. It should have been reused or mirrored rather than reinvented as a string.

With a typed error carrying `maxSendable`, the whole block becomes:

```ts
const err = await wallet.send({ address: ADDR, amount: 40_700 }).catch((e) => e);
expect(err).toBeInstanceOf(AssetChangeStrandedError);
expect(err.maxSendable).toBe((await wallet.getBalance()).maxSendable);
```

That asserts the invariant that actually matters, that the send path's ceiling and the balance figure agree, instead of hardcoding 39700 in two places. Minimum fix if the typed error is out of scope here: match only the number, and compute it from the balance rather than hardcoding it.

---

packages/ts-sdk/test/recipient-address-binding.test.ts
```diff
@@ -444,6 +452,155 @@ describe("send with caller-selected vtxos", () => {
...
+
+    it("lets a send of `maxSendable` through the accounting", async () => {
+        const wallet = await makeWallet();
+        const err = await wallet.send({ address: ADDR, amount: 39_700 }).catch((e: unknown) => e);
+        // It fails later, at the submit this mock cannot serve. What matters is
+        // that the carrier was found and the send got past coin selection.
+        expect(String(err)).not.toMatch(/cannot carry/);
+        expect(String(err)).not.toMatch(/Insufficient funds/);
+    });
+
```
tiero, line 566: This test asserts two absences and its own comment admits it relies on the send failing later at a submit the mock cannot serve. It therefore passes if `send` throws a network error, and it passes if the method is replaced with `throw new Error("nope")`. It cannot fail for the right reason.

Assert the expected downstream failure instead, so a harness break fails loudly rather than passing quietly:

```ts
expect(String(err)).toMatch(/submit|fetch|network/);
```

Zero extra lines.

---

packages/ts-sdk/src/wallet/balance.ts
```diff
@@ -10,7 +10,27 @@ import { canRecoverOnchain, canSpendOffchain, hasTerminalSpend } from "./vtxo";
...
+     * whenever the coins it selects carry assets — and after the first asset
+     * receive that is every wallet, since assets arrive on a dust carrier that
+     * `available` counts. A ceiling, not a tight bound: an amount the plain
+     * coins alone cover goes through above it, but only this figure holds
+     * whatever coins selection picks. Zero when less than dust could leave —
+     * a send pads its recipient to dust before selecting, so nothing below
+     * dust ever leaves on its own — and so zero when the carriers are all
+     * there is.
+     */
+    maxSendable: number;
```
tiero, line 33: Worth asking whether this needed to be permanent public API. `maxSendable` is derivable from two fields already on the balance, and adding it costs a field on `OffchainBalance`, a field on `WalletBalance`, a `BalanceCapabilities.dust`, the service-worker balance path, and two READMEs.

A pure exported helper would have fixed the same user-facing bug with far less surface:

```ts
export function sendCeiling(
    b: Pick<WalletBalance, "available" | "availableAssets">,
    dust: bigint,
): number {
    const c = b.available - (b.availableAssets.length ? Number(dust) : 0);
    return c >= Number(dust) ? c : 0;
}
```

Counter-argument, and I think it holds: a Max button needs the number *before* attempting a send, to prefill the field, so a balance field genuinely earns its place. But then the floor rule should live in one place. Right now `computeOffchainBalance` and `_sendImpl` each implement `raw >= dust ? raw : 0` independently, and the only thing asserting they agree is a hardcoded `39700` inside a regex in `recipient-address-binding.test.ts`. If the two drift, nothing typed catches it.

---

packages/swap/src/watch.ts
```diff
@@ -131,74 +165,221 @@ export async function watchOfferSwaps({
...
         if (!isContractVtxoEvent(event) || event.type !== "vtxo_spent") return;
-        enqueue(() => handleSpend(event));
+        enqueue(() => handleSpendEvent(context, event));
     });
 
+    // After subscribing, never before: registration is what starts the
+    // manager's own watch on a script, and an event it delivers during the
+    // sweep must find a handler already attached.
+    const uncovered = await liveOfferRecords(repository);
+    await coverLiveOffers(wallet, arkServerUrl, uncovered);
```
tiero, line 178: Two code paths now register the same covenants: `restoreAssetSwaps({ cover })` does it when the record is rebuilt, and this sweep does it again at every watcher start. The README added by this PR says so out loud, which is the tell.

Register where the record is born or rebuilt, and let the watcher only reconcile. Concretely: make coverage unconditional inside `restoreAssetSwaps` (drop the `cover` option, keep the few lines that call `ensureOfferContracts`), delete `coverLiveOffers` and its call site here, and keep the reconcile pass, which is the part that actually fixes the reload defect.

That removes roughly 30 lines of source and takes `retireUntouchedScripts` with it, since that only exists to undo a race this sweep creates.

---

packages/swap/src/watch.ts
```diff
@@ -131,74 +165,221 @@ export async function watchOfferSwaps({
...
+}
+
+/**
+ * A settlement that landed mid-sweep retired its script before the sweep put it
+ * back in the watched set, and nothing would retire it again: re-derive
+ * liveness from the records as they stand now, for the scripts the sweep
+ * touched and left watched. `retireOfferContract` leaves a live script alone; a
+ * row the reconcile already retired is not written twice.
+ */
+async function retireUntouchedScripts(
```
tiero, line 374: This helper exists only to undo a race the cover sweep itself creates: the sweep re-watches scripts, so a fill that lands mid-sweep would otherwise be left watched forever. It also forced `OfferWatchManager` to grow a `getContracts` member, which then needed a stub in the test harness.

If the cover sweep goes (see the comment above), this goes with it and the sweep can end with the existing `retireSettledOfferContracts`, which already no-ops on a live script and already swallows a failed write. Roughly 20 lines of source plus the manager type narrowing plus the harness stub.

---

packages/swap/test/watch.test.ts
```diff
@@ -317,6 +372,8 @@ describe("watchOfferSwaps", () => {
                     arkServerUrl: "http://ark",
                     repository,
                 });
+                // the start-up sweep's own "watched" write is not what these assert on
+                setContractWatchState.mockClear();
```
tiero, line 376: These five `mockClear()` calls are the clearest evidence that this suite pins the implementation's call sequence rather than its behaviour. They exist only because the new start-up sweep writes `watched` before the pre-existing tests' assertions, so the spy history had to be reset to keep those assertions passing.

Nothing the sweep does before its last write matters to any caller. Have the harness track state rather than calls:

```ts
watchState: (script: string) => watchOf.get(script) ?? "watched"
```

then assert `expect(watchState(script)).toBe("retained")`. All five `mockClear` lines disappear, the pre-existing tests go back to their master form, and the later race test collapses to a single line.

---

packages/swap/test/watch.test.ts
```diff
@@ -447,6 +512,306 @@ describe("watchOfferSwaps", () => {
...
+                            virtualStatus: { state: "spent" },
+                            arkTxId: fill.txid,
+                        },
+                    ],
+                }),
+            },
+        );
+    });
+
+    it("marks a deposit swept before the sweep covered it recoverable, and keeps its script", async () => {
```
tiero, line 698: This test and its two neighbours (649, 735) are one table written out three times: deposit state in, record status and watch state out. Roughly 120 lines for three rows.

```ts
it.each([
    ["spent",   "fulfilled",   "retained"],
    ["swept",   "recoverable", "watched"],
    ["settled", "pending",     "watched"],
])(...)
```

About 35 lines over one body, with the fill's `arkTxId` added only for the spent row. A fourth state then costs one row rather than another 40 lines.

---

packages/swap/test/restore.test.ts
```diff
@@ -390,6 +397,166 @@ describe("restoreAssetSwaps", () => {
         }
     });
 
+    // AI-generated, and to be redone: written by Claude, kept for the coverage
+    // it gives the fix rather than for its shape. Rewrite by hand before
+    // reading it as the specification.
+    describe("with `cover`, a live restored deposit is the wallet's to watch again", () => {
```
tiero, line 403: This block is ~115 lines re-testing `ensureOfferContracts` end to end, through a full offer build and server stubs, when `restore.ts` itself only added about eight lines: pick the live records, pass the funding-time key, keep going when registration fails.

Test what this file decides, in two tests of roughly 40 lines total:

1. Seed a pending, a recoverable and a fulfilled record, then assert `createContract.mock.calls.map(c => c[0].script)` equals the pending and recoverable scripts. That is the `LIVE_DEPOSIT` filter and nothing else.
2. Make `getInfo` reject, and assert the record is still returned and the txid still marked scanned.

The registration behaviour itself already has its own suite in `register.test.ts`, so exercising it again here is duplicate coverage at triple the setup cost.

---

packages/swap/test/e2e/swap.test.ts
```diff
@@ -338,3 +365,693 @@ describe("maker-side swap loop (regtest)", () => {
...
+        // the solver has to be running to take it.
+        await ensureRgt(1_500);
+        await ensureSats(Number(dust));
+        const plan = rgtToBtc(BigInt(1_500));
+        const restoredRepository = new InMemoryAssetSwapRepository();
+        const restoredWallet = await createWallet(keyHex);
+        await historyOf(restoredWallet);
+        const satsBefore = (await restoredWallet.getBalance()).available;
+        let restoredWatcher: OfferSwapWatcher | undefined;
+        execCommand("docker pause solver");
```
tiero, line 863: `docker pause` puts a global, timing-bound side effect into a job whose stack is shared with every other case in the file. The window is bounded by arkd dropping a stream client that stays silent for about 50 seconds, and the case carries a 240 second ceiling, so the budget contradicts itself: if the pause ever runs long the failure is a stack artifact, not a defect.

The same claim is reachable without touching the container. The suite already has `wrongPrice`. Publish at a want the solver declines, restore the wallet while it sits pending, bring its watcher up, then cancel and republish at a fair want. The restored wallet's watcher then hears a fill it was not present for, which is the actual assertion, with no container manipulation and no race.

That is about 89 lines removed.

---

packages/swap/test/e2e/swap.test.ts
```diff
@@ -338,3 +365,693 @@ describe("maker-side swap loop (regtest)", () => {
...
+
+        // the asset is in the wallet, spendable, on the solver's carrier; the
+        // change from the funding is still here too
+        await waitFor(async () => (await rgtHeld(maker)) >= plan.receive.atomic);
+        expect((await maker.getBalance()).available).toBe(
+            SOLVER_FAUCET_SATS - 10_000 + SOLVER_CARRIER,
+        );
+    }, 180_000);
+
+    it("swap ALL BTC -> RGT: the balance the wallet reports sendable funds the offer, and it fills", async () => {
```
tiero, line 703: This case was the centrepiece of the defect report. At 0f3f6db it funded with `available` and failed with `Insufficient funds`, which is exactly the arkade.money Max button. When the fix landed, the input was changed to `maxSendable` so it would pass.

That inverts the wrong thing. The regression it was written to catch is now unguarded end to end: nothing here fails if `available` becomes fundable again by accident.

Invert the assertion rather than the input:

```ts
await expect(publish(maker, btcToRgt(available), ...)).rejects.toThrow(/maxSendable/);
```

then keep the `maxSendable` happy path as a second, cheaper assertion. One line changed, and the case goes back to pinning the defect.

Separately, this whole case may not need a solver at all. The defect is arithmetic in `_sendImpl` and reproduces with one issued asset and no market, which would move it to `packages/ts-sdk/test/e2e/asset.test.ts` at 60 seconds instead of 180 in a solver-dependent suite.

---

packages/swap/test/e2e/swap.test.ts
```diff
@@ -338,3 +365,693 @@ describe("maker-side swap loop (regtest)", () => {
...
+        "the solver's price feed",
+    );
+    const value = found.price_feed_schema.price_path
+        .split("/")
+        .filter(Boolean)
+        .reduce<unknown>((node, key) => (node as Record<string, unknown>)?.[key], body);
+    return { market: found, feedValue: Number(value) };
+};
+
+describe("the swap, against solverd (regtest)", () => {
```
tiero, line 484: Structural note on the block as a whole, since it is the bulk of the diff.

Nine cases run against one wallet, one key, one RGT balance and a module-level watcher that two cases reassign. That coupling caused three of this branch's commits on its own: an RGT top-up helper after cases starved each other, distinct covenant amounts after three offers derived the same script, and a rewritten assertion after the shared balance moved. Each case also carries a 180 to 240 second ceiling, roughly 33 minutes of budget in total, for coverage the critique puts at about two minutes.

Two things would have avoided most of it. Give each case its own key and its own faucet from a shared `test/e2e/helpers.ts`, so every case is runnable alone with `-t` and no case can starve another. And split by file, `fill.test.ts` and `swapAll.test.ts` alongside a helpers module, rather than appending 717 lines to a file whose own header declares the fill path out of scope. That layout existed on the earlier working branch and was discarded when the suites were merged into this one file.

Minimum subset that still demonstrates all three defects: one ts-sdk asset case for the send ceiling, the restore-and-cancel case, and the first half of the reload case. Roughly 95 lines against the current 723.

---

packages/swap/test/register.test.ts
```diff
@@ -335,3 +344,129 @@ describe("an offer at a script an earlier offer retired", () => {
...
+        expect(state.created[0]).toMatchObject({
+            type: "arkade",
+            script,
+            label: OFFER_CONTRACT_LABEL,
+            metadata: { genericallySpendable: false, kind: OFFER_CONTRACT_KIND },
+        });
+        expect(state.watched).toEqual([[script, "watched"]]);
+    });
+
+    it("sets no issuance mark, so a settled record can retire the script", async () => {
```
tiero, line 387: "A landed deposit sets no issuance mark" is asserted three times at three levels for what is a one-line branch: here, in `coverage.test.ts`, and again in `watch.test.ts` through a full watcher start.

Keep the `coverage.test.ts` unit, which tests the mark at the module that owns it with no offer construction, no watcher and no network stubs, and delete the other two. About 60 lines. The only thing lost is the wiring check that `registerOfferContract` really passes `"landed"`, which is one assertion in the test that already builds the row here, not a dedicated case.

---

tiero (commented):
Review of my own work, at the maintainer's request, after an adversarial pass with independent verification of each finding. Short version: the defects are real and the fixes are right, but this PR is roughly three times the size it needed to be, and the tests are the weaker half.

## Why there is non-test code here at all

The original ask was "fix swap-all: doesn't work in arkade.money". The fix was always the deliverable. Mid-flight the maintainer asked to see the failures first, so the branch was force-pushed down to master plus one test file; the fix then came back as c321a429. The PR did not grow non-test code, it regained it, which is why the body now tells two stories.

## The scope problem: three defects fixed, one asked for

| Defect | Asked for | Src lines |
|---|---|---|
| D1 swap-all, the Max button | yes | ~110 |
| D2 restore never re-covers the covenant | no, found while writing the tests | ~330 |
| D3 reload misses an unwatched fill | no, found while writing the tests | ~190 |

The reported bug cost 110 lines. The two found along the way cost 521, and `watch.ts` went from 204 lines to 385. The reasoning at the time was that the restore and reload tests stay red otherwise, so the fix had to ship with them. That is right about the tests and wrong about the PR: D2 and D3 should have been reported as findings for the maintainer to fund, not fixed inside a swap-all PR.

The correct split, all of which was available at the time: #854 stays tests-only and red, as asked, and is closed by two follow-ups; one ts-sdk PR for the send ceiling; one swap PR for restore coverage and the reload reconcile.

## The API miss behind the ugliest tests

`_sendImpl` computes the ceiling and embeds it in a sentence, so three tests read it back by regex, em dash included. The SDK already had `DustChangeError` in `wallet/ramps.ts`, same domain, exported from the root, documented for exactly this UX ("offering to exit the full balance"). Mirroring it would have given a typed error carrying the ceiling, made the tests three lines each, and given arkade.money the number without parsing prose. Details inline.

## On the tests

About 1663 added lines, 723 e2e and 940 unit. The unit half over-specifies: five `mockClear` calls exist only to keep pre-existing assertions passing against the new sweep, three sweep-reconcile tests are one table written out three times, one claim is asserted at three levels, and a 115-line block re-tests a function that has its own suite. The strongest block is `computeOffchainBalance.maxSendable` in `spendability-gate.test.ts`, pure input to output with no mocks, and it is what the rest should have looked like.

The e2e block appends 717 lines to a file whose header declares the fill path out of scope, couples nine cases to one wallet and one asset balance, and spends roughly 33 minutes of timeout budget on coverage reachable in about two. Three of its commits exist only to work around that coupling.

I agree with b11291f6's banners. These tests earn their keep as coverage for the fixes and should not be read as the specification.

## What was worth keeping

The three defects are real and were invisible before. `maxSendable` as a balance field is justified, since a Max button needs the number before attempting a send. The boundary cases chosen for the send ceiling are the right three. The restore and reload cases found D2 and D3, which no unit test would have. And the reconcile pass in the watcher is the part that actually fixes D3; it is the coverage sweep around it that is redundant.

Happy to do the reduction as a follow-up if useful.

---

packages/swap/src/watch.ts
tiero: Correcting myself: I had this backwards, and following it would have deleted the more robust of the two paths.

The redundancy is real, but the sweep is the one to keep. The watcher's start-up registration is the only idempotent retry in the design, and it reaches records the incremental scan will never revisit, because `restoreAssetSwaps` skips txids already in `existingIds`/`scanned` (restore.ts:280). That includes records whose registration failed the first time, since restore's cover is best-effort. Delete the sweep and a covenant that failed to register once is never retried.

So the reduction runs the other way: drop the restore-side `cover` option and keep `coverLiveOffers`. That is roughly 45 lines out of `restore.ts` (the import churn at 19-25, the doc at 249-266, the opts and destructure at 273-278, the body at 428-438, and `LIVE_DEPOSIT` plus its doc at 440-452) plus about 6 README lines. The only case `cover` uniquely serves is a consumer that restores and never starts a watcher, and `ensureOfferContracts` is already exported from `index.ts:74` for exactly that, so the cost is one line in the `restoreFromKey` test helper.

My comment below on `retireUntouchedScripts` is therefore also wrong in its premise. If the sweep stays, that helper's race is real and the cheaper fix is to run its liveness re-read inside the existing per-script `serialize` queue in `coverage.ts`, rather than as a separate pass.

---

packages/swap/test/e2e/swap.test.ts
tiero: Correction to the second paragraph. The first paragraph stands, and verification sharpened it: as written this case asserts `maxSendable < available`, which is near-tautological given the `held > 0` assertion two lines above, so every proof that master was broken is now mock-backed. The fix is three lines added, nothing deleted:

```ts
await expect(publish(maker, btcToRgt(available), makerRepository)).rejects.toThrow(/maxSendable/);
```

before the `maxSendable` funding. Red on master, green here, and it costs no solver round trip because the send is refused before the deposit goes out. While there, the try/catch at 715-723 is nine lines re-wrapping a throw vitest already reports with the same context.

Where I was wrong: I implied this case could move to ts-sdk and save time. Three corrections. `180_000` is a timeout ceiling, not runtime. Solverd is already up on the swap profile, since `packages/swap/.env.regtest` sets no `REGTEST_PROFILES` and the regtest runner defaults to every tier, so this case adds no profile cost. And cases 3, 4 and 5 take their RGT from the carry-over this case and case 1 create, so deleting them breaks three siblings and the repair reintroduces the same solver fill as unasserted fixture plumbing.

The ts-sdk e2e case is still worth adding on its own merits, as an addition rather than a move: `packages/ts-sdk/test/e2e/asset.test.ts` zeroes all four fee knobs and already issues assets, so a send of `available` reaches the asset-change branch there and reproduces the defect in the package that owns it.

---

packages/swap/test/e2e/swap.test.ts
tiero: Correcting the last paragraph. "Roughly 95 lines against the current 723" was my own estimate and it is wrong. Measured against the branch, the minimum that still proves all three defects is **four cases, about 495 lines**, not one plus two halves.

Four survive: a swap-all corridor case carrying the one red line the branch lacks (about 20 lines), the restore case at 842-930 as-is, the reload case's first half at 965-992 (28), and the corpus case at 1015-1056 (42). Cut: case 1 whose RGT purchase folds into `beforeAll`, cases 3, 4, 5 and 7, and the second reload cycle at 993-1012. All of those are green on master and none touches a line this PR adds, which is the test for whether an e2e case is earning its keep here.

The coupling and file-split points stand unchanged, and the commit archaeology behind them is the strongest evidence in this review.

For scale on the whole PR rather than just this file: the minimal version is about +1333/−150 against the actual +2250/−120, so roughly 917 added lines, 41%. Source drops 52%, tests 37%. The two biggest single items are not in this file: `restore.ts`'s `cover` option with its test block at 223 lines, and `watch.test.ts` at 142.

---

packages/swap/test/e2e/swap.test.ts
tiero: Withdrawing this one. The criticism of the mechanism stands, but the conclusion does not.

`docker pause` is still a global side effect in a shared-stack job with a self-contradicting budget, and that is worth knowing. But when the whole PR was reduced to its minimum, this case survived: it is the **only proof of D2 against the real stack**. The replacement I proposed, publish at a declined price then republish at a fair one, changes what is being tested. It shows a restored watcher hearing a *later* fill, not a fill of the deposit that existed before the restore, which is the actual defect.

So keep it as-is. If the pause window ever proves flaky in practice, the fix is to tighten the window rather than to swap the mechanism, since the restored wallet is already brought up and synced before the pause.

The line saving I quoted, 89, was also double-counted: it appears again in the structural comment's subset, and the two are not additive.

---

PR #857, opened 2026-09-07
feat(contracts): watch-only scripts on ContractWatcher

Kukks:
Adds watch-only scripts to `ContractWatcher`: a way to be told when VTXOs appear at or leave a script the wallet does **not** own, without registering that script as a contract and without any of it entering the wallet's balance, recovery or history.

The motivating consumer is a swap solver watching a counterparty's lockup. Today its only option is to register the lockup as a real contract, which buys the event stream at the price of pulling a foreign script into the wallet's owned set.

## The finding that shaped the design

**The failsafe poll does not poll the indexer — it re-reads the wallet repository.**

`getContractVtxos` takes `const repo = this.config.walletRepository` (`contractWatcher.ts:299`), and the call site says so outright: *"Load all the virtual outputs for these contracts, from DB"*. The watcher never calls the indexer for the poll.

That makes the obvious implementation — put watch-only scripts in the existing map and let the existing poll serve them — **actively destructive rather than merely inert**. For a script the wallet does not own, the repository read is permanently `[]`, so `currentKeys` is empty; the spent-diff then walks every entry the *subscription* just wrote into `lastKnownVtxos`, emits each as spent, and deletes it. The next subscription event re-adds it. A funded lockup would oscillate received → spent on a 20-second cycle (`failsafePollIntervalMs: 20_000`).

Watch-only entries therefore get an **indexer-backed diff**.

> **To be clear about what this is not:** it is not a second poll loop. It is a second *data source inside the same poll* — same timer, same `pollAllContracts()`, same reconnect-triggered call, same `forcePoll()`. One subscription, one interval, one backoff regime. The uniformity is preserved exactly; only the source of truth for the diff differs, because for a foreign script the repository is not and can never be one.

## API

On `ContractWatcher` (a class, so purely additive):

```ts
addWatchedScript(script: string, options?: { label?: string }): Promise<void>
removeWatchedScript(script: string): Promise<void>
getWatchedScripts(): WatchedScript[]
```

On `IContractManager`, as **optional** methods `watchScript?` / `unwatchScript?` / `getWatchedScripts?`.

Optional is forced, not a preference: `ServiceWorkerWallet.getContractManager()` builds a fully-typed `const manager: IContractManager = { … }` object literal (`serviceWorker/wallet.ts:1249`). A required method would break that file, the partial mock in `test/arkade-contract.test.ts`, and every external implementor. This follows the precedent already set by `assertSpendableNow?` / `unspendableNowReasons?` in the same interface.

**Both shipped implementations implement them for real** — core `ContractManager` and the service-worker proxy — and the service-worker handler now *refuses* rather than acking when a manager cannot serve them, so the optionality is never a silent no-op.

## Event shape

Watch-only activity **reuses the existing `vtxo_received` / `vtxo_spent` types**. There are no new event variants.

An earlier revision of this PR added `script_vtxo_received` / `script_vtxo_spent`. That was reverted: widening a published discriminated union is paid for by every consumer with an exhaustive switch, including consumers that never call `watchScript` and get no benefit. One of the two in-repo breaks (`examples/contract-manager.ts`) was invisible to CI, because `examples/` is excluded from the package tsconfig — so CI could not have caught the class of break at all.

The two forms are told apart by **shape, not a flag**: the owned variant requires `contract`; the watch-only variant has no such property.

```ts
export type ContractEvent =
    | { type: "vtxo_received"; contractScript: string; vtxos: ContractVtxo[]; contract: Contract; timestamp: number }
    | { type: "vtxo_spent";    contractScript: string; vtxos: ContractVtxo[]; contract: Contract; timestamp: number }
    | { type: "vtxo_received"; contractScript: string; vtxos: VirtualCoin[]; timestamp: number }
    | { type: "vtxo_spent";    contractScript: string; vtxos: VirtualCoin[]; timestamp: number }
    | { type: "connection_reset"; timestamp: number };
```

with an exported narrowing helper:

```ts
export type ContractVtxoEvent = Extract<ContractEvent, { contract: Contract }>;
export function isContractVtxoEvent(event: ContractEvent): event is ContractVtxoEvent;
```

`contractScript` is present on both forms, so a consumer that only needs the script keeps compiling untouched. Note the absent fields on the watch-only form: no `contract`, and no `extendVirtualCoinForContract` annotation. You cannot derive forfeit tapscripts for a script you do not own, and the event does not pretend otherwise.

## Isolation is now a guard, not a type — and here is where it lives

**This is the material change from the previous revision, and it is the part that deserves scrutiny.**

Previously the isolation was structural: distinct `script_`-prefixed variants meant no case in `ContractManager.handleContractEvent` matched, so a watch-only event *could not* reach `syncContracts` → `saveVtxosForContract` → the wallet repository. With reuse **the case matches**, and a guard is the thing that stops it.

That guard lives at **`packages/ts-sdk/src/contracts/contractManager.ts:2179` and `:2184`** — one line at the head of each of the `vtxo_received` and `vtxo_spent` cases, ahead of `syncContracts`:

```ts
case "vtxo_received":
    if (!isContractVtxoEvent(event)) break;
    await this.syncContracts({ contracts: [event.contract] });
```

It is **compile-enforced**, not merely conventional. Because `contract` is *absent* rather than optional on the watch-only variant, `event.contract` does not type-check without narrowing, so deleting the guard is not a silent regression — it fails the build:

```
error TS2339: Property 'contract' does not exist on type
'{ type: "vtxo_received"; contractScript: string; vtxos: ContractVtxo[]; contract: Contract; timestamp: number; }
 | { type: "vtxo_received"; contractScript: string; vtxos: VirtualCoin[]; timestamp: number; }'
```

**Defence in depth — being accurate rather than flattering.** The guard is the first line, not the only one. `extendVirtualCoinForContract` (`wallet/utils.ts:175`) refuses to annotate a VTXO whose script matches no registered contract, and `validateVtxosForScript` (`contracts/vtxoOwnership.ts:61`) refuses to persist a script mismatch. A mutation that synthesises a `Contract` purely to satisfy the case still writes nothing — it dies at annotation. That was found by mutation testing, not assumed.

## The consumer the type system would *not* have caught

`VtxoManager`'s renewal/delegation subscriber (`wallet/vtxo-manager.ts:2768`) reads `event.vtxos` and never `event.contract`. Under reuse it therefore **compiles silently** while now matching watch-only events — and would have fired a `settle()` on a foreign script's funding and passed foreign outputs to `delegateManager.delegate()`.

This is why the consumer table below was re-derived from scratch rather than carried forward, and why a type-level widening alone was not sufficient: the compiler flags the consumers that read `contract`, and says nothing about the ones that read `vtxos`.

## Where the entries live, and what every consumer sees

Watch-only entries live in a second map (`watchedScripts`), **not** in `this.contracts`. That boundary is still enforced by type: `getAllContracts()` and `getWatchedContracts()` return `Contract[]` and read only `this.contracts` — a watch-only entry cannot appear in either, because it is not a `Contract` and has no `Contract` to be. `getWatchedContracts()` is left byte-identical, so `syncContracts` and `reconcileWatched` are untouched and cannot regress.

| # | Consumer | Watch-only entry |
|---|---|---|
| 1 | `ContractWatcher.getAllContracts()` | **absent** — reads `this.contracts` only |
| 2 | `ContractWatcher.getWatchedContracts()` | **absent** — filters (1) |
| 3 | `updateSubscription()` | **PRESENT** — the point of the feature |
| 4 | `pollAllContracts()` | **PRESENT, indexer-sourced** — owned branch unchanged |
| 5 | `getContractVtxos()` | **absent** — never reads the repository for them |
| 6 | `processSubscriptionVtxos()` | **PRESENT** — routed by the ownership precedence rule |
| 7 | `emitVtxoEvent()` | **absent** — watch-only emits via its own path, unannotated |
| 8 | `ContractManager.syncContracts({})` | **absent** ⇒ never fetched, never `saveVtxosForContract` |
| 9 | `reconcileWatched()` / `reconcilePendingFrontier` | **absent** |
| 10 | **`handleContractEvent`** | **case now MATCHES** ⇒ isolated by the guard at `contractManager.ts:2179/2184`, then forwarded |
| 11 | `ContractManager.onVtxosSpent` hook | **absent** — sits after the same guard |
| 12 | `ReadonlyWallet` indexer event stream (`wallet.ts:1920`) | **absent** — guarded; would not have compiled otherwise |
| 13 | **`VtxoManager` renewal / delegation (`vtxo-manager.ts:2768`)** | **absent** — guarded; *would have compiled silently* (see above) |
| 14 | `WalletReceiveRotator` (`walletReceiveRotator.ts:432`) | **absent, unchanged** — filters to `defaultContractScript`, and an active contract always out-ranks a watch-only registration |
| 15 | `RfqSwapManager.subscribe` (`swap/swapManager.ts:1290`) | **absent** — guarded |
| 16 | Offer spend watcher (`swap/watch.ts:194`) | **absent** — guarded; would not have compiled otherwise |
| 17 | Service-worker `CONTRACT_EVENT` forward (`wallet-message-handler.ts:2256`) | **forwards verbatim, unchanged** — never reads `contract`, names no variant |
| 18 | `demoteFundedAwaitingContracts` | **absent** — no watch state to demote |
| 19 | `promoteLookAheadHits` | **absent** |
| 20 | `ContractManager.getContracts()` | **absent** — reads `contractRepository`, never written |
| 21 | `getContractsWithVtxos()` | **absent** |
| 22 | `refreshOutpoints()` | **absent even if explicitly asked** — existing ownership filter drops it |
| 23 | `Wallet.contractSnapshot()` | **absent** |
| 24 | `Wallet.getVtxos()` / `getSpendableVtxos()` / `getBalance()` | **absent** |
| 25 | **`VtxoManager.recoverVtxos()`** | **absent** — the untimelocked-sweep hazard, closed |
| 26 | Transaction history | **absent** |
| 27 | `annotateVtxos` | **not annotatable, never asked** |
| 28 | Service-worker `getVtxosFromRepo` | **absent** — reads `manager.getContracts()` |

Rows 8–9 and 18–28 are absent for one structural reason: ownership flows from `contractRepository` rows plus `walletRepository` VTXO rows, and a watch-only script gets neither. Rows 3, 4, 6 are the intended presences. **Rows 10–16 are the ones this revision changes**, and each was re-checked against the code rather than inherited from the previous table.

**Precedence.** If a script is in both maps the contract path wins — *unless* the contract is `retained`, in which case routing it as a contract event would answer it with `syncContracts` and undo the owner's opt-out from background channels. A `retained`-and-watched script therefore routes watch-only.

## Failsafe poll: retained-and-watched had no coverage

Raised independently by @arkana-ai-bot and @coderabbitai. `pollWatchedScripts` filtered on `!this.contracts.has(s)`, which excluded *any* script with a contract row. A `retained` contract is also excluded from the contract sweep (`getWatchedContracts()` drops it), so a retained-and-watched script had **no failsafe poll at all** — the subscription was its only delivery path.

The poll filter now mirrors the subscription routing predicate exactly, and a regression test drives the failsafe timer for that combination.

## Seeding: no baseline, and a restart re-announces

`seedLastKnownVtxos` exists so the first poll after a restart doesn't treat every persisted VTXO as new. For a watch-only script **nothing is persisted anywhere** — that is the feature — so there is no baseline to seed *from* without inventing one.

The real fork is whether to suppress the first indexer read (edge-only) or announce it (at-least-once). This announces, because the failure modes are not symmetric:

- **Suppressing** means: the consumer restarts, the counterparty funded the lockup while it was down, the script's state has not changed since the seeding read, and **no event ever fires**. A silently missed funding.
- **Announcing** means: a duplicate `vtxo_received` per restart, always dedupable downstream by outpoint.

A duplicate is recoverable; a missing event can never be invented. There is deliberately no `emitExisting` option — one behavior, one test. The contract is documented on the method as *at-least-once, re-announces on restart*.

Re-registering an already-watched script is idempotent: it keeps the baseline and only updates the label, so a caller may re-derive its whole watched set on a timer without a re-announce storm.

**Lifetime:** watch-only entries are in-memory and process-lifetime, never persisted. That is what makes "nothing about a foreign script is written anywhere" literally true, and it adds zero `ContractRepository` surface.

## Known limitation: `spendableOnly`

The watch-only poll queries with `spendableOnly: true`, matching the known consumer's own lockup lookup. A preconfirmed output is spendable in Arkade, so this does **not** miss a fresh funding — the case that matters.

It does mean an output in a recoverable or swept state is at the script but will not be reported by the poll.

**Known residual, narrowed by review but not closed.** Admission now excludes outputs that are already terminally spent, swept or batch-expired when the subscription delivers them (after @pietro909's review), which closes the arrival-time half of the phantom-spend class. Two gaps remain, both deliberate:

1. **Aging.** An output spendable when delivered that later ages into recoverable — a lockup nobody claims before its batch expires — drops out of the next `spendableOnly` poll and still yields a spurious `vtxo_spent`. This cannot be fixed at admission time, because the transition happens afterwards.
2. **Height-based expiry.** The admission check uses `isPastExpiry` with no chain tip, so it evaluates `expiresAt` but not `expiresAtHeight`. Giving the watcher a tip source is out of scope here.

Closing (1) means dropping `spendableOnly`, fetching the full set, and deriving spentness from each output's own terminal state rather than from absence. That widens the documented contract — recoverable and swept outputs would then be reported as present — so it is an event-contract decision, not a bug fix, and is left for review rather than taken unilaterally. For the known consumer it is a briefly-wrong latency hint, not a correctness problem: its lockup lookup stays the money-path authority and contradicts the event immediately.

## Compatibility

Additive at runtime. Reuse removes the exhaustive-switch break entirely — a consumer that switches on `event.type` needs no new case, and `examples/contract-manager.ts` needs no watch-only branch at all (the branch the previous revision added has been deleted).

**One break remains, deliberately.** Because `contract` is *absent* rather than optional on the watch-only variant, a consumer that reads `event.contract` after narrowing only on `event.type` will stop compiling, and must add `isContractVtxoEvent`. That is a smaller surface than an exhaustive-switch break — it touches only code that reads `contract` — and it is the mechanism that makes the isolation compile-enforced.

**Open question for the reviewer.** @pietro909 — you approved the previous shape, so this one should get your read too. `contract: Contract` absent buys the compile-enforced guard, but it still breaks external consumers who read `event.contract` and never call `watchScript`, which is the same objection that killed the new variants. The alternative is `contract?: Contract` on a single merged variant: nobody's build breaks, but an unguarded read silently yields `undefined` at runtime and `syncContracts([undefined])` becomes reachable. I chose the compile error because this boundary keeps a foreign script's outputs out of the wallet repository, but I would rather you decide it than have me decide it quietly. Happy to switch.

## What `intent-solver` can delete once this lands

Named here for coordination; **nothing in that repo is changed by this PR**.

- `registerLiveLockups` / `lockupContractRegistration` — the lockup-as-contract path. Its own docblock notes that registration buys the gate, the balance accounting and a recovery path, and *in the same stroke arms the untimelocked sweep*. With `watchScript` the fast path no longer costs that.
- **The recovery guard in `runVtxoLifecycle`** — the pre-CLTV check that skips a whole recovery pass because one immature lockup would fail an all-or-nothing settlement. Its stated cause is registration; remove the cause and the guard is dead weight. This is the biggest win: today a live swap can block the solver's own float recovery.
- `LockupWatcher`'s coverage-reconciliation half — `checkCoverage`, `readCoverage`, `seenUncovered`, `reportedUnwatched`, `onUnwatched` and the `ContractSource` dependency. Watching now implies coverage, so there is no gap left to reconcile.

Not deletable: `findLockups`. It remains the money path's authority, and these events stay latency hints.

## Test plan

Unit-only, in `test/contracts/` and `test/vtxo-manager.test.ts`.

1. Watch-only script receives over the subscription → `vtxo_received` with the right `contractScript` and VTXOs, asserted as an **exact** object so a `contract` smuggled onto a watch-only event fails here.
2. Still served when the subscription is down and only the failsafe poll runs.
3. Survives a reconnect — the second `subscribeForScripts` still carries the script.
4. **Isolation.** A funded watch-only script appears in none of `getAllContracts`, `getWatchedContracts`, `manager.getContracts`, `manager.getContractsWithVtxos`, `wallet.getVtxos({ withRecoverable: true })`, `getSpendableVtxos`, `getBalance` (every bucket), or the recoverable set — while an *owned* contract's VTXO of known value is asserted present in the same test, so the zero-assertions cannot pass vacuously.
5. **Fail-closed poll.** A rejected indexer fetch leaves `lastKnownVtxos` untouched rather than reading as an empty result.
6. Contract precedence — a script in both maps emits exactly one event, and `isContractVtxoEvent` is true for it.
7. A `retained` contract does **not** out-rank a watch-only registration — the event goes out with no `contract`.
8. **Retained-and-watched is polled** — the failsafe timer delivers for the combination that previously had no coverage.
9. `removeWatchedScript` drops it from the next `subscribeForScripts` and stops its events.
10. Admission drops already-spent / swept / batch-expired subscription outputs from the watch-only baseline.
11. Re-registering an already-watched script does not re-announce it.
12. **The guard, both directions.** An owned contract's subscription event *does* reach `walletRepository.getVtxosForScript`, and a watch-only one does *not*, in the same test. The owned side is delivered strictly after boot through a test-driven subscription, because `createContract` hydrates from the indexer on its own and an earlier version of this test passed vacuously for that reason.
13. **`onVtxosSpent` is not called for a watch-only spend** — a real wallet-state prune hook sitting after the same guard, which still catches a leak when the sync itself fails harmlessly further down.
14. **`VtxoManager` does not settle on a watch-only event** — the silent-compile consumer above.
15. **`RfqSwapManager` ignores a contract-less event on a script it tracks** — the same type on the same script, minus `contract`, must not cause an indexer read.

Service-worker parity ships in this PR (`WATCH_SCRIPT` / `UNWATCH_SCRIPT` / `GET_WATCHED_SCRIPTS`, with a handler round-trip test and a refusal test); the event already rides `CONTRACT_EVENT` and is forwarded verbatim.

## Verification

Measured on this branch, not inherited:

- `pnpm -r build`, `pnpm -C packages/ts-sdk run typecheck`, `packages/boltz-swap`, `packages/swap` — all exit 0.
- `pnpm run test:unit` — **233 files, 4389 passed, 1 skipped, 0 failed, exit 0.** Baseline at the previous head (`2ea51dc3`) was 233 / 4385 / 1 / 0; the delta is exactly the four tests added here.
- `pnpm run lint` is not runnable on a Windows checkout — biome reports 461 of 462 `ts-sdk` files as unformatted purely because they are CRLF on disk — so each changed file is format-checked against an LF copy instead. All twelve pass.
- The regtest e2e suite runs in CI (the `integration` matrix, `needs: check`), not locally.

Mutation results against the finished branch:

| Mutation | Result |
|---|---|
| Remove the guard | killed **by the compiler** (`TS2339`, quoted above) |
| Remove it via a cast | killed — watch-only event never delivered |
| Invert the guard | killed — `expected [] to not deeply equal []` (owned contract never persisted) |
| Guard placed *after* `syncContracts` | killed by two tests |
| Synthesise a `Contract` to satisfy the case | **survived at first.** Test 13 was added for it; now killed — `expected "spy" to not be called at all, but actually been called 1 times` |
| Revert the poll-filter fix | killed by test 8 |

## Caveats, stated rather than buried

- **`RfqSwapManager`'s guard (`swapManager.ts:1290`) is defensive, not load-bearing.** `pollSwap`'s own `monitored` check is the actual backstop; the guard is not the thing preventing a mis-drive. It is now covered by a test (`85c775da`) — the pre-existing `lockupEvent` factory sets `contract: {}`, which is truthy and passes `isContractVtxoEvent`, so the defensive branch had never actually been taken by any test.
- **`swap/watch.ts:194` has no test, and deliberately so.** `handleSpend` reads `event.contract.metadata` on its first line, so with the guard removed it throws into `enqueue`'s swallowing `catch` and the store is unchanged *either way*. A test there passes with the guard deleted — I wrote one, confirmed it did not fail under mutation, and removed it rather than ship false coverage. That guard is type-driven and has no distinguishable runtime behaviour.
- **`wallet.ts:1920` has no test of its own.** It would fail to compile if removed, which is a weaker guarantee than the `VtxoManager` case — that one *is* covered precisely because it would not have failed to compile.
- `WalletReceiveRotator` (row 14) is unchanged. Its safety follows from the contract-precedence rule, which is tested (test 6), but the rotator itself is not exercised against a watch-only event.

## Summary by CodeRabbit

- **New Features**
  - Added support for monitoring arbitrary scripts without registering them as contracts or affecting wallet balances.
  - Added APIs to watch, unwatch, label, and list monitored scripts.
  - Added service-worker wallet support for managing watched scripts.
  - Watched scripts now report received and spent VTXO activity, including activity recovered through polling or reconnection.

- **Bug Fixes**
  - Contract-only wallet actions now ignore activity from watch-only scripts.
  - Unsupported script-watching requests now return a clear error instead of silently failing.


packages/ts-sdk/src/contracts/contractWatcher.ts
```diff
@@ -773,20 +903,27 @@ export class ContractWatcher {
...
+        const byWatchedScript = new Map<string, VirtualCoin[]>();
         let unknownScript = 0;
         for (const vtxo of vtxos) {
-            if (!this.contracts.has(vtxo.script)) {
+            // Contract precedence: registered-and-watched yields one event.
+            const target = this.contracts.has(vtxo.script)
+                ? byContract
+                : this.watchedScripts.has(vtxo.script)
+                  ? byWatchedScript
+                  : undefined;
```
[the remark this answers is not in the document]
Kukks: Good catch — accepted and fixed in `8f584d8`.

The underlying defect was that the two data sources admitted different sets: `pollWatchedScripts` asks the indexer with `spendableOnly: true`, while `processSubscriptionVtxos` routed every `newVtxo` straight into `lastKnownVtxos`. Anything the subscription admitted but the poll would not return was read as missing on the very next tick and reported as `script_vtxo_spent` for an output nobody spent.

The watch-only received path now drops outputs that are terminally spent or swept, using `hasTerminalSpend` plus `isSwept`. Those are the closest client-side approximation available here — `canSpendOffchain` would be the exact predicate, but it needs a `TimeHeight` and the watcher has no chain-tip source. Spend notifications stay unfiltered, as you said. Covered by a new test (`keeps an already-spent subscription output out of the watch-only baseline`), which I verified fails when the guard is reverted to `false`.

## One residual case this does not close

Being explicit rather than letting it look handled: an output that is spendable when the subscription delivers it and *later transitions out* of the spendable set — a lockup whose batch expires and becomes recoverable before anyone claims it — will still produce a spurious `script_vtxo_spent` on the following poll. No client-side filter at admission time can prevent that, because the transition happens after admission.

Closing it properly means changing what the poll asks for: drop `spendableOnly`, fetch the full set, and derive "spent" from the outputs' own terminal state rather than from absence. That is a deliberate widening of the documented contract (recoverable and swept outputs would then be reported, which the current design explicitly says they are not), so I have left it as an open question on this PR rather than making the call inside a review fix.

For the one known consumer this is a latency hint that would be briefly wrong, not a correctness problem — its lockup lookup remains the money-path authority and would immediately contradict the event. But if a reviewer wants the event contract to be exact, the widening above is the change to make, and I am happy to do it here.

Kukks:
Thanks — all four verified against the code before acting, and three are now fixed in `00b356dd`.

**1. Stale example — confirmed, and worse than "won't fail CI".** I checked rather than took it: `examples` is in the package tsconfig's `exclude`, so `pnpm typecheck` skips it entirely. Running `tsc` against the file directly reproduces it exactly where you said:

```
examples/contract-manager.ts(131,66): error TS2339: Property 'contractScript' does not exist on type ...
```

That is the same negative-guard shape the widening broke in `SwapManager`, so the example was quietly demonstrating the pattern that now fails. Fixed with a positive guard that also shows the new variants, and I re-ran `tsc` against the file to confirm it compiles.

**2. Double `normalizeVtxo` — confirmed and fixed**, structured as you suggested:

```ts
if (target === byWatchedScript && eventType === "vtxo_received") {
    const n = normalizeVtxo(vtxo);
    if (hasTerminalSpend(n) || n.isSwept) continue;
}
```

**3. Orphaned `script_vtxo_spent` — confirmed, documented rather than suppressed.** You are right that it is the sibling of `8f584d8`. I did not try to suppress it, because suppressing a spend is the more dangerous direction: a spend notification for an outpoint we never admitted is still true information, and dropping it would be worse than delivering it unpaired. The method contract now says so explicitly — callers must tolerate a `script_vtxo_spent` with no preceding `script_vtxo_received`, since an output can be created and spent inside one gap in the stream.

**4. Silent-success on an absent optional method — trade-off confirmed as understood.** Both shipped implementations provide the methods for real, so the branch is unreachable in-tree. It exists only so an embedder's own partial `IContractManager` does not crash the worker.

## Still open, and deliberately not decided here

There is a residual case neither your review nor CodeRabbit's named, which I have written into the PR body: because the poll diffs a *filtered* set, an output that is spendable when the subscription delivers it and later transitions out of that set — a lockup whose batch expires into recoverable before anyone claims it — still yields a spurious `script_vtxo_spent`. It cannot be fixed at admission time, since the transition happens afterwards.

Fixing it means dropping `spendableOnly`, fetching the full set, and deriving spentness from each output's own terminal state instead of from absence. That widens the documented contract, so I have left it as an explicit decision for a human reviewer rather than folding it into a review-response commit. If you have a view on which way that should go, it would be useful.

Kukks:
@coderabbitai review

Two commits landed after your last pass (`8f584d8`, `00b356dd`) — the received-path guard plus the example and doc fixes from arkana's review. Please re-check the head.

packages/ts-sdk/src/contracts/contractWatcher.ts
```diff
@@ -773,20 +903,35 @@ export class ContractWatcher {
...
             }
-            let bucket = byContract.get(vtxo.script);
+            // The watch-only baseline has to admit only what the spendable
+            // poll would return, or the next tick reads the difference as a
+            // spend and emits `script_vtxo_spent` for an output nobody spent.
+            // Spend notifications stay unfiltered.
+            if (target === byWatchedScript && eventType === "vtxo_received") {
+                const n = normalizeVtxo(vtxo);
+                if (hasTerminalSpend(n) || n.isSwept) continue;
+            }
```
pietro909, line 949: The filter excludes `hasTerminalSpend(n) || n.isSwept` but not recoverable (batch-expired-but-unswept) outputs. The spendable poll these baselines diff against uses server-side `spendableOnly: true`, which excludes recoverable outputs — and `IContractManager.watchScript`'s own doc promises "one already recoverable or swept is not reported." If a recoverable VTXO ever announces over the subscription (`newVtxos` on handshake replay, or one that ages into recoverability while streaming), it enters `lastKnownVtxos`, misses the next spendable poll, and we emit `script_vtxo_spent` for an output nobody spent — the exact phantom-spend class this filter exists to prevent. Suggest extending the admission check to exclude batch-expired VTXOs (compare `expiresAt` against `Date.now()` — `normalizeVtxo` already surfaces it), or documenting why the indexer's `newVtxos` can't surface recoverable outputs.

packages/ts-sdk/src/contracts/contractWatcher.ts
```diff
@@ -773,20 +903,35 @@ export class ContractWatcher {
...
+        eventType: "vtxo_received" | "vtxo_spent",
         timestamp: number,
     ): void {
         const byContract = new Map<string, VirtualCoin[]>();
+        const byWatchedScript = new Map<string, VirtualCoin[]>();
         let unknownScript = 0;
         for (const vtxo of vtxos) {
-            if (!this.contracts.has(vtxo.script)) {
+            // Contract precedence: registered-and-watched yields one event.
+            const target = this.contracts.has(vtxo.script)
```
pietro909: Note: `this.contracts.has(vtxo.script)` includes `retained` contracts, which `getWatchedContracts()` deliberately keeps out of every background channel. A script that is both a retained contract and a watched script enters the subscription via the watch-only path but routes here as a contract event — and `ContractManager.handleContractEvent` answers `vtxo_received` with `syncContracts([event.contract])`, a background sync for a contract the owner explicitly retired from background channels. Narrow edge case and "registered wins" is a defensible precedent, but it silently overrides the retained opt-out. Worth a comment, or filter the contract-sided routing to `isWatchedContract`.

✅ Confirmed as addressed by @Kukks

packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts
```diff
@@ -1247,6 +1281,34 @@ export class WalletMessageHandler
...
+                case "WATCH_SCRIPT": {
+                    const manager = await this.readonlyWallet.getContractManager();
+                    await manager.watchScript?.(message.payload.script, {
+                        label: message.payload.label,
+                    });
+                    return this.tagged({
+                        id,
+                        type: "SCRIPT_WATCHED",
+                        payload: { script: message.payload.script },
+                    });
```
pietro909, line 1300: If an embedder's `IContractManager` doesn't implement the optional `watchScript`, this silently no-ops and still returns `SCRIPT_WATCHED` — the caller believes the script is being watched when nothing is subscribed. Given the interface comment says both shipped implementations provide it for real, throwing a "not supported" error here would be less misleading than a false ack. Same applies to `UNWATCH_SCRIPT` below.

packages/ts-sdk/src/contracts/contractWatcher.ts
```diff
@@ -287,6 +296,49 @@ export class ContractWatcher {
...
+        }
+        return Array.from(scripts);
+    }
+
+    /** @see IContractManager.watchScript for the delivery and filter contract. */
+    async addWatchedScript(script: string, options?: { label?: string }): Promise<void> {
+        this.watchedScripts.set(script, {
+            label: options?.label,
+            lastKnownVtxos: new Map(),
+        });
```
pietro909, line 329: Re-calling `addWatchedScript` for an already-watched script resets `lastKnownVtxos` to empty, so the next poll re-announces every live output on that script. That's consistent with the documented at-least-once semantics, but unlike `addContract` — which re-seeds from the repository and avoids re-announcement — there's no guard here. If duplicate registrations are expected (e.g., two plugin features watching the same lockup), a short comment confirming the re-announce is deliberate would save a future reader a chase.

pietro909 (commented):
**Doc follow-up (AGENTS.md, not in this diff):** the ownership doc states "`ContractWatcher` is event-only … never reads VTXO state from `IndexerProvider`." `pollWatchedScripts` is now the first watcher code that reads VTXO state from the indexer (the contract poll path diffs the wallet repository instead). The design is right — the watch-only baseline has nowhere else to live — but AGENTS.md should be updated to carve out the watch-only path in the same PR or a fast follow-up.

packages/ts-sdk/src/contracts/contractWatcher.ts
Kukks: Confirmed and taken in `2ea51dc3` — and you are right that the docstring was the tell. `IContractManager.watchScript` already promised "one already recoverable or swept is not reported", so the code was not honouring its own stated contract. That is the part of this I am least comfortable having shipped.

Admission now also excludes batch-expired outputs, using the repo's own `isPastExpiry` rather than a raw `expiresAt` comparison, so the height-based case degrades honestly instead of silently:

```ts
if (hasTerminalSpend(n) || n.isSwept || isPastExpiry(n, now)) continue;
```

`isPastExpiry` takes a `TimeHeight`. The watcher has no chain-tip source, so I pass `{ timestamp: new Date() }` only — **time-based expiry is closed, height-based expiry is not**. `isPastExpiry` returns `false` when `now.height` is undefined, so a VTXO carrying `expiresAtHeight` rather than `expiresAt` still slips through admission and is left to the poll. Giving the watcher a tip source is a bigger change than this fix deserved; say the word if you want it.

## Two different cases, and this closes one of them

Worth separating explicitly, because they look alike and only one is fixed:

**(a) Already recoverable when it arrives** — handshake replay, or an output that aged before the stream delivered it. **This is your case, and admission-time filtering closes it** (modulo the height caveat above). Covered by a new `it.each` case, `"batch-expired"`, which I verified fails when the `isPastExpiry` clause is removed: `expected "spy" to not be called with arguments: [ ObjectContaining{…} ]`.

**(b) Spendable when delivered, ages into recoverable while we hold it** — a lockup nobody claims before its batch expires. **This is not closed, and cannot be closed at admission time**, because the transition happens after admission. It is the residual I flagged in the PR body before your review. The output sits in `lastKnownVtxos`, drops out of the next `spendableOnly` poll, and we emit `script_vtxo_spent` for an output nobody spent.

So: your narrower fix is strictly better than what I had, and I have taken it — but please do not read it as closing the phantom-spend class. It closes the arrival-time half. The aging half survives.

Closing (b) means changing what the poll asks for: drop `spendableOnly`, fetch the full set, and derive spentness from each output's own terminal state instead of from absence. That widens the documented contract — recoverable and swept outputs would then be reported as present, which is the opposite of what the docstring above promises — so it is a deliberate event-contract decision rather than a bug fix, and I have not taken it unilaterally. It is the thing on this PR I would most like a second opinion on. For the one known consumer it is a briefly-wrong latency hint rather than a money-path error, since its lockup lookup stays authoritative and contradicts the event immediately.

packages/ts-sdk/src/contracts/contractWatcher.ts
Kukks: Confirmed and fixed in `2ea51dc3`. I went with the filter rather than the comment, because once I tried to write the comment I could not defend the behaviour it would have been excusing.

You framed it exactly right: "registered wins" is a defensible precedent, but `retained` is the one case where it is least obvious — and it is precisely the case where the override is most harmful. `retained` is the owner explicitly saying this contract stays out of every background channel; answering its VTXOs with `syncContracts([event.contract])` is a background sync, so the precedence rule was silently undoing an opt-out that the rest of the watcher goes out of its way to honour.

Precedence now applies only among *actively watched* contracts:

```ts
const state = this.contracts.get(vtxo.script);
const watchOnly = this.watchedScripts.has(vtxo.script);
const preferContract =
    state !== undefined && (!watchOnly || isWatchedContract(state.contract));
const target = preferContract ? byContract : watchOnly ? byWatchedScript : undefined;
```

Deliberately narrow: when `watchOnly` is false this reduces to the old `this.contracts.has(...)` test, so the pre-existing owned path is untouched. The only behaviour that changes is the case you found — both retained-contract *and* watch-only registered — where the explicit watch-only registration now wins and the event goes out as `script_vtxo_received`, which reaches no case in `handleContractEvent` and therefore syncs nothing.

New test `does not let a retained contract out-rank a watch-only registration`, mutation-verified: reverting `preferContract` to plain `state !== undefined` fails it with `expected [ 'vtxo_received' ] to deeply equal [ 'script_vtxo_received' ]`.

packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts
Kukks: Taken in `2ea51dc3` — you caught the interface comment contradicting the code one layer out. The comment claims optionality is a type-level courtesy and never a silent no-op; a `SCRIPT_WATCHED` ack from a manager that cannot watch is exactly that silent no-op, just relocated to the message bus. Refusing is the honest answer.

All three messages, not just the two you named:

- `WATCH_SCRIPT` and `UNWATCH_SCRIPT` now throw rather than ack.
- `GET_WATCHED_SCRIPTS` was worse than I first credited — it answered `?? []`, and an empty list reads as "nothing is watched", which is a *different claim* from "this manager cannot say". It throws too.

```ts
if (!manager.watchScript) throw unsupportedByManager("watchScript");
```

`handleMessage` already wraps the switch in `catch (error) { return this.tagged({ id, error }) }`, so this surfaces as an error response on the bus rather than an unhandled rejection in the worker — same shape as the existing `throw new Error("Unknown message")`. No new error plumbing.

New test `refuses watch-only messages when the manager cannot serve them` drives all three against a `{}` manager. Mutation-verified: restoring the optional-call-and-ack shape fails it with `expected undefined to be defined`.

Unreachable in-tree, since both shipped implementations provide the methods for real — it only fires for an embedder's own partial `IContractManager`, which is the exact case the optionality exists for and therefore the exact case that should get a truthful answer.

packages/ts-sdk/src/contracts/contractWatcher.ts
Kukks: You offered a comment as the likely answer; I checked the consumer first and it changed the verdict, so this got a guard instead.

`LockupWatcher.sync` in the solver does not add incrementally — it replaces its whole watched set on every sweep:

```ts
sync(scripts: readonly string[]): void {
  this.watched = [...new Set(scripts)].sort()
  void this.checkCoverage()
}
```

and its own docblock puts that sweep at three seconds. So the natural migration onto this API is a `watchScript` call per script per sweep, for the same scripts. Under the old code that reset `lastKnownVtxos` every time, which means re-announcing every live output at the script every three seconds — not the once-per-process re-announce the at-least-once semantics were chosen to buy. That is a re-announce storm, not a deliberate duplicate, and it would have been a bad surprise for the first consumer to migrate.

Registration is now idempotent: re-registering an already-watched script keeps its baseline and only updates the label. Restart still re-announces, which is the documented behaviour and the whole point.

I also put it in the `watchScript` contract, so a caller can rely on it rather than infer it:

> Re-registering a watched script is a no-op, so a caller may re-derive its whole set on a timer.

New test `re-registering an already-watched script does not re-announce it`, mutation-verified: forcing the `existing` lookup to `undefined` fails it with `expected "spy" to not be called at all, but actually been called 1 times`.

Nothing in `intent-solver` is changed by this PR — I only read it to settle which of comment-or-guard was right.

pietro909 (approved):

packages/ts-sdk/src/contracts/contractWatcher.ts
Kukks: Still stands at `d7a8e858` — flagging only what moved under it, since the head changed shape after you approved.

The admission filter itself is byte-identical:

```ts
if (hasTerminalSpend(n) || n.isSwept || isPastExpiry(n, now)) continue;
```

What changed is the event it guards. Watch-only activity no longer has its own `script_`-prefixed variants; it reuses `vtxo_received` / `vtxo_spent` and is told apart by the absence of `contract`. So the phantom-spend this filter prevents is now a spurious `vtxo_spent` rather than a spurious `script_vtxo_spent` — same defect, same fix, different name in the docs and comments. The `it.each` baseline-admission suite (including your `batch-expired` case) is unchanged and still passes.

Both residuals I named above are also unchanged and still open: the **aging** half (spendable when delivered, recoverable before anyone claims it) cannot be closed at admission time, and **height-based expiry** still slips through because `isPastExpiry` gets no chain tip. Neither is made better or worse by the event-shape change. They are written up in the PR description under *Known limitation: `spendableOnly`* so they do not quietly become someone else's surprise.

packages/ts-sdk/src/contracts/contractWatcher.ts
Kukks: **Correcting something I wrote in this thread**, because the head has moved under it and the old claim no longer holds.

I said above:

> the explicit watch-only registration now wins and the event goes out as `script_vtxo_received`, which reaches no case in `handleContractEvent` and therefore syncs nothing.

**That second half is now false.** `d7a8e858` drops the `script_`-prefixed variants entirely — watch-only activity reuses `vtxo_received` / `vtxo_spent`, so the case in `handleContractEvent` *does* match, and "syncs nothing" is no longer a property of the type system. It is now a property of a guard, and I would rather say so plainly than leave the old sentence standing.

The guard is at `contractManager.ts:2179` and `:2184`, one line at the head of each case, ahead of `syncContracts`:

```ts
case "vtxo_received":
    if (!isContractVtxoEvent(event)) break;
    await this.syncContracts({ contracts: [event.contract] });
```

It is compile-enforced rather than conventional: `contract` is *absent* (not optional) on the watch-only variant, so `event.contract` does not type-check without narrowing. Deleting the guard fails the build rather than silently regressing — I verified that by deleting it:

```
error TS2339: Property 'contract' does not exist on type
'{ type: "vtxo_received"; ... contract: Contract; timestamp: number; }
 | { type: "vtxo_received"; contractScript: string; vtxos: VirtualCoin[]; timestamp: number; }'
```

**Your finding itself is unaffected and the fix is unchanged.** `preferContract` still routes a `retained`-and-watched script down the watch-only path, and the test still pins it — though the assertion had to change, because the event *type* no longer distinguishes the two routes. It now asserts on the discriminator that does:

```ts
expect(vtxoEvents.map((e) => e.type)).toEqual(["vtxo_received"]);
expect(vtxoEvents.map(isContractVtxoEvent)).toEqual([false]);
```

## Your finding had a second half I had missed

Chasing this, @arkana-ai-bot and @coderabbitai both landed on the same gap, and it is the polling mirror of exactly what you found: `pollWatchedScripts` filtered on `!this.contracts.has(s)`, which excludes *any* script with a contract row. A `retained` contract is also excluded from the contract sweep, because `getWatchedContracts()` drops it. So a `retained`-and-watched script had **no failsafe poll at all** — the subscription was its only delivery path, and the 20-second backstop was inert for precisely the combination you flagged.

Fixed in `d7a8e858`; the poll filter now mirrors the routing predicate exactly:

```ts
const scripts = candidates.filter((s) => {
    const state = this.contracts.get(s);
    return (
        this.watchedScripts.has(s) &&
        (state === undefined || !isWatchedContract(state.contract))
    );
});
```

New test `polls a retained-and-watched script the contract sweep leaves uncovered` drives the failsafe timer for the combination. Mutation-verified: reverting the filter to `!this.contracts.has(s)` fails it with `expected "spy" to be called with arguments: [ ObjectContaining{…} ]`.

packages/ts-sdk/src/wallet/serviceWorker/wallet-message-handler.ts
Kukks: Unchanged and still correct at `d7a8e858`. Confirming rather than leaving you to diff it, since the head was reworked after you approved.

`unsupportedByManager` and all three refusals (`WATCH_SCRIPT`, `UNWATCH_SCRIPT`, `GET_WATCHED_SCRIPTS`) are untouched, and `refuses watch-only messages when the manager cannot serve them` still passes.

The event-shape rework does not reach this file at all, which is worth stating explicitly because it was on my list of things to re-check: the service worker forwards `ContractEvent` verbatim —

```ts
this.contractEventsSubscription = manager.onContractEvent((event) => {
    emitter.emit(this.tagged({ type: "CONTRACT_EVENT", broadcast: true, payload: { event } }));
});
```

— and never reads `event.contract` nor names a variant, on either the worker side or the page-side proxy. So dropping the `script_`-prefixed variants required no change here.

One detail that matters for the bus specifically: the watch-only form is distinguished by the *absence* of a `contract` property, and absence survives both structured clone and a JSON round-trip (`JSON.stringify` drops `undefined` keys rather than materialising them). The narrowing helper also tests the value, not just the key —

```ts
return "contract" in event && event.contract !== undefined;
```

— so a transport that did materialise `contract: undefined` would still be classified as watch-only rather than silently treated as owned.

packages/ts-sdk/src/contracts/contractWatcher.ts
Kukks: Unchanged at `d7a8e858` — the idempotent registration and its test both survive the event-shape rework untouched, and `re-registering an already-watched script does not re-announce it` still passes.

Only the documented contract line moved, because the event it names moved:

> Re-registering a watched script is a no-op, so a caller may re-derive its whole set on a timer.

That sentence is unchanged; the surrounding docstring now says the activity arrives as `vtxo_received` / `vtxo_spent` without the `contract`, rather than as the `script_`-prefixed variants.

Worth closing the loop on the consumer reasoning that decided this one, because it got *more* relevant rather than less. I argued for the guard over the comment on the grounds that `LockupWatcher.sync` replaces its whole watched set every few seconds, so the natural migration is a `watchScript` call per script per sweep. That is still the shape of the migration — and the poll-filter bug fixed in this same commit (see the `retained` thread) was in the other half of that same "which registrations actually get served" question. Both were cases where a script was registered and quietly not covered the way the caller would assume.

Kukks:
Force-pushed `d7a8e858` (was `2ea51dc3`). The watch-only events no longer get their own `script_`-prefixed variants — they reuse `vtxo_received` / `vtxo_spent` and are told apart by the absence of `contract`. The PR description has been rewritten to match; the consumer isolation table was re-derived from scratch rather than carried forward, because several of its rows read "absent" only *because* no case matched.

Three things reviewers should look at specifically:

1. **The isolation moved from the type system to a guard**, at `contractManager.ts:2179`/`:2184`. It is compile-enforced (deleting it is a `TS2339`, not a silent regression), but it is a guard now, and that is the whole risk of this revision.
2. **`VtxoManager` (`vtxo-manager.ts:2768`) would have compiled silently** and settled + `delegate()`d on a foreign script's funding. It reads `event.vtxos` and never `event.contract`, so the compiler had nothing to say. That is the case that justifies re-deriving the table.
3. **An open question I did not want to decide alone** — `contract` absent vs `contract?: Contract`. See the *Compatibility* section; @pietro909 approved the previous shape so I would like his read on this one.

Also fixes the retained-and-watched failsafe-poll gap that @arkana-ai-bot and @coderabbitai both raised — the poll filter now mirrors the subscription routing predicate, with a regression test.

Gate: 233 files / 4388 passed / 1 skipped / 0 failed, exit 0, against a re-measured baseline of 4385 on the previous head. Unit gate only — regtest e2e was not run.

@coderabbitai review

Kukks:
@coderabbitai review

Kukks:
Pushed `85c775da` for @arkana-ai-bot's pass-4 advisories. Two taken, two declined with reasons — CI re-run on `d7a8e858` was 7/7 green before this.

**1. JSDoc pointing external consumers at `isContractVtxoEvent` — taken.** Added to the helper, since it is the pattern a consumer has to adopt the moment they read `event.contract`.

**2. Swap guard paths untested — taken for `swapManager`, declined for `watch.ts`, and you were more right than you knew.**

You spotted that `lockupEvent` uses `contract: {}`, which is truthy and passes the guard. So the defensive branch had never been taken by any test. There is now one that emits the same type on the same script with no `contract` and asserts the manager does not read the indexer. Mutation-verified — reverting the guard to the old `event.type` check fails it with `expected 2 to be 1`.

`watch.ts` I deliberately left uncovered, and I want to be explicit rather than quietly skip it. I wrote the equivalent test, ran the mutation, and **it survived** — the test passed with the guard deleted. The reason is that `handleSpend` reads `event.contract.metadata` on its first line, so without the guard it throws into `enqueue`'s swallowing `catch`:

```ts
queue = queue.then(task).catch(() => {
    // a handler must never reject into the manager's dispatch loop
});
```

The store is unchanged either way, and nothing else is observable. A test there would have been green and worthless, so I removed it instead of shipping it. That guard is type-driven — without it the file does not compile — and it has no distinguishable runtime behaviour. This is now recorded in the PR description rather than left implicit.

**3. `examples/contract-manager.ts` no longer demonstrates watch-only handling — declined, with a reason I think is the right one.**

The example never calls `watchScript`, so it never receives a watch-only event. Under the old shape it needed a branch anyway, because the union had variants its switch did not handle — that branch existed to satisfy the type system, not to teach anything. Under reuse it needs nothing, and adding a guard for a case the example cannot produce would teach the wrong lesson.

There is also a specific reason not to put instructional code there: `examples/` is excluded from the package tsconfig, which is exactly why the stale branch survived CI in the first place. Code that cannot be typechecked is a poor place to demonstrate a narrowing pattern. The JSDoc example on the exported helper reaches the same reader through their editor instead.

**4. AGENTS.md — agreed, follow-up.** `pollWatchedScripts` is genuinely the first `ContractWatcher` path that reads VTXO state from `IndexerProvider`, and the doc says it never does. That is a real doc bug and @pietro909 raised it first. It is not in this diff and I would rather not widen the PR to carry it.

On your note about `event.vtxos` narrowing to `ContractVtxo[] | VirtualCoin[]` — agreed, and that is the intended cost. A consumer that wants the annotated fields has to prove ownership first, which is the same gate that protects the repository.

Kukks:
Pushed `5fba4d3d` — I broke CI on `85c775da` and this fixes it.

The failure was `pnpm run lint`: biome joins an arrow body onto one line when it fits, and `watchedScriptEvent` omits `contract: {}`, so unlike `lockupEvent` it is under the width and must not be wrapped. Formatting only, no behaviour change.

The miss is worth naming: my local gate was build + all three typechecks + the unit suite, and **did not include `pnpm run lint`**, which CI runs. So when I said "full gate" earlier, it was not. `pnpm run lint` is not directly runnable on this checkout — biome reports 461 of 462 `ts-sdk` files as unformatted purely because they are CRLF on disk on Windows — so I now format-check each changed file against an LF copy instead. All twelve files touched by this PR pass that check.

Also worth stating because a skipped job is not a passing job: the `integration` matrix has `needs: check`, so the lint failure gated it off entirely and **the regtest e2e did not run on `85c775da`**. It did run on `d7a8e858` and was 7/7 green, including `exit-providers-rotation` on a re-run after one `waitFor` timeout that I re-ran on the same commit to confirm it was environmental rather than mine.

Local gate on `5fba4d3d`: build 0, typechecks 0/0/0, unit **233 files / 4389 passed / 1 skipped / 0 failed**, exit 0. Against the `d7a8e858` baseline of 233 / 4388 / 1 / 0 that is +1 — the swap watch-only rejection test.

Kukks:
CI is 7/7 green on `5fba4d3d`.

**`Integration (ts-sdk / exit-providers-rotation)` is flaky, and it is not this PR.** Recording the evidence rather than just re-running until green:

| commit | first run | re-run |
|---|---|---|
| `d7a8e858` | FAIL — `Error: timeout in waitFor` (utils.ts:323) | pass |
| `5fba4d3d` | FAIL — `INVALID_ARK_PSBT (1): duplicated offchain tx` (submitTx) | pass |

Same test both times (`unilateralExitDeepChain.test.ts > prepare() emits a deep offchain send chain in physical-input topological order`), **two different failure modes**, both cleared by a re-run on the identical commit. A deterministic break would not change its symptom between runs.

Three further reasons it is not this change:

- The test has zero references to `onContractEvent`, `ContractEvent` or `watchScript`. This PR touches no send, PSBT, forfeit or unilateral-exit path.
- `5fba4d3d` is functionally identical to `d7a8e858` for `ts-sdk` runtime — a JSDoc line, one deleted comment, and a formatting fix in a `swap` test — yet one passed first time and the other did not.
- `master` has its own recent integration flake (`ts-sdk / ark-core`).

I also checked the one way this change *could* have produced a hanging `waitFor`: if `isContractVtxoEvent` ever returned false for a genuinely owned event, `wallet.ts`'s subscriber would stop emitting. It cannot — `emitVtxoEvent` sets `contract: state.contract` on both branches after an early return that guarantees `state` exists (`contractWatcher.ts:1023`), so an owned event always carries a contract.

Worth a separate issue for the flake if it keeps costing re-runs; I have not opened one, to avoid guessing at its cause from two samples.

---

**State:** head `5fba4d3d`, CI 7/7, `mergeable: MERGEABLE`, @pietro909 approved, @arkana-ai-bot pass-5 says "nothing blocks merge from my side".

`reviewDecision` is still `CHANGES_REQUESTED` — that is arkana's pass-3 review on `2ea51dc3`, which its own pass-4 and pass-5 have since closed out. A COMMENTED re-review does not clear a blocking one, so it needs dismissing to unblock. I have not dismissed it: that is the repo owner's call, not mine, and I am not merging this.

Still open for a human: the `contract` absent vs `contract?: Contract` question in the *Compatibility* section — @pietro909 approved the previous event shape, so I would like his read before this lands.

PR #859, opened 2026-09-07
fix(sdk): gate transaction history the way balance is gated

tiero:
@gringokiwi for review.

## The defect

`getBalance()` applies the spendability gate. `getTransactionHistory()` applied none — it flat-mapped every contract VTXO in, whatever its `genericallySpendable` marker.

`@arkade-os/swap` registers an offer covenant as a contract of the wallet that funds it, so the covenant's output landed in history's own-coin set and counted as change. Both legs cancelled: `spentAmount - changeAmount` went to zero, and `subtractAssets` deleted every key that reached `0n`. The ghost-row guard (`if (txAmount !== 0 || assets)`) then dropped the row entirely.

Result: funding, fill and cancel each produced **no row at all**. An `offerAsset` swap, whose whole movement is asset units on a dust carrier, disappeared completely. And since `getActivityHistory()` is `buildActivities(getTransactionHistory(), resolvers)` and resolvers contribute no rows of their own, asset swaps could not appear in activity at all.

## The fix

`buildTransactionHistory` takes the gate — exactly what `gatedContracts()` returns — and reads those VTXOs as an external counterparty.

| Step | Before | After |
|---|---|---|
| Fund the covenant | no row | SENT, the deposit |
| Solver fills | SENT of the escrowed amount | RECEIVED, the bought asset on its carrier |
| Cancel | no row | RECEIVED, the deposit coming back |

Both directions, with asset movements carried on the row instead of cancelled away.

Both read paths derive the gate off a single snapshot: `ReadonlyWallet.getTransactionHistory` from `contractSnapshot()`, and the worker's `buildTransactionHistoryFromCache` from its own `repoSnapshot()` — which it now takes itself, so neither of its two callers can hand it coins without the gate that judges them.

## Notes for review

**The covenant movement stays attributable.** Rows facing a gated contract are tagged `"gated"`, through the open string arm of `TxTag`, so a consumer can tell "sent to my own escrow" from "sent to a stranger". `"gated"` rather than a contract-flavoured name because every VTXO here belongs to a contract row — the wallet's own receive address is a `default` contract — and because it points at `WalletBalance.gated`, which is the same money. `BuiltinTxTag` is not in the package's export map, so `TxTag` stays the only public surface.

**Timestamps.** Removing the covenant coin also removed the funding tx's only local timestamp source when the deposit leaves no change — an indexer round-trip per deposit on the main thread, and on the worker's offline read a fallback to the *input* coin's timestamp, filing a fresh deposit at the wrong end of the list. `localCreatedAtByTxid` states the rule once over every coin, before the gate partitions them, so the covenant output dates the row directly and `collectArkTxidsNeedingCreatedAt` is exactly that map's complement.

**One invariant changed on purpose.** `spendability-gate.test.ts` listed history beside `getVtxos` as an ungated read. Treating it as one was the defect. That test is split: the recovery read still keeps escrowed coins, the history read now gates them. Worth a look — it's the one place this contradicts something the repo previously asserted.

**One spelling of the gate.** `isGatedVtxo`/`gatedTypeOf` now live in `contracts/spendability.ts` next to `gatedContracts`, and `gateExclusion` uses them, instead of a private fourth copy in the utils module. `gatedFrom(snapshot)` replaces four hand-kept copies of `gatedContracts(snapshot.map((_) => _.contract))`.

## Follow-ups, deliberately not in scope

`gatedContracts()` is a single predicate that also covers VHTLC lockups, so corridor rows shift as a side effect. Not carved out — a swap-only variant would leave history and balance disagreeing for lockups, which is the bug being fixed — but not tested or tuned here either:

- A gated coin settled in a **batch** changes the exit arm's `forfeitAmount` vs `settledAmount` arithmetic; batch and exit rows carry no tag. Tracked in #860.
- One ark tx spending **both** a gated coin and a wallet coin can drive the amount negative or flip an asset sign. Unreachable in the offer flow (the covenant is always the sole wallet-side input), reachable via `sendBitcoin({ selectedVtxos })`.
- A default-closed `arkade` row funded **from outside the wallet** loses the receive it used to report. Direct consequence of the external-counterparty reading; pinned in a test rather than left latent.
- SQLite and Realm wallet repositories drop `tag` on write. Pre-existing for all tags.
- The gate arrives as a fifth positional parameter carrying the contracts map. Passing a predicate instead — the shape `BalanceCapabilities.isGenericallySpendable` already uses — would let `utils/transactionHistory.ts` drop its `contracts/spendability` import entirely. Not breaking (`buildTransactionHistory` is not exported from `src/index.ts` and has no subpath), just kept out of a bugfix diff.

## Tests

17 cases in a new `Gated contracts (swap covenants)` block covering both directions, plus two wiring tests in `spendability-gate.test.ts`. Stripping the gate argument from either production call site turns exactly those two red — without them the whole feature was revertible while staying green, since the unit block never builds a wallet.

Full suite green: ts-sdk 2987, swap 785. Lint, typecheck and build clean.

https://claude.ai/code/session_01XCGopokSkmbcbNZMKnYVEG

---
_Generated by [Claude Code](https://claude.ai/code/session_01XCGopokSkmbcbNZMKnYVEG)_

## Summary by CodeRabbit

- **New Features**
  - Transaction history now recognizes gated contract activity and labels it **“gated.”**
  - Gated contract outputs are excluded from ordinary wallet-owned history while remaining available in recovery views.
  - Gated deposits, fills, and cancellations receive consistent send/receive classifications and timestamps.
  - Balance and transaction-history views apply the same gating rules across main-thread and worker paths.

- **Bug Fixes**
  - Prevented contract-controlled funds from being incorrectly reported as wallet change or ordinary wallet activity.

### DIFF
--- a/packages/ts-sdk/src/contracts/spendability.ts
+++ b/packages/ts-sdk/src/contracts/spendability.ts
@@ -46,15 +46,23 @@ export type GatedContracts = ReadonlyMap<string, string>;
  * generic spending is open to it.
  *
  * The single spelling of the per-VTXO question, so the gate cannot be asked two
- * subtly different ways. Strict on `script`, matching `isVtxoForScript`: a VTXO
- * with a missing or empty script belongs to no contract row at all, so it is the
- * wallet's own coin.
+ * subtly different ways. A VTXO with no script belongs to no contract row, so it
+ * is the wallet's own coin.
+ *
+ * The `undefined` check is deliberately not a truthiness check, so that an
+ * empty-string script still asks the map rather than being read as "no
+ * contract". The two spellings are indistinguishable on both SDK read paths —
+ * `isVtxoForScript` keeps a `""`-scripted coin out of every snapshot bucket, and
+ * `saveVtxosForContract` refuses to persist one — so this is not a guarantee to
+ * lean on, only the pre-existing semantics preserved verbatim. It is kept
+ * because this predicate now decides coin selection and the `available` balance
+ * as well as history, which is not the place to take on a gratuitous difference.
  */
 function gatedTypeOf(
     vtxo: Pick<ExcludableVtxo, "script">,
     gated: GatedContracts,
 ): string | undefined {
-    return vtxo.script ? gated.get(vtxo.script) : undefined;
+    return vtxo.script === undefined ? undefined : gated.get(vtxo.script);
 }
 
 /** {@link gatedTypeOf} as a predicate, for callers with no use for the type. */
--- a/packages/ts-sdk/src/contracts/spendability.ts
+++ b/packages/ts-sdk/src/contracts/spendability.ts
@@ -31,7 +31,7 @@ export function gatedContracts(contracts: readonly Contract[]): Map<string, stri
  * actually holds, so the four callers that need the gate do not keep
  * `gatedContracts(snapshot.map((_) => _.contract))` in step by hand.
  */
-export function gatedFrom(snapshot: readonly { contract: Contract }[]): Map<string, string> {
+export function gatedFrom(snapshot: readonly { contract: Contract }[]): GatedContracts {
     return gatedContracts(snapshot.map((entry) => entry.contract));
 }
 
@@ -50,7 +50,7 @@ export type GatedContracts = ReadonlyMap<string, string>;
  * with a missing or empty script belongs to no contract row at all, so it is the
  * wallet's own coin.
  */
-export function gatedTypeOf(
+function gatedTypeOf(
     vtxo: Pick<ExcludableVtxo, "script">,
     gated: GatedContracts,
 ): string | undefined {
--- a/packages/ts-sdk/src/contracts/spendability.ts
+++ b/packages/ts-sdk/src/contracts/spendability.ts
@@ -26,6 +26,15 @@ export function gatedContracts(contracts: readonly Contract[]): Map<string, stri
     return gated;
 }
 
+/**
+ * {@link gatedContracts} over the contract+VTXO snapshot every read path
+ * actually holds, so the four callers that need the gate do not keep
+ * `gatedContracts(snapshot.map((_) => _.contract))` in step by hand.
+ */
+export function gatedFrom(snapshot: readonly { contract: Contract }[]): Map<string, string> {
+    return gatedContracts(snapshot.map((entry) => entry.contract));
+}
+
 /** The minimum a VTXO must carry to be matched against an exclusion set. */
 export type ExcludableVtxo = { txid: string; vout: number; script?: string };
 
--- a/packages/ts-sdk/src/contracts/spendability.ts
+++ b/packages/ts-sdk/src/contracts/spendability.ts
@@ -29,6 +29,30 @@ export function gatedContracts(contracts: readonly Contract[]): Map<string, stri
 /** The minimum a VTXO must carry to be matched against an exclusion set. */
 export type ExcludableVtxo = { txid: string; vout: number; script?: string };
 
+/** {@link gatedContracts}' answer, as callers that only read it should take it. */
+export type GatedContracts = ReadonlyMap<string, string>;
+
+/**
+ * The type of the gated contract this VTXO belongs to, or `undefined` when
+ * generic spending is open to it.
+ *
+ * The single spelling of the per-VTXO question, so the gate cannot be asked two
+ * subtly different ways. Strict on `script`, matching `isVtxoForScript`: a VTXO
+ * with a missing or empty script belongs to no contract row at all, so it is the
+ * wallet's own coin.
+ */
+export function gatedTypeOf(
+    vtxo: Pick<ExcludableVtxo, "script">,
+    gated: GatedContracts,
+): string | undefined {
+    return vtxo.script ? gated.get(vtxo.script) : undefined;
+}
+
+/** {@link gatedTypeOf} as a predicate, for callers with no use for the type. */
+export function isGatedVtxo(vtxo: Pick<ExcludableVtxo, "script">, gated: GatedContracts): boolean {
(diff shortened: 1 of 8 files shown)


tiero:
Thanks — three things, two of them actioned.

**Batch-settlement undercount:** filed as #860 with the concrete failure shape (a lockup recovered via `settle({ inputs })` alongside an ordinary coin drops the exit row entirely, or reports the wrong amount, because `forfeitVtxos` loses the gated inputs while `changes` keeps a leaf whose value still includes them). Agreed it should be tracked rather than left in a PR body. The issue also carries the deeper question from this review — that `isGenericallySpendable` is fail-closed, which is right for spending but means history drops a *row* rather than a *tag* for a contract this build cannot interpret — since a fail-open custody predicate is what would give the exit arm something better to key on.

**The `scriptless` cast:** fixed in 5617eac, comment only. You're right that it reads as suppressing the check; it's actually modelling the shape a legacy repository row arrives in, which is why `isVtxoForScript` and the gate both guard on an absent script. Now says so.

**`tag` persistence:** confirmed pre-existing and unchanged — `saveTransactions` never wrote `tag` for any of the four existing values either, so `"gated"` inherits the gap rather than widening it. Left alone deliberately; adding a column is a repository-schema change with a `version` bump on `WalletRepository`, which doesn't belong in this fix. Worth noting your framing is the right one for consumers: branch on `tag` from `getTransactionHistory()`, not from the cached path.

---

tiero:
Both remaining items were answered above before this pass landed: the batch-settlement undercount is tracked in #860 (with the failure shape and the fail-open-custody-predicate question), and the `tag` persistence gap is confirmed pre-existing for all five values, left for a separate schema change. Nothing outstanding on my side.

---

pietro909:
@tiero @gringokiwi please don't merge this PR yet. It could affect other work.

bordalix:
Possibly related:
PR merged on wallet
[Fixed an issue where funding activity for Arkade swap offers could be missing from transaction history](https://github.com/arkade-os/wallet/pull/956)

tiero:
> Possibly related:
> 
> PR merged on wallet
> 
> [Fixed an issue where funding activity for Arkade swap offers could be missing from transaction history](https://github.com/arkade-os/wallet/pull/956)
> 
> 

Yes this is the fix in the wallet doing an additional loop 

Whereas this works on refactor of the SDK itself so all other implementation beside our wallet can have the same fix

I reckon it may need some more review so holding off for now

gringokiwi (approved):

gringokiwi:
Confirming the link, and answering the question it raises — because I think the thread is currently holding for the wrong reason.

## It is the same defect, not a related one

Wallet #956 quotes the code this PR changes, verbatim:

```js
const changes = fromOldestVtxo.filter((_) => _.txid === vtxo.arkTxId)
txAmount = spentAmount - changeAmount        // 110_000 − (50_000 + 60_000) = 0
if (txAmount !== 0 || assets) { sent.push(...) }   // never fires
```

Same cause — `createOffer` registers the covenant, so its output counts as the wallet's own change and the funding tx nets to zero — and same symptom, a resting offer that never appears. Two fixes for one bug at two layers.

## Merging both is safe: they cannot double-count

This is the part nobody has stated, and it is the coordination question worth answering.

The wallet's fallback is conditional on the SDK row being absent. `assetSwapResolver.prepare()` indexes `swap.fundingTxid → swap.id`, and #956 synthesizes only for swaps whose id is missing from the grouped set:

```ts
const groupedSwaps = new Set(activities.flatMap((activity) => swapIdOf(activity) ?? []))
for (const swap of swaps) {
  if (!groupedSwaps.has(swap.id)) rows.push(ungroupedOfferTx(swap, { network, assetDisplay }))
}
```

So:

- **Today.** Funding tx yields no history row → the resolver never sees the txid → the id is not grouped → the wallet synthesizes. One row.
- **With this PR.** Funding tx yields a real `SENT` row tagged `"gated"` → the resolver matches `fundingTxid` → the activity carries `swap:<id>` → the id *is* grouped → the wallet does not synthesize. One row, the real one.

Mutually exclusive by construction. #956's own docstring says it was built that way ("keyed as the resolver would key it, so the real row replaces this one"), and it holds. The wallet's loop degrades to a fallback covering the window where a record exists but history has not produced the row yet.

## What #956 landing actually changes here

It removes the urgency, not the reason to merge. #956 merged at 08:35 UTC — before the hold at 09:01 — so the user-facing bug is already fixed in production for our wallet. That makes this PR a parity change for every other consumer of `buildTransactionHistory`, not a hotfix. Waiting costs nothing user-visible, which is a stronger argument for holding than "may need more review", and it means this can be sequenced properly against the v0.5 work rather than rushed.

Worth keeping in view that the SDK fix is strictly wider: #956 iterates swap *records*, so a funding tx present in history with no record in `assetSwapRepository` — a wiped store before the restore scan lands — still shows nothing there. This PR shows the row from history alone.

## @pietro909 — the direct question

You asked us not to merge because it "could affect other work". Given the above, the collision risk with the wallet is nil. Could you name what it does collide with, and when the window opens? Happy to hold as long as needed, but "could affect other work" is not something I can schedule around or design against.

## Two follow-ups, for whoever picks them up

1. **Wallet, on the SDK bump past this PR:** `ungroupedOfferTx` becomes dead for the resting-offer case. Keep it deliberately as a belt-and-braces fallback or drop it, but it should not be left unowned. Note the two rows are not identical in shape — the synthesized amount is `fromAsset === 'btc' ? Number(fromAmount) : 0`, the SDK's is `spentAmount − changeAmount` — so if the fallback ever does fire post-bump, the amount will differ from what the same offer showed a moment earlier.
2. **Here:** a line on `BuiltinTxTag` stating that `"gated"` coverage is offchain-only pending #860. It is the last loose end from the review passes and costs nothing.

PR #861, opened 2026-09-08
feat(vtxo-manager): let renewVtxos settle into several outputs

Kukks:
## The gap this closes

A renewal has always asked for exactly **one** output. So a wallet that renews its balance gets it back as a single coin — and spending one coin pins the whole balance, which serialises anything that funds more than one thing at a time. Splitting afterwards is not equivalent: it costs a second intent fee and leaves a window where the whole balance sits on one coin.

`Wallet.settle` has always taken an *array* of outputs. What was missing is a way to reach it without also reimplementing the pricing:

- **`Wallet.settle(params)` does no fee arithmetic at all** once `params` are supplied. All of it lives in the `if (!params)` branch (`wallet.ts:3932-4042`); with params, control goes straight to output construction at `wallet.ts:4044`.
- **The three helpers that get it right are module-private and unexported** — `priceSettlementInputs`, `subtotalOf`, `deductOffchainOutputFee` (`vtxo-manager.ts:335/361/392`). They are absent from `index.ts`, so an external caller literally cannot reuse them.
- The one correctly-priced multi-input path, no-arg `Wallet.settle()`, hard-codes a single output to the wallet's own address and sweeps everything.

Net: a caller who wants to renew into a chosen shape today has to re-derive the operator's intent-fee arithmetic from scratch. Getting it wrong in the cheap direction means `INTENT_INSUFFICIENT_FEE` and a float that never renews; getting it wrong in the other direction silently overpays out of user funds.

## What this adds

`RenewVtxosOptions.split?: RenewalSplit`. The SDK hands the caller the **already-priced** subtotal, an output-fee oracle, the per-output ceiling and the dust floor, and takes back piece amounts:

```ts
await manager.renewVtxos(undefined, {
    split: {
        maxOutputs: 8,
        plan: ({ subtotal, outputFeeOn, maxAmount, dust, address }) => [ /* amounts */ ],
    },
});
```

Selection, pricing, capping and validation stay in the SDK. Only the *shape* is the caller's, because the shape is a policy the wallet cannot know.

**Plans are refused, not corrected.** Over budget, over the ceiling, more pieces than `maxOutputs`, or none at all — each throws. Quietly reshaping a bad plan would renew the float into a shape nobody asked for, which is the worst way for a caller's bug to surface. Under-claiming *is* allowed: it only leaves a residue as extra fee, the direction `deductOffchainOutputFee` already deliberately errs in.

**The batch cap had to move with it.** `vtxoMaxAmount` bounds one *output*, so judging the whole batch against a single one capped every renewal at one ceiling's worth however many pieces it emitted:

```ts
const capacity =
    vtxoMaxAmount < 0n || !options?.split
        ? vtxoMaxAmount
        : BigInt(options.split.maxOutputs) * vtxoMaxAmount;
```

Raised by the output count alone and deliberately *not* by the output fee too: the cap weighs **gross** input value while the pieces are paid from the **post-fee** subtotal, so this stays strictly under what they can absorb and defers the remainder to the next cycle.

## A hazard found and closed on the way

`RenewVtxosOptions` crosses the service-worker MessageBus — `serviceWorker/wallet.ts` posts `payload: options` straight through. **A `plan` callback is not structured-cloneable**, so passing `split` there would have died as an opaque `DataCloneError` thrown from inside `postMessage`, with nothing naming the actual constraint. It is now refused by name at that boundary, with a test asserting no `RENEW_VTXOS` message is posted.

This is why `split` is a callback rather than a data description, and why it is honestly unavailable on the service-worker wallet: the plan depends on a budget that is not known until the SDK has priced and capped the inputs, so it cannot be precomputed and shipped as data.

## Assets

Renewal selects through `getSpendableVtxos`, which applies **no asset filter**, so an asset-bearing VTXO can be renewed. `Wallet.settle` does not burn those assets — it bags every input asset and assigns the whole lot to the **first** output whose script matches the destination (`findDestinationOutputIndex` is a plain `findIndex`, `delegate.ts:478`).

With **one** output that is total and unambiguous, and it is what the unsplit path has always done — this PR does not introduce asset handling. With a **split** it is not: every piece lands on the same wallet address, so the entire holding goes to **piece 0** while the caller believes it received interchangeable pieces. A pool built that way has one contaminated piece and no way to know which — and in a consumer that sizes pieces as funding rungs, piece 0 is typically the largest one.

So `RenewalSplitContext` reports `hasAssets`, and **a plan that returns more than one piece while it is set is refused**.

**Refusing the SPLIT rather than the renewal is deliberate, and I would push back on refusing outright.** Renewal is what stops a float expiring. Blocking it whenever the wallet holds an asset would leave that float drifting to batch expiry and into `recoverable`, which is strictly worse than one fat coin — and asset-bearing floats are a real, occurring state, not a hypothetical. A single piece is exactly what the unsplit path already produces and carries the assets unambiguously, so a caller reading `hasAssets` has a correct path that keeps renewing, while one that ignores it fails closed instead of silently poisoning a pool.

`hasAssets` is read off the *final* selection, after pricing and capping, so a cap that drops the only asset-bearing input frees the plan to split again.

**The batch cap moves with the guard, or the guard strands the float.** Refusing a multi-piece split while still admitting `maxOutputs` ceilings' worth leaves an asset-bearing batch with no valid plan at all under any operator that sets a per-output ceiling: split it and the asset guard refuses, return the single piece the guard demands and it exceeds `vtxoMaxAmount`, so the ceiling guard refuses instead — renewal fails every pass and the float expires. An asset-bearing batch therefore takes the unsplit path's cap, deferring the overflow to the next cycle like every other cap here. Judged *before* `capSettlementBatch`, so it cannot admit a batch it then strands; the two reads can only disagree in the safe direction. Caught by writing the test first, which failed with `Renewal split output 8000 exceeds the per-output limit 5000`.

This mirrors the existing `has-assets` idiom (`wallet.ts:5272-5276`), which refuses asset-bearing VTXOs from the BTC-only cash sweep *"up front rather than left to a server rejection"*.

Confirmed by reading, not executed: no test harness exists anywhere in this repo for `Wallet.settle`'s asset routing, so the piece-0 consolidation is established from straight-line code (`wallet.ts:4078-4122`) rather than observed. The guard itself is tested and mutation-verified.

## Compatibility

**Omitted, the default path is byte-identical** — same single output, same `deductOffchainOutputFee`, same dust message. `deductOffchainOutputFee` is now a one-line wrapper over a new `offchainOutputFee`, which is the same arithmetic factored so the split can price each piece at its own size (a proportional policy makes `gross - n * fee` and a single evaluation at the subtotal both wrong).

## Testing

`packages/ts-sdk/test/vtxo-manager.test.ts` — 12 new cases: the pieces actually settled, the subtotal offered net of input fees, per-piece output-fee pricing, the ceiling/dust context, the raised batch admission, and one refusal per rejection reason. Plus 1 in `test/serviceWorker/wallet.test.ts` for the MessageBus guard.

Plus 6 asset cases: the flag reported both ways, the multi-piece refusal, an asset-bearing float still renewing into one piece, and the flag clearing when a cap drops the only asset-bearing input.

**11 of the 12 original split tests were confirmed failing with the source change stashed**, and both halves of the asset guard are mutation-verified: disabling the refusal fails 1 test, hardcoding `hasAssets: false` fails 2. The twelfth is the deliberate regression guard that the no-`split` path still emits exactly one output, so it passes either way by design.

Gate on this branch's base (`102ff3b4`): **179 files / 2997 passed + 1 skipped, no type errors**. On the branch: **179 files / 3016 passed + 1 skipped, no type errors** — +19, exactly the tests added. Rebased onto `102ff3b4` (#857) with no conflicts; #857's `vtxo-manager.ts` changes are in the contract-event subscription, not in the renewal path.

## Not covered

No regtest run — this is exercised by unit tests only. The arithmetic is the same as `runPeriodicSettle`'s and the existing intent-fee tests, but a multi-output renewal has not been observed being accepted by a live arkd.

Downstream consumer, and the change that motivated this: arkade-os/intent-solver#78

## Summary by CodeRabbit

- **New Features**
  - Renewal transactions can be split into multiple outputs using configurable plans.
  - Added public SDK types for configuring renewal splits and accessing planning details.
  - Existing renewals continue to produce a single output by default.

- **Bug Fixes**
  - Split renewals now account for output limits, fees, dust thresholds, and invalid configurations.
  - Renewals involving assets remain restricted to a single output.

- **Known Limitations**
  - Renewal splitting is not supported through the service-worker wallet interface.

### DIFF
--- a/packages/ts-sdk/src/index.ts
+++ b/packages/ts-sdk/src/index.ts
@@ -173,6 +173,8 @@ import { isVtxoExpiringSoon, VtxoManager } from "./wallet/vtxo-manager";
 import type {
     IVtxoManager,
     RenewVtxosOptions,
+    RenewalSplit,
+    RenewalSplitContext,
     SettlementConfig,
     MigrateDeprecatedSignerOptions,
     DeprecatedSignerMigrationReport,
@@ -976,6 +978,8 @@ export type {
     SettlementConfig,
     IVtxoManager,
     RenewVtxosOptions,
+    RenewalSplit,
+    RenewalSplitContext,
     MigrateDeprecatedSignerOptions,
     DeprecatedSignerMigrationReport,
     DeprecatedSignerReport,
--- a/packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
+++ b/packages/ts-sdk/src/wallet/serviceWorker/wallet.ts
@@ -2135,6 +2135,15 @@ export class ServiceWorkerWallet
                 eventCallback?: (event: SettlementEvent) => void,
                 options?: RenewVtxosOptions,
             ): Promise<string> {
+                // `options` is posted to the worker and `split.plan` is a function,
+                // which structured clone refuses. Named rather than left to surface
+                // as a DataCloneError from inside postMessage.
+                if (options?.split) {
+                    throw new Error(
+                        "renewVtxos: split is not supported over the service worker — " +
+                            "its plan callback cannot cross the message bus",
+                    );
+                }
                 const message: RequestRenewVtxos = {
                     tag: messageTag,
                     type: "RENEW_VTXOS",
--- a/packages/ts-sdk/src/wallet/vtxo-manager.ts
+++ b/packages/ts-sdk/src/wallet/vtxo-manager.ts
@@ -420,10 +420,17 @@ function offchainOutputFee(amount: bigint, estimator: Estimator, arkAddress: str
  * direction {@link deductOffchainOutputFee} already errs in.
  */
 function planSplitOutputs(split: RenewalSplit, context: RenewalSplitContext): bigint[] {
-    const { subtotal, outputFeeOn, maxAmount } = context;
+    const { subtotal, outputFeeOn, maxAmount, hasAssets } = context;
     const amounts = [...split.plan(context)];
     if (amounts.length === 0) {
         throw new Error(`Renewal split produced no outputs for a subtotal of ${subtotal}`);
+    }
+    if (hasAssets && amounts.length > 1) {
+        throw new Error(
+            `Renewal split produced ${amounts.length} outputs while the selected inputs carry assets: ` +
+                "settle assigns every input asset to the first matching output, so they would all land on " +
+                "piece 0. Return a single piece when the split context reports hasAssets.",
+        );
     }
     if (amounts.length > split.maxOutputs) {
         throw new Error(
@@ -792,6 +799,14 @@ export interface RenewalSplitContext {
     dust: bigint;
     /** The address every piece lands on — the wallet's own, after any signer rotation. */
     address: string;
+    /**
+     * Whether any selected input carries assets — when it does, a plan MUST
+     * return a single piece. `Wallet.settle` bags every input asset onto the
+     * FIRST output matching the destination script, and every piece is on that
+     * script, so a split would silently put the whole holding on piece 0. The
+     * SPLIT is refused rather than the renewal: an unrenewed float expires.
+     */
+    hasAssets: boolean;
 }
 
 /**
@@ -1870,6 +1885,9 @@ export class VtxoManager implements AsyncDisposable, IVtxoManager {
                       maxAmount: vtxoMaxAmount,
                       dust: dustAmount,
                       address: arkAddress,
+                      // Read off the FINAL selection, so a cap that dropped the only
+                      // asset-bearing input frees the plan to split again.
+                      hasAssets: vtxos.some((vtxo) => (vtxo.assets?.length ?? 0) > 0),
                   })
                 : // The fee an intent offers IS `sum(inputs) - sum(outputs)`, so the
                   // output has to come back short by what the operator's programs
--- a/packages/ts-sdk/src/wallet/vtxo-manager.ts
+++ b/packages/ts-sdk/src/wallet/vtxo-manager.ts
@@ -800,11 +800,11 @@ export interface RenewalSplitContext {
     /** The address every piece lands on — the wallet's own, after any signer rotation. */
     address: string;
     /**
-     * Whether any selected input carries assets — when it does, a plan MUST
-     * return a single piece. `Wallet.settle` bags every input asset onto the
-     * FIRST output matching the destination script, and every piece is on that
-     * script, so a split would silently put the whole holding on piece 0. The
-     * SPLIT is refused rather than the renewal: an unrenewed float expires.
+     * Whether any selected input carries assets — then a plan MUST return one
+     * piece. `Wallet.settle` bags every input asset onto the FIRST output
+     * matching the destination script, and every piece is on it, so a split
+     * would silently put the whole holding on piece 0. The SPLIT is refused
+     * rather than the renewal: an unrenewed float expires.
      */
     hasAssets: boolean;
 }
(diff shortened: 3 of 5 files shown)


Kukks:
@coderabbitai review

Kukks:
Thanks — M1 is a fair hit and I'd defended the old wording too weakly. Addressed in `d7e4c3a4`.

**M1 + M2 — fixed.** You're right, and the sharp part isn't that the number was untrue in isolation: `vtxoMaxAmount` really is the per-output ceiling. It's that the *count* is measured against `capacity`, so the sentence as a whole doesn't cohere — an operator reading "3 VTXO(s) exceed the per-output limit 5000" would work out which of their VTXOs exceed 5000 and get a different set from the three being reported.

Both sites now render one `limitLabel`:

```ts
const limitLabel =
    capacity === vtxoMaxAmount
        ? `per-output limit ${vtxoMaxAmount}`
        : `combined split capacity ${capacity} (${options?.split?.maxOutputs} x ${vtxoMaxAmount})`;
```

Close to your suggestion, computed next to `capacity` so it needs no non-null assertions and the warning and the throw cannot drift apart. With no split it renders `per-output limit 10000` — byte-identical to before, and the existing warn test passes unchanged.

Covered by a new test that exercises both in one path (a 22000 VTXO against `vtxoMaxAmount: 5000, maxOutputs: 4` overflows `capacity`, warns, and then empties the batch), and mutation-verified: restoring the stale text fails it.

**Low (dust inside `planSplitOutputs`) — not taking it, and here's why.** The dust check today is one loop over `amounts` covering *both* paths:

```ts
const amounts = options?.split ? planSplitOutputs(...) : [deductOffchainOutputFee(...)];
for (const amount of amounts) { if (amount < dustAmount) throw ... }
```

Moving it inside `planSplitOutputs` doesn't relocate the rule, it *duplicates* it — the single-output path still needs its own check — and leaves two copies of a money-path threshold to drift. There's a second reason: that throw's wording (`Total amount N is below dust threshold M`) is load-bearing downstream. At least one consumer classifies renewal outcomes by matching `is below dust threshold` as benign, and having the split and single-output paths emit the same sentence is what keeps that working.

You're right that the *docblock* overstated the contract, though, so I narrowed it: `planSplitOutputs` validates a plan's SHAPE — count, ceiling, budget — and dust is explicitly the caller's, applied to every output whichever path produced it. Happy to reconsider if you'd rather have the duplication.

**Observation (no live arkd coverage) — agreed, and it stands.** Multi-output renewal has not been executed against a real server, and the asset-to-first-output routing is established from straight-line code (`delegate.ts:478`), not observed. It's in the PR body and I'd rather it stay visible than be quietly dropped. Worth adding: there is currently *no* test harness anywhere in this repo for `Wallet.settle`'s asset routing, which is why I couldn't assert it directly.

Gate on this branch: **179 files / 3017 passed + 1 skipped, no type errors**, against **2997** on `102ff3b4` — +20, exactly the tests added.

One correction to the summary, since it matters for anyone reading this later: the `hasAssets` guard as first written was **not** sufficient. Refusing a multi-piece split while still admitting `maxOutputs` ceilings' worth left an asset-bearing batch with no valid plan at all under any operator setting a ceiling — split and the asset guard refuses, return the single piece it demands and the ceiling guard refuses. Renewal would have failed every pass and the float expired. Fixed in `20c6308f` by giving an asset-bearing batch the unsplit path's cap, judged before `capSettlementBatch` so it cannot admit a batch it then strands.

Issue #872, opened 2026-09-08
Pre-commit hook passes while running no lint and no tests, in any git worktree


Kukks:
`.husky/pre-commit` is:

```
pnpm exec biome format config
pnpm --filter="...[HEAD]" run --if-present lint
pnpm --filter="...[HEAD]" run --if-present test:unit
```

In a **git worktree**, both filtered steps print

```
No projects matched the filters in "<worktree path>"
```

and exit 0. The hook therefore passes having run only `biome format config` — no lint, no tests. It is a green signal that carries no information, and anyone committing from a worktree is shipping ungated.

I hit this committing a change to `packages/swap` and only noticed because the hook finished in seconds instead of the usual several minutes.

## Reproduction

pnpm 10.25.0, Windows.

| Checkout | Change | `pnpm --filter="...[HEAD]"` |
|---|---|---|
| main clone | unstaged edit in `packages/boltz-swap` | matches `boltz-swap` — correct |
| worktree | unstaged edit in `packages/swap` | **no projects matched** |
| worktree | same edit, staged | **no projects matched** |
| worktree | `...[HEAD~1]`, previous commit touched `packages/swap` | **no projects matched** |

Reproduced in two independent worktrees, one created fresh from `origin/feat/v0.5` for the test, so it is not an artefact of one working copy.

Project discovery is fine in the same worktrees — `pnpm --filter="@arkade-os/swap" exec pwd` resolves correctly. It is specifically the changed-since detection that returns an empty set. A linked worktree has `.git` as a *file* containing a `gitdir:` pointer rather than a directory, which is the usual cause of this class of failure.

## Why this is worth fixing here and not only upstream

Even with pnpm fixed, the hook cannot distinguish "no project changed" from "the filter selected nothing" — both are an empty set and both exit 0. A gate that silently selects nothing should fail, not pass.

Suggested hardening, independent of the upstream bug:

- fail the hook when the filter resolves to zero projects while `git diff --cached --name-only` is non-empty, or
- drop the filter and run lint/`test:unit` across the workspace, or
- select from the staged file list directly rather than from a commit range, since at pre-commit time the changes are in the index and not in any commit.

Happy to open a PR for whichever shape is preferred.

Kukks:
Two more occurrences today, in unrelated work, which I think makes this self-evidencing.

[...]

Issue #871, opened 2026-09-08
coverage.ts docblock overstates what contract retirement does


Kukks:
`retireOfferContract`'s docblock in `packages/swap/src/coverage.ts` says:

> `retained`, not deleted: the row is what keeps the deposit's VTXOs annotatable and its history readable, while `retained` is what drops it from the subscription, the failsafe poll **and every sync**.

The last clause is wrong. Retirement removes a contract from `watcher.getWatchedContracts()`, so the subscription and the failsafe poll do narrow — but it does not narrow every sync.

Measured directly with 8 registered contracts, 4 set to `watch: "retained"`:

```
BEFORE retirement -> scripts per pre-send indexer call: [ 8 ]
AFTER  retirement -> scripts per pre-send indexer call: [ 8 ]
watcher.getWatchedContracts(): 4
```

Mechanism in 0.5.0-rc.4: `Wallet.getVtxos()` -> `contractSnapshot()` -> `getContractsWithVtxos()` is called with no filter; `buildContractsDbFilter({})` leaves `watch` undefined, so retained rows still match; and it passes `contracts` explicitly to `syncContracts({ contracts })`, bypassing the `options.contracts ?? this.watcher.getWatchedContracts()` fallback that is the only place `isWatchedContract` is applied.

This is the sentence that seeded the assumption that retirement would shrink the pre-send synced set. It cost at least two people a wrong conclusion, so it is worth correcting even though it is only a comment.

#870 has the full write-up of the mechanism.

Suggested fix: narrow the clause to the subscription and the poll, and say plainly that `contractSnapshot()` reads retained rows regardless.

Issue #870, opened 2026-09-08
Every send syncs every registered contract, and retirement does not narrow that set


Kukks:
Every offchain send syncs **every registered contract** against the indexer, and contract retirement does not narrow that set. Reported with measurements, plus a design and the hazard that constrains it.

## The path

`Wallet._sendImpl` opens with `this.getVtxos({ withRecoverable: false })`, and in `@arkade-os/sdk@0.5.0-rc.4` that reaches the indexer:

```
Wallet.getVtxos(filter)
  -> contractSnapshot()
  -> contractManager.getContractsWithVtxos()      // no filter
  -> getContracts(undefined)
  -> buildContractsDbFilter({})                   // { script, state, type, watch } all undefined
  -> syncContracts({ contracts })                 // network
```

So the cost of a spend scales with how many contracts the wallet has ever registered, not with how many coins it is about to select from.

`getAllNormalizedVtxos` then chunks the scripts at `SCRIPT_QUERY_CHUNK_SIZE = 32` and walks the chunks **serially**, with a hard `await new Promise(r => setTimeout(r, 500))` between pages within a chunk.

## Measurements

Against `mutinynet.arkade.sh`, from Node, `@arkade-os/sdk@0.5.0-rc.4`.

**The pre-send sync, on a wallet with only 4 registered contracts** — `getVtxos()` timed on its own, immediately before a `send()` that internally repeats it (n=6):

```
preSendSync(getVtxos)=1231ms   full send()=3901ms
preSendSync(getVtxos)= 878ms   full send()=2633ms
preSendSync(getVtxos)= 820ms   full send()=5391ms
preSendSync(getVtxos)= 972ms   full send()=1617ms
preSendSync(getVtxos)= 938ms   full send()=1326ms
preSendSync(getVtxos)= 866ms   full send()=2142ms

avg preSendSync=951ms   avg send=2835ms
```

Roughly a third of a send, at four contracts.

**Serial-chunk cost**, 3 reps each, freshly generated (empty) scripts so this is pure per-request overhead:

```
scripts=  1 chunks= 1  min=173 p50=178 max=320 ms
scripts=  8 chunks= 1  min=255 p50=407 max=716 ms
scripts= 16 chunks= 1  min=142 p50=143 max=676 ms
scripts= 32 chunks= 1  min=178 p50=194 max=605 ms
scripts= 64 chunks= 2  min=314 p50=524 max=2087 ms
[...]

PR #873, opened 2026-09-08
fix(swap): retire a corridor lockup when it is SPENT, not merely terminal

Kukks:
## The money-safety half

`finalize()` calls `retireContract()` for **any** terminal state, `failed` included. But `failed` means an action missed its window — it says nothing about the covenant. A swap can fail with its lockup **still funded and still refundable**, and retiring that drops a live lockup off the watcher: the subscription and the failsafe poll stop seeing money that is still the user's, while the refund is still the thing that ends the swap.

It is the same hazard `RETIRABLE` already excludes `recoverable` for on the offer side, and the docblock there states the rule outright — *"a swept deposit is still the user's money at that script, and unwatching it is how it goes missing."*

`LOCKUP_RETIRABLE` is read off `RfqSwapState`'s own definitions rather than copied from `RETIRABLE`, whose members are offer statuses:

- `settled` — "the lockup was spent by a hash-verified claim"
- `refunded` — "the lockup was spent by something other than a claim"

Both say **spent**, which is the only thing that makes a script safe to stop watching. `failed`, `needs_counterparty` and every live state are excluded.

## The growth half

`RfqSwapManager.retireContract` already existed, so a corridor lockup did leave the watcher once its swap ended. What it did not do was distinguish a swap that **ended** from a covenant that was **spent**. With the guard, the watch set still shrinks on the two states where shrinking is safe.

## What this does NOT do — please do not re-derive a benefit from the diff

**The pre-send sync is unaffected.** Retirement narrows `watcher.getWatchedContracts()` and nothing else. Measured directly with 8 registered contracts, 4 set to `watch: "retained"`:

```
BEFORE retirement -> scripts per pre-send indexer call: [ 8 ]
AFTER  retirement -> scripts per pre-send indexer call: [ 8 ]
watcher.getWatchedContracts(): 4
```

`Wallet.getVtxos()` -> `contractSnapshot()` -> `getContractsWithVtxos()` runs unfiltered — `buildContractsDbFilter({})` leaves `watch` undefined, so retained rows still match — and it passes `contracts` explicitly to `syncContracts({ contracts })`, bypassing the `options.contracts ?? this.watcher.getWatchedContracts()` fallback that is the only place `isWatchedContract` is applied. So there is **no latency or scaling benefit here, and no effect on the 33-contract chunking cliff**. See #870 for the mechanism, and #871 for the `coverage.ts` docblock that overstates what retirement does.

A consequence worth stating for anyone worried this abandons coins: because `getSpendableVtxos()` -> `contractSnapshot()` is also the selection path `renewVtxos` uses, **a retained contract's VTXOs remain renewal candidates**. Retirement is less consequential than the word suggests — it is a watcher change, not an eviction.

## Tests

The state table, plus the case that matters: a receive whose claim kept throwing until its window shut, which ends `failed` with its lockup still funded and must stay watched. Removing the guard turns that one red.

## What I actually exercised

The merge brought in `packages/swap/src/reveal.ts` and the covclaimd Reveal client, so **"gate green at `784cab82`" is a statement about the union and not an endorsement of the base's new code**. What I exercised is the retirement path.

Gated against the merged base checked out and measured directly, rather than subtracting:

| | base `29d9138e` alone | tip `784cab82` | delta |
|---|---|---|---|
| ts-sdk | 177 files / 2963 passed, 1 skipped | 177 / 2963, 1 skipped | 0 |
| swap | 62 files / **1370** passed | 62 / **1377** passed | **+7** |

`pnpm -r build` exit 0, zero failing, +7 exactly the tests added here.

**Those numbers come from manual runs, not from the pre-commit hook.** The hook gated nothing: `pnpm --filter="...[HEAD]"` matches no projects in a git worktree, so both the lint and `test:unit` steps printed "No projects matched the filters" and exited 0, leaving only `biome format config` to run. Filed as #872 — worth reading, since it means a passing hook is not evidence for anyone committing from a worktree.

### DIFF
--- a/packages/swap/src/coverage.ts
+++ b/packages/swap/src/coverage.ts
@@ -29,6 +29,7 @@
  * the same backstop every other best-effort step here relies on.
  */
 import type { IContractManager } from "@arkade-os/sdk";
+import type { RfqSwapState } from "./rfqSwapState";
 import type { AssetSwapStatus } from "./store";
 
 /**
@@ -60,6 +61,11 @@ export const RETIRABLE: readonly AssetSwapStatus[] = ["fulfilled", "cancelled"];
 
 /** The one contract-manager capability changing coverage needs. */
 export type OfferContractRetirer = Pick<IContractManager, "setContractWatchState">;
+
+/** Corridor states after which the LOCKUP no longer holds funds: both mean it
+ *  WAS SPENT. `failed` is terminal too and excluded on purpose — an action that
+ *  missed its window says nothing about a covenant that can still be funded. */
+export const LOCKUP_RETIRABLE: readonly RfqSwapState[] = ["settled", "refunded"];
 
 /**
  * Scripts whose address has been handed out, by the time it was handed out.
--- a/packages/swap/src/swapManager.ts
+++ b/packages/swap/src/swapManager.ts
@@ -116,6 +116,7 @@ import {
     type RfqSwapRecord,
 } from "./rfqRecord";
 import { RefundNotLocallyPossibleError } from "./refundBlocked";
+import { LOCKUP_RETIRABLE } from "./coverage";
 import { isRfqSwapTerminal, type RfqSwapState } from "./rfqSwapState";
 
 // ── Records ──────────────────────────────────────────────────────────────────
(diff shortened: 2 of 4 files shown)


Kukks:
@coderabbitai review

Issue #875, opened 2026-09-08
Provider tests mock baseFetch with objects that are not Responses


Kukks:
The provider tests mock `baseFetch` with object literals that are not `Response`s. It reads as a harmless shortcut and is not: it blocked a correct implementation of an unrelated change, and permissive mocks of this shape have hidden real bugs here before.

## The shape

Sixteen top-level test files mock the fetch boundary:

```
delegate-provider  esplora-watch-addresses  esplora  fetch  hdWalletCapable
indexer-provider  provider-availability  provider-build-version-header
rate-gate  recipient-address-binding  serviceWorkerContractSecrets  wallet
walletBoardingRotation  walletBoardingTxs  walletHdRotation  walletNewAddresses
```

and the mock is typically:

```js
mockFetch.mockResolvedValueOnce({
    ok: true,
    json: () => Promise.resolve({ vtxos: [] }),
});
```

There are 82 such `ok: true` literals across the top-level test files. A real `Response` also has `status`, `statusText`, `headers`, `text()`, `arrayBuffer()`, `body`, and `clone()`. The mock has two properties, so any production code that reaches for a third is untested — and, worse, cannot be written without breaking sixteen files at once.

## How it blocked a change

While adding in-flight coalescing of identical concurrent indexer reads, the natural seam was `indexerFetch` — one funnel, every indexer call passes through it, and it already separates GET/HEAD from other methods, so "reads only, never `subscribe`" falls out of a line that exists.

Coalescing requires giving each waiter its own body, because a single `Response` cannot be read twice. That needs `arrayBuffer()` or `clone()`. Neither exists on the mocks, so the options were:

1. change sixteen test files as part of a load-only performance fix;
2. sniff for the capability and behave one way under test and another in production;
3. narrow the change to a seam that never touches a `Response`.

(2) is disqualifying on its own terms. (1) turns a small perf change into a test refactor reviewed as a footnote. So the change went in at (3) — correct, and measurably effective, but at a seam chosen by the test mocks rather than by the design.

`indexerFetch` already calls `res.clone().text()` on the non-ok path, so the mocks are *already* only viable because most tests never take that branch.

## Why it is worth fixing on its own

[...]

PR #876, opened 2026-09-08
perf(indexer): serve an identical in-flight vtxo read from one request

Kukks:
A send that leaves change makes the indexer emit `vtxo_spent` and `vtxo_received` for the wallet's own contract milliseconds apart, and `ContractManager.handleContractEvent` delta-syncs the named contract on **both** arms. The same `GET /v1/indexer/vtxos` goes out twice, concurrently.

**This is a load fix, not a latency one**, and the measurements below are included partly to show that.

## Traced, not guessed

The first read of this was "duplicate queries", which invites the wrong fix. Capturing the synchronous caller at every `syncContracts` entry during one send gives the real shape:

```
t+3ms   getContractsWithVtxos()      at Wallet.contractSnapshot
                                     at Wallet.getSpendableVtxos
                                     at Wallet._sendImpl
t+3ms   syncContracts(contracts=2)   at ContractManager.getContractsWithVtxos

t+72ms  syncContracts(contracts=1)   at ContractManager.handleContractEvent  (case "vtxo_received")
t+74ms  syncContracts(contracts=1)   at ContractManager.handleContractEvent  (case "vtxo_spent")
```

Not a background poll, and not a redundant foreground read — two event arms, each doing its job:

```js
case "vtxo_received":
  await this.syncContracts({ contracts: [event.contract] });
  ...
case "vtxo_spent":
  await this.syncContracts({ contracts: [event.contract] });
```

Both are individually correct, so there is no call site to delete. Two callers want the identical answer at the same instant and neither needs a distinct result, which is what an in-flight coalescer is for.

## What changed

`fetchVtxosPage` shares the wire read behind a per-URL in-flight map on the provider instance.

The conversion stays **per caller**. The shared value is the validated JSON; each caller then runs its own `data.vtxos.map(convertVtxo)`, so two callers served from one request get their own array *and* their own coins. Isolation is structural rather than a claim about what callers happen to do — a survey of today's callers cannot bind a future one.

## Measurements

Same build, same server, toggling only this change.

| | vtxos calls | redundant | redundant indexer time | wall |
|---|---|---|---|---|
| local regtest, 5 sends, before | 13 | 4 (31%) | 60 ms | **276 ms** |
| local regtest, 5 sends, after | 9 | **0** | **0 ms** | **288 ms** |
| mutinynet, 4 sends, before | 10 | 3 (30%) | 4,112 ms | 12,407 ms |
| mutinynet, 4 sends, after | 6 | **0** | **0 ms** | 9,878 ms |

About **1.0 s of redundant indexer work per send** on mutinynet.

The local wall figures are the honest read on latency: **276 ms before, 288 ms after — no user-visible change**, which is expected, because the duplicate requests overlap. The mutinynet wall column moved 12,407 → 9,878 ms and **that is not claimed as a win**: per-send time there spans 1.2–5.4 s from network variance alone, and this is n=4 with one sample per arm. Nothing here should make a send feel faster; it removes work the indexer was doing twice.

## Reads only — and why `subscribe` must never be added

`POST /v1/indexer/script/subscribe` looks duplicated by endpoint and is not. Each call carries a **larger script set** and the `subscriptionId` the previous one returned:

```
t+   17   subscribe  n=1  sub=-         scripts=[a6d055a2]
t+   39   subscribe  n=2  sub=d203ce22  scripts=[a6d055a2,fda32198]
```

They are the watcher incrementally extending its watch set. Serving the second from the first would drop a contract from it. Keying on the vtxo read's own URL means this coalescer cannot reach them, and it should stay that way.

## Tests, and a bug in my own test worth reading

Three cases: one request serves two identical concurrent reads; each caller gets an independently mutable result; a later read after the shared one settles re-requests.

Both were mutation-checked, and the second one **failed the check on its first version**:

```js
a.vtxos.reverse();
a.vtxos[0].value = 999;              // after the reverse, this is the OTHER coin
expect(b.vtxos[0].value).toBe(1000); // so it never observed the shared write
```

With isolation deliberately removed — sharing the converted array between callers — that test **still passed**. It would have shipped green proving nothing. Rewritten to mutate by identity (`find(v => v.txid === "aa")`), it goes red under exactly that mutation.

Confirmed both directions:

| mutation | result |
|---|---|
| share the converted array between callers | independent-mutability test RED |
| bypass the coalescer entirely | one-request and mutability tests RED |
| as submitted | 17/17 green |

Suite: `packages/ts-sdk` 179 files, 2997 → 3000 passed, 1 skipped, 0 failed. The +3 are these. `typecheck` and `biome` clean.

## Related

The seam is narrower than it could be. The natural place is `indexerFetch` — one funnel for every indexer call, already separating GET/HEAD from other methods. Coalescing there needs each waiter to get its own body, so it needs `arrayBuffer()` or `clone()`, and sixteen test files mock `baseFetch` with `{ ok, json }` object literals that have neither. Rather than turn a load-only fix into a sixteen-file test refactor, that is filed separately as #875.

## Summary by CodeRabbit

* **Performance**
  * Improved handling of concurrent identical VTXO reads by sharing a single request while it is in progress.
  * Subsequent requests after completion continue to fetch fresh data.

* **Bug Fixes**
  * Ensured results from concurrent callers remain independently mutable.

### DIFF
--- a/packages/ts-sdk/src/providers/indexer.ts
+++ b/packages/ts-sdk/src/providers/indexer.ts
@@ -401,6 +401,12 @@ export class RestIndexerProvider implements IndexerProvider {
     /** Overrides {@link configureEventSource} for this provider's subscription. */
     protected readonly eventSource?: EventSourceFactory;
 
+    /** @see fetchVtxosJson */
+    private readonly inFlightVtxoReads = new Map<
+        string,
+        Promise<{ vtxos: Vtxo[]; page?: PageResponse }>
+    >();
+
     constructor(
         public serverUrl: string = DEFAULT_ARKADE_SERVER_URL,
         options: EventSourceCapable = {},
@@ -799,18 +805,47 @@ export class RestIndexerProvider implements IndexerProvider {
         if (params.toString()) {
             url += "?" + params.toString();
         }
-        const res = await indexerFetch(url);
-        if (!res.ok) {
-            throw new Error(`Failed to fetch vtxos: ${res.statusText}`);
-        }
-        const data = await res.json();
-        if (!Response.isVtxosResponse(data)) {
-            throw new Error("Invalid vtxos data received");
-        }
+        const data = await this.fetchVtxosJson(url);
+        // Mapped per caller, not shared: `convertVtxo` builds a fresh coin, so
+        // two callers served from one request still get independent results.
         return {
             vtxos: data.vtxos.map(convertVtxo),
             page: data.page,
         };
+    }
+
+    /**
+     * The wire read behind {@link fetchVtxosPage}, with an identical read
+     * already in flight served from that one request instead of repeated.
+     *
+     * Two callers can want the same page at the same instant without either
+     * being redundant, so there is no single call site to remove: a send that
+     * leaves change makes the indexer emit `vtxo_spent` and `vtxo_received` for
+     * the wallet's own contract milliseconds apart, and `handleContractEvent`
+     * delta-syncs that contract on both arms.
+     */
+    private async fetchVtxosJson(url: string): Promise<{ vtxos: Vtxo[]; page?: PageResponse }> {
+        const joined = this.inFlightVtxoReads.get(url);
+        if (joined) return joined;
+
+        const shared = (async () => {
+            const res = await indexerFetch(url);
+            if (!res.ok) {
+                throw new Error(`Failed to fetch vtxos: ${res.statusText}`);
+            }
+            const data = await res.json();
+            if (!Response.isVtxosResponse(data)) {
+                throw new Error("Invalid vtxos data received");
+            }
+            return data;
+        })();
+
+        this.inFlightVtxoReads.set(url, shared);
+        try {
+            return await shared;
+        } finally {
+            if (this.inFlightVtxoReads.get(url) === shared) this.inFlightVtxoReads.delete(url);
+        }
     }
 
     async getAssetDetails(assetId: string): Promise<AssetDetails> {
--- a/packages/ts-sdk/test/indexer-provider.test.ts
+++ b/packages/ts-sdk/test/indexer-provider.test.ts
@@ -90,6 +90,80 @@ describe("RestIndexerProvider", () => {
             expect(requestUrl.searchParams.get("before")).toBe("0");
         });
 
+        const wireVtxo = (txid: string) => ({
+            outpoint: { txid, vout: 0 },
+            createdAt: "1700000000",
+            expiresAt: "1700003600",
+            amount: "1000",
+            script: "51200a",
+            isPreconfirmed: false,
+            isSwept: false,
+            isUnrolled: false,
+            isSpent: false,
+            spentBy: null,
+            commitmentTxids: [],
+        });
+
+        it("serves two identical concurrent reads from one request", async () => {
+            let resolveFetch: (v: unknown) => void = () => {};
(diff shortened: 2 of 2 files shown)


Kukks:
Audited the coalescing key myself rather than waiting on a bot, since this is the one default-on change in this batch and a keying bug here would hand caller B an answer to caller A's question on a read path that feeds spend decisions.

## The parameter key is complete

The key is the fully-built URL from `fetchVtxosPage`, and the map is `private readonly` on the provider instance. Every dimension that can make two reads different is in one or the other:

| dimension | in the key |
|---|---|
| `scripts` | yes, appended in order |
| `outpoints` | yes, appended in order |
| `spendableOnly` / `spentOnly` / `recoverableOnly` / `pendingOnly` / `renewableOnly` | yes, each appended when set |
| `after` / `before` | yes |
| `page.index` / `page.size` | yes |
| `serverUrl`, and therefore network | yes — URL prefix, and the map is per-instance besides |
| request headers | n/a — `indexerFetch(url)` is called with no `init` on this path |

Script order is significant, so `["a","b"]` and `["b","a"]` do not coalesce. That is a missed opportunity, not a bug, and it errs in the safe direction.

## The dimension the key cannot see is time, and it lands on this PR's own motivating case

The docstring cites the right scenario: a send leaving change makes the indexer emit `vtxo_spent` and `vtxo_received` for the wallet's own contract milliseconds apart, and `handleContractEvent` delta-syncs on both arms.

I checked that those two arms really do build the same URL, and they do. `syncContracts({ contracts: [event.contract] })` is a targeted sync, so it never advances the cursor; the window from `computeSyncWindow(cursor)` is therefore identical across the pair, `fetchContractVtxosBulk` defaults `pageSize` the same way both times, and the contract is the same one. Identical URL, and the second arm lands well inside the first's flight window (~830 ms against mutinynet, ~7 ms locally).

So the pair coalesces — which is the point. The question this raises is whether it *should*, given the two reads were triggered by **different events**:

- arm 1 fires on `vtxo_received` and issues request A;
- arm 2 fires on `vtxo_spent` — an event whose whole purpose is to say state changed again — and is served A's response;
- if the indexer produced A's answer from a view that predates the spend, the spent state never reaches the repository from this event pair.

**Two things make this narrower than it first looks**, and I want them on the record rather than implied:

1. For the wallet's own send — the exact case above — `updateDbAfterOffchainTx` has already written the spent inputs and the change VTXO to the repository directly. The repo is correct regardless of what the sync returns. The residual exposure is a spend of the wallet's coins that the wallet did not perform: another instance, or a delegate.
2. Today the pre-send sync runs on every send, so any missed transition self-heals on the next spend. That is **not** true for a consumer that adopts the staleness budget in #882, which is one more reason that PR's default is `0`.

**What I could not determine from here** is arkd's read-snapshot semantics — specifically whether a query issued before an event is emitted can be served from a view that excludes the change the event announces. If it cannot, this is a non-issue and the coalescing is strictly a win.

## If a reviewer wants it closed rather than argued

The cheap fix is to add an epoch to the key: bump a counter whenever a contract event is handled, and include it, so a read triggered by a later event can never join one triggered by an earlier one. That keeps the load saving for genuinely simultaneous identical reads (two UI components asking at once) while never merging across an event boundary. Happy to add it — I have not, because it costs the exact saving this PR exists for in the one case it was written for, and that trade should be a reviewer's call rather than mine.

Kukks:
@coderabbitai review

PR #881, opened 2026-09-08
test(sdk): mock the fetch boundary with real Responses

### LINKED ISSUE #875: Provider tests mock baseFetch with objects that are not Responses

Kukks:
Closes #875.

## The shape

The provider tests mocked the fetch boundary with object literals:

```js
mockFetch.mockResolvedValueOnce({
    ok: true,
    json: () => Promise.resolve({ vtxos: [] }),
});
```

81 of these across 25 files — 79 in the top-level provider tests the issue names, plus one each in `test/helpers/restoreWallet.ts` and `test/wallet/unsignableBoardingInput.test.ts`. A double with two members answers only the two its own call site reads, so production code could not reach for a third — `clone()`, `arrayBuffer()`, `headers` — without breaking two dozen files at once. That is why in-flight coalescing of identical indexer reads (#876) had to go in at a seam chosen by the test doubles rather than by the design: coalescing needs to give each waiter its own body, and neither `clone()` nor `arrayBuffer()` existed on the mocks.

`indexerFetch` already calls `res.clone().text()` on the non-ok path, so the mocks were only viable because most tests never take that branch.

## The change

`test/helpers/response.ts` builds a real `Response`:

```ts
export const jsonResponse = (body: unknown, init?: ResponseInit): Response =>
    new Response(JSON.stringify(body, wireSafe), {
        status: 200,
        headers: { "content-type": "application/json" },
        ...init,
    });
```

plus `textResponse` for the plain-text endpoints. Every `ok: true` literal now goes through one of them. `Response` is global in every runtime the SDK targets and the production code already calls `res.json()`, so this needs no polyfill.

`test/jsonResponse.test.ts` pins what the helper is for: `instanceof Response`, `clone()`, `arrayBuffer()`, a second body read rejecting, and `init` overriding the defaults.

## Two things the fakes were hiding

Both are fixed here rather than papered over, and both are exactly the failure mode the issue describes — a mock more permissive than the real thing does not merely fail to catch bugs, it makes tests green on behaviour the real object refuses.

**A single `Response` handed to every call.** `mockResolvedValue(oneResponse)` is not what a fetch does: a real body can be read once. Seven tests in `esplora-watch-addresses` and one in `rate-gate` failed on the first migration with `Body is unusable: Body has already been read` — their poll loops had been re-reading one body indefinitely, and with a real `Response` the loop dies on cycle two. 28 sites now build a fresh response per call (`mockImplementation(async () => ...)`), and the three `deferred` fetches were restructured to gate on a `deferred<void>` and construct their own response, which preserves "this fetch is suspended" without sharing a body.

**`BigInt` in a JSON body.** `mockArkInfo` carried `batchExpiry: BigInt(144)`, which `JSON.stringify` refuses and no HTTP response can contain. The helper serializes bigint as a decimal string — which is what grpc-gateway puts on the wire for int64, and what `getInfo()` already parses back with `BigInt(fromServer.batchExpiry ?? 0)`. So the fixtures are now closer to the wire, not further from it.

## Deliberately out of scope

The 21 `ok: false` doubles. They are fault injection rather than ordinary responses, and at least one of them cannot be a real `Response` at all: `arkProviderDigest.test.ts` returns a `clone()` whose `text()` throws mid-body. Worth its own pass; mixing it in would have made this one unreviewable.

Also untouched: three `{ ok: true }` literals under `test/worker/` and `test/repositories/` that are message payloads, not fetch responses.

## Gate

Run manually — the pre-commit hook selects no projects in a worktree (#872), which is visible in this branch's own commit output.

`pnpm -r build`, then `pnpm -r test:unit`:

| | Test files | Tests |
|---|---|---|
| `origin/master` baseline | ts-sdk 179, swap 33, boltz-swap 23 | 2997 + 1 skipped / 810 / 617 |
| this branch | ts-sdk 180, swap 33, boltz-swap 23 | 3001 + 1 skipped / 810 / 617 |

Delta is `+1` file and `+4` tests, all of them `test/jsonResponse.test.ts`. Zero failures either side, no type errors. `pnpm --filter=@arkade-os/sdk lint` and `biome format config` both clean.

Net `-103` lines: 269 insertions against 372 deletions.

### DIFF
--- a/packages/ts-sdk/test/ark-provider.test.ts
+++ b/packages/ts-sdk/test/ark-provider.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 import { RestArkProvider } from "../src/providers/ark";
 import { MockEventSource } from "./mocks/eventSource";
+import { jsonResponse } from "./helpers/response";
 
 describe("RestArkProvider.getEventStream", () => {
     beforeEach(() => {
@@ -192,7 +193,7 @@ describe("RestArkProvider.getInfo transaction limits", () => {
     const infoResponse = (body: Record<string, unknown>) => {
         vi.stubGlobal(
             "fetch",
-            vi.fn(async () => ({ ok: true, json: async () => body })),
+            vi.fn(async () => jsonResponse(body)),
         );
         return new RestArkProvider("http://ark.test").getInfo();
     };
--- a/packages/ts-sdk/test/arkProviderDigest.test.ts
+++ b/packages/ts-sdk/test/arkProviderDigest.test.ts
@@ -1,10 +1,11 @@
 import { describe, it, expect, vi, afterEach } from "vitest";
 import { RestArkProvider, DigestMismatchError } from "../src/providers/ark";
+import { jsonResponse } from "./helpers/response";
 
 const SIGNER = "0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798";
 
 function okInfo(digest: string, signerPubkey: string = SIGNER) {
-    return { ok: true, json: async () => ({ signerPubkey, digest }) };
+    return jsonResponse({ signerPubkey, digest });
 }
 
 /** A rejected non-/info response whose body carries the given marker. */
--- a/packages/ts-sdk/test/cachingArkProvider.test.ts
+++ b/packages/ts-sdk/test/cachingArkProvider.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi, afterEach } from "vitest";
 import { CachingArkProvider } from "../src/providers/cachingArk";
 import { ArkInfo, ArkProvider, RestArkProvider } from "../src/providers/ark";
 import { extractArkProviderUrl } from "../src/wallet/wallet";
+import { jsonResponse } from "./helpers/response";
 
 const SIGNER = "0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798";
 const ROTATED = "02c6047f9441ed7d6d3045406e95c07cd85c778e4b8cef3ca7abac09b95c709ee5";
@@ -176,7 +177,7 @@ describe("CachingArkProvider", () => {
         let signerPubkey = SIGNER;
         vi.stubGlobal(
             "fetch",
-            vi.fn(async () => ({ ok: true, json: async () => ({ signerPubkey, digest }) })),
+            vi.fn(async () => jsonResponse({ signerPubkey, digest })),
         );
         const provider = new CachingArkProvider(new RestArkProvider("http://ark.test"), 60_000);
         await provider.getInfo();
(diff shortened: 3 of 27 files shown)


Kukks:
@coderabbitai review

PR #884, opened 2026-09-08
fix(providers): bound reads with a deadline so the retries above them can run

Kukks:
There is no `AbortSignal.timeout` anywhere in `packages/ts-sdk/src`, and `fetch` has no default timeout. Every provider read — indexer, Esplora, delegate, emulator, `getInfo` — is unbounded. This gives reads a deadline at the one chokepoint they all share.

## Why this is a correctness argument, not hygiene

**A timeout is not one resilience measure among several. It is the precondition for the others working at all.** Retry counts, circuit breakers, bounded queues and backpressure all assume the call underneath them terminates.

This repo has a live instance of that. `indexerFetch` already carries `INDEXER_MAX_ATTEMPTS = 3` with exponential backoff — and **that ladder is unreachable if the first attempt never settles.** The retry logic reads as though the case is handled; it is not.

The blast radius is not confined to the hung call. `_withTxLock` (`wallet/wallet.ts:3224`) is a bare promise-chain mutex with no timeout and no abort, and `send` (:5467), `settle` (:3898) and two further spend paths (:3802, :5831) all serialize through it. The pre-send contract sync is a **read** that runs inside `_sendImpl`, inside that lock. So one hung indexer GET stalls **every subsequent send and settle on that wallet**, not just its own caller.

## Prior art, same defect class, three independent arrivals

- **arkade-os/solver#30** (Go) adds `rpcTimeout`/`submitTimeout` to fulfillment RPCs, opened against a production symptom — a swap pending for hours with `read tcp ...: connection timed out` after the call had already blocked.
- **arkade-os/intent-solver#82** fixes the notification sink with `AbortSignal.timeout`. Its failure mode is the sharpest illustration of the argument above: a hung request meant the drain never resolved, so **the bounded queue and the bounded retries it had already built were never reached.** That work is not touched here.
- This PR: the TypeScript SDK's provider reads.

## What changed

One chokepoint. `baseFetch` is what the indexer, Esplora, the delegate provider and the Ark-server `fetch` wrapper (and therefore `getInfo`) all funnel through, so a default deadline there covers those without touching each provider.

**Reads only, and deliberately so:**

- **GET and HEAD get a deadline.** Bounded, retry-eligible, no state at risk. The scope matches `indexerFetch`'s existing retry eligibility exactly — the methods already considered safe to send twice are the ones bounded here, which is also why the subscribe/unsubscribe POSTs keep their current behaviour and cannot be orphaned by an abort.
- **Every write is left unbounded.** Aborting `submitTx` or `finalizeTx` does **not** mean the submission did not land — it means the outcome is now unknown. A caller with no reconcile path would have to choose between re-sending (double-submit) and marking a spend failed that actually succeeded. A stall is the safer failure until that reconcile path exists, and pretending otherwise would turn a wasted wait into a potential loss of funds. **This is a stated gap, not an oversight.**
- **A caller's own signal always wins.**
- **A `Request` input is left alone — and this rule must not be "tidied up" later.** The tempting version is `input instanceof Request && input.signal`, i.e. "only skip if the caller brought a signal". That condition is **always true**: the WHATWG spec gives every `Request` a `signal`, populated whether or not its author passed one. So the property "did this caller bring a lifetime?" is **unfalsifiable by inspection** — there is no observation that distinguishes a caller-supplied signal from the default one. The only safe reading is to treat any `Request` as owning its own lifetime and leave it unbounded. A future reader who re-adds the `.signal` check will not be making it stricter; they will be writing a condition that can never be false, and the tests will not tell them, which is exactly how this got in.

**Long-lived streams are unaffected, but not for the reason I first wrote**, and the correction is in the diff.

On web, `getEventStream` (`providers/ark.ts:746`), `getTransactionsStream` (:835) and `getSubscription` (`providers/indexer.ts:550`) run on `EventSource` and never reach `baseFetch`. **On Expo they can.** `getExpoFetch` (`providers/expoUtils.ts:7-35`) falls back to this module's `fetch` whenever `expo/fetch` will not import, and on that path an SSE stream does come through `baseFetch`.

What actually protects it is that `sseStreamIterator` passes `signal: fetchController.signal` (`expoUtils.ts:64`) and `readDeadline` bails on a caller-supplied signal. So the invariant is **"a long-lived request opts out by bringing its own signal"**, not "streams are EventSource". My original docstring said the latter, which would have told the next reader that dropping that signal is safe — it would truncate the subscription at 30 s. The docstring now states the real rule, the signal is marked as required where it is passed, and a test fails if it is removed.

Same class as the `Request.prototype.signal` catch below: a claim that reads as true and is not falsifiable by looking at the line it describes.

### Not covered: the emulator

I claimed in the first revision that the emulator's read was covered. **It is not.** `providers/emulator.ts` imports only `Intent` — its five `fetch(...)` calls (`:60, :81, :118, :154, :198`) resolve to the **global** `fetch`, never `baseFetch`. So `RestEmulatorProvider.getInfo()` is still unbounded after this PR.

Left out on purpose rather than fixed in-flight: the right move is `baseFetch` (not the Ark-server `fetch` wrapper, whose `X-Build-Version` / `X-SDK-VERSION` headers would fail CORS preflight against a non-arkd origin), and it changes the emulator's error type from a raw `TypeError` to `FetchError`. That is a behaviour change worth its own review rather than an amendment to a PR already being looked at. Happy to fold it in here if a reviewer would rather.

## Effect on failure handling

A fired deadline rejects with `TimeoutError`, which `baseFetch` wraps as `FetchError`, which `toProviderUnavailable` maps to `ProviderUnavailableError`, which `isRetryableProviderError` already treats as retryable. So a timed-out indexer read now **enters** the existing ladder instead of hanging. Worst case for an indexer read becomes roughly three attempts plus backoff (~91 s) rather than unbounded — still long, but bounded, and the lock is released.

`READ_TIMEOUT_MS` is exported so the value is inspectable rather than implicit.

## On `_withTxLock` itself

I considered giving the lock its own bound and **recommend against it.** A lock timeout would let a second spend begin while the first is still in flight — the same double-submit hazard in different clothing, and worse than the stall it replaces because the first spend may yet succeed. Bounding the calls *inside* the lock is the safe half of the problem; bounding the lock is not, and should not be bundled here.

## Test plan

Baseline captured on clean `master` first, then diffed:

| | files | tests | failed |
|---|---|---|---|
| baseline (`origin/master` @ `5ea42a35`) | 180 | 3004 passed, 1 skipped | 0 |
| with change | 181 | 3013 passed, 1 skipped | 0 |

+1 file, +9 tests — exactly the 9 added. Rebased onto current master after #876 and #881 landed, and re-baselined against it rather than reusing the pre-merge numbers. `tsc --noEmit` clean, `biome format` clean.

Mutation-tested:

- remove the GET/HEAD gate → the "leaves a POST unbounded" test fails
- ignore the caller's signal → the "never overrides a caller's own signal" test fails
- drop `sseStreamIterator`'s signal → the stream-opt-out guard fails

Mutation also caught a bug in my own first draft, and it is the kind review does not catch. I had written `input instanceof Request && input.signal` to mean "the caller brought their own signal". Because `Request.prototype.signal` is always populated, that condition is always true, so **no `Request` ever received a deadline** — and my test named "reads the method off a Request" passed for entirely the wrong reason. Removing the GET/HEAD gate failed the plain-POST test but *not* the Request-POST test, which is what exposed it. The failure was a DOM invariant, not a logic slip, so it read as correct on the page. The rule is now stated with its reason, and the test asserts the real behaviour.

Note for reviewers: this repo's pre-commit hook printed `No projects matched the filters` and ran nothing (#872). The gate above was run manually.

## Summary by CodeRabbit

- **New Features**
  - Added a 30-second read deadline for SDK GET and HEAD requests when no cancellation signal is provided.
  - Exposed the read-timeout value for applications that need to reference or configure related behavior.
  - Preserved caller-provided cancellation signals and uninterrupted streaming behavior.

- **Reliability**
  - Requests that exceed the read deadline now surface a retryable fetch error, helping applications recover from stalled reads.

### DIFF
--- a/packages/ts-sdk/src/index.ts
+++ b/packages/ts-sdk/src/index.ts
@@ -538,7 +538,7 @@ import type {
 import type { ScanResult, ScanContractsOptions, HandlerError } from "./contracts/contractManager";
 import { timelockToSequence, sequenceToTimelock } from "./utils/timelock";
 import { toXOnly } from "./utils/keys";
-import { buildVersion, sdkVersion, FetchError } from "./utils/fetch";
+import { buildVersion, sdkVersion, FetchError, READ_TIMEOUT_MS } from "./utils/fetch";
 import {
     closeDatabase,
     openDatabase,
@@ -638,6 +638,7 @@ export {
     CachingArkProvider,
     DigestMismatchError,
     FetchError,
+    READ_TIMEOUT_MS,
     RestIndexerProvider,
     RestEmulatorProvider,
     DEFAULT_VTXO_PAGE_SIZE,
--- a/packages/ts-sdk/src/providers/expoUtils.ts
+++ b/packages/ts-sdk/src/providers/expoUtils.ts
@@ -61,6 +61,9 @@ export async function* sseStreamIterator<T>(
                 Accept: "text/event-stream",
                 ...headers,
             },
+            // Required, not just for cancellation: `getExpoFetch` falls back to
+            // `baseFetch`, which bounds any read that brings no signal of its
+            // own. Drop this and the stream is truncated at READ_TIMEOUT_MS.
             signal: fetchController.signal,
         });
 
--- a/packages/ts-sdk/src/utils/fetch.ts
+++ b/packages/ts-sdk/src/utils/fetch.ts
@@ -65,8 +65,14 @@ function readDeadline(input: RequestInfo | URL, init?: RequestInit): AbortSignal
  * preflight.
  *
  * Reads without a caller-supplied signal are bounded by {@link READ_TIMEOUT_MS}.
- * Long-lived streams do not come through here — they are `EventSource` — so the
- * deadline cannot truncate a subscription.
+ *
+ * **A long-lived request opts out by supplying its own signal — that is the only
+ * thing keeping this deadline off a stream.** On web, SSE happens to use
+ * `EventSource` and never arrives here at all; on Expo it does arrive here
+ * whenever `expo/fetch` fails to import and `getExpoFetch` falls back to this
+ * module, and `sseStreamIterator` passing `signal: fetchController.signal` is
+ * what prevents a 30-second truncation. That signal is load-bearing, not
+ * incidental: a streaming caller that omits one gets cut off here.
  *
  * Transport-level rejections are re-thrown as a {@link FetchError}.
  */
--- a/packages/ts-sdk/src/utils/fetch.ts
+++ b/packages/ts-sdk/src/utils/fetch.ts
@@ -27,10 +27,46 @@ export class FetchError extends Error {
 }
 
 /**
+ * Deadline applied to a read that carries no `AbortSignal` of its own.
+ *
+ * `fetch` has no default timeout, so a connection the network dropped silently
+ * stays pending until the runtime gives up — and every bound built on top of it
+ * waits with it. A retry ladder, a bounded queue or a mutex only work if the
+ * call underneath them terminates.
+ */
+export const READ_TIMEOUT_MS = 30_000;
+
+/**
+ * A deadline for this request, or `undefined` to leave it unbounded.
+ *
+ * Only GET and HEAD get one. A write that is aborted has not necessarily failed
+ * — its outcome is unknown — so bounding one converts a stall into a state
+ * question its caller may have no way to answer.
+ *
+ * A `Request` input is always left alone. Every `Request` carries a `signal`
+ * whether or not its author supplied one, so "did the caller bring a signal?"
+ * cannot be read back off it, and guessing wrong would cancel someone's
+ * lifetime for them.
+ */
+function readDeadline(input: RequestInfo | URL, init?: RequestInit): AbortSignal | undefined {
+    if (init?.signal || input instanceof Request) return undefined;
+    const { method } = describeRequest(input, init);
+    const verb = method.toUpperCase();
+    if (verb !== "GET" && verb !== "HEAD") return undefined;
+    return typeof AbortSignal !== "undefined" && typeof AbortSignal.timeout === "function"
+        ? AbortSignal.timeout(READ_TIMEOUT_MS)
+        : undefined;
+}
+
+/**
  * Guarded passthrough to the platform `fetch` with no Arkade-specific headers.
  * Use for any service that is NOT the Ark server (delegate, Esplora, …): those
  * origins reject unknown request headers such as `X-Build-Version` in the CORS
  * preflight.
+ *
+ * Reads without a caller-supplied signal are bounded by {@link READ_TIMEOUT_MS}.
+ * Long-lived streams do not come through here — they are `EventSource` — so the
+ * deadline cannot truncate a subscription.
  *
  * Transport-level rejections are re-thrown as a {@link FetchError}.
  */
@@ -38,7 +74,8 @@ export function baseFetch(input: RequestInfo | URL, init?: RequestInit): Promise
     if (typeof globalThis.fetch !== "function") {
         throw new Error("Fetch API is not available in this environment.");
     }
-    return globalThis.fetch(input, init).catch((cause) => {
+    const signal = readDeadline(input, init);
+    return globalThis.fetch(input, signal ? { ...init, signal } : init).catch((cause) => {
         const { url, method } = describeRequest(input, init);
         throw new FetchError(`Network request failed: ${method} ${url}`, { url, method, cause });
     });
--- a/packages/ts-sdk/src/utils/fetch.ts
+++ b/packages/ts-sdk/src/utils/fetch.ts
@@ -36,8 +36,10 @@ export class FetchError extends Error {
  */
 export const READ_TIMEOUT_MS = 30_000;
 
+let warnedNoTimeoutSupport = false;
+
 /**
- * A deadline for this request, or `undefined` to leave it unbounded.
+ * The signal bounding this request, or `undefined` to leave it unbounded.
  *
  * Only GET and HEAD get one. A write that is aborted has not necessarily failed
  * — its outcome is unknown — so bounding one converts a stall into a state
@@ -53,9 +55,31 @@ function readDeadline(input: RequestInfo | URL, init?: RequestInit): AbortSignal
     const { method } = describeRequest(input, init);
     const verb = method.toUpperCase();
     if (verb !== "GET" && verb !== "HEAD") return undefined;
-    return typeof AbortSignal !== "undefined" && typeof AbortSignal.timeout === "function"
-        ? AbortSignal.timeout(READ_TIMEOUT_MS)
-        : undefined;
+    if (typeof AbortSignal !== "undefined" && typeof AbortSignal.timeout === "function") {
+        return AbortSignal.timeout(READ_TIMEOUT_MS);
+    }
+
+    // `AbortSignal.timeout` is absent on some React Native runtimes while
+    // `AbortController` is not, so fall back rather than silently leaving the
+    // read unbounded — which is the exact state this exists to prevent.
+    if (typeof AbortController === "function") {
+        const controller = new AbortController();
+        // Deliberately never cleared, so the bound covers the body read and not
+        // just the headers, matching the native path. Aborting an already
+        // settled request is a no-op; `unref` keeps it from holding Node open.
+        const timer: unknown = setTimeout(() => controller.abort(), READ_TIMEOUT_MS);
+        (timer as { unref?: () => void })?.unref?.();
+        return controller.signal;
+    }
+
+    if (!warnedNoTimeoutSupport) {
+        warnedNoTimeoutSupport = true;
+        console.warn(
+            "Neither AbortSignal.timeout nor AbortController is available in this runtime: " +
+                "provider reads are UNBOUNDED and READ_TIMEOUT_MS is not being applied.",
+        );
+    }
+    return undefined;
 }
 
 /**
--- a/packages/ts-sdk/test/fetch-read-deadline.test.ts
+++ b/packages/ts-sdk/test/fetch-read-deadline.test.ts
@@ -69,8 +69,62 @@ describe("baseFetch read deadline", () => {
         expect(isRetryableProviderError(err)).toBe(true);
     });
 
-    it("states a deadline rather than leaving it implicit", () => {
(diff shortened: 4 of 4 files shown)


Kukks:
@coderabbitai review

Kukks:
Both reviews landed together; addressing them in one pass. Pushed as `1ae2e940`.

## @coderabbitai — both taken, and the first made the second testable

**Fallback when `AbortSignal.timeout` is absent — fixed.** You were right that returning `undefined` leaves the read in exactly the state this PR exists to prevent, with `READ_TIMEOUT_MS` still readable and no longer true. `AbortController` is present on the runtimes where `AbortSignal.timeout` is not, so the fallback is a controller plus a timer. Only when *neither* exists does it go unbounded, and then it warns once.

One deliberate detail: the fallback timer is **not** cleared when the request settles. Clearing it would bound only the time to headers, whereas `AbortSignal.timeout` keeps running and bounds the body read too — so not clearing is what keeps the fallback semantically equal to the native path. Aborting an already-settled request is a no-op, and `unref()` stops it holding a Node process open.

**Test that only proved attachment, not firing — fixed, and your first finding is what made it possible.** With the native `AbortSignal.timeout` there is no way to do this: it runs on an internal timer that `vi.useFakeTimers()` does not control, so advancing fake timers would never have fired it. Now that the fallback path uses a real `setTimeout`, the test forces that path, has `fetch` stay pending until `init.signal` aborts, advances by `READ_TIMEOUT_MS`, and asserts the resulting `FetchError` is retryable. Mutation-checked: removing the fallback timer fails that test alone.

**`READ_TIMEOUT_MS` assertion widened to a range** (`>= 5_000`, `<= 120_000`), so a dropped zero or an extra one fails rather than passing on any positive number. Agreed the old assertion caught almost nothing.

**Wording nit on `readDeadline` returning a signal rather than a deadline** — taken, renamed in the docstring.

## @arkana-ai-bot

**Independent confirmation of the Expo invariant is the part I most wanted checked**, since my first version of that docstring was wrong in a way that read as true. Thank you for tracing `getExpoFetch → fetch → baseFetch` and `sseStreamIterator`'s signal rather than taking the comment's word.

**`getExpoFetch` fallback sending Ark headers to the indexer — agreed, real, and out of scope here.** It is the same defect class as the emulator gap: the wrong wrapper for what may be a distinct origin. In practice `RestArkProvider` and `RestIndexerProvider` are constructed with the same `serverUrl` in every usage I can find, so it is latent rather than live, but an embedder splitting the two hosts would hit a preflight failure. Filed separately rather than folded in.

**Worst-case bound** — confirming your reading: ~91 s for an indexer read (3 attempts x 30 s plus backoff) against unbounded today, with the lock released either way.

## Note on the commit

The pre-commit hook now refuses to run in a git worktree instead of passing silently — that is #880 fixing #872, landed an hour ago, and it caught this commit. I ran the gate it asks for by hand and committed with `--no-verify`, as its own message instructs:

| | files | tests | failed |
|---|---|---|---|
| `origin/master` baseline | 180 | 3004 passed, 1 skipped | 0 |
| this branch | 181 | 3015 passed, 1 skipped | 0 |

`tsc --noEmit` clean, `biome format` clean.

Kukks:
@arkana-ai-bot — the warn-once note was right, and it is fixed in `8986ca4d`.

Your reading of why it did not bite was correct: the only path that sets `warnedNoTimeoutSupport` is that same test, so the "exactly once" assertion held. But it held *by accident of ordering*, not by construction — which means it was asserting something it could not actually observe. Any later test that stripped both primitives would have silently turned it into an assertion about a flag some earlier test had already set.

The fix takes the module state out of the equation rather than resetting the flag: `vi.resetModules()` and a fresh `import("../src/utils/fetch")` inside the test, so both `baseFetch` calls run against a module instance whose flag starts false. The once-ness is now measured rather than inherited. Test renamed to say so ("— and only once").

Mutation check: setting the flag's initial value to `true` fails this test and nothing else, which is the behaviour I wanted from it.

No production code changed — `8986ca4d` touches only `fetch-read-deadline.test.ts`.

On CI: `Integration (ts-sdk / ark-core)` came back red on this commit with `AssertionError: expected 0 to be greater than 0` in `test/e2e/ark.test.ts > Cross-contract spending`. That is a pre-existing flake on master, not this change — master run [34267318645](https://github.com/arkade-os/ts-sdk/actions/runs/34267318645) (`56319987`, unrelated to fetch) failed on the same file with the *same* assertion, and run [34197565949](https://github.com/arkade-os/ts-sdk/actions/runs/34197565949) failed on the neighbouring `Delegate Lifecycle` test the same way. Re-running the job.

Kukks:
@arkana-ai-bot — the incremental finding is right and is fixed in `aa7e93bb`. The "prior findings still open" section is not: it is reading a stale diff range, and I want to put the evidence here so nobody re-opens work that is already in the branch.

## The accepted one

**`expoUtils.ts` comment names the wrong wrapper — correct, fixed.** `getExpoFetch`'s catch branch does `return fetch`, and line 1 imports `fetch` from `../utils/fetch` — the Ark wrapper, not `baseFetch`. The deadline reasoning survives (the wrapper calls `baseFetch`), but eliding it hides exactly the layer that attaches `X-Build-Version` / `X-SDK-VERSION` to what may be a non-arkd origin, which is your finding 3. The comment now reads "falls back to the Ark `fetch` wrapper, which routes through `baseFetch` and so bounds any read that brings no signal of its own."

That is the same class of defect this PR exists to correct, so thank you for catching it in a comment I wrote *about* not doing that.

## The branch was not rewritten

> The branch has been rewritten since the commits I last verified (1ae2e940 / 8986ca4d). … None of those changes were carried forward into the current branch.

There has been no force-push. `1ae2e940` and `8986ca4d` are not missing from the branch — they **are** the branch's two most recent commits:

```
8986ca4d test(providers): isolate the warn-once assertion from module state
1ae2e940 fix(providers): bound the read on runtimes without AbortSignal.timeout
d98fc22d fix(providers): state the real reason a stream escapes the deadline
58258816 fix(providers): bound reads with a deadline so the retries above them can run
f6129a3b (master) feat(payment): state a quote's validity on RouteQuote (#874)
```

Linear, no rewrite. Your review is recorded against `commit_id=8986ca4d`, but the body diffs `b1e1c1d → d98fc22` — a range that ends **two commits before** head. That explains all four: you diffed a range that predates the fixes, then read their absence from that range as their removal from the branch.

Checked against the actual head, at the time of your review:

- **Finding 1 — fallback: present.** `fetch.ts:65-79` — `if (typeof AbortController === "function")`, controller plus timer, `unref()`, and the one-time `console.warn` below it.
- **Finding 2 — assertion: already widened.** `fetch-read-deadline.test.ts:75-76` is `toBeGreaterThanOrEqual(5_000)` and `toBeLessThanOrEqual(120_000)`. `toBeGreaterThan(0)` does not appear in the file.
- **Finding 4 — docstring: already renamed.** `fetch.ts:42` reads "The signal bounding this request, or `undefined` to leave it unbounded." The text you quote is not in the branch.
- **Finding 5 — warn-once: exists.** `warnedNoTimeoutSupport` at `:39`, set at `:75-76`. The isolation fix you prompted is `8986ca4d`, and it is mutation-checked: initialising the flag to `true` fails that test alone — 1 failed / 3014 passed across 181 files, against a clean 3015.

## `indexer.ts` is not in this PR

The `fetchVtxosJson` coalescing review is thorough, but it is reviewing #876, which is already on master. This PR's diff is four files:

```
packages/ts-sdk/src/index.ts                     |   3 +-
packages/ts-sdk/src/providers/expoUtils.ts       |   3 +
packages/ts-sdk/src/utils/fetch.ts               |  69 +++++++++-
packages/ts-sdk/test/fetch-read-deadline.test.ts | 155 +++++++++++++++++++++++
```

Your second-order point still lands, though, and is worth recording where it belongs: `inFlightVtxoReads` holds live promises, so a hung read wedges that URL for the provider's lifetime. That is precisely the argument in this PR's description — deduplication and retry ladders are amplifiers without a deadline underneath them — and it is why #876 wanting #884 is the right ordering rather than a coincidence.

**Finding 3 (`getExpoFetch` CORS) remains open and separately filed**, unchanged.

Gate on `aa7e93bb`, run by hand (the pre-commit hook refuses in a worktree — #872, fixed by #880): `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit`. ts-sdk 181 files / 3015 passed, swap 33 / 811, boltz-swap 23 / 617, zero failures, matching baseline.

Kukks:
Correction to my own comment above: I wrote it without noticing this PR had **already merged** as `d27dc82e` (20:32 UTC), about 35 minutes before that review landed. Two things follow.

**It probably explains the "branch has been rewritten" reading.** The merge was a squash, so on `master` the individual commits `1ae2e940` and `8986ca4d` genuinely no longer exist *as commits*. A diff that looks for them there finds nothing and can reasonably report them as not carried forward. That is a kinder explanation than the one I gave, and I'd rather record it than leave the review looking simply wrong.

**The content did carry forward, though — verified on `master`, not on the branch:**

- fallback: `packages/ts-sdk/src/utils/fetch.ts:65` — `if (typeof AbortController === "function")`, controller plus timer, `unref()`
- warn-once: `fetch.ts:39`, set at `:75-76`
- docstring: `fetch.ts:42` — "The signal bounding this request, or `undefined` to leave it unbounded."
- range assertion: `test/fetch-read-deadline.test.ts:75-76` — `toBeGreaterThanOrEqual(5_000)` / `toBeLessThanOrEqual(120_000)`
- the warn-once isolation: `test/fetch-read-deadline.test.ts:121` — `vi.resetModules()`

So findings 1, 2, 4 and 5 are closed on `master`, not merely on a branch.

**One thing did not make it, and it is the finding that was right.** The `expoUtils.ts` comment correction is `aa7e93bb`, pushed to `fix/provider-read-deadlines` *after* the squash. `master` still reads:

```ts
// Required, not just for cancellation: `getExpoFetch` falls back to
// `baseFetch`, which bounds any read that brings no signal of its own.
```

`getExpoFetch` returns `fetch` — the Ark wrapper — which reaches `baseFetch` one layer down. Comment-only, no behaviour change, but it is exactly the layer that attaches `X-Build-Version` / `X-SDK-VERSION` to a possibly non-arkd origin, which is finding 3. `aa7e93bb` is sitting on the branch ready to be a one-line follow-up; I have not opened one.

PR #888, opened 2026-09-08
fix(boltz-swap): bound Boltz API reads with a deadline

Kukks:
`BoltzSwapProvider.doRequest` calls `globalThis.fetch` directly, and this package never imports the SDK's fetch utilities, so **every request to the Boltz API is unbounded**. `fetch` has no default timeout, so a connection the network dropped silently stays pending until the runtime gives up.

This is the sibling of #884, which does the same for the `ts-sdk` providers. It does nothing for this package — different module, different entry point — so the gap needs closing here.

## Why this is a correctness argument, not hygiene

**A timeout is not one resilience measure among several. It is the precondition for the others working at all.** Retry counts, circuit breakers, bounded queues and backpressure all assume the call underneath them terminates.

There is a live instance of that a few lines above the change. `request()` deduplicates concurrent GETs through `inflightGets`, joining later callers to the in-flight promise. So **a single hung GET does not stall one caller — it parks every later caller of that path behind a request that will never settle.** The deduplication is a good optimisation that becomes an amplifier without a deadline underneath it.

The same shape has now been hit three times independently across the org: `arkade-os/solver#30` (Go RPC deadlines, opened against a swap that sat pending for hours), `arkade-os/intent-solver#82` (a hung notification request meant `flush()` never resolved, so the bounded queue and bounded retries that PR had already built **were never reached**), and #884 here.

## What changed

**GET gets a deadline.** Bounded, safe to repeat, no state at risk.

**POST deliberately does not.** An aborted POST to Boltz has not necessarily failed — it may already have created a swap. Bounding it would convert a stall into a state question the caller cannot answer, and a retry could create a *second* swap. A stall is the safer failure until a reconcile path exists.

I looked for a POST here whose outcome is reconcilable from Boltz's own API and did not find one, so all of them are left alone. If someone knows of one, it is worth arguing separately rather than quietly including.

**Runtime support.** `AbortSignal.timeout` where available; `AbortController` plus a timer where it is not — this package ships `./expo` and `./expo/background`, and some React Native runtimes have the latter but not the former. If neither exists the read stays unbounded and says so once, rather than leaving `BOLTZ_READ_TIMEOUT_MS` readable and untrue.

The timer on the fallback path is deliberately not cleared, so the bound covers the body read and not merely the headers, matching what `AbortSignal.timeout` does. Aborting an already-settled request is a no-op, and `unref()` keeps it from holding a Node process open.

## Test plan

Baseline captured on `origin/master` first, then diffed:

| | files | tests | failed |
|---|---|---|---|
| baseline (`94c9a1fb`) | 23 | 617 passed | 0 |
| with change | 24 | 623 passed | 0 |

+1 file, +6 tests — the 5 added here plus the warn-path test added in `d1d26134` after review. `tsc --noEmit` clean, `biome format` clean.

Mutation-tested: applying the deadline to writes as well fails the "leaves a POST unbounded" test, alone; removing the fallback timer fails the fallback test, alone.

**One existing-test change, and it is a real behaviour change rather than a test fix.** Six assertions in `boltz-swap-provider.test.ts` pin the exact `fetch` init for GET calls, so adding `signal` broke them. They now also assert `signal: expect.any(AbortSignal)`; method and headers are still matched exactly, so the assertions are no weaker than before.

Worth noting how that surfaced: the new tests passed when run alone and the **existing** suite failed only in the full run. Running the whole package gate rather than just the new file is what caught it.

Note for reviewers: the pre-commit hook now correctly refuses to run in a git worktree (#872, fixed by #880) and instructs running the gate by hand. I did, with the numbers above, and committed with `--no-verify` as it directs.

### DIFF
--- a/packages/boltz-swap/src/boltz-swap-provider.ts
+++ b/packages/boltz-swap/src/boltz-swap-provider.ts
@@ -49,6 +49,48 @@ export type BoltzSwapStatus =
     | "transaction.server.mempool"
     | "transaction.server.confirmed";
 
+/**
+ * Deadline for a Boltz API read.
+ *
+ * `fetch` has no default timeout, so a connection the network dropped silently
+ * stays pending until the runtime gives up. That is not one missing safeguard
+ * among several — it is the one the others rest on: the `inflightGets`
+ * deduplication above and every retry around this client assume the call
+ * underneath them terminates.
+ */
+export const BOLTZ_READ_TIMEOUT_MS = 30_000;
+
+let warnedNoTimeoutSupport = false;
+
+/**
+ * A signal bounding a Boltz **read**, or `undefined` when the runtime offers no
+ * way to build one.
+ *
+ * Reads only. An aborted POST to Boltz has not necessarily failed — it may have
+ * created a swap — so bounding one would turn a stall into a state question the
+ * caller cannot answer, and a retry could create a second swap.
+ */
+const readDeadline = (): AbortSignal | undefined => {
+    if (typeof AbortSignal !== "undefined" && typeof AbortSignal.timeout === "function") {
+        return AbortSignal.timeout(BOLTZ_READ_TIMEOUT_MS);
+    }
+    // Absent on some React Native runtimes, where `AbortController` is not.
+    if (typeof AbortController === "function") {
+        const controller = new AbortController();
+        const timer: unknown = setTimeout(() => controller.abort(), BOLTZ_READ_TIMEOUT_MS);
+        (timer as { unref?: () => void })?.unref?.();
+        return controller.signal;
+    }
+    if (!warnedNoTimeoutSupport) {
+        warnedNoTimeoutSupport = true;
+        console.warn(
+            "Neither AbortSignal.timeout nor AbortController is available in this runtime: " +
+                "Boltz API reads are UNBOUNDED.",
+        );
+    }
+    return undefined;
+};
+
 /** Returns true if the status indicates a failed submarine swap. */
 export const isSubmarineFailedStatus = (status: BoltzSwapStatus): boolean => {
     return ["invoice.failedToPay", "transaction.lockupFailed", "swap.expired"].includes(status);
@@ -1741,6 +1783,8 @@ export class BoltzSwapProvider {
                 method,
                 headers: { "Content-Type": "application/json" },
                 body: body ? JSON.stringify(body) : undefined,
+                // GET only — see `readDeadline`.
+                ...(method === "GET" ? { signal: readDeadline() } : {}),
             });
 
             if (!response.ok) {
--- a/packages/boltz-swap/test/boltz-read-deadline.test.ts
+++ b/packages/boltz-swap/test/boltz-read-deadline.test.ts
@@ -69,9 +69,8 @@ describe("BoltzSwapProvider read deadline", () => {
         ) as any;
 
         try {
-            const pending = (provider() as any)
-                .request("/v2/slow", "GET")
-                .catch((e: unknown) => e);
+            const p = provider() as any;
+            const pending = p.request("/v2/slow", "GET").catch((e: unknown) => e);
             await vi.advanceTimersByTimeAsync(BOLTZ_READ_TIMEOUT_MS);
 
             expect(await pending).toBeInstanceOf(Error);
--- a/packages/boltz-swap/test/boltz-read-deadline.test.ts
+++ b/packages/boltz-swap/test/boltz-read-deadline.test.ts
@@ -1,0 +1,83 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { BoltzSwapProvider, BOLTZ_READ_TIMEOUT_MS } from "../src/boltz-swap-provider";
+
+describe("BoltzSwapProvider read deadline", () => {
+    let seen: RequestInit[] = [];
+    const original = globalThis.fetch;
+
+    beforeEach(() => {
+        seen = [];
+        globalThis.fetch = vi.fn(async (_input: any, init?: RequestInit) => {
+            seen.push(init ?? {});
+            return new Response(JSON.stringify({}), {
+                status: 200,
+                headers: { "content-type": "application/json", "content-length": "2" },
+            });
+        }) as any;
+    });
+
+    afterEach(() => {
+        globalThis.fetch = original;
+        vi.restoreAllMocks();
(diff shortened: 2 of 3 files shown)


Kukks:
## Two findings on where this fix belongs

CI is green here (7/7). This is not a request to change anything in the PR — it is the collision with #886 (`feat/v0.5` → `master`), and an answer to the question that collision raises. Sequencing #886 is not mine to call, so I am recording evidence rather than proposing a disposition.

### 1. `packages/boltz-swap` does not exist on `feat/v0.5`

Confirmed three ways:

- `git ls-tree --name-only origin/feat/v0.5 packages/` returns `packages/swap` and `packages/ts-sdk` only.
- A recursive listing filtered for `boltz` returns zero files.
- `git grep -l "class BoltzSwapProvider" origin/feat/v0.5` returns nothing.

Deleted wholesale, not moved. So this PR bounds reads in a file that #886 removes.

### 2. The same defect is in `packages/swap`, which survives on v0.5 — nobody had looked

I checked. It is there, on **both** branches, in two of the package's three HTTP clients.

**`rfq.ts` — `httpTransport`** (`master:607`, `v0.5:649`)

```ts
const fetchImpl = options.fetchImpl ?? fetch;   // the global
```

`status()` (`GET /v1/rfq/<rfq_id>` — `master:641`, `v0.5:683`) passes no `signal`. Unbounded read, structurally identical to `doRequest` before this PR. `requestQuote()` is a `POST`, so by this PR's own rule it stays unbounded deliberately.

**`markets.ts` — `makeCachedFeedFetch` / `discoverMarkets`** (`master:50` and `:150`; `v0.5:56` and `:161`)

Unbounded, and it carries the same structural argument this PR's docstring makes about `inflightGets`. `makeCachedFeedFetch` keeps an `inflight` map so that a burst of debounced keystrokes collapses to one upstream call. A read that never settles never resolves that entry — and every later caller for the same URL `await`s the same never-settling promise, with no `catch` or `finally` reachable to evict it. The deduplication rests on the call underneath it terminating.

The gap is already acknowledged in-tree at `markets.ts:146-150`:

> `ponytail:` no request deadline here — a caller that needs one wraps its own `fetchImpl` with an `AbortSignal`; add one if a hung registry ever strands discovery in practice.

**v0.5 adds more of the same shape**, in the new `client/` subtree: `client/corridors/chainSource.ts:156`, where `esploraTip` does `await (fetchImpl ?? fetch)(\`${esploraUrl}/blocks\`)` — a GET with no signal; and `client/quoteOffer.ts:70`, `feedFetch(base: typeof fetch = fetch)`. Both `client/client.ts:170` and `client/drive.ts:1141` state the position in prose: "nothing in this package takes an `AbortSignal`".

**The precedent is already inside the same package.** `reveal.ts:111` (`covclaimdClient`) does exactly what this PR does:

```ts
signal: AbortSignal.timeout(timeoutMs),   // timeoutMs ?? 30_000
```

Same 30 s as `READ_TIMEOUT_MS` (#884) and `BOLTZ_READ_TIMEOUT_MS` here. So `packages/swap` is not uniformly unbounded — one of its three HTTP clients is bounded and two are not, which reads as an oversight rather than a decision.

### One constraint on whoever fixes `packages/swap`

It cannot route through `baseFetch` the way #887 does. `baseFetch` is not exported: `packages/ts-sdk/src/index.ts` imports only `FetchError` and `READ_TIMEOUT_MS` from `./utils/fetch`, and the `exports` map in `package.json` has no `utils` subpath. A fix there needs either its own `readDeadline` (the pattern in this PR) or a new SDK export.

Kukks:
@arkana-ai-bot — thank you, and one of the three observations turned out to be worth more than "record-only". Acted on it.

## `warnedNoTimeoutSupport` and the untested warn path — fixed in `d1d26134`

You are right that the flag is never mutated during a test run today, and right about why: the fallback test leaves `AbortController` in place, so it takes the controller branch and the warning is never reached.

The consequence is sharper than the cross-test-contamination risk you flagged, though. It means **the one branch that admits `BOLTZ_READ_TIMEOUT_MS` is not being applied was the only branch with no test at all.** Of the two failure modes this PR reasons about, "bounded by the fallback" was proven and "honestly reports that it cannot bound anything" was not.

So rather than the comment you suggested, I added the test — which then needs the isolation your note anticipated. It strips both `AbortSignal.timeout` and `AbortController`, and asserts the once-ness against a **freshly imported module** (`vi.resetModules()` + `await import(...)`) rather than a shared instance, so what is measured is the flag rather than test ordering. Same shape as `8986ca4d` on #884, where you raised the identical point.

Mutation-checked before claiming it: initialising `warnedNoTimeoutSupport` to `true` fails this test **alone** — 1 failed / 622 passed across the package. Package now 24 files / 623 passed.

## The two I am not acting on, with reasons

**`AbortError` wrapped as `NetworkError`** — agreed, and agreed it is not urgent. A distinct `TimeoutError` is the right eventual shape, but it is a behaviour change for every current `catch`, and this PR deliberately leaves write-path error types untouched so that nothing catching a Boltz error changes shape. Folding an error-taxonomy change into a deadline PR would blur exactly the boundary the PR is arguing for. Worth its own change once a caller actually wants to branch on it.

**`BOLTZ_READ_TIMEOUT_MS` as a public export** — agreed, and deliberate for the reason you give: it makes the value greppable, and a constant a caller can read but that is silently not applied is the failure this PR treats as a defect. Confirming your cross-repo check matches mine: no consumer imports it today.

## Confirming your correctness section

Your `inflightGets` trace matches mine at `boltz-swap-provider.ts:1768-1773` — the `.finally()` sits downstream of the await, so a rejection evicts the entry and joined callers all receive it; a later caller gets a fresh request. That cleanup is precisely what a never-settling fetch would deny, which is the argument for the deadline.

## One thing outside your scope that bears on this PR's fate

Recorded in the comment above: `packages/boltz-swap` **does not exist on `feat/v0.5`** (#886), so this fixes code that branch removes — while `packages/swap`, which survives, carries the same unbounded-read defect in `rfq.ts` (`httpTransport`) and `markets.ts`, including the same dedup-without-a-deadline amplifier. Sequencing #886 is not mine to call, so that is evidence rather than a proposal.

Gate on `d1d26134`, run by hand (pre-commit hook refuses in a worktree — #872, fixed by #880): `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit`. boltz-swap 24 / 623, swap 33 / 811, ts-sdk 180 / 3023, zero failures.

Issue #889, opened 2026-09-08
getExpoFetch's fallback sends Ark-server headers to the indexer origin


Kukks:
`getExpoFetch` (`providers/expoUtils.ts:7-35`) returns the Ark-server `fetch` wrapper when `expo/fetch` will not import. That wrapper sets `X-Build-Version` and `X-SDK-VERSION`, which arkd expects and a non-arkd origin does not.

`ExpoIndexerProvider.getSubscription` (`providers/expoIndexer.ts:25`) passes that function to `sseStreamIterator` with an **indexer** URL. If the indexer is served from a different origin with a strict CORS policy, those custom headers fail the preflight and the subscription never opens.

The correct fallback is `baseFetch`, which exists precisely to omit them — its own docstring says "Use for any service that is NOT the Ark server … those origins reject unknown request headers such as `X-Build-Version` in the CORS preflight."

## Severity: latent, not live

Every usage I can find constructs `RestArkProvider` and `RestIndexerProvider` with the same `serverUrl`, so in practice the headers reach the origin that expects them. An embedder splitting the two hosts would hit it.

Pre-existing; not introduced by #884. Raised by `arkana-ai-bot` reviewing that PR, and kept out of it because it is a different file and a different failure mode.

## Same class as two other findings

This is the third instance of one shape — *the wrong fetch wrapper for what may be a distinct origin*:

- #887 — `providers/emulator.ts` used the global `fetch` and reached no SDK boundary at all; the emulator is a distinct origin, so the fix there is `baseFetch` rather than the Ark wrapper, for exactly this reason.
- #884 — the `baseFetch` chokepoint, which only helps callers that actually route through it.
- this one.

Worth deciding whether `getExpoFetch` should take the wrapper as a parameter rather than choosing one, since its two callers want different origins.

PR #891, opened 2026-09-08
docs(providers): name the wrapper the Expo fallback actually returns

Kukks:
`sseStreamIterator`'s comment (`expoUtils.ts:64-66`) says `getExpoFetch` falls back to `baseFetch`. It falls back to `fetch` — the **Ark-server wrapper** — which reaches `baseFetch` one layer further down.

```ts
// what the comment claims
// `getExpoFetch` falls back to `baseFetch`, which bounds any read …

// what getExpoFetch actually does (expoUtils.ts:1, :33)
import { fetch, buildVersion, sdkVersion } from "../utils/fetch";
...
return fetch;   // the Ark wrapper
```

Comment-only. No behaviour change, no test change.

## Why this is worth a PR: the comment asserts #889 is already fixed

**#889 is the open issue arguing that this fallback should be `baseFetch`. The comment says it already is.**

So the failure is not that the comment is vague. A reader auditing the Expo path — the reader most likely to be here, because this is where a distinct-origin subscription would break — finds a note stating the very fix that has not been made. It does not merely fail to warn them. **It closes the question.**

That is the general shape, and it is why this earns a PR rather than a drive-by: **a comment that is 90% right is worse than no comment, because it reads as a question already asked and answered.** An absent comment invites the reader to go look. A confident, slightly wrong one ends the enquiry before it starts.

## The elided layer is the one that matters

The difference between `fetch` and `baseFetch` is not incidental — it is the whole of it, and it is the security-relevant part. `fetch` sets `X-Build-Version` and `X-SDK-VERSION`; `baseFetch` exists precisely to omit them, because a non-arkd origin rejects unknown request headers in CORS preflight. `baseFetch`'s own docstring says so.

So the comment is wrong in exactly the place a reader would be checking, three lines from the fallback most likely to be audited.

## The same shape, once more, elsewhere

There is a second instance in one of our corridors this week, which is what makes it a pattern rather than a one-off. A comment opened with a genuinely correct and non-obvious warning — that a thrown error is not proof nothing was sent — and then asserted that a particular check owned that ambiguity. It did not; the check was an indexer read and could not settle the question. The *correct* first clause is what carried the wrong second clause past review, and the pairing was then copied into another corridor. Being mostly right is what gave it the range.

## Provenance

Found by `arkana-ai-bot` reviewing #884, in the sync posted after that PR had already been squash-merged, so it did not make the merge. I verified it independently before acting: `getExpoFetch`'s catch branch is `return fetch`, and line 1 imports `fetch` from `../utils/fetch`.

Related: **#889** — same class, *the wrong fetch wrapper for what may be a distinct origin*, in this same function. #889 changes the behaviour; this only stops the comment from claiming that change is already made. Independent, and they can land in either order.

## Test plan

Comment-only, so the bar is that nothing moved. Gate run by hand against `ab87f555`:

| | files | tests |
|---|---|---|
| master `ab87f555` (from CI) | ts-sdk 183 · swap 33 · boltz-swap 23 | 3047 passed, 1 skipped · 811 · 617 |
| this branch | ts-sdk 183 · swap 33 · boltz-swap 23 | 3047 passed, 1 skipped · 811 · 617 |

Identical across all three packages, zero failures. `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit` all clean.

The pre-commit hook refuses to run in a git worktree (#872, fixed by #880) and directs running the gate by hand and committing with `--no-verify`. I did both.

### DIFF
--- a/packages/ts-sdk/src/providers/expoUtils.ts
+++ b/packages/ts-sdk/src/providers/expoUtils.ts
@@ -62,8 +62,9 @@ export async function* sseStreamIterator<T>(
                 ...headers,
             },
             // Required, not just for cancellation: `getExpoFetch` falls back to
-            // `baseFetch`, which bounds any read that brings no signal of its
-            // own. Drop this and the stream is truncated at READ_TIMEOUT_MS.
+            // the Ark `fetch` wrapper, which routes through `baseFetch` and so
+            // bounds any read that brings no signal of its own. Drop this and
+            // the stream is truncated at READ_TIMEOUT_MS.
             signal: fetchController.signal,
         });
 


Kukks:
@arkana-ai-bot — the `#889` section and the core confirmation are right and useful. The `READ_TIMEOUT_MS` finding is not: it was measured against a tree that predates #884, and acting on it would put a new error into the comment.

## `READ_TIMEOUT_MS` exists

> `grep -r READ_TIMEOUT_MS` finds nothing.

It is defined on `master` at `packages/ts-sdk/src/utils/fetch.ts:37`:

```ts
export const READ_TIMEOUT_MS = 30_000;
```

with ten further references, including a public re-export at `index.ts:641` and four in `test/fetch-read-deadline.test.ts`.

## `baseFetch` does bound reads

> `baseFetch` … and the Ark `fetch` wrapper … both pass `init` straight to `globalThis.fetch`; neither inserts an abort signal or a deadline.

`fetch.ts:103-112`:

```ts
export function baseFetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
    ...
    const signal = readDeadline(input, init);
    return globalThis.fetch(input, signal ? { ...init, signal } : init).catch(...)
}
```

`readDeadline` (`:53`) returns `AbortSignal.timeout(READ_TIMEOUT_MS)` for a GET or HEAD that carries no signal of its own, with an `AbortController` fallback at `:70`. Inserting that signal is the whole of #884.

## The line numbers say which tree was read

The review cites `baseFetch (utils/fetch.ts:37-45)` and `the Ark fetch wrapper (utils/fetch.ts:53-58)`. On the tree this PR is built from, `:37` is `READ_TIMEOUT_MS`, `:53` is `readDeadline`, and the two functions are at **`:103`** and **`:120`**.

Those offsets are the pre-#884 file. #884 merged as `d27dc82e` at 20:32Z; this PR branches from `ab87f555`, which contains it. So the finding is accurate about a tree that no longer exists, and the `grep` returning nothing is the clearest confirmation of which one was searched.

Flagging it plainly because this is the second time today — the second #884 review reported four findings as "still open" against commits that were the branch's two newest. Both times the substance was recorded against the right `commit_id` and derived from the wrong tree. Worth knowing for anyone weighing these reviews later.

## The `expoIndexer.ts:81` citation is real, but is about the other path

That comment — "these timeouts are set by expo/fetch function" — sits in the `AbortError`/timeout handler and concerns the **expo/fetch** path, where `expoFetchModule.fetch` is used and never reaches `baseFetch`. The comment this PR edits is scoped to the **fallback** ("`getExpoFetch` falls back to…"), which is the path where `baseFetch` and `READ_TIMEOUT_MS` do apply.

Both are true at once: expo/fetch imposes its own idle deadline on its path; `READ_TIMEOUT_MS` bounds a signal-less read on the fallback path.

## Why I am not taking the suggested rewrite

```ts
// expo/fetch imposes an internal read deadline (READ_TIMEOUT_MS) when no signal is provided.
```

This attributes an SDK constant defined in `utils/fetch.ts` to a third-party module that cannot see it, and it moves the claim onto the path where it is false. That is the same failure this PR exists to correct — a confident sentence that reads as settled and sends the next reader to the wrong layer.

The current text is accurate for the path it names, so the code stays as-is. If a reviewer wants the expo/fetch deadline documented too, that belongs next to `expoIndexer.ts:81` where that path actually lives.

## `--no-verify`

#880 is merged — `f5f9f895`, 19:22:30Z, before this commit. The hook refusing **is** #880 working: it declines to run in a worktree because `pnpm --filter="...[HEAD]"` selects no projects there, and it prints the manual gate to run instead. I ran that gate in full (`pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit`) with results matching `ab87f555` exactly across all three packages.

**#889 stands, unchanged, and your restatement of it is right** — `expoIndexer.ts:33` calls `getExpoFetch()` without `requireExpo: true`, so a distinct-origin indexer meets the Ark wrapper's headers in preflight. Stopping the comment from claiming that is already fixed is the entire point of this PR.

Kukks:
Thank you — and the most useful line in this review is the one about your own setup:

> the local clone on this VPS is behind the PR's base (`94c9a1f` vs `ab87f555`) and lacks these additions

`94c9a1fb` is the commit immediately before #884, #887 and #882 merged, which is exactly why `READ_TIMEOUT_MS` was invisible: it arrives with #884 (`d27dc82e`). That one fact explains both of today's stale reads — this PR's first pass, and the second #884 review that reported four findings as "still open" against commits that were then the branch's two newest. I had it as two coincidences; you've given it a single cause, and naming it is worth more than either finding.

Re-fetching the file over the API when the clone looks behind is the right instinct. The generalisation, if it helps: the tell is cheap and available before the review starts — a symbol that should exist returning nothing from `grep` is a statement about the tree, not about the code.

On `--no-verify`: agreed it should not become routine, and worth being precise about what happened. The hook refusing **is** #880 working — it is merged (`f5f9f895`, 19:22:30Z), and it declines in a git worktree because `pnpm --filter="...[HEAD]"` selects no projects there, printing the manual gate rather than passing silently, which is the bug #872 described. So `--no-verify` here follows the hook's own instruction rather than bypassing it, and the gate was run in full. The residual item is worktree ergonomics, not the hook.

Everything else confirmed as you have it: the comment is accurate for the fallback path it names, and #889 remains the actual bug.

Issue #878, opened 2026-09-08
swap: restoreAssetSwaps treats "pending" as an answer, so a swap it rebuilds as pending stays pending forever


tiero:
`unscannedSwapCandidates` (`restore.ts:95-101`) drops any tx whose txid is in `existingIds` or in `scanned`. A run that finds the deposit still unspent returns the record with `status: "pending"` *and* puts its funding txid in `scannedTxids`: that list is `fetchedTxids` minus `unresolved`, and a pending record is not unresolved (`restore.ts:391`).

So the natural call, passing every stored record id as `existingIds` and the stored cursor as `scanned`, asks about a funding tx exactly once and never again. `pending` is the absence of an answer, but it is persisted as one.

## Why it matters

For a record whose covenant is registered, the watcher covers the gap. For a record the scan itself rebuilt there is no covenant at all (#877), so the scan is the only route that swap has to an outcome, and it has closed itself off. The record is stuck `pending` for the life of the wallet.

The same applies to a swept deposit: `restoreAssetSwaps` is the only thing in the package that reports `recoverable`, and it reports it once, on the run that happens to catch the sweep.

## Workaround, and why it should not be the caller's job

A consumer can pass only *terminal* ids as `existingIds` and subtract open records' funding txids from the cursor. arkade-os/wallet#962 does exactly that. But every consumer has to rediscover it against a docblock that reads the other way ("Rebuild swap records after a wallet restore ... nothing is ever fetched twice"), and getting it half right (relaxing one skip list but not the other) silently changes nothing.

It is also the wrong shape for a resolve path: re-asking re-fetches the funding psbt to re-decode an offer the caller's record already carries in `offerHex`.

## Suggested fix

Take the open records directly, e.g. a `reopen: AssetSwap[]` option: decode each offer from its stored `offerHex`, skip the funding-tx fetch entirely, and go straight to the vtxo lookup and `classifyDepositSpend`. That makes re-answering one indexer call instead of two and removes the skip-list reasoning from the caller.

Failing that, do not mark a txid scanned while the record it produced is non-terminal, and document that a caller must keep open ids out of `existingIds`.

Issue #864, opened 2026-09-08
swap: watchOfferSwaps has no pass behind its subscription, so a spend it does not receive live is never learned


tiero:
`watchOfferSwaps` (`packages/swap/src/watch.ts`) is a pure event sink. Its entire wiring is one subscription:

```ts
// packages/swap/src/watch.ts:193-196 (master d652ab6) / :280-283 (feat/v0.5 6fb72c2)
    const unsubscribe = manager.onContractEvent((event) => {
        if (event.type !== "vtxo_spent") return;
        enqueue(() => handleSpend(event));
    });
```

There is no start-up pass and no failsafe of its own — the function returns immediately after subscribing (`watch.ts:198-201` master, `:285-288` v0.5). A spend it does not receive as a live, well-formed `vtxo_spent` event is never learned, and the record keeps `status: "pending"` with no `spentTxid` indefinitely.

Two SDK paths discover such a spend and neither can tell it.

## 1. The boot sync runs before any subscriber can exist

`ContractManager.initialize` reconciles first and installs the watcher callback afterwards:

```ts
// packages/ts-sdk/src/contracts/contractManager.ts:899 and :909 (master) / :889 and :899 (v0.5)
            await this.reconcileWatched();
...
        this.stopWatcherFn = await this.watcher.startWatching((event) => {
```

`create` is the only construction path and it awaits `initialize()` (`contractManager.ts:808-810` master, `:798-800` v0.5), so no caller can hold the manager — and therefore none can have called `onContractEvent` — until the boot reconcile has already run. `emitEvent` is private with exactly one call site, the last line of `handleContractEvent` (`contractManager.ts:2100` and `:2162` master, `:2088` and `:2150` v0.5; `git grep emitEvent` returns only those two lines on each branch). So everything the boot sync writes to the repository is invisible to subscribers.

The ordinary case is a wallet closed with an offer funded and open, the solver filling it, and the user reopening.

## 2. The failsafe poll emits, but with a payload that cannot carry a spend txid

`ContractWatcher` seeds its baseline from the repository before the boot sync, skipping spent rows (`contractWatcher.ts:191`, `:215`). `pollContracts` then reads the post-sync repository with `includeSpent: false` (`:556`) and reports the difference using the **cached** rows, not fresh ones:

```ts
// contractWatcher.ts:578-593
                for (const [key, vtxo] of state.lastKnownVtxos) {
                    if (!currentKeys.has(key)) {
[...]

PR #892, opened 2026-09-09
fix(swap): restore coverage for the covenants the scan rebuilds

### LINKED ISSUE #877: swap: restoreAssetSwaps rebuilds records without restoring coverage, so a restored wallet's covenants are never watched
`createOffer` registers the offer covenant as a wallet contract: `offer.ts:633` calls the private `registerOfferContract`, which does `contract.register({ label, metadata: { genericallySpendable: false, kind: OFFER_CONTRACT_KIND } })` and then `promoteOfferContract`.

`restoreAssetSwaps` does not. It writes records and nothing else. Grep `packages/swap/src/restore.ts` for `register`, `createContract` or a contract manager and the only hit is a comment.

## Consequence

A record the restore scan rebuilt has no contract row behind it. So:

* `ContractManager` never subscribes to the covenant script.
* `watchOfferSwaps` is a pure sink on `manager.onContractEvent` (`watch.ts:193`), so it can never fire for that swap. No amount of waiting resolves the record.
* Any consumer reading contract rows for the script (a reconciliation pass, a balance view, a coverage check) finds nothing.
* The deposit is not one of the wallet's own coins, so the funding tx surfaces in history as an ordinary `sent` row rather than netting away against the covenant output.

The last two combine into the user-visible failure: a filled swap renders as a pending row plus a bare received row, and a cancel returns funds under a "Received" row with nothing tying it to the swap.

## Evidence

From a real mainnet wallet restored from seed: 35 asset-swap records spanning three weeks, every one with `swapAddress: ""` (only `restore.ts:386` writes that, so every record came from the scan), and zero contract rows for any of their covenant scripts. Six unrelated contract rows sit in the same store, so the store itself is healthy.

## Why the existing note does not cover it

`coverage.ts`'s docblock says a funded-but-unwatched row is left to "the restore scan [which] still finds [it] — the same backstop every other best-effort step here relies on". That backstop restores the *record*, not the *coverage*, so it does not hold. The two other best-effort steps it excuses (a second wallet context issuing while this one retires, an address issued before a restart and funded after) inherit the same hole.

## Suggested fix

`registerOfferContract` is private and `restoreAssetSwaps` is deliberately indexer-only, so this needs a seam. Either:

* export `restoreOfferCoverage(wallet, arkServerUrl, network, records)`, which re-registers and promotes each rebuilt covenant, or
* take an optional `coverage: { wallet, network }` on `restoreAssetSwaps` and do it inline.

The scan already holds the decoded `Offer` and the `serverPubkey`, which is everything `registerOfferContract` needs beyond the wallet and the network.

Whichever shape: do not promote a covenant whose record is already `RETIRABLE`, or the restore undoes `retireSettledOfferContracts` and puts settled scripts back in the subscription.

Related: #878 (the scan will not re-answer a record it left pending, which is what makes this permanent), #866 (four modules document a restore-scan backstop nothing runs), #864 (`watchOfferSwaps` has no pass behind its subscription).

Kukks:
Closes #877.

## The bug

`createOffer` registers an offer's covenant as a wallet contract (`offer.ts:633`). `restoreAssetSwaps` writes records and nothing else — grep `restore.ts` for `register`, `createContract` or a contract manager and the only hit is a comment.

So a record the restore scan rebuilt has **no contract row behind it**. `ContractManager` never subscribes to the script, no `vtxo_spent` is ever emitted for it, and `watchOfferSwaps` — a pure sink on `onContractEvent` — can never resolve that swap. No amount of waiting helps.

The evidence in the issue is a mainnet wallet restored from seed: 35 asset-swap records spanning three weeks, every one with `swapAddress: ""` (only the scan writes that), and zero contract rows for any of their covenant scripts — in a store holding six healthy unrelated rows.

### The backstop three modules lean on does not catch this

`coverage.ts`'s docblock excuses two other best-effort gaps on the grounds that both "leave a funded-but-unwatched row that the restore scan still finds — the same backstop every other best-effort step here relies on."

That backstop restores the **record**, not the **coverage**. It does not hold here, and it does not hold for the two gaps it was written to excuse.

## The fix

`restoreOfferCoverage(wallet, arkServerUrl, swaps)` re-registers and promotes each rebuilt covenant, deriving it from the record's own `offerHex` against the server's current signer key, once per script.

A separate export rather than an option on `restoreAssetSwaps`: that scan is deliberately indexer-only, and a coverage failure must not cost a caller the records it just recovered. Best-effort per script — a covenant that no longer derives, because the server key rotated since funding, is skipped rather than registered at the wrong script.

## Three things it must not do

All three are tested, and each mutation is caught by exactly one test.

**1. Promote a script whose every record is already `RETIRABLE`.** Named in the issue. That would undo `retireSettledOfferContracts` and put settled scripts back in the subscription and the failsafe poll.

**2. Leave the restored script pinned watched.** Not in the issue; found while building this. `promoteOfferContract` records an issuance mark so a script whose address is out but not yet funded is never retired, and `addressOutstanding` clears it only when a record at the script is *newer* than the mark. A restored record is older than `Date.now()` by construction — so marking now blocks every later retire **for the life of the process**. `promoteOfferContract` now takes the issuance time as a parameter, defaulting to `Date.now()`, and restore passes the record's own funding time.

**3. Cost a round trip per script.** `Arkade.connect` does a live `/v1/info` with no cache, so building a client per record puts N sequential requests on the **boot path** — at 110-215 ms each that is seconds for a wallet with the history this bug was found on. One client is built for the batch and handed to each registration, making `/v1/info` a constant cost. A test pins it; reverting the hoist turns it red.

## Test plan

- `packages/swap/test/restoreCoverage.test.ts`, 7 tests. Its own file because `coverage.ts` keeps the issuance mark in module state — a `createOffer` promotion in the same file would decide the retire assertion.
- As in `register.test.ts`, the covenant derivation, `Arkade.connect`, `ArkadeContract` and `register()` are all real; only the network seam and the contract manager are stubbed.
- Gate on this branch: `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit` all exit 0. swap `33 files / 811 tests` -> `34 / 818`; ts-sdk `183 / 3047 +1 skipped` and boltz-swap `23 / 617` unchanged.

## Notes for review

- The pre-commit hook refuses to run in a git worktree (#880) — its filter selects no projects, so lint and `test:unit` would test nothing. The gate above was run manually and the commit uses `--no-verify`, as that hook's own message instructs.
- Independent of #864 and #878, not stacked. Merge order does not matter, though #877 and #878 are what make a restored wallet's swaps resolvable at all.
- Conflict risk: the export list in `packages/swap/src/index.ts`, against the in-flight onchain-receive work. One line; keep both sides.

## Summary by CodeRabbit

* **New Features**
  * Added automatic restoration of coverage for eligible, unsettled swap records.
  * Restored swaps are tracked using their original funding time.
  * Added support for reusing an existing connection during swap registration.
  * Exposed swap coverage restoration through the package API.

* **Bug Fixes**
  * Prevented restored deposits that have already landed from being incorrectly marked as watched.
  * Settled and stale swap records are skipped safely during restoration.

## Failure contract (added after review)

Setup for the batch — `getInfo`, `Arkade.connect`, `getContractManager` — **rejects** rather than being caught, and the docblock now says so. A server it could not reach covered *nothing*, which is exactly the unwatched-covenant state this issue is about, and a caller told it succeeded never retries. `promoteOfferContract` already draws the same line ("Throws, unlike retiring..."). A single unusable record — a covenant that no longer derives — is skipped instead, so one bad record cannot block the other thirty-four. Guard the call if your boot path must survive an outage.

### DIFF
--- a/packages/swap/src/coverage.ts
+++ b/packages/swap/src/coverage.ts
@@ -103,12 +103,16 @@ function addressOutstanding(swaps: AssetSwap[], script: string): boolean {
 export async function promoteOfferContract(
     manager: OfferContractRetirer,
     script: string,
+    /** When the address went out. A restore passes its record's funding time:
+     * that deposit has landed, and a mark dated after every record at the
+     * script pins it watched for the life of the process. */
+    issued: number = Date.now(),
 ): Promise<void> {
     await serialize(script, async () => {
         await manager.setContractWatchState(script, "watched");
         // after the write: a promotion that failed hands out no address, so
         // there is nothing outstanding to protect
-        issuedAt.set(script, Date.now());
+        issuedAt.set(script, issued);
     });
 }
 
--- a/packages/swap/src/index.ts
+++ b/packages/swap/src/index.ts
@@ -1,6 +1,7 @@
 export {
     createOffer,
     cancelOffer,
+    restoreOfferCoverage,
     encodeOffer,
     decodeOffer,
     offerVtxoScript,
--- a/packages/swap/src/offer.ts
+++ b/packages/swap/src/offer.ts
@@ -40,9 +40,14 @@ import {
 
 import wantAssetProgram from "./swap-want-asset.program.json";
 import wantBtcProgram from "./swap-want-btc.program.json";
-import { promoteOfferContract, retireOfferContract } from "./coverage";
+import { promoteOfferContract, retireOfferContract, RETIRABLE } from "./coverage";
 import type { AssetSwapRepository } from "./repository";
-import { getAssetSwapsOrThrow, updateAssetSwap, updateAssetSwapBestEffort } from "./store";
+import {
+    getAssetSwapsOrThrow,
+    updateAssetSwap,
+    updateAssetSwapBestEffort,
+    type AssetSwap,
+} from "./store";
 
 // json imports widen "type": "pubkey" to string; parseArtifact validates at runtime
 type Artifact = Parameters<typeof arkade.parseArtifact>[0];
@@ -477,19 +482,22 @@ async function registerOfferContract(
     binding: Omit<Offer, "swapPkScript">,
     serverPubkey: Uint8Array,
     expectedPkScript: Uint8Array,
+    opts: { issued?: number; client?: arkade.Arkade } = {},
 ): Promise<void> {
     const { program, args, keys } = swapProgramBinding(binding, serverPubkey);
     const contractManager = await wallet.getContractManager();
-    const client = await arkade.Arkade.connect({
-        arkade: new RestArkProvider(arkServerUrl),
-        indexer: new RestIndexerProvider(arkServerUrl),
-        identity: wallet.identity,
-        // without this the row's `address` would be derived against the SDK's
-        // default network while its script is right — a row that disagrees with
-        // the address the user is about to fund
-        network: getNetwork(network),
-        contractManager,
-    });
+    const client =
+        opts.client ??
+        (await arkade.Arkade.connect({
+            arkade: new RestArkProvider(arkServerUrl),
+            indexer: new RestIndexerProvider(arkServerUrl),
+            identity: wallet.identity,
+            // without this the row's `address` would be derived against the SDK's
+            // default network while its script is right — a row that disagrees with
+            // the address the user is about to fund
+            network: getNetwork(network),
+            contractManager,
+        }));
     const contract = new arkade.ArkadeContract(client, program, args, keys);
     // the row is keyed by script: registering anything but the script being
     // funded would leave the real deposit unwatched and unmarked, which is the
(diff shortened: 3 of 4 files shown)


Kukks:
Thanks — the observation is accurate: the batch setup (`getInfo`, `Arkade.connect`, `getContractManager`) sits outside the per-script `try`/`catch`, so a server outage rejects the whole call.

I've documented that contract rather than swallowing it, because I think rejecting is the right behaviour here and swallowing would be actively harmful.

`coverage.ts` already draws this line explicitly, in `promoteOfferContract`:

> Throws, unlike retiring: an address returned without coverage is the failure this exists to prevent, and nothing is at stake yet, so the caller can retry.

`restoreOfferCoverage` *is* promotion, in bulk. If the server is unreachable, **nothing** was covered — that is precisely the state this function exists to prevent, and a caller told it succeeded never retries. It would trade a loud, retryable failure for the silent unwatched-covenant state that #877 is about.

The asymmetry with the per-script `catch` is deliberate rather than an oversight: one record with a rotated server key must not block the other thirty-four, whereas a dead server means none of them can be done at all and the caller needs to know.

What was genuinely missing is that none of this was written down, so a boot caller could be surprised by it. Fixed in `91fa31ec` — the docblock now states that setup rejects while a single bad record is skipped, so anyone wiring this into a boot path knows to guard it.

Kukks:
Thanks — Issue 1 was real and it was mine. Fixed in `48be1e90`.

## Issue 1 — split contract manager: confirmed and fixed

You're right that `registerOfferContract` fetched `wallet.getContractManager()` unconditionally, so on the restore path `contract.register()` went through `opts.client`'s embedded manager while `promoteOfferContract` used a separately-fetched one.

Two things I checked before acting:

- `Wallet.getContractManager()` memoises (`wallet.ts:2057-2069` — returns `this._contractManager`, or the in-flight init promise), so the SDK's own wallet returns the same instance either way and nothing breaks there.
- But `IWallet` is a public interface and a consumer implementation need not memoise — and, more to the point, **before my client hoist a single manager did both halves**. So this was a regression I introduced, not a pre-existing quirk, which settles it regardless of the memoisation.

The manager now threads through `opts` alongside the client, fetched once for the batch. `createOffer` passes no override and is unchanged. The existing cost test pins the call count too: reverting the override makes it 5 for a four-script batch instead of 1.

## Issue 2 — sequential loop: confirmed, and worse than you put it. Not fixed here.

You framed this as local store fanout. It's heavier than that: `createContract` does `await this.fetchContractVxosFromIndexer([contract])` (`contractManager.ts:1264`) — an **indexer round trip per contract**. So restoring N covenants costs N indexer calls on the boot path, on top of the `/v1/info` this PR already made constant.

I'm deliberately not fixing it here, because the right fix is not `Promise.all`:

- Parallelising `createContract` still makes N calls; `rateGate` caps concurrency at 6 per origin, so it trades latency for a burst and leaves the call count untouched. It also makes partial-failure semantics worse on a boot path.
- The SDK **already has the right shape** — `persistAndWatchContract` (`contractManager.ts:1275-1291`), which skips the per-contract fetch precisely so "a scan that finds N contracts costs one batched indexer call instead of N + 1", paired with a bulk `refreshVtxos`. It is `private`, used only by `scanContracts`.

So the real fix is exposing a batch registration path in `@arkade-os/sdk`, not restructuring the loop in `@arkade-os/swap`. That is a public API change to another package with its own design question (a contract registered without hydration is a footgun for consumers), and it does not belong as a rider on a correctness fix. Flagging it for a follow-up rather than smuggling it in.

## Minor note

Good catch that `state.created` isn't reset between the two calls in the cost test — added a one-line comment saying `5 = 1 + 4` so the next reader doesn't re-derive it.

Kukks:
> Ready to merge once the sequential-loop follow-up is tracked against the SDK.

Tracked, and built: **#897**, stacked on this branch.

One deliberate departure from the shape you suggested there. Making `persistAndWatchContract` public is the footgun I'd want to avoid: it skips hydration entirely, and `scanContracts` gets away with that only because the *caller* (`Wallet.restore`) follows up with a bulk `refreshVtxos`. Exposed raw, it hands consumers a contract that is registered and watched but unhydrated — and because `watcher.addContract` seeds its baseline from the repository, the next poll then reports that contract's whole history as newly received.

So #897 adds `createContracts` instead, which is `createContract`'s own step order with only the fetch batched: persist all → hydrate once → watch all. Same guarantees per contract, one indexer round trip instead of N.

Two things it turned up that are worth knowing here:

- **`createContract`'s cost is an indexer round trip, not a store write** (`contractManager.ts:1264`) — so the sequential loop was heavier than either of us first framed it, and squarely in the same class as the `/v1/info` hoist this PR already made.
- **Declaring the method as *required* on `IContractManager` is a breaking change**, not an additive one. It broke the build immediately: `serviceWorker/wallet.ts:1260` implements the interface by proxying every method over a message protocol, where a new method means new wire. That's in-tree evidence external implementers would break too, so it is optional with a per-item fallback — the service-worker path keeps today's behaviour rather than regressing.

Kukks:
Consumed by arkade-os/wallet#976: the wallet now restores coverage for records loaded at startup and records rebuilt by the current scan. The vendored integration tarball is built from branch \integration/wallet-restore\.

PR #893, opened 2026-09-09
fix(swap): let restoreAssetSwaps re-ask a record it left pending

### LINKED ISSUE #878: swap: restoreAssetSwaps treats "pending" as an answer, so a swap it rebuilds as pending stays pending forever

Kukks:
Closes #878.

## The bug

`unscannedSwapCandidates` (`restore.ts:90-101`) drops any tx whose txid is in `existingIds` or in `scanned`. A run that finds the deposit still unspent returns the record with `status: "pending"` **and** puts its funding txid in `scannedTxids` — that list is `fetchedTxids` minus `unresolved`, and a pending record is not unresolved.

So the natural call — every stored id as `existingIds`, the stored cursor as `scanned` — asks about a funding tx exactly once and never again. **`pending` is the absence of an answer, but it is persisted as one.**

For a record the scan itself rebuilt there is no covenant behind it either (#877), so the scan was that swap's only route to an outcome and it had closed itself off. The same goes for a swept deposit: `restoreAssetSwaps` is the only thing in the package that reports `recoverable`, and it reported it once — on the run that happened to catch the sweep.

## The fix

`reopen: AssetSwap[]` takes the open records directly. Each offer is decoded from its stored `offerHex`, so the funding-tx fetch is skipped entirely and the record goes straight to the vtxo lookup and `classifyDepositSpend`. `existingIds` and `scanned` stay exactly as they were — the skip-list reasoning leaves the caller.

Two properties worth stating, both tested and both mutation-checked:

- **Only a definite outcome comes back.** A reopened record whose deposit is still unspent contributes nothing. Emitting `pending` would overwrite a `cancelling` record and lose the cancel txid that `watchOfferSwaps`'s exact classifier matches on.
- **A resolved reopened record keeps its stored fields.** It is returned as the stored record with the new status, not rebuilt from chain data — which would blank the `swapAddress` a `createOffer` record carries.

## If you are looking at arkade-os/wallet#962, read this

**wallet#962 keeps working unchanged, and it must not be combined with `reopen` naively.**

#962 works around this bug caller-side, by keeping open ids out of `existingIds` and subtracting them from the cursor. That is still correct on its own.

But a consumer that adopts `reopen` **without** reverting #962 would hand the same funding tx to *both* paths, and the two disagree — the scan path writes `swapAddress: ""` where `reopen` preserves it. One id would come back twice, with two different records, and whichever the caller's upsert applied last would win. That is worse than the redundancy it looks like.

So a reopened txid is dropped from the scan's candidate set here, which makes the combination merely redundant rather than harmful, in either order. There is a test for exactly that shape (`answers a funding tx once when it is also a scan candidate`). **The clean end state for a consumer is: keep every stored id in `existingIds`, keep the cursor whole, and pass the open records as `reopen`.** No change is required in that repo for this PR to land.

## Test plan

- Six tests in `packages/swap/test/restore.test.ts`, under `restoreAssetSwaps — reopening records the scan left pending`.
- Gate on this branch: `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit` all exit 0. swap `33 files / 811 tests` -> `33 / 817`; ts-sdk `183 / 3047 +1 skipped` and boltz-swap `23 / 617` unchanged.

## Notes for review

- The module docblock's "nothing is ever fetched twice" is still true and now points at `reopen`; the full contract is on `restoreAssetSwaps` itself, which is the doc a caller's editor surfaces at the call site.
- The pre-commit hook refuses to run in a git worktree (#880) — its filter selects no projects, so lint and `test:unit` would test nothing. The gate above was run manually and the commit uses `--no-verify`, as that hook's own message instructs.
- Independent of #864 and #877, not stacked.

### DIFF
--- a/packages/swap/src/restore.ts
+++ b/packages/swap/src/restore.ts
@@ -10,7 +10,7 @@
  *
  * The scan is incremental: txids checked with an authoritative answer are
  * remembered, so late-synced history is picked up by later scans and nothing
- * is ever fetched twice.
+ * is ever fetched twice. `reopen` re-asks one it left `pending`.
  */
 import { base64, hex } from "@scure/base";
 import {
@@ -107,6 +107,8 @@ const unscannedSwapCandidates = (
  * caller decides whether to retry or accept a default.
  */
 export type SpendKind = "cancelled" | "fulfilled" | "indeterminate";
+
+type Found = { fundingTx: Tx; offer: Offer; offerHex: string; existing?: AssetSwap };
 
 /**
  * Classify a spend by the covenant leaf it took.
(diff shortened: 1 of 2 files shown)


Kukks:
Thanks — and the one gap you named was the right one to name.

**`cancelling` → `cancelled` through `reopen` now has a dedicated test** (`8deed3fc`). You're right that it was the operationally important path: the suppression side was covered (a `cancelling` record whose deposit is still unspent is left alone), but not the positive resolution, which is the whole reason the status exists — the user cancels, the wallet closes before the spend lands, and `reopen` is what finishes it.

Mutating the emitted status to a constant `fulfilled` turns the new test red, so it discriminates rather than decorating. It also pins that a cancel gets no `completedAt` — the rule `spendUpdate` states, which the reopen path inherits through the shared `completion`.

Two of your observations I want to confirm back, because both are load-bearing and I'd rather they be on the record than assumed:

- **`spentTxid: undefined` in the spread for `recoverable`.** Your reading matches mine: the `pending` guard means only `fulfilled`, `cancelled` and `recoverable` reach that line, and for a swept deposit overriding a stale `spentTxid` to `undefined` is the correct outcome, not a loss.
- **`unresolved.add(...)` for reopened records is inert.** Also matches: a reopened txid is never in `fetchedTxids`, so the `filter` at the return has nothing to remove. Harmless, and it keeps the branch uniform with the scan path rather than special-casing it.

Gate after the new test: `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit` all exit 0; swap goes 817 → 818, ts-sdk and boltz-swap unchanged.

Kukks:
@coderabbitai review

Kukks:
Consumed by arkade-os/wallet#976: open records now travel through \
estoreAssetSwaps({ reopen })\, while the normal existing/scanned skip lists remain intact. The vendored integration tarball is built from branch \integration/wallet-restore\.

PR #894, opened 2026-09-09
feat(swap): drive the onchain-receive corridor, L1 refund included

Kukks:
## Summary

`onchain:BTC->arkade:BTC` was excluded from `RfqSwap` by name (`swapManager.ts:291` on master). Its Arkade half is a `LightningReceiveSwap`'s solver-funded lockup, but it adds an L1 HTLC **the trader funds and must take back themselves** — a second deadline, a second observation seam and a second action callback. The comment's own argument was that wiring the lockup half alone would be worse than refusing. This wires both halves and removes the refusal.

## The failure mode is not the one the old comment described

The comment said the manager could "silently let that L1 refund window pass". **That window does not close.** `buildHtlcRefund` sets `nLockTime` (`onchainHtlc.ts`), which consensus reads as a *floor*, and the only spend that can remove the leaf is the solver's claim — which needs `P`, held by the trader.

So the hazard is not a missed deadline but **the manager finishing the swap while that output still holds the trader's money**, and `runPass` step 1 is what would cause it: on this leg `LockupFate.returned` means the *solver reclaimed the lockup*, which is precisely when the trader's funding must come home — yet it maps to the terminal state `refunded` and finalizes, which unwatches the HTLC forever.

The receive arm is therefore dispatched **before** that terminal shortcut, and a record here is terminal only once **both** halves have a verdict:

| Arkade half | L1 half | terminal? |
|---|---|---|
| `claimed` | — (the output is the solver's payment) | `settled` |
| `returned` | our refund landed, or solver took it | `refunded` |
| `returned` | never funded, never seen funded | `refunded` |
| `returned` | **still funded** | **no — keep driving the refund** |

A mutation moving the dispatch back after the shortcut turns four tests red on `isRfqSwapTerminal`.

**The Arkade arm must not finalize either.** `driveReceiveClaim` ends a receive swap at `refundLocktime + REFUND_MTP_LAG_SECONDS` — right for the lightning leg, which owes nothing else, and wrong here because this leg's own refund leaf opens *after* that moment. The reachable trigger is the mainline default rather than an outage: `readLockupFate` returns `unknown` for a lockup with no VTXOs (`refund.ts:446`), which is what a solver that never funds leaves behind. So the swap is held open until the L1 half reports `resolved`. That refusal is applied *after* the Arkade arm, because that arm calls `unblock` on an empty lockup and would otherwise erase it.

## No new constant, and no clock

`nextOnchainReceiveAction` mirrors `nextOnchainAction` with one member fewer — there is no `refund_window_closed` — and takes **neither a clock nor a margin**. `refundable` *is* the median-time-past verdict `classifyOnchainHtlc` already computed, so a second time source could only disagree with the phase it was handed. A refund publishes nothing and races nobody, so there is no counterpart to `ONCHAIN_CLAIM_MARGIN_SECONDS`. Correspondingly there is **no give-up branch**, unlike `driveArkadeRefund` — the safety property expressed structurally rather than as a number.

## Nothing about the phases is inverted

`claimed` versus `swept` turns purely on whether the spending witness revealed a preimage, **never on who spent**. What differs between the legs is which leaf the trader owns, so the same fact calls for a different actor. `OnchainHtlcPhase`'s docs carried only the send leg's reading — a receive-leg reader arrived at `refundable` and was told *"must not try to claim… the claim was missed"*, which is the opposite of what to do there. Amended to name both directions.

## Refusing to drive blind

Following `driveOnchain`'s missing-`ChainSource` precedent, split by whether the gap can be closed later:

- **No `ChainSource` → `addSwap` throws `OnchainReceiveNeedsChainSource`.** `chain` is constructor-injected and can never arrive late, and `addSwap` normally precedes the trader's L1 funding — the last moment refusing costs nothing.
- **No `refundOnchain` callback → `needs_counterparty`, non-terminal, re-checked every pass.** Exactly `claimOnchain`'s precedent: `setCallbacks` exists so a callback *can* arrive late, and nothing is expiring while it does. Auto-actions-off reports the same way, because unlike the send leg there is no `claimable` state to fall back on and silence would strand a live refund.
- **`restoreFromRepository` gets the same refusal.** It rebuilds and `track`s directly rather than going through `addSwap`, so without its own check a restored swap could be monitored with no way to see or make its refund. Inside the existing `try`, so it lands in `RfqRestoreResult.failed` and the record stays.

## Neither leg may be taken twice

Whichever leg the trader takes first forecloses the other, and both directions are explicit refusals rather than missing arms — so neither is later "fixed" into a theft.

| pairing | what closes it |
|---|---|
| claim, then refund (later pass) | `claimArkTxid` suppresses the refund |
| refund, then claim (either pass) | `refundTxid` blocks the claim |
| refund **attempted**, then claim | the attempt itself blocks the claim |
| restored record, either order | `hydrate` carries both txids |

The attempt case is the subtle one: a `refundOnchain` that broadcasts and then throws — a parse failure on the response suffices — leaves no txid, so keying on the txid alone let the same pass go on and claim. Being wrong about an attempt costs one poll interval on a claim whose window is hours.

A claim that throws is deliberately **not** symmetric: it leaves the trader with nothing on the Arkade side, so the L1 refund is still owed and must stay available.

`claimArkTxid` cannot ride on the `settled` flag, which drives the terminal early return — folding it in would skip the refund entirely and strand the funding if the claim never landed.

This matters here and not on the send leg because `assertFundable` applies its timelock-order gate to `direction === "send"` only (`rfq.ts:539`), so on this corridor both legs can be live at once.

## Restore

`OnchainReceiveCorridor` makes the record round-trip — a handler plus one `register()` call, as `rfqCorridors.ts` documents. `hydrate` throws rather than defaults on missing L1 keys, an uncheckable `minConfirmations`, a non-finite `expectedAmount`, and inputs deriving a different HTLC address.

`refundPkScript` is **required**, where the send leg's `payoutPkScript` is optional: no record predates this corridor, and without it no refund can be built at all.

## Breaking changes

Two, for consumers typing against the strict `RfqSwapManagerCallbacks`:

1. **`refundOnchain` is a new required member.** Deliberately not widened away — a monitored receive swap with nothing wired to refund it is the exact lie that type exists to prevent.
2. **`claimLockup`'s parameter widens to `ReceiveLockupSwap`**, so both receive legs share one claim path and one `expectedAmount` gate rather than duplicating it. Under `strictFunctionTypes` this rejects a callback whose parameter is explicitly annotated `LightningReceiveSwap`.

Inline lambdas are contextually typed and unaffected. `AvailableRfqSwapManagerCallbacks` users are unaffected.

**Known downstream:** `wallet/src/test/providers/lnReceive.test.tsx` casts a mock to the strict `RfqSwapManagerCallbacks`; that cast needs `refundOnchain` added when the wallet bumps `@arkade-os/swap`. The provider itself uses `setCallbacks` and is safe.

## Test plan

Gate run against a recorded baseline on `d4228ede`, whole repo, not narrowed:

| | baseline | this branch |
|---|---|---|
| ts-sdk | 183 files / 3047 + 1 skipped | unchanged |
| boltz-swap | 23 files / 617 | unchanged |
| swap | 33 files / **811** | 33 files / **844** |
| **total** | **239 / 4475 / 0 failing** | **239 / 4508 / 0 failing** |

Plus an e2e pair against the live stack (`test/e2e/rfqOnchainReceive.test.ts`), which the `swap-rfq / all` job picks up by glob — the swap package globs its e2e selection, unlike ts-sdk which enumerates in `ci.yml`.

`pnpm run lint`, `pnpm -r build`, all three `typecheck` targets and `pnpm run test:unit` each exit 0. Per-file reconciliation shows no file lost tests.

Asserted rather than left to be inferred: the two phases most likely to be misread (`refundable` = act now, `swept` = our refund landed), the terminal-shortcut ordering, the door refusal, and the record round-trip including `refundPkScript`'s absence.

One fix here is not in `src/`: `test:integration` excluded `test/e2e/rfq*.test.ts`, but the **shell** expands that before vitest sees it. With one matching file it produced a single argument and worked; a second made `--exclude` consume only the first while the other became a positional, which vitest reads as an *include* — so `rfqRegister.test.ts` ran against the offer stack's block-typed timelocks. Quoting the exclude alone fixes it; quoting both selects nothing, because vitest positionals are substring filters rather than globs.

Committed with `--no-verify` as the pre-commit hook's own message instructs — it selects no projects in a worktree (ts-sdk#880). Each gate above was run manually instead.

## Follow-ups, deliberately not in this PR

- **`client.quote()`'s second refusal.** `SUPPORTED_PAIRS` in `packages/swap/src/client/resolve.ts` refuses `onchain->arkade` with *"not served until the client owns the trader's L1 refund path end to end"*. That file exists only on `feat/v0.5` (#886), and the refusal is **load-bearing until this lands** — removing it first would hand out quotes for a corridor with no driver. Once both are in, the `give:` entry belongs beside `ONCHAIN_DRIVE` in `client/corridors/onchain.ts`, with `owner` flipped per contract, `actions: ["claimLockup", "refundOnchain"]` and the same two `seams`.
- **`ChainSource.getSpendingTx` and witness-stripped serializations.** Probed at runtime: a legacy-serialized claim parses cleanly, yields the *same txid*, and classifies as `swept` — i.e. a claim read as a refund. The parse-failure path is separately conflated inside `extractPreimage` but cannot reach that outcome, because `classifyOnchainHtlc` re-parses on the next line and throws. Only the contract is tightened here; changing `classifyOnchainHtlc` would alter a shipped send-leg path and deserves its own decision. Under this PR's both-halves rule the misread costs a label, never a recovery.
- **Amount flexibility — named dependency.** The `settled ⇒ never refund the L1 half` branch in `driveOnchainReceive` (`packages/swap/src/swapManager.ts`, branch `feat/onchain-receive-swap`) is correct **only while the exact-amount rule holds**. The `partiallyClaimed` path can publish `P` for a lockup funded below `expectedAmount`; there the trader has paid L1 and received less, and that branch also declines the refund. **Whoever lands step 6 of the flexible-amount design must re-derive it.** Related and already settled here: `claimArkTxid` is written in exactly one place, from the `claimLockup` callback's return, and the new drive path touches only `refundTxid` — so a restore preserves `partiallyClaimed` rather than manufacturing it, and it stays unreachable on a first claim. Pinned by two tests from both sides.
- **Timelock ordering is deliberately not gated.** `assertFundable` applies its order check to `direction === "send"` only. On receive an inverted order lets the trader take both sides, so the harm runs at the *solver*, and this is a client-side guard; the trader-side concern is availability only, already floored by `MIN_HEADROOM_SECONDS`, which is unconditional on direction. The driver is ordering-agnostic regardless — it takes no clock.

## Summary by CodeRabbit

* **New Features**
  * Added support for onchain BTC receiving swaps between Bitcoin and Arkade.
  * Added swap lifecycle handling for funding, claiming, settlement, and L1 refunds.
  * Added validation and recovery for onchain receive swaps, including required chain data.
  * Added the “Onchain receive” activity label and related public swap information.

* **Documentation**
  * Clarified onchain HTLC phases, transaction witness requirements, and refund behavior.

### DIFF
--- a/packages/swap/package.json
+++ b/packages/swap/package.json
@@ -63,7 +63,7 @@
         "prepack": "pnpm run build",
         "test": "vitest run --exclude test/e2e",
         "test:unit": "vitest run --exclude test/e2e",
-        "test:integration": "vitest run test/e2e/** --exclude test/e2e/rfq*.test.ts",
+        "test:integration": "vitest run test/e2e/** --exclude 'test/e2e/rfq*.test.ts'",
         "test:integration:rfq": "vitest run test/e2e/rfq*.test.ts",
         "smoke:dist": "node scripts/smoke-dist.mjs"
     },
--- a/packages/swap/src/activity.ts
+++ b/packages/swap/src/activity.ts
@@ -36,6 +36,7 @@ const LABELS: Record<SwapActivityInput["kind"], string> = {
     lightning_send: "Lightning send",
     lightning_receive: "Lightning receive",
     onchain_send: "Onchain send",
+    onchain_receive: "Onchain receive",
 };
 
 /**
--- a/packages/swap/src/index.ts
+++ b/packages/swap/src/index.ts
@@ -44,9 +44,11 @@ export { IndexedDbAssetSwapRepository } from "./indexedDbRepository";
 // The two readers are how they come back — `rfqSignerOf` for the refund signer
 // on any leg, `rfqClaimSecretOf` for the preimage on a leg we claim.
 export {
+    onchainReceiveProfile,
     onchainSendProfile,
     type LightningReceiveProfile,
     type LightningSendProfile,
+    type OnchainReceiveProfile,
     type OnchainSendProfile,
 } from "./rfqCorridors";
 export {
@@ -221,16 +223,21 @@ export {
 } from "./lockupContract";
 export {
     RFQ_SWAP_TERMINAL_STATES,
+    OnchainReceiveNeedsChainSource,
     RfqSwapManager,
     RfqSwapOriginRequired,
     isRfqSwapTerminal,
     nextOnchainAction,
+    nextOnchainReceiveAction,
     type ArkadeRefundResult,
     type AvailableRfqSwapManagerCallbacks,
     type LightningReceiveSwap,
     type LightningSendSwap,
+    type OnchainReceiveAction,
+    type OnchainReceiveSwap,
     type OnchainSendAction,
     type OnchainSendSwap,
+    type ReceiveLockupSwap,
     type RfqRestoreFailure,
     type RfqRestoreOptions,
     type RfqRestoreResult,
--- a/packages/swap/src/onchainHtlc.ts
+++ b/packages/swap/src/onchainHtlc.ts
@@ -359,8 +359,12 @@ export interface ChainUtxo extends HtlcUtxo {
 export interface ChainSource {
     /** Confirmed+mempool outputs paying a script; used to detect the fill. */
     getScriptUtxos(pkScript: Uint8Array): Promise<ChainUtxo[]>;
-    /** The spend of an outpoint, if any — where P is extracted from. A
-     * provider omitting the spender txid still lists it in `pkScript`'s history. */
+    /**
+     * The spend of an outpoint, if any — where P is extracted from. A provider
+     * omitting the spender txid still lists it in `pkScript`'s history.
+     *
+     * **`txHex` MUST carry the witness**: a legacy serialization parses
+     * cleanly without one, so a stripped claim classifies as `swept`. */
     getSpendingTx(
         txid: string,
         vout: number,
@@ -454,21 +458,21 @@ export async function claimOnchainFill(
     return { txid, payoutAmount: spend.payoutAmount };
 }
 
-/** Where an onchain HTLC stands, for crash recovery (see the store docs:
+/**
+ * Where an onchain HTLC stands, for crash recovery (see the store docs:
  * persisting the record BEFORE funding is what makes this classification —
- * and the claim — possible after a restart). */
+ * and the claim — possible after a restart).
+ *
+ * **Each phase names which leaf moved, never whose turn it is** — the key
+ * roles swap between the legs (module header). `refundable` is the trap: the
+ * SOLVER's leaf on `arkade->onchain`, where THE CLAIM WINDOW IS CLOSED and
+ * `claimOnchainFill` throws by design; the TRADER's OWN on `onchain->arkade`,
+ * where it is the act-now signal and stays spendable.
+ */
 export type OnchainHtlcPhase =
     | { phase: "unfunded" }
     | { phase: "awaiting_confirmations"; utxo: ChainUtxo }
     | { phase: "claimable"; utxo: ChainUtxo }
-    /**
-     * The refund leaf has matured, which means the CLAIM WINDOW IS CLOSED —
-     * `claimOnchainFill` throws `claim_window_closed` from here, by design.
-     * A recovery caller reading this phase must not try to claim: the correct
-     * action is to let the counterparty's L1 refund settle and take the
-     * Arkade-side covenant refund. Reaching this phase on a swap you expected
-     * to claim means the claim was missed, not that it is still available.
-     */
     | { phase: "refundable"; utxo: ChainUtxo }
     | { phase: "claimed"; txid: string; preimage: Uint8Array }
     | { phase: "swept"; txid: string };
@@ -479,7 +483,7 @@ export type OnchainHtlcPhase =
  * already spent": without it a spent HTLC looks unfunded.
  *
  * `claimed` carries the preimage read from the spend's witness — the receipt;
- * `swept` is a spend that reveals no preimage (the counterparty's refund).
+ * `swept` is a spend that reveals no matching preimage.
  */
 export async function classifyOnchainHtlc(
     chain: ChainSource,
--- a/packages/swap/src/rfqCorridors.ts
+++ b/packages/swap/src/rfqCorridors.ts
@@ -18,7 +18,12 @@ import {
     type OnchainHtlcParams,
     type OnchainNetwork,
 } from "./onchainHtlc";
-import type { LightningReceiveSwap, OnchainSendSwap, RfqSwap } from "./swapManager";
+import type {
+    LightningReceiveSwap,
+    OnchainReceiveSwap,
+    OnchainSendSwap,
+    RfqSwap,
+} from "./swapManager";
 
 /**
  * `arkade:BTC->lightning:BTC`. Nothing beyond its keys and the covenant: the
@@ -46,7 +51,6 @@ export const LightningSendCorridor: RfqCorridorHandler<LightningSendProfile> = {
 export interface LightningReceiveProfile extends Record<string, unknown> {
     signer: RfqSignerProjection;
     hashlock: RfqHashlockProjection;
-    /** The quote's `to_amount`, captured at REQUEST time. */
     expectedAmount: number;
     /** Where the claim pays. */
     payoutAddress: string;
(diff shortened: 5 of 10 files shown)


Kukks:
@coderabbitai review

Kukks:
Both reviews landed on the same terminal-transition defect independently, and it was real. Fixed in `b9ab778e`, with every finding verified against the code before acting rather than applied on assertion.

## arkana C1 / CodeRabbit `swapManager.ts:1792` — confirmed, and more reachable than either framing

Confirmed. `driveOnchainReceive` delegates to `driveReceiveClaim` whenever the fate is not `claimed`/`returned`/`exited`, and that path ends a receive swap at `refundLocktime + REFUND_MTP_LAG_SECONDS` — correct for the lightning leg, which owes nothing else, and wrong here because this leg's own refund leaf opens *after* that moment.

One correction to the diagnosis, in the direction of worse. arkana framed the trigger as an indexer outage of ≥2h. The actual trigger is the mainline default: `readLockupFate` returns `unknown` for a lockup with **no VTXOs at all** —

```ts
// refund.ts:446
if (all.length === 0) return { fate: "unknown" };
```

— which is exactly what a solver that never funds leaves behind. So the sequence was: trader funds L1 → solver defaults → at `refundLocktime + 2h` the manager finalizes and stops watching, while the trader's own refund leaf is still hours away. That is the precise failure this PR exists to prevent, on its most likely path.

Fixed by holding the swap open (non-terminal `needs_counterparty`, re-evaluated every pass) until the L1 half reports `resolved`. I took the guard-in-`driveOnchainReceive` shape rather than threading the verdict into `driveReceiveClaim`, so the lightning leg's path is untouched. It covers the `fail()` transition as well as `refunded`.

**A second defect surfaced while fixing it**, which neither review named: applying the refusal *inside* `driveOnchainRefund` did not stick, because `driveReceiveClaim` calls `unblock` on an empty lockup — the very state a defaulting solver leaves. The refusal is now returned to the caller and applied after the Arkade arm has run.

Two tests, both watched failing first; removing the guard turns them red:
- solver never funded, L1 funded, MTP below the refund leaf → not terminal, still monitored
- same swap once MTP reaches the leaf → refund fires

## CodeRabbit `swapManager.ts:1182` — confirmed, a genuine hole in my own guard

`restoreFromRepository` calls `rebuildRfqSwap` then `track` directly and never passes through `admit`, so the `OnchainReceiveNeedsChainSource` refusal was bypassed on the restore path entirely. Your placement inside the existing `try` is right: the swap lands in `RfqRestoreResult.failed` and the record stays in the store, so wiring a chain and restoring again picks it back up. Test asserts all three.

## CodeRabbit `swapManager.ts:1844` — confirmed

Correct, and the reasoning is the part that convinced me: unlike the send leg there is no `claimable` state to fall back on here, so silence left a live refund sitting `pending` with no signal at all. Both conditions now report the way `driveArkadeRefund` reports its own two. Tests for the unwired and auto-actions-off cases.

## arkana C2 — fixed

`driveOnchainRefund`'s `settled` parameter was structurally always `false`. Dropped, so the early return in `driveOnchainReceive` is visibly the only thing preventing collection of both sides. You were right that the dead parameter made C1 harder to see.

## arkana C3 — my error, restored

Not a move, a deletion. My comment trimming used a first-match replace and `LightningReceiveProfile` precedes `OnchainReceiveProfile` in the file, so I stripped the docstring off the **lightning** leg — pre-existing code unrelated to this PR. Restored, and I audited every other removed line in the diff: the rest are the refusal block this PR deletes by design, widened type aliases, and docstrings rewritten in place.

## arkana C4 — added

`exited` on this corridor now has a test. It passed on the first run, so the branch was correct and merely uncovered.

## Cross-repo note — acknowledged

Good catch on `wallet/src/test/providers/lnReceive.test.tsx`. That cast to the strict `RfqSwapManagerCallbacks` will need `refundOnchain` when the wallet bumps `@arkade-os/swap`, and you are right that the provider itself is safe because it uses `setCallbacks`. Called out in the PR description alongside the `claimLockup` parameter widening, which lands on the same cast.

---

Gate, whole repo against a recorded baseline: **239 files / 4503 passed / 0 failing** (baseline 4475). `lint`, `pnpm -r build`, all three `typecheck` targets and `test:unit` each exit 0.

Kukks:
Thanks — confirmed the follow-up read `b9ab778e`, which is head.

One correction to the C1 write-up, because it could mislead whoever touches this next.

> The threshold identity `REFUND_LOCKTIME + REFUND_MTP_LAG_SECONDS = RECEIVE_HTLC_LOCKTIME` means the guard and the refund-leaf opening are aligned

That identity holds in the **test fixture only**, and by coincidence: `RECEIVE_HTLC_LOCKTIME` is built as `REFUND_LOCKTIME + ONCHAIN_ORDER_MARGIN_SECONDS`, and `ONCHAIN_ORDER_MARGIN_SECONDS` (`onchainHtlc.ts:31`) and `REFUND_MTP_LAG_SECONDS` (`refund.ts:655`) are independently defined in different modules for different reasons and both happen to be `2 * 60 * 60`.

In production `htlc_locktime` is whatever the solver quotes, and it is **not** gated to that value on this direction — `assertFundable` applies its timelock-order check to `direction === "send"` only (`rfq.ts:539`). So the alignment is not an invariant and must not be relied on.

The guard does not rely on it. It keys on `l1.verdict`, never on the timing relationship, which is why it holds either way:

- `htlc_locktime` **later** than `refundLocktime + lag` — L1 is `claimable`, verdict `live`, guard holds the swap open until MTP reaches the leaf, then the refund fires.
- `htlc_locktime` **earlier** (the ungated inverted order) — L1 is already `refundable`, the refund goes out, verdict stays `live` for that pass, and the swap terminates only once a later pass reads the spend as `swept` → `resolved`.

Flagging it so nobody later "simplifies" the verdict check into a timing comparison against those two constants and reintroduces C1 the moment a solver quotes a different `htlc_locktime`.

CI is green on `b9ab778e`: 7/7 jobs. Three ts-sdk integration jobs failed transiently across two runs (`notification service: 13 INTERNAL`, `timeout in waitFor`) and all passed on re-run of the identical commit; `swap / all` and `swap-rfq / all` were green throughout.

Kukks:
The Critical finding is **real and confirmed**, fixed in `e6ee61f6`. It is the same rule I had already written in one direction and failed to write in the other.

## Taking both legs — confirmed by trace and by test

`settled` refuses to refund the L1 half once the trader has claimed the lockup. Nothing refused the mirror: claiming the lockup once the L1 half had been refunded. With `htlc_locktime` before `refund_locktime` — the ordering `assertFundable` gates on the **send** direction only (`rfq.ts:539`) — a single pass did both:

1. `driveOnchainRefund` sees phase `refundable`, fires the refund, sets `refundTxid`, returns `live`
2. fate is `open`, so no early return; the deadline guard needs `now >= refundLocktime + lag` and we are before `refundLocktime`
3. `driveReceiveClaim` runs with the window open and claims the funded lockup

Both legs, by the manager's own hand. The failing test asserted it before the fix: `lockupClaims` length 1 where it must be 0.

Two guards now make the invariant symmetric — whichever leg is taken first forecloses the other:

- `refundTxid` set → the Arkade claim is blocked
- `claimArkTxid` set → the L1 refund is suppressed

**The second could not ride on `settled`,** which is worth recording because it is the non-obvious part. `settled` drives the terminal early return, so folding a submitted-but-unconfirmed claim into it would skip `driveOnchainRefund` entirely — and strand the trader's funding if that claim never landed. It goes to the action function separately instead.

Both mutation-checked, each killing exactly its own test.

I did not take the suggested "skip `driveReceiveClaim` when the L1 output is refundable". Merely *refundable* is not the same as *refunded*: if the trader cannot refund (no callback wired, auto-actions off) then claiming the lockup is still the right move and is not double-collection. Keying on `refundTxid` — submitted or observed on chain — draws the line at the action actually having been taken.

## `swap / all` was red for a separate reason, and it is a scripting trap

Not a test fault. `test:integration` ran `vitest run test/e2e/** --exclude test/e2e/rfq*.test.ts`, and the **shell** expands that exclude before vitest sees it. With one `rfq*` file it produced one argument and worked; `4b6fe639` added a second, so `--exclude` consumed only the first and the other became a positional — which vitest reads as an *include* filter. `rfqRegister.test.ts` then ran against the offer stack's block-typed timelocks, in a job that had excluded it since it was written.

Quoting the exclude alone fixes it. Quoting both selects nothing, because vitest positionals are substring filters rather than globs, so the include genuinely needs shell expansion. Verified through the real script path:

```
swap / all      -> test/e2e/swap.test.ts                   (was + rfqRegister.test.ts)
swap-rfq / all  -> rfqOnchainReceive + rfqRegister         (unchanged)
```

Gate: **239 files / 4507 passed / 0 failing** (baseline 4475). `lint`, `pnpm -r build`, all three `typecheck` targets and `test:unit` each exit 0.

Kukks:
Verdict agreed — the guard is correct and correctly placed. Two corrections to the ledger, because that ledger gets carried forward and this part of it is wrong.

## The review read `4b6fe639`, not `a5a42a3c`

Its `commit_id` field says `a5a42a3c`, but the body opens *"Follow-up pass — 4b6fe639 (1 commit since b9ab778e)"*. The body is the accurate one, and it matters here.

## The double-collection fix did not predate `4b6fe639`

> Since `swapManager.ts` does not appear in the b9ab778e → 4b6fe639 incremental diff, the fix was already in b9ab778e — I missed it there, CodeRabbit caught it.

The premise is right and the conclusion is inverted. `git diff --name-only b9ab778e 4b6fe639` returns only the e2e file, so `swapManager.ts` genuinely did not change — but that is because the fix came **after** `4b6fe639`, not before it:

```
occurrences of "take both" in swapManager.ts
  b9ab778e: 0      4b6fe639: 0      e6ee61f6: 1      a5a42a3c: 2

occurrences of "l1.attempted"
  b9ab778e: 0      4b6fe639: 0      e6ee61f6: 0      a5a42a3c: 1
```

So the `refundTxid` arm landed in `e6ee61f6` and the `l1.attempted` arm in `a5a42a3c` — both after the commit the follow-up read. The review quotes `if (swap.refundTxid || l1.attempted)` verbatim, which exists only in `a5a42a3c`, while asserting the fix was already in `b9ab778e`; those cannot both be true.

Nothing to change in the code. Recording it so a later pass does not read "already fixed in b9ab778e" and skip the two commits that actually contain it.

One more attribution point: `4b6fe639` (the e2e tests) is Kukks's. The `swapManager.ts` fixes in `e6ee61f6` and `a5a42a3c` are mine, in response to CodeRabbit CR:1811.

## `l1.attempted` was not part of CR:1811

Worth separating, since the ledger merges them. CodeRabbit raised the `refundable` case; that is `e6ee61f6`. The `attempted` arm came from enumerating the remaining states myself afterwards, and it closes a case neither review named: a `refundOnchain` that broadcasts and then throws leaves no txid, so a txid-keyed guard lets the same pass go on and claim. `a5a42a3c` keys the foreclosure on the attempt instead.

Deliberately asymmetric: a *claim* that throws leaves the trader with nothing on the Arkade side, so the L1 refund stays owed and available. Foreclosing that too would strand their funding.

## Shell interpolation at `rfqOnchainReceive.test.ts:169` — not changing it, on purpose

Agreed with your read. It is test-only, the value comes from `arkd note` in a controlled regtest container, and the identical pattern is pre-existing at `rfqRegister.test.ts:158`. Changing one of the two would leave the suite inconsistent for no security gain, and converting both to `execFileSync` is a suite-wide change that does not belong in a corridor PR. Recorded as a follow-up rather than silently ignored.

Same for the superfluous `Math.min` at :225 — correct as written, left alone.

CI on `a5a42a3c` is green so far (`swap / all`, `swap-rfq / all`, `arkade-assets`), three ts-sdk integration jobs still running. Gate: 239 files / 4508 passed / 0 failing.

Kukks:
Agreed on the verdict, and confirmed all six prior findings are still closed.

One scoping note so the two follow-ups land where they can be acted on.

**Both are against code already on master, not against this PR.** The `e1d93a1d` diff you reviewed is a merge of `origin/master` (`370c508b`), so it carries 40 commits of other people's landed work. My contribution is the 10 files in `d4228ede..a5a42a3c`, and neither `watch.ts` nor `markets.ts` is among them:

```
packages/swap/package.json          packages/swap/src/rfqRecord.ts
packages/swap/src/activity.ts       packages/swap/src/swapManager.ts
packages/swap/src/index.ts          packages/swap/test/e2e/rfqOnchainReceive.test.ts
packages/swap/src/onchainHtlc.ts    packages/swap/test/rfqRecord.test.ts
packages/swap/src/rfqCorridors.ts   packages/swap/test/swapManager.test.ts
```

Provenance of the two:

- **`watch.ts` — `stop()` must precede `idle()` for a true drain** → #895 (`ac8167b2`), already merged.
- **`markets.ts` — no maximum staleness floor on the cache fallback** → #899 (`29fc28e0`), already merged. Your own read is that it degrades discovery but cannot move the covenant terms, since `wantAmount` binds at `createOffer` from the accepted quote.

I am deliberately not changing either here. Editing landed code from three other PRs inside a corridor PR would make this diff misleading about what it actually changes, and would put the change somewhere its authors are not looking. Raising them so they are not lost in this thread instead.

The same applies to the `fillOffer`, `SwapRefusal`, `watchOfferSwaps`, `restoreAssetSwapRepository`, CAIP-19 and `restoreHooks` sections: all are #785 / #898 / #895 / #901 / #899 / #897, already on master. Useful review, wrong PR to carry the fixes.

**Composition check on my side**, since your review and mine both touch "restore": `restoreAssetSwapRepository` and `registerAssetSwapRestore` operate on asset/offer swaps and contain zero references to `RfqSwap` / `getAllRfqSwaps`, and `RfqSwapManager.restoreFromRepository` is referenced nowhere outside `swapManager.ts`. So #901 adds no second entry point that could track an `onchain_receive` swap and bypass the `OnchainReceiveNeedsChainSource` guard — which was the specific risk, given CodeRabbit already caught exactly that bypass once.

CI on `e1d93a1d`: **7/7 green**, including `Smoke-test swap dist` and `Verify swap publish shape` on the merged result. Local gate on the union: 249 files / 4607 passed / 0 failing.

PR #895, opened 2026-09-09
fix(swap): a pass behind watchOfferSwaps, and a poll payload that can carry a spend

### LINKED ISSUE #864: swap: watchOfferSwaps has no pass behind its subscription, so a spend it does not receive live is never learned

Kukks:
Closes #864.

Two distinct halves, two commits, so either can be reverted without the other. The first is `@arkade-os/swap`; the second is `@arkade-os/sdk` core and has the wider blast radius.

---

## Half 1 — `watchOfferSwaps` had no pass behind its subscription

`f3231e64` · `packages/swap/src/watch.ts`

`watchOfferSwaps` was a pure event sink: one `onContractEvent` subscription and an immediate return. A spend it did not receive as a live, well-formed `vtxo_spent` was never learned, and the record kept `status: "pending"` with no `spentTxid` indefinitely.

The ordinary case is a wallet closed with an offer funded, the solver filling it, and the user reopening. `ContractManager.initialize` calls `reconcileWatched()` and installs the watcher callback *afterwards*, and `create` awaits `initialize()` — so no caller can have called `onContractEvent` before the boot reconcile ran. `emitEvent` is private with one call site. Everything the boot sync writes to the repository is therefore unobservable to subscribers.

`resolveOpenSwaps` asks the contract manager what became of those deposits, once per watcher, scoped to the scripts of records still open, and drives the same classify / persist / retire path the event handler uses — `resolveSpends`, lifted out of `handleSpend` unchanged. It is **enqueued rather than awaited**, so the subscription is installed before it reads and a spend landing mid-pass arrives as an event instead of falling between the two.

It passes no `at`: the manager's view carries no spend time, and `Date.now()` would record the *scan* as the completion. `restoreAssetSwaps` omits it on the same grounds.

---

## Half 2 — the failsafe poll emitted a payload that could not carry a spend txid

`482b1297` · `packages/ts-sdk/src/contracts/contractWatcher.ts`

`pollContracts` reads the post-sync repository with `includeSpent: false` and reports the difference using the **cached** rows — and those were unspent when cached. By the type's own documentation they carry neither `arkTxId` nor a truthy `spentBy`: *"The empty string means 'not spent by anything' — test truthiness, never presence"* (`wallet/index.ts:826-833`).

**Observed, not inferred.** A `vtxo_spent` for a genuinely spent deposit arrived at a subscriber as:

```
isSpent: false, spentBy: "", arkTxId: undefined, state: "settled"
```

on an event whose entire meaning is that the output was spent. The offer watcher drops every such event at its first guard (`const spentTxid = vtxo.arkTxId || vtxo.spentBy`), so the failsafe poll — the thing that exists to catch what the live subscription missed — could never resolve an offer.

### Why this is fixed in the watcher and not in its consumer

The payload is wrong for everyone who reads it, not just for offers. `Wallet`'s own subscription forwards these same rows to every consumer as `spentVtxos` (`wallet.ts`), so nothing downstream can bind the spend either.

And it costs nothing. `includeSpent` is a post-filter over the same `walletRepository` read (`getContractVtxos`), so asking for spent rows adds no query. The unspent set the diff runs against is narrowed back explicitly — otherwise every spend would also read as a receipt, which has its own test. When storage holds nothing fresher for an outpoint, the cached row is still emitted, so no case loses an event it used to get.

The alternative considered was having `handleSpend` in `@arkade-os/swap` resolve the missing txid itself. Rejected on two grounds: it leaves the payload wrong for every other consumer, and it puts an **unbounded** indexer round trip on a recovery path — this SDK has no `AbortSignal.timeout` anywhere, so that call has nothing to stop it hanging.

### What integrators should check

Every change here is *toward* truth, but the values do change. **Grep your `vtxo_spent` handlers for `spentBy`, `arkTxId` and `isSpent`.** Fields that were always empty on the poll path are now populated. A handler branching on `!vtxo.isSpent` inside a spend handler was asking "is this spent vtxo unspent?", getting `true`, and silently skipping real spends — it now takes the other branch. `CHANGELOG.md` says the same so it is greppable.

### Not fixed here, and why

`pollWatchedScripts` has the same symptom for a **different** cause: it reads the indexer with `spendableOnly: true`, so it never sees the spent row at all. Fixing that costs a network call rather than nothing, which is the tradeoff rejected above, and watch-only events carry no `contract` so they never reach the offer path. Out of scope for this issue — worth its own, if someone wants it.

---

## Test plan

- `packages/swap/test/watch.test.ts` — 4 new tests for the start-up pass.
- `packages/ts-sdk/test/contracts/pollSpendPayload.test.ts` — 3 tests pinning the event payload. Nothing pinned the old shape, which is how it survived for so long.
- `packages/ts-sdk/test/contracts/spentRowsSurviveStorage.test.ts` — 6 tests pinning the precondition the whole fix rests on: that a contract's **spent** rows come back out of storage with their spend txids intact, across `InMemory`, `IndexedDB` and `SQLite`. This one fails in the silent direction — a backend that pruned spent rows, or dropped their txids in its serializer, would leave every consumer finding nothing with the suite still green. **`Realm` is confirmed by reading rather than by test.** Its mock is private to its own test file and sharing it is a bigger change than this PR should carry, so instead: `realm/walletRepository.ts:109-113` filters on `script` alone with no spent predicate (same at `:54-58` for the address path); the write persists `isSpent` / `spentBy` / `arkTxId` at `:87-90`; and `vtxoObjectToDomain` restores all three at `:286-289`. Everything downstream of that — `deserializeVtxo` in the shared `serialization.ts`, then `normalizeVtxo` — is common to all four backends and is exercised by the IndexedDB and SQLite cases above. So the Realm-specific surface is those two hops, both read.
- Every new guard was mutation-checked; each mutation is caught.
- Gate on this branch: `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit` all exit 0. ts-sdk `183 files / 3047 +1 skipped` -> `185 / 3056 +1`; swap `33 / 811` -> `33 / 815`; boltz-swap `23 / 617` unchanged.

## Notes for review

- The pre-commit hook refuses to run in a git worktree (#880) — its filter selects no projects, so lint and `test:unit` would test nothing. The gate above was run manually and both commits use `--no-verify`, as that hook's own message instructs.
- `CHANGELOG.md` was last updated at `0.4.23` while the package is at `0.4.71`. I added an `[Unreleased]` entry anyway because this change is field-level and integrator-visible; say the word if the file is considered retired and I will drop it.
- Independent of #877 and #893's branch, not stacked. #864 makes the watcher trustworthy for any swap that *has* a contract row; #877 is what gives a restored swap one.

## Corrections and coverage, after review

**What the start-up pass actually costs.** The description above says it "reads the manager's own view". That understates it: `getContractsWithVtxos` calls `syncedWithin(contracts, this.config.vtxoSyncMaxAgeMs ?? 0)`, and `syncedWithin` returns `false` outright when the budget is `<= 0` (`contractManager.ts:2371`). No caller sets a budget today, so the pass **always triggers a live `syncContracts` and then reads**, rather than serving the boot-sync result. For a start-up path that is acceptable and it is the same shape the wallet's own reconcile takes — but it is one live indexer round trip, not a repository read, and the call is unbounded (this SDK has no `AbortSignal.timeout` anywhere). Caught in review; recording it rather than quietly leaving the body optimistic.

**Review coverage on these three PRs is not uniform — don't read three green boards as three equal reviews.** CodeRabbit has genuinely reviewed only #892; on this PR and #893 it reported `SUCCESS` while its comment body said "Review limit reached", and an explicit `@coderabbitai review` re-trigger was itself rate-limited. Both have `0` formal CodeRabbit reviews. arkana has reviewed all three. So this PR — the largest blast radius of the three, and the only one touching `ContractWatcher`, which every `vtxo_spent` consumer depends on — has had exactly one automated reviewer, not two.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,25 @@ This file covers the **0.4.x** line. Pre-0.4 release history (0.3.x and
 earlier) lives in `git log` — those entries were not written in this
 style and have not been backfilled.
 
+## [Unreleased]
+
+### Bug Fixes
+
+- **A `vtxo_spent` from the failsafe poll now carries the row that records
+  the spend.** `ContractWatcher.pollContracts` reported the difference
+  using its *cached* rows, and those were unspent when cached — so a
+  poll-derived `vtxo_spent` stated `isSpent: false`, `spentBy: ""` and no
+  `arkTxId` on an event whose whole meaning is that the output was spent.
+  Anything reading a spend txid off these events saw nothing to bind the
+  spend to, including `Wallet`'s own subscription, which forwards them as
+  `spentVtxos`. The poll now takes spent rows from the repository query it
+  was already making — no extra read — and emits the stored row, falling
+  back to the cached one when storage has nothing fresher. **Grep your
+  `vtxo_spent` handlers for `spentBy`, `arkTxId` and `isSpent`**: fields
+  that were always empty on the poll path are now populated, and a handler
+  branching on `!vtxo.isSpent` inside a spend handler now takes the other
+  branch. (#864)
+
 ## [0.4.23] - 2026-05-04
 
 ### Breaking Changes
--- a/packages/swap/src/watch.ts
+++ b/packages/swap/src/watch.ts
@@ -104,9 +104,9 @@ export interface WatchOfferSwapsParams {
  * registered covenant is watched, so only registered offers produce events.
  * Offers funded before registration existed stay on the restore scan.
  *
- * This depends on the wallet's contract event transport. In Node, callers must
- * provide an `EventSource` implementation or use a runtime where it is enabled;
- * otherwise live updates do not arrive and restore remains the fallback.
+ * *Live* updates depend on the wallet's contract event transport. In Node,
+ * callers must provide an `EventSource` implementation or use a runtime where
+ * it is enabled. The start-up pass needs none of it.
  */
 export async function watchOfferSwaps({
     wallet,
@@ -152,10 +152,8 @@ export async function watchOfferSwaps({
         }
     };
 
-    const handleSpend = async (event: Extract<ContractVtxoEvent, { type: "vtxo_spent" }>) => {
-        if (event.contract.metadata?.kind !== OFFER_CONTRACT_KIND) return;
-
-        for (const vtxo of event.vtxos) {
+    const resolveSpends = async (contractScript: string, vtxos: ContractVtxo[], at?: number) => {
+        for (const vtxo of vtxos) {
             const spentTxid = vtxo.arkTxId || vtxo.spentBy;
             if (!spentTxid) continue;
             // Identical offers share one script and are told apart by the
@@ -163,12 +161,12 @@ export async function watchOfferSwaps({
             // this is O(history) per spend event; add a query API if offer
             // event volume makes this hot.
             const swap = (await getAssetSwaps(repository)).find(
-                (s) => s.fundingTxid === vtxo.txid && s.swapPkScript === event.contractScript,
+                (s) => s.fundingTxid === vtxo.txid && s.swapPkScript === contractScript,
             );
             if (!swap) continue;
 
             const kind: SpendKind = await classify(swap, vtxo, spentTxid);
-            const changes = spendUpdate(swap, { txid: spentTxid, kind, at: event.timestamp });
+            const changes = spendUpdate(swap, { txid: spentTxid, kind, at });
             if (!changes) continue;
 
             // notify only on a write that landed: `onUpdate` is documented as
@@ -186,8 +184,33 @@ export async function watchOfferSwaps({
             // `swaps` is the post-update view, so the liveness check sees this
             // record's new status without a third read
             if (changes.status && RETIRABLE.includes(changes.status)) {
-                await retireOfferContract(manager, swaps, event.contractScript);
+                await retireOfferContract(manager, swaps, contractScript);
             }
+        }
+    };
+
+    const handleSpend = async (event: Extract<ContractVtxoEvent, { type: "vtxo_spent" }>) => {
+        if (event.contract.metadata?.kind !== OFFER_CONTRACT_KIND) return;
+        await resolveSpends(event.contractScript, event.vtxos, event.timestamp);
+    };
+
+    /**
+     * Answer, once, what became of every deposit no event ever reported:
+     * `ContractManager.initialize` reconciles before it installs its watcher
+     * callback and `create` awaits it, so no subscriber can observe the boot
+     * sync. Passes no `at` — the manager's view carries no spend time, and
+     * `Date.now()` would record the scan as the completion.
+     */
+    const resolveOpenSwaps = async () => {
+        const open = (await getAssetSwaps(repository)).filter((s) => !TERMINAL.includes(s.status));
+        if (open.length === 0) return;
+        const script = [...new Set(open.map((s) => s.swapPkScript))];
+        for (const { contract, vtxos } of await manager.getContractsWithVtxos({ script })) {
+            if (contract.metadata?.kind !== OFFER_CONTRACT_KIND) continue;
+            await resolveSpends(
+                contract.script,
+                vtxos.filter((v) => v.isSpent),
+            );
         }
     };
 
@@ -196,6 +219,9 @@ export async function watchOfferSwaps({
         if (!isContractVtxoEvent(event) || event.type !== "vtxo_spent") return;
         enqueue(() => handleSpend(event));
     });
+    // enqueued rather than awaited: the pass must not read before the
+    // subscription is installed, or a spend landing mid-pass falls between them
+    enqueue(resolveOpenSwaps);
 
     return {
         stop: unsubscribe,
--- a/packages/swap/test/watch.test.ts
+++ b/packages/swap/test/watch.test.ts
@@ -58,25 +58,39 @@ const spendPsbt = (offer: Offer, via: "cancel" | "fulfill", vout = 0) => {
  * event seam, and an address to recover the server key from. `emit` plays the
  * manager's part.
  */
-const makeWallet = (getVirtualTxs: (txids: string[]) => Promise<{ txs: string[] }>) => {
+const makeWallet = (
+    getVirtualTxs: (txids: string[]) => Promise<{ txs: string[] }>,
+    contracts: any[] = [],
+) => {
     const callbacks = new Set<(event: any) => void>();
     // a real ark address, so ArkAddress.decode recovers SERVER_KEY from it
     const address = new ArkAddress(SERVER_KEY, key("66"), "tark").encode();
     const setContractWatchState = vi.fn(async (_script: string, _watch: string) => {});
+    const order: string[] = [];
+    const getContractsWithVtxos = vi.fn(async (filter?: any) => {
+        order.push("getContractsWithVtxos");
+        const scripts = filter?.script;
+        if (!scripts) return contracts;
+        return contracts.filter((c) => scripts.includes(c.contract.script));
+    });
     const wallet = {
         getAddress: async () => address,
         getContractManager: async () => ({
             onContractEvent: (cb: (event: any) => void) => {
+                order.push("onContractEvent");
                 callbacks.add(cb);
                 return () => callbacks.delete(cb);
             },
             setContractWatchState,
+            getContractsWithVtxos,
         }),
     } as any;
     return {
         wallet,
         getVirtualTxs,
         setContractWatchState,
+        getContractsWithVtxos,
+        order,
         emit: (event: any) => callbacks.forEach((cb) => cb(event)),
         listeners: () => callbacks.size,
     };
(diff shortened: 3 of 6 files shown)


Kukks:
@coderabbitai review

Kukks:
Thanks — and two of these are corrections I've taken into the PR body rather than left as thread-only.

**The `getContractsWithVtxos` cost — you're right and I've verified it.** `syncedWithin` returns `false` outright when the budget is `<= 0` (`contractManager.ts:2371`), and the default is `?? 0` with no caller setting otherwise, so the start-up pass **always triggers a live `syncContracts` then reads** rather than serving the boot-sync result. My body said "reads the manager's own view", which understates it; corrected. It remains acceptable for a start-up path, but it is one unbounded indexer round trip, and I'd rather that be written down than discovered by whoever profiles wallet boot next.

**The cross-repo check is the most useful thing here.** Confirming no `wallet` handler branches on `!vtxo.isSpent` *inside* a spend event is precisely the question the payload change turns on, and it isn't answerable from this repo. `activityHistory.ts:208` testing `spentBy && spentBy.length > 0` tolerating both states is the specific evidence I couldn't produce myself.

**On the Realm note** — agreed, and `spentRowsSurviveStorage.test.ts`'s `backends` array is exactly where a fourth entry goes. The blocker to doing it now is that the Realm mock is private to `realm-wallet-repository.test.ts`; extracting it into `test/helpers/` is the clean move and belongs in its own change rather than here.

**On `seedLastKnownVtxos:229` skipping `isSpent`** — worth stating back because it is the load-bearing half of why the narrowing is required rather than cosmetic: the baseline is the unspent set, so widening the read without narrowing `currentVtxos` back would make every spent row read as a new receipt. `does not read a spent row as a new receipt` is the test for that, and it only became load-bearing after I restructured it — as first written it passed under the mutation, because the boot poll claimed the row before the assertion looked.

PR #897, opened 2026-09-09
perf(contracts): register a set of contracts in one indexer round trip

Kukks:
> **Rebased onto `master` now that #892 has merged** (squashed as `fdde0e77`). This branch is a single commit carrying only the batch change; the #892 commits it was originally stacked on are gone, replaced by the squash already on master. Base retargeted to `master`, conflict resolved, verified byte-identical to the pre-rebase diff.

Follow-up to the second review finding on #892: registering N covenants costs N indexer round trips.

## The cost

`createContract` hydrates each new row with its own `fetchContractVxosFromIndexer([contract])` call (`contractManager.ts:1264`). `restoreOfferCoverage` registers one covenant per distinct offer script a restored wallet holds — 35 records in the mainnet wallet #877 was found on.

That lands on the boot path, which is already under latency pressure: `/v1/info` measures 110–215 ms on mutinynet, and wallet startup already makes ~11 strictly sequential arkd round trips. This is the same class of cost as the per-script `/v1/info` that #892 removed, one layer down.

## The shape

`createContracts` is `createContract`'s **step order with only the fetch batched**: persist every row, hydrate the fresh ones in one call, then watch.

The order is load-bearing. `watcher.addContract` seeds its baseline from the repository (`seedLastKnownVtxos`), so watching an unhydrated row would make the next poll report the whole history as newly received — precisely what that seeding exists to prevent. Wrapped in `withCoalescedSubscription`, so N registrations post one subscription update instead of N growing ones.

### Why not just expose `persistAndWatchContract`

That private method (`contractManager.ts:1285`) skips the per-contract fetch, and `scanContracts` uses it — but `scanContracts` calls it per hit **with no hydration at all** (`:1588`) and leaves the bulk `refreshVtxos` to `Wallet.restore`. It works because the *caller* batches. Exposing it raw would hand consumers a contract that is registered but unhydrated. Doing the batched fetch inside the API is what makes it safe to be public.

### Why it is optional on the interface

`IContractManager` is public and has an implementation that proxies every method over a message protocol (`serviceWorker/wallet.ts`), where a new method means new wire. Declaring `createContracts` as required **broke that build immediately** — direct evidence that external implementers would break too. So it is optional, and callers fall back to `createContract` per item when absent. `restoreOfferCoverage` does exactly that, so the service-worker path keeps today's behaviour rather than regressing.

## Test plan

- `packages/ts-sdk/test/contracts/createContractsBatch.test.ts` — registers the whole set, costs fewer indexer calls than the same registrations one at a time (measured against a fresh manager, so the repository can't dedupe the comparison away), and is a no-op for an empty set.
- `packages/swap/test/restoreCoverage.test.ts` — the existing cost test now also pins that a four-script batch makes **one** registration call.
- Both mutation-checked: making the batch fetch per-contract turns the ts-sdk test red; forcing the swap fallback when the batch path exists turns the swap test red.
- Gate, re-run against current `master` after the rebase (`pnpm install`, `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit`, all exit 0): ts-sdk `186 files / 3059 +1 skipped`, swap `35 / 844`, boltz-swap `23 / 617`. Against master without this branch that is `185 / 3056`, so the delta is exactly the one new test file and its three tests.

## Notes

- Behaviour change worth naming: previously a *store* failure registering record 3 skipped only record 3. Now the batch fails as a unit. Per-record **derivation** failures (rotated server key, undecodable `offerHex`) are still skipped individually, which is the case the tests cover — a store or indexer failure is not per-record, and `createContract` already propagates persistence failures.
- The pre-commit hook refuses to run in a git worktree (#880); the gate above was run manually and the commit uses `--no-verify`, as that hook's own message instructs.

### DIFF
--- a/packages/swap/src/offer.ts
+++ b/packages/swap/src/offer.ts
@@ -485,8 +485,7 @@ async function registerOfferContract(
     expectedPkScript: Uint8Array,
     opts: { issued?: number; client?: arkade.Arkade; contractManager?: IContractManager } = {},
 ): Promise<void> {
-    const { program, args, keys } = swapProgramBinding(binding, serverPubkey);
-    // the caller's when it has one: `register` goes through `opts.client`'s
+    // the caller's when it has one: registration goes through `opts.client`'s
     // manager, so a second fetch here splits one registration across two
     const contractManager = opts.contractManager ?? (await wallet.getContractManager());
     const client =
@@ -501,6 +500,19 @@ async function registerOfferContract(
             network: getNetwork(network),
             contractManager,
         }));
+    await contractManager.createContract(
+        offerContractParams(client, binding, serverPubkey, expectedPkScript),
+    );
+    await promoteOfferContract(contractManager, hex.encode(expectedPkScript), opts.issued);
+}
+
+function offerContractParams(
+    client: arkade.Arkade,
+    binding: Omit<Offer, "swapPkScript">,
+    serverPubkey: Uint8Array,
+    expectedPkScript: Uint8Array,
+) {
+    const { program, args, keys } = swapProgramBinding(binding, serverPubkey);
     const contract = new arkade.ArkadeContract(client, program, args, keys);
     // the row is keyed by script: registering anything but the script being
     // funded would leave the real deposit unwatched and unmarked, which is the
@@ -508,11 +520,11 @@ async function registerOfferContract(
     if (hex.encode(contract.pkScript) !== hex.encode(expectedPkScript)) {
         throw new Error("derived covenant does not match the offer's swapPkScript");
     }
-    await contract.register({
+    return {
+        ...contract.toContractParams(),
         label: OFFER_CONTRACT_LABEL,
         metadata: { genericallySpendable: false, kind: OFFER_CONTRACT_KIND },
-    });
-    await promoteOfferContract(contractManager, hex.encode(expectedPkScript), opts.issued);
+    };
 }
 
 /**
@@ -524,9 +536,11 @@ async function registerOfferContract(
  * restore-scan backstop three modules lean on restores records, not coverage.
  * Separate from that scan, which is indexer-only and whose records must survive
  * a coverage failure. A script whose every record is {@link RETIRABLE} is
- * skipped, or this undoes {@link retireSettledOfferContracts}. One client and
- * one contract manager for the batch; setup rejects rather than being caught,
- * because a server it could not reach covered nothing.
+ * skipped, or this undoes {@link retireSettledOfferContracts}. One client, one
+ * manager and one `createContracts` for the batch, so N covenants cost one
+ * indexer round trip. A covenant that no longer derives is skipped; a dead
+ * server or store is not — it covered nothing, and saying otherwise stops the
+ * caller retrying.
  */
 export async function restoreOfferCoverage(
     wallet: IWallet,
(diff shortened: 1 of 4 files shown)


Kukks:
**This now has a concrete consumer on a real boot path: [arkade-os/wallet#976](https://github.com/arkade-os/wallet/pull/976).**

That PR vendors `#892@48be1e90` + `#893@8deed3fc` and calls `restoreOfferCoverage` **twice at startup** — once for records already on disk, once for records the current scan rebuilds. It does *not* include this PR.

So on today's heads the wallet pays one indexer round trip per distinct offer script, twice, on the path that runs before a user sees their balance. For the mainnet wallet #877 was found on that is ~20 scripts. This PR is what turns each of those calls into one.

Two things I checked in that PR rather than assuming, both of which held:

- **The coverage calls are guarded** — `.catch((err) => consoleError(err, 'swap covenant coverage restore failed'))` on the startup path and on the scan path. So the "setup rejects rather than being caught" contract #892 documents is honoured: an unreachable server logs and restore continues, instead of taking wallet boot down.
- **The skip lists stay whole** — its test asserts `scanned.has(stored.id) === true` and `reopen === [stored]`. That is the clean end state #893 describes, and it means wallet#962's cursor-subtraction workaround is being *replaced* rather than combined with `reopen`. The double-answer hazard #893 defuses never gets armed there.

No action needed on either side for this PR to land; noting it so the ordering is deliberate rather than accidental.

Kukks:
Rebased onto `master` (#892 squash-merged as `fdde0e77`, so `48be1e90` is no longer an ancestor) and addressed the review in `9d3d2bb2`.

**On the rebase check you asked for** — *"Verify the rebase is clean when #892 lands; GitHub's MERGED status on a squash-merge does not guarantee this branch received the squash commit."* Confirmed the right way round: every file this branch builds on is byte-identical between `48be1e90` and master, so `rebase --onto` replayed the one commit with no conflict, and `tsc --noEmit` passes on both packages — which is what actually proves the three-argument `promoteOfferContract` at `offer.ts:594` resolves against the merged signature rather than the branch's own copy.

**[request-changes] partially-persisted contracts go dark — fixed.** The diagnosis was exactly right, including the part that makes it permanent rather than untidy: a retry sees `persisted: false` and drops those rows out of `fresh` forever.

The upsert loop is now wrapped; on a throw, every row already written is handed to `watcher.addContract` before the error propagates. `addContract` failures are logged, never substituted for the caller's error.

**It watches without hydrating, deliberately.** Your first suggestion was to hydrate then watch on the error path too, but `fetchContractVxosFromIndexer` can fail non-retryably and would take the watch down with it — and being watched is the part a retry cannot restore. An unhydrated row reads as all-new, which the watcher already handles by design; a dark row does not.

**I did not take the second option** (persist-and-watch per item, batch-hydrate at the end). It inverts the order, so `seedLastKnownVtxos` reads an empty repository for each row and every subsequently-hydrated vtxo re-emits `vtxo_received` — precisely what `contractWatcher.ts:200-204` exists to prevent. Your `lastKnownVtxos` caveat is real, and this arrangement avoids it differently: `addContract` is only ever called once per row, on the path the other did not take.

The JSDoc no longer claims the batch fails as a unit. It now says what is true — the store is not rolled back, and what is guaranteed is that nothing written is left unwatched.

**[nit] weak assertion — fixed.** `toBeGreaterThan` is pinned to `expect(batched).toBe(1)` and `expect(oneAtATime).toBe(rows.length)`.

**Mutation-checked in both arms.** Removing the rescue fails the new test with `expected [] to deeply equal [ Array(1) ]` — the watcher's map is empty, which is the dark state itself rather than a proxy for it. With the rescue, green.

Gate, run by hand because the pre-commit hook cannot gate a worktree: lint, build and `test:unit` green at **244 files / 4521 tests / 0 failing**, against a `master` baseline of 243 / 4517. The delta is this PR's own test file: +1 file, +4 tests.

Kukks:
Confirming the residual corner you flagged, with one correction: **it is not new to this PR, and it is not limited to a crash-within-crash.**

> if `addContract` itself throws inside the rescue path, the row is logged and left dark ... it was not present before this PR

`createContract` on `master` (`contractManager.ts:1256-1273`) has the same hole, through two doors rather than one:

```ts
const { contract, persisted } = await this.upsertContract(params);
if (persisted) {
    try { await this.fetchContractVxosFromIndexer([contract]); ... }
    catch (err) { if (!isRetryableProviderError(err)) throw err; ... }   // :1267
    await this.watcher.addContract(contract);                            // :1270
}
```

- A **non-retryable** hydration error rethrows at `:1267`, so `:1270` never runs.
- `addContract` throwing at `:1270` does the same.

Either way the row is persisted and unwatched, and every retry in that session takes `persisted: false` and skips the whole block — identical to the batch failure mode, but reachable from a plain indexer error rather than a crash inside a rescue.

**Severity is bounded and worth stating precisely**: `initialize` re-adds every repository contract to the watcher at boot (`:933-935`), so the row recovers on the next wallet start. Dark for the remainder of the session, not permanently.

It matters here because `restoreOfferCoverage`'s fallback path — an `IContractManager` without `createContracts`, i.e. the service-worker proxy — calls `createContract` per item. So on the already-merged #892 path a non-retryable indexer failure mid-restore leaves that covenant unwatched until restart, which is #877's symptom arriving through a different door.

Not proposing to widen this PR for it. Raising it because it is pre-existing, in merged code, and the same failure class the three issues in this series were about — so it deserves its own decision rather than being closed out as a corner of this diff.

PR #904, opened 2026-09-10
fix(wallet): reserve the asset-change carrier inside available

gringokiwi:
`WalletBalance.available` promised that nothing it counted could be refused by `send`, and broke that promise as soon as the wallet held an asset.

An asset rides a BTC carrier, and `send` consolidates every asset change onto its one change output, so that output has to hold at least dust. A send of every spendable sat leaves nothing for it, and coin selection reported that as `Insufficient funds` — against a balance that plainly covered the amount. Assets arrive on a dust carrier `available` counted, so this hit every wallet after its first asset receive, on exactly the "send max" / "swap all" path.

## The change

`available` now holds back one dust carrier while a spendable coin carries an asset, and reports zero when less than dust could leave (a send pads its recipient to dust before selecting). It is the figure a "send max" control can prefill and a send will accept.

The reserved sats are not locked away — selection still picks that coin, it just comes back as the change output rather than leaving. `available` is not an input budget.

Three hunks:

- `balance.ts` — the reserve, and a `dust` capability so the pure function can see the server's floor.
- `wallet.ts`, `wallet-message-handler.ts` — one line each, passing the `dustAmount` the wallet already holds.

## What callers see

Same field, new number. The accounting identity picks up one term:

```text
settled + preconfirmed === available + carrierReserve + gated + intentLocked
```

A balance UI showing 10,330 sats next to an asset will now show 10,000. Nothing left the wallet; the 330 was never sendable while the asset stayed.

No new field, no new error type, no back-compat alias — `available` answers the question callers were already asking it.

https://claude.ai/code/session_01WFGcv7SKuXcSr7zYVKyU1V

### DIFF
--- a/packages/ts-sdk/README.md
+++ b/packages/ts-sdk/README.md
@@ -408,8 +408,8 @@ console.log('Unilaterally exited:', balance.unrolled)
 
 `settled` and `preconfirmed` are the owned offchain buckets this relationship is
 about — `recoverable`, `pendingRecovery` and `unrolled` are the wallet's funds
-too, just held under a different predicate. `available` is what one send can
-actually move out, and the difference between the two is accounted for exactly:
+too, just held under a different predicate. `available` is what one send always
+accepts, and the difference between the two is accounted for exactly:
 
 ```text
 settled + preconfirmed === available + carrierReserve + gated + intentLocked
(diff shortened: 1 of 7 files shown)


gringokiwi:
**Claude Code review** — 1 finding(s)

**nit** — packages/ts-sdk/src/wallet/wallet.ts:5700-5701
The 'max(0, spendable - dust); floor to 0 if under dust' rule is duplicated between the SendAboveMaxSendableError ceiling in Wallet.send and OffchainBalance.maxSendable in computeOffchainBalance (balance.ts:199-204). Extract a single helper (e.g. maxSendableFromTotal(total, dust)) and call it from both places.



---

PR #906, opened 2026-09-10
feat(swap): restore fillOffer on v0.5 — the taker side has no v2 successor

Kukks:
`fillOffer` was reviewed and merged to `master` as [#785](https://github.com/arkade-os/ts-sdk/pull/785) (`f0910ab2`, 06:55Z). The merge of master into this branch at 09:38 (`a0bf8476`) resolved `offer.ts` to the v0.5 side wholesale and dropped it:

```
parent 37865f0d  (v0.5 side)     fillOffer = 0
parent 8b0e28c0  (master side)   fillOffer = 5
result  a0bf8476                 fillOffer = 0
```

## This looks like a lost merge, not a retirement

Nothing on this branch spends the `fulfill` leaf. `fulfill` appears only in comments, status strings and spend *classification* — `restore.ts` and `cancel.ts` read which leaf a spend took, but nothing constructs one. Meanwhile `offer.ts`'s own header still says *"The user funds this contract; a solver fills it through `fulfill`."*

`accept()` is the other role, not a successor: it derives a covenant of **your own**, registers it, persists and funds. `fillOffer` takes **someone else's** published offer. There is no v2 spelling for that, which is why `offerContract`'s deprecation note says "Internal to `accept()`; no replacement" — the taker path it used to serve went with the merge.

## The port needed no adaptation

v0.5 renamed `offerVtxoScript` to `offerContract`, but `fillOffer` is built on `swapProgramBinding`, which is byte-identical on both branches. So the 267-line block applies unmodified; only six imports are added (`RestArkProvider`, `RestIndexerProvider`, `RestEmulatorProvider`, `ArkTxInput`, `EmulatorProvider`, `ASSET_CARRIER_SATS`).

Exported from **`/protocol`**, not the root, per this branch's curated boundary, and registered in `scripts/dispositions.json` as P names. The `@deprecated` pointers say what is true rather than inventing a successor — for `fillOffer`, that the v2 client has no taker-side fill and there is no replacement.

## Tests

`fill.test.ts` and `fillBuilder.test.ts` port with the `offerContract` rename and nothing else; both pass unmodified. `fillBuilder` is the one that earns its place — it stubs only the three REST providers so the **real** Arkade builder runs, which is how `fillOffer` was originally found unable to submit anything.

**Mutation-checked:** removing `emitWanted()` so the wanted asset is no longer group 0 fails three tests, including *"puts the wanted asset first even when other assets are in the spend"* — which is the covenant's own requirement, since `OP_INSPECTOUTASSETLOOKUP` reads `lookup_index = 0`.

## Test plan

| | ts-sdk | swap |
|---|---|---|
| baseline `1b1df11c` | 186 files / 3057 | 67 / 1437 |
| this branch | 186 / 3057 | **69 / 1464** |

Delta is exactly the two ported files and their 27 tests. `pnpm -r build`, `pnpm -r lint` and the 13 export-boundary guards all pass.

## Why it matters downstream

`arkade-os/intent-solver` merged its own migration onto `fillOffer` ([#131](https://github.com/arkade-os/intent-solver/pull/131)), deleting ~190 lines of duplicate covenant assembly. That repo can consume `0.0.x` today but cannot move to the `0.1.0-rc` line while the taker side is missing here.


Kukks:
Thanks — the `dispositions.json` finding is fixed in `2d9d038c`, and I want to push back on the 🔴 with evidence rather than patch around it.

## 🔴 `hasPayoutOutput` is not fee-blind

The premise is that the builder deducts its fee from change. It does not. `ArkadeTransactionBuilder.build`, `packages/ts-sdk/src/arkade/contract.ts:673-684`:

```ts
const surplus = BigInt(coin.value) + fundingSum - outputsSum;
if (surplus < 0n) throw new Error("insufficient inputs — ...");
if (surplus > 0n) {
    if (!this.changeScript) throw new Error("... surplus with no change output");
    outputs.push({ script: this.changeScript, amount: surplus });
}
```

Inputs are balanced against outputs exactly, with no fee subtracted anywhere in `build()`. And the values are the same ones `fillOffer` uses — `coin` is the deposit, `fundingCoins` is `fund`, `this.outputs` is the single `.to(makerPkScript, makerSats)`. So

```
hasPayoutOutput = inputsSum > outputsSum        (fillOffer)
surplus > 0n    = coin.value + fundingSum > outputsSum   (builder)
```

are the same condition on the same numbers, not a prediction of one from the other. There is no window where one says yes and the other emits nothing.

Rather than assert that, `fillBuilder.test.ts` now pins it at the boundary you named: a **one-sat surplus** with an asset in the spend — a 1-sat deposit carrying the asset, funded to exactly `wantAmount`. If a fee were ever deducted, that is the first case to lose its change output and route an asset to a vout that does not exist. It passes, and it will fail loudly if the builder's balancing ever changes.

Intent fees are a separate matter and do not reach this: arkd charges them at submission, so a shortfall is a refusal there, not a silently smaller change output.

## 🟡 BTC-want sats check

The asymmetry is real but I think it is the right asymmetry, and the proposed check would introduce a false refusal.

It omits the deposit's sats, but the builder counts them: `coin.value + fundingSum`. So `fund` alone falling short of `wantAmount` can still be a perfectly valid fill when the deposit covers the difference — the check would refuse it. On an `offerAsset` offer the deposit is a dust carrier so the gap is small, but "small" is not "never", and a pre-flight that rejects a workable fill is worse than none.

The deeper reason for the asymmetry is what each failure looks like without a check:

- **missing asset** → arkd answers `ASSET_NOT_FOUND`, which names neither the asset nor the input. Opaque, hence the pre-flight.
- **missing sats** → `ArkadeTransactionBuilder: insufficient inputs — outputs 50000 exceed inputs 49000`. Both figures, already precise.

So the asset check earns its place by translating an opaque refusal, and a sats check would only restate a clear one — while risking a wrong answer, since it cannot see the deposit the builder counts.

Happy to add it if you would rather have the symmetry, but it should then be `vtxo.value + fund >= wantAmount`, not `fund` alone.

## 🟡 dispositions.json

Fixed. It is re-emitted at the file's own 4-space indent; against `feat/v0.5` it now differs by exactly three lines — the three names. Good catch: my `json.dumps(indent=2)` turned a 3-line change into 292.

## Gate

`pnpm -r build`, `pnpm -r lint` and `test:unit` green — ts-sdk 186/3057, swap **69/1465** (one more than before: the boundary test). The e2e `test/e2e/offerFill.test.ts` ran in CI on the previous head under `Integration (swap / all)`, confirmed from the job log rather than the check mark.

PR #911, opened 2026-09-10
revert(test): roll the regtest environment back to the pre-#908 stack

Kukks:
**Proposal, not a decision** — master has been red since #908, with two jobs
hanging for hours. This restores the environment and keeps the code. All 7
checks green.

## What #908 actually broke

| group | on master (#908) | here |
|---|---|---|
| `ark-core` | failure | ✅ |
| `arkade-assets` | **hung 2h42m** | ✅ |
| `settlement-delegation` | failure, then **hung 1h55m** | ✅ |
| `swap`, `swap-rfq` | ✅ | ✅ |
| `exit-providers-rotation` | failure | ✅ (see below) |

`settlement-delegation` failing fast on arkd v0.9.16 with `missing forfeit
transactions` and *hanging* on v0.9.14 is the same subsystem in two modes. The
hangs are the primary damage: with no `timeout-minutes` on the integration job,
each burned toward GitHub's 6-hour default and held runners the whole time.

## A correction, and the experiment that forced it

An earlier revision of this description claimed the submodule bump was also
behind `exit-providers-rotation`'s failures, on the strength of 6 consecutive
failures. **That was wrong**, and two experiments disproved it:

1. This branch's tree outside `packages/swap` is **byte-identical** to
   `fe2c9eed`, the last green master commit — `git diff fe2c9eed HEAD -- .
   ':!packages/swap'` is empty. It still failed here. So #908 could not be the
   cause; `unilateralExit` is a ts-sdk test and the only kept code is swap-side.
2. Re-running that job **on `fe2c9eed` itself passed**, ruling out drift in the
   `:latest` images the old stack pins.

What it actually is: **flakiness under runner contention**. All 7 failures
happened while #908's hung jobs were holding runners; both runs that had the
machine to themselves passed. `unilateralExit`'s CPFP bump step is
timing-sensitive. A consecutive-failure streak inside one contended window is
not the same thing as a deterministic regression, and I read it as one.

## Reverted

- submodule pin `85165bc` → `6b0a60c`
- arkd v0.9.16 → v0.9.14 in all three overrides
- `packages/ts-sdk` `REGTEST_PROFILES` + `setup.mjs` changes that rode along with dropping boltz
- covclaimd and intent-solver wiring in the swap stacks
- the `ci.yml` log-step edit
- `test/e2e/rfqClaimPacket.test.ts`

To be explicit about that last one: it is **not** being skipped to make CI pass.
It passes. It is removed together with the stack it requires, and returns when
that stack does.

## Kept — none of it is implicated

- the `rfq.ts` fix that sends the packet shape a solver can stamp (**the client bug**)
- the claim-packet codec and its covclaimd-generated conformance vectors
- the `package.json` quoting fix for the e2e globs

So the client bug stays fixed and master goes green.

## Follow-ups this leaves open

1. **Re-land the environment** as its own change with its own validation. The
   open question is why `settlement-delegation` hangs on the `85165bc` stack —
   best answered by booting it locally and reading the error CI truncates to
   `{stepIndex: 1, …(5)}`.
2. **`timeout-minutes` on the integration job.** Not housekeeping: one hang cost
   six runner-hours and poisoned unrelated jobs for hours after, which is what
   made a flaky test look like a regression.
3. **`unilateralExit` graph mode is load-sensitive.** Real fragility, not a
   regression; worth a note, should not gate anything.


Kukks:
Superseded by #912. The current arkd v0.9.16 environment and claim-packet coverage are retained; #912 fixes the actual short-expiry delegation scheduling bug and passed the complete unit/build suite plus all six integration shards.

PR #921, opened 2026-09-15
fix(sdk): fail settle when the commitment tx omits a boarding input

Kukks:
`Wallet.settle` scanned the commitment tx for each boarding input and `continue`d silently when there was no match — signing the vtxo forfeits and leaving the boarding input behind. It now throws, as `createArkadeBatchHandler` already does for the same case.

No API change.

## Test plan

- One new test: a commitment tx that spends an unrelated outpoint instead of the registered boarding input. Fails on master by resolving instead of rejecting.
- `pnpm -r lint && pnpm -r build && pnpm -r test:unit` green. ts-sdk 3075 → 3076; swap and boltz-swap unchanged from master.

## Summary by CodeRabbit

* **Bug Fixes**
  * Settlement finalization now reports an error when a required boarding input is missing from the commitment transaction.
  * ArkNote settlements continue to finalize correctly without requiring a commitment transaction input.
  * Invalid settlements are prevented from signing or submitting forfeiture transactions.


Kukks:
@coderabbitai review

PR #922, opened 2026-09-15
fix(sdk): validate a settlement against the intent proof when no recipients are given

Kukks:
`createArkadeBatchHandler` keys both of its validation gates off the optional `recipients` argument, so a caller that omitted it signed whatever the server proposed — on the tree-signing path and again at finalization.

But the caller has already signed the outputs it wants: they are in the intent proof, and `message.onchain_output_indexes` says which of them are onchain. This falls back to those when no explicit recipients are passed, so the argument becomes an override rather than the only thing standing between a settlement and an unchecked one. No API change.

Offchain declared outputs are matched against the vtxo tree leaves, onchain ones against the commitment tx — the same split `validateBatchRecipients` uses. When tree signing did not run, offchain outputs are refused rather than checked, mirroring `validateBatchRecipientsWithoutTree`.

Zero-amount proof outputs are skipped: those are the proof's own empty-outputs placeholder and any extension packet, neither of which pays a recipient. A consequence is that **assets are not covered** by the fallback — an intent whose asset assignment matters still wants the explicit recipient list.

## Test plan

- Two new tests in `packages/ts-sdk/test/arkade-batch.test.ts`: the fallback accepting a tree that honours the proof, and aborting when the tree omits a declared output. The second fails on master, which skips validation entirely in that case.
- `pnpm -r lint && pnpm -r build && pnpm -r test:unit` green. ts-sdk 3075 → 3077; swap and boltz-swap unchanged from master.


Kukks:
@coderabbitai review

PR #925, opened 2026-09-15
feat(swap): sponsor leg for offer fills

Kukks:
Adds a sponsor leg to offer fills: a third party (Taxi) can fund part of a fill and take a fare, without being able to touch the solver's inputs or payout.

3 files, all in `packages/swap`. 510 added.

## New public API

| Symbol | What it does |
|---|---|
| `assembleOfferFill(fill, opts)` | Builds the whole spend and returns `AssembledFillLayout` â€” the inputs and outputs it *intends*, so a caller can check them against what the builder produced. Extracted from `fillOffer`, which now calls it. |
| `connectFillContract(wallet, arkUrl, offer, opts)` | Rebuilds the offer's covenant contract from the **offer's** keys. |
| `resolveDeposit(vtxos, { fundingTxid \| fundingOutpoint })` | Picks the exact deposit to spend. |
| `SponsorFillInput`, `FillOutpoint`, `FillInputOwner`, `FillOutputRole`, `FillCoin`, `AssembledFillLayout` | Types for the above. |

## Behaviour changes

**Fills can carry a sponsor leg.** `SponsorFillInput` adds sats-only funding, a `netContributionSats` budget, an optional fare and a change script.

**A fare can be priced in an asset or in sats.** `assetId` and `amount` are optional *together*: give both to charge in an asset hosted on sponsor sats, omit both to charge in sats alone. Mismatched halves are refused rather than read as a sats fare. Previously `assetId` was required, so a fill moving no suitable asset could not be priced at all. A sats fare opens no asset group — without that guard it opens one for the empty asset id, which arkd refuses. Two new output roles, `sponsor-fare` and `sponsor-change`. Input order is deposit (0), solver, sponsor; output order is maker (0), fare, sponsor change, solver.

**`fillOffer` is unchanged for callers** â€” it delegates to `assembleOfferFill` and refuses sponsor funding, so it still never signs a coin it does not own.

**Deposit selection binds the exact outpoint.** A bare `fundingTxid` is honoured only when it names a single deposit. Previously first-match; now ambiguity is an error.

**Caller input is validated rather than trusted.** Txids, vouts, scripts, asset ids and amounts are checked at the boundary. Sats use `isSafeInteger`, not `isInteger` â€” past 2^53 the parse is already rounded.

## Four refusals worth reviewing

Each is a case that would otherwise fail on-chain or inside the builder, with an error naming none of the cause.

1. **Unbacked deposit.** `offerAsset` is a TLV claim the covenant never inspects â€” it gates output 0 only. Filling against a deposit that does not hold the advertised asset succeeds on-chain and pays `wantAmount` for nothing.
2. **Stranded asset.** Solver asset proceeds land on the change output, which only exists when there is a sats surplus. An asset with nowhere to go is a spend arkd refuses for an unrelated-sounding reason.
3. **Sats underfund.** When every asset balances, no stranding guard fires and the shortfall reaches `fill.build()` as `ArkadeTransactionBuilder: insufficien...`. Reachable via a sponsor fare priced above what the inputs cover.
4. **Same coin funding two legs.** Self-double-spend within one transaction; txid compared case-insensitively, vout exactly.

## Scope

This PR was 4,693 lines across 14 files. The joint-signing protocol and the fill *planner* moved to the taxi client (ArkLabsHQ/arkade-taxi#2) â€” they had one consumer and it was not this repo.

The seven PSBT boundary helpers went with them, for the same reason: after the signer left, `grep` for each across `packages/*/src` and `packages/*/test` returned zero hits. Exporting them here would mean versioning public API with no caller. If a second consumer appears, upstreaming them then will have a caller to justify the shape.

What is left has an in-tree caller in `fillOffer`.

## Tests

`ts-sdk` 3075 passed +1 skipped Â· `boltz-swap` 617 Â· `swap` 911.

The two refusals added in review (1 and 3 above) are each confirmed by mutation: with the guard removed the test reports what actually happens â€” `promise resolved instead of rejecting`, and `ArkadeTransactionBuilder: insufficien...` â€” rather than asserting the guard works.


Kukks:
@coderabbitai review

Kukks:
Thanks — addressed in `d06c1991`. Taking the three findings in turn.

**`setTapScriptSigEntries` merges rather than replaces — valid, fixed.**

Correct, and worth stating plainly: the function is documented as replacing wholesale and does not. No in-tree caller was affected, because the only production call site writes `[...base, ...fresh]` — always a superset of what is already on the input, so merge and replace coincide there. But it is exported, and an external caller passing a shorter set would silently retain stale signatures.

Now clears with `undefined` first, which is the same delete path `unsignedPsbtBytes` already relies on a few lines above. The boundary test writes an empty set and asserts the entries are gone; removing the clear makes that test fail, so the behaviour is pinned rather than assumed.

**`changeScript` doc contradicts the type — valid, fixed.**

The field is required and `assembleOfferFill` validates it unconditionally, so it is never omitted, only unused when the contribution spends every sponsor sat. Doc corrected to say that. I kept the field required rather than making it conditional: the caller always knows where its change should land, and an optional field would push a "was change produced?" decision onto every caller.

**Skip the live suite unless `TAXI_FILL_LIVE=1` — declining, and the CI failure it refers to is already fixed differently.**

The failure was real, but a conditional skip is the wrong remedy here. This test is the only coverage of a real two-owner fill against arkd and the emulator, including a signature-acceptance path that cannot be exercised with unit fixtures. Skipping it by default means it silently never runs and rots.

`55a4b40b` instead gave the job the environment it needs. That job already starts the right stack: `scripts/regtest.sh` brings swap up from `packages/swap/.env.regtest`, which zeroes intent fees and disables automine — without zero fees the faucet cannot redeem a note at all. Container names and ports are the regtest defaults the sibling swap e2e test already assumes. So the suite now runs for real in CI rather than being excluded from it.

The remaining failure after that was a genuine bug in the test, not the environment: it funded three wallets with `Promise.all`, and since each faucet redeems a note and drives a settlement round, the concurrent rounds collided with `missing forfeit transactions`. Fixed in `927e1bd5` by funding them in sequence.

The loud throw is deliberate and stays: if someone runs this suite without a stack, failing is the correct answer, and a skip would let an absent stack read as success.

Kukks:
Thanks — this was a genuinely useful review. All seven findings are addressed in `a42ea6e9`; details and my reasoning below.

Noting up front that I have **not** treated the protocol-critical gate as satisfied. These are hardening changes; the human protocol review you asked for is still outstanding and nothing here substitutes for it.

**Accumulated signatures not pinned to the selected leaf — fixed.**

This was the one worth catching. New signatures were leaf-pinned; signatures already present in an incoming partial only had to name a leaf that existed and contained the key. Your failure mode is the right one to worry about: nothing is stolen, but the emulator fails to finalise and the failure arrives as *ambiguous* — the single outcome that forbids a retry, so it is operationally worse than a clean rejection.

Accumulated entries now must match the canonical leaf for their key, and the bound leaf as well wherever a binding is known. `prepareJointSubmission` gets the canonical check only, since there are no signing bindings at that stage to pin against. A test builds a graph whose input carries a second valid leaf with a second solver key, signs via that leaf, and asserts rejection; disabling the comparison fails exactly that test.

**`containsBytes` substring matching — fixed, and it was more than theoretical.**

Replaced with push-data parsing that only considers 32-byte push operands. The new test embeds the solver key inside a `PUSHDATA1` 64-byte constant: the old code raised a spurious "multiple candidate leaves" rejection on a perfectly valid graph. So the false positive was reachable, not just conceivable.

**Covenant checkpoint liveness — fixed.** A zero-signature covenant checkpoint now throws, mirroring the guard the covenant ark input already had.

**`covenantCosignerKey` does not re-verify — fixed by guarding, not renaming.** It now runs `verifyOfferFillPlan` on entry. A rename would document the hazard; the guard removes it, and the check is a cheap self-hash. Honest callers are unaffected because they already pass verified graphs.

**`JointOwnerKeys` JSDoc** now records that `"offer-covenant"` is never valid — rejected at runtime, unrepresentable in the type.

**Multi-taxi-input coverage** added, symmetric to the existing multi-solver test.

**`honestEmulator` fixture fidelity — fixed, and this was the sharpest testing point.** The fixture signed every checkpoint with the server key, so the covenant-checkpoint allowed-key set could have been collapsed to a single key with the suite still green. It now signs the covenant checkpoint with the tweaked key the live run actually observed. Collapsing the allowed set now fails three tests, so the set cannot be narrowed silently.

One thing I deliberately did **not** do: narrow the covenant-checkpoint allowed set itself (`emulator`/`server`/`covenant`). The live run only tells us what this emulator build does today, and tightening acceptance is a protocol decision that belongs with your review rather than inside a hardening pass.

Gate after the change: `pnpm -r lint`, `-r build`, `-r test:unit` all exit 0 — ts-sdk 188 files / 3082 passed + 1 pre-existing skip, boltz-swap 23 / 617, swap 41 / 980 (974 + 6 new).

Separately, `Integration (swap / all)` is still red and I do not yet have its root cause. The concurrency explanation in `927e1bd5` was wrong — serialising the faucet changed nothing. I have since ruled out round count, cold-versus-warm stack, and stack configuration, none of which reproduce locally. The latest commit adds a bounded, loudly-logged retry around the funding rounds so the next run reports whether the failure is transient or deterministic. That is instrumentation, not a fix, and I will not claim the job is healthy until it is.

Kukks:
Root cause of the `Integration (swap / all)` failure found, plus the two open nits. Fixed in `a7881603`, `589ba52a`, `5b5e898e`.

**The live-fill CI failure was never a funding-amount problem.**

Five previous attempts treated it as one — more sats, mint more than we send, serialise the faucet, re-read after funding. The last of those proved the funding was fine and the diagnosis wrong:

```
00:10:25.5973  maker ready:  1 coins, 1000000 sats
00:10:25.6086  solver ready: 1 coins, 1000000 sats
00:10:25.6193  taxi ready:   1 coins, 1000000 sats
00:10:25.6858  → Insufficient funds        ← 66ms later, at issuance
```

`assetManager.issue` selects its inputs with `getSpendableVtxos({ withRecoverable: false })` and targets only `dustAmount`, so the selector can only throw when that list is **empty**. The faucet's readiness gate called the no-arg accessor, which defaults to `withRecoverable: true` (`wallet.ts:422`) and so counts exactly the recoverable coins `issue()` refuses.

The gate was asserting a strictly weaker predicate than the code under test enforces. A wallet passed it holding a million sats while handing the selector nothing — which is why it failed in 66ms with no network round-trip, and why every amount-based fix missed.

Two changes:
- Gate on `withRecoverable: false`, so the faucet waits for what the test actually spends. When it can't, it now fails *at the faucet* naming the branch — `isSwept` / `expiresAt` / `expiresAtHeight` / `canSpendOffchain` — instead of surfacing a step later as "Insufficient funds".
- One note redemption for all three wallets instead of three. `redeem-notes` drives a settlement round; three serially stretched funding to 40s and logged `timeout waiting for forfeit txs` / `missing forfeit transactions`. The sibling `swap.test.ts`, which passes in the same job, does one redemption and spends immediately.

To be explicit about what is and isn't established: the predicate mismatch is read directly from source at the failing call site. *Which* branch made the coins ineligible is not yet confirmed — the bug has never reproduced locally, so the instrumentation is what settles it.

---

**CodeRabbit — serial waits can overrun the hook: valid, fixed in `589ba52a`.**

Three 120s waits is 360s inside a 300s `beforeAll`. This mattered more than it looks: it would have killed the hook mid-wait and suppressed the diagnostic above, so the run would have told us nothing. One 240s budget is now shared across all three.

**arkana — `isBoundedDecimal` unbounded BigInt parse: not applicable, no action.**

The symbol does not exist anywhere in the tree (`grep -rn isBoundedDecimal packages/` is empty) and `offerFillPlan.ts` has no regex-to-BigInt parse. Your `8381756a` pass retired this as a ghost finding and that call was right; the `fad03fac` pass re-raised it.

**arkana — dead second leaf check in `assertJointInputSigs`: keeping the check, message differentiated in `5b5e898e`.**

The real problem you identified is that both branches emitted an identical string, so a reader can't tell which fired and the second looks dead. That's now `is not on its bound leaf`.

I'm not removing it. Unreachability rests on `leafForKey` and the caller's binding always agreeing — true of today's call paths, but that's a property of the callers, not of this function, and it guards a signing path for the price of a string compare. Happy to drop it if you think the invariant is structural rather than incidental.

Gate unchanged: ts-sdk 3083 passed +1 skipped, boltz-swap 617, swap 982.

The protocol-critical human review is still outstanding; none of this substitutes for it.

Kukks:
Correction on two things I said above, both in my own favour to admit and one that changes a review conclusion.

**The "dead" bound-leaf check is not dead — the suite already reaches it.**

I kept it on the grounds that unreachability was a property of the callers rather than of the function. That was the right call for a better reason than I gave: `signJointGraphForOwner > rejects an accumulated signature on a different valid leaf` reaches it today. After `5b5e898e` gave the two branches distinct messages, that test failed with

```
Received: "joint signing: incoming partial input 1 signature is not on its bound leaf (not submitted)"
```

`bound leaf`, not `selected leaf` — so the canonical-leaf check above it passed, and only the binding check rejected the foreign-leaf signature. Removing it as dead code would have deleted a guard the test suite actively exercises.

The duplicated message is what hid this: with both branches emitting the same string the test passed either way, and there was no way to tell which fired. @arkana-ai-bot — that is why I'd rather differentiate than delete; the reachability argument was load-bearing and it doesn't hold.

**I was wrong that no test asserted the old string.**

I grepped for the literal `"not on its selected leaf"`. The assertion is the regex `/selected leaf/`, which that search never matched, and I asserted it in a commit message as though I'd checked properly. CI caught it in `Lint, Test & Build`. Fixed in `85b3ef31`, which asserts `/bound leaf/` and so now pins *which* check fires — stricter than before.

**Why my local run didn't catch it, which is the more useful finding.**

`packages/swap` depends on `@arkade-os/sdk` as `workspace:*`, which resolves to the SDK's built `dist`. Editing `ts-sdk/src` and running swap's unit tests without `pnpm run build` first tests the *previous* SDK build — silently, and green. My local gate was structurally unable to see the change I was verifying. Re-run after a rebuild: ts-sdk 3083 passed +1 skipped, boltz-swap 617, swap 982, and the rebuilt dist does now carry the new string.

Worth noting CI's baseline is not mine: it reports swap `0.0.19` / 987 tests against my `0.0.18` / 982, because it tests the branch merged with `master`.

The live-fill instrumentation from `a7881603` has not had a run yet — `Lint, Test & Build` failing skipped the whole integration matrix, so the faucet diagnostic is still unexercised. That's what the current run is for.

Kukks:
All six addressed — `de6f7603`, `8816ec9f`, `bc19ab01`, plus the PR body. Both of the substantive ones I confirmed by mutation rather than by reading, because each turned on a reachability question I'd have got wrong by inspection.

**1. Critical — wanted-asset plan with no asset group. Valid, fixed in `de6f7603`.**

You're right, and the guard read exactly as you describe:

```ts
if (wantedAssetId !== undefined && (packet?.groups.length ?? 0) > 0)
```

Absence is the failure here. `toJointGraph` re-derives outputs from the transaction precisely to catch a builder that stopped emitting what the offer binds, so "no groups at all" is the case it should be loudest about, not the one it skips. Now throws before the group-0 comparison.

Not reachable through today's builder — `offer.ts` always calls `withAsset` when a wanted asset is present — so I drove it from the packet. With the short-circuit restored the new test reports `promise resolved "{ …(4) }" instead of rejecting`: the plan really is accepted with no group. That's the finding demonstrated, not argued.

(Minor: the citation was `~line 226`, which is `verifyOfferFillPlan`. The code is at 203.)

**2. Medium — `explicitSats > inputsSum`. Valid, fixed in `8816ec9f`.**

Confirmed your prediction exactly: with the guard removed the case surfaces as `ArkadeTransactionBuilder: insufficien...`, naming neither the amounts nor which leg to raise.

Reaching it took three attempts, and the two failures are the interesting part. A sats-only deposit trips the `offerAsset` backing check first; an asset-carrying one strands the deposit asset. Both shadow this guard, so the "pure-sats underfund" framing isn't quite the reachable shape. What does reach it is **a sponsor fare priced in sats above what the inputs cover, with every asset still balancing** — which is a caller error the taxi can actually make, so this is worth having rather than defence-in-depth.

**3. Low — `covenantCosignerKey` in the PR body.** Fixed; the description now says `providerCosignerKey` in all three places.

**4. Low — `pushData32`. Comment added.**

The reason is narrower than "limited coverage is safe": this scans attacker-supplied leaves, where a leaf that cannot be fully parsed must not abort the scan of the rest. `Script.decode` throws instead — verified rather than assumed, a truncated 32-byte push raises `Reader(): readBytes: Unexpected end of buffer`. That's now in the comment.

On your `tweak.ts` sub-point: agreed it's protocol-correct and worth a note, but `arkade/tweak.ts` isn't in this diff and I'd rather not widen the PR to annotate an untouched file. Happy to do it separately.

**5. Low — deposit-in-sponsor-fund. Test added** in `de6f7603`. The dedup guard already caught it, as you said; it just had no coverage.

**6. Low — `build()` vs `send()`. Comment added** at the `fulfill()` site.

---

One correction to your previous pass that matters more than any of the above. The bound-leaf check you flagged as unreachable dead code is reached by `rejects an accumulated signature on a different valid leaf` — after `5b5e898e` gave the branches distinct messages, that test failed with `not on its bound leaf`, meaning the canonical-leaf check passed and only the binding check rejected. Removing it as dead would have deleted a guard the suite exercises. Detail in the comment above.

Gate after rebuild: ts-sdk 3083 passed +1 skipped, boltz-swap 617, swap 985 (was 982; +3 tests). The live fill gate went green on `35065225998` — root cause was a readiness check reading `getSpendableVtxos()` while `issue()` selects with `withRecoverable: false`.

Protocol-critical human review still outstanding.

Kukks:
**Scope change — this PR is now 858 lines across 6 files, down from 4,693 across 14.** Title and description updated to match.

The joint-signing protocol, the offer-fill planner and their suites moved to arkade-taxi's client package (`c877d39c`). What stays is what stands on its own without them: the transaction-boundary helpers in `utils/arkTransaction`, and `offer.ts`'s sponsor roles, which the offer format owns.

The move was cheap, and the reason is worth recording for anyone reviewing the seam. Every symbol the joint code reached for was **already public API on master** — `Transaction`, `Extension`, `isBatchSignable`, `toXOnly`, `EmulatorProvider`, `arkade.computeArkadeScriptPublicKey`, plus the boundary helpers this PR adds. Verified against the built package rather than by reading the barrel. And `offer.ts` never referenced the joint layer at all, so only six files touched it and four moved whole.

Newly exported: `assembleOfferFill`, `connectFillContract`, `resolveDeposit`. The five layout types were already public.

@arkana-ai-bot — your CHANGES_REQUESTED findings were all addressed before the move, and the two substantive ones travelled with the code:

- **#1 (critical, wanted-asset group)** and **#2 (medium, sats underfund)** are fixed in `offer.ts` / the planner, each confirmed by mutation. #1's guard is in the planner and now lives in arkade-taxi; #2's is in `offer.ts` and is still here.
- **#3** PR body, **#4** `pushData32`, **#5** dedup test, **#6** `build()` comment — all done. #4's comment moved with `jointSigning.ts`.
- The bound-leaf check you asked me to delete as dead code is **reached by the test suite** — detail in my earlier comment. It moved with the signer, still intact.

Two corrections to my own work in this increment, both caught from evidence rather than review:

- `c877d39c` deleted `packages/swap/test/fillBuilder.test.ts` whole. That file already existed on master with 209 lines of `fillOffer` coverage unrelated to joint signing; the PR had only added a describe to it. The diff stat reading `209 ----------` against master is what exposed it. Restored in `019df4eb` — swap went 907 → 911.
- `submitJointFill` calls exactly one provider method, so its parameter is now `Pick<EmulatorProvider, "submitTx">`. That is the honest signature; the old one made every caller build a whole provider for one call.

CI is green on all 8 checks. `exit-providers-rotation` failed once with `VTXO_ALREADY_SPENT` in `unilateralExitDeepChain` — unrelated to anything left here and green on the prior head, so I re-ran it rather than call it flake on inspection; it passes.

Gate: ts-sdk 3082 passed +1 skipped, boltz-swap 617, swap 911.

The taxi side is built and passing locally against these exact packages and lands separately. Protocol-critical human review still outstanding.

Kukks:
Taxi side is open for review: **ArkLabsHQ/arkade-taxi#2** (draft).

It carries the joint-signing protocol this PR gave up, plus the taxi service half — quoting, submission, reconciliation, and the cross-flow reservation fence.

It is draft because it depends on this PR: its `package.json` files point `@arkade-os/sdk` and `@arkade-os/swap` at local tarballs, since the `arkTransaction` helpers and the three new `offer.ts` exports here are unreleased. It unblocks once this merges and both packages publish.

Kukks:
Cut again — **3 files, 510 lines**, all in `packages/swap`. Title and description rewritten.

The 858 was hiding something. After the signer moved out, `grep` for each of the seven boundary helpers this PR added to `utils/arkTransaction` — `unsignedPsbtBytes`, `assertUnsignedPsbt`, `assertSameUnsignedTx`, `assertDefaultTapScriptSigs`, `tapScriptSigEntries`, `setTapScriptSigEntries`, `tapLeavesOfInput` — returns **zero hits** across `packages/*/src` and `packages/*/test`, excluding their own file and the barrel. Every caller left with the signer.

Keeping them would mean this SDK versions and supports seven exported functions for one out-of-tree consumer. They moved to the taxi client with their boundary test (`2f8a4c71`). If a second consumer turns up, upstreaming them then costs nothing and will have a caller to justify the shape. `assertAllowedSighashTypes`, `verifyTapscriptSignatures` and `matchServerCheckpoints` were public before this PR and stay.

Everything left has an in-tree caller in `fillOffer`.

Also fixed two things that made it read as vague:

- **Comments were 17% of added lines in `offer.ts`, now 12%.** What went was the prose that restated the line beneath it. The traps stay — `offerAsset` being a TLV claim the covenant never inspects, the emulator reporting only that the covenant said no, output 0 not being cosmetic.
- **The description was an essay.** It is now the new API as a table, the behaviour changes, and the four refusals with the concrete failure each prevents.

One self-inflicted bug caught on the way: reverting `arkTransaction.ts` used `origin/master`'s tip, but the PR diff measures against the merge-base, so it silently pulled in master's `deprecatedServerPubkeys` rotation work and showed 45 lines of someone else's change as mine. Restored to the merge-base in `18ba4998`; the file now reads as untouched.

Gate: ts-sdk 3075 passed +1 skipped, boltz-swap 617, swap 911.

Kukks:
All five addressed — `ecd807c7` here, `bdb3b04` in arkade-taxi.

**1. Sponsor leg had no unit test — blocking, and you're right. Fixed.**

This was my doing. The suite that covered it moved to arkade-taxi with the planner, and it tested `buildOfferFillPlan` rather than the assembly underneath, so a money-routing public API was left with "the code being correct" as the only thing holding it up. Exactly the gap you describe.

`fillOffer` refuses sponsor funding, so the new tests drive `assembleOfferFill` directly and assert what the builder is asked for. Your list, all seven:

- sponsor funding carrying an asset → refused (sats-only)
- contribution above the sponsor's own inputs → refused
- full contribution → no `sponsor-change` output emitted
- change returns to the sponsor's script, not the solver's
- same-asset fare → folds into group 0, `required` includes the fare
- cross-asset fare → own group, behind the wanted one
- sats shortfall → named before `fill.build()` sees it

swap goes 911 → 923.

**2. `cancelOffer` inherited the txid bugs. Fixed here rather than deferred.**

You're right that it is the natural place — `resolveDeposit` was written for this and is exported now. `cancelOffer` uses it, and its error strings already matched, so the existing assertions hold unchanged.

Covered by a new cancel test for the shared-txid case plus four `resolveDeposit` unit tests: case folding, shared txid, outpoint precedence over txid, and a txid/outpoint pair that disagree. Worth noting the case-folding assertion sits on `resolveDeposit` rather than through `cancelOffer` — routed through the full call it passes deposit selection and then dies on unrelated mock plumbing, which would have made it pass for the wrong reason.

**3. `assertDefaultTapScriptSigs` footgun — valid, but it left this repo.**

The helper moved to arkade-taxi in `2f8a4c71`, along with the other six, because `grep` showed zero consumers here once the signer left. Documented there in `bdb3b04`, in your words: omitting `requiredPubKeys` requires every allowed key to have signed, so `requiredPubKeys: []` is what verifies signatures without demanding a particular signer.

**4. `changeScript` validated when unused.** Kept required, comment strengthened to say why: the caller always knows where its change belongs, and an optional field pushes a "was change produced?" decision onto every caller. Validated either way so a bad script fails on every fill rather than only the one that happens to emit change.

**5. Solver sats not fee-adjusted.** Comment added at the push — Ark pays fees through the P2A anchor, so VTXO values balance and the figure is exact.

Thanks for checking the other SDK repos for consumers; that matched what I found here and is part of why the helpers moved rather than staying as unused public API.

Gate: ts-sdk 3075 passed +1 skipped, boltz-swap 617, swap 923.

Kukks:
Both addressed in `00243db8`. Thanks for the cross-repo consumer sweep on the seven helpers — that's the check I'd have wanted before removing them and I only ran it in this repo.

**Duplicate-input guard has no test — correct, and it's a coverage regression I caused.**

Not an oversight in what to test: I *wrote* that test in `de6f7603`, deleted the file in `c877d39c` when the planner moved, then restored it from master in `019df4eb` — and master never had it. It went out with the restore and I didn't notice. It now lives in the sponsor describe in `fill.test.ts`, where the rest of that coverage ended up.

**`cancelOffer` has no `fundingOutpoint` escape hatch — taking this now rather than as a follow-up.**

Refusing an ambiguous `fundingTxid` is only safe if the caller can be exact, and I'd left them with no way to be: `fillOffer` gained `fundingOutpoint` in this PR, `cancelOffer` didn't, so two deposits sharing a txid went from *miscancelled* to *uncancellable*. That's a smaller harm but it's a regression in reach, and it's three lines to close since `resolveDeposit` already takes the parameter.

Confirmed by mutation rather than asserted: dropping `fundingOutpoint` from the call makes the new test report `multiple spendable deposits share ...`.

---

One thing your pass predates, flagged rather than buried: `b70556ef` **makes the fare's asset half optional**, so a sponsor can charge in sats alone. Previously `sponsor.fare.assetId` was required, and a fill moving no suitable asset could not be priced at all.

It affects output composition, so it wants the same protocol attention as the rest. The part I'd look at hardest: a sats fare pays the sponsor script carrying no assets — **exactly as change does** — so nothing about the output distinguishes them. In this repo that only affects grouping (a sats fare opens no asset group; without the guard it opens one for the empty asset id, which arkd refuses, and the test pins it). On the Taxi side the same ambiguity had two real consequences, both now fixed there: the wire role mapper labelled it a second `sponsor-change` and disclosed a fare of zero, and the reservation check counted it as change.

Gate: ts-sdk 3075 passed +1 skipped, boltz-swap 617, swap 927. CI was green on all 8 for `b70556ef`.

Protocol-critical human review still outstanding, and now covers the fare change too.

Kukks:
Thanks — nothing to action on `00243db8`, and the mutation note landing as intended is good to know.

One correction for the record, so a human reader doesn't go hunting. Both "still open" nits cite files that **no longer exist in this repo**:

```
packages/swap/src/offerFillPlan.ts    ABSENT
packages/swap/src/jointSigning.ts     ABSENT
packages/ts-sdk/src/jointSigning.ts   ABSENT
grep -rn isBoundedDecimal packages/ --include=*.ts   → 0 hits
```

@arkana-ai-bot your own 10:12 pass retired both as *"No longer applicable — moved to the taxi service and absent. Verified: `ls` → no such file."* This pass reopens them as "not in this diff", which reads as outstanding work. They aren't in the diff because they aren't in the repo.

On the second one specifically, this is the third pass asserting the leaf check is unreachable, and it isn't. `signJointGraphForOwner > rejects an accumulated signature on a different valid leaf` reaches it: after `5b5e898e` gave the two branches distinct messages, that test failed with `not on its bound leaf` — meaning check (1) passed and only the binding check rejected. Deleting it as dead would remove a guard the suite exercises. Detail in my earlier comment; the code and its test now live in ArkLabsHQ/arkade-taxi#2.

Not a complaint about the review — the incremental analysis has been consistently useful, and the cross-repo consumer sweep on the seven helpers was a check I hadn't run. Just flagging that the open-findings ledger has now flip-flopped twice on files that moved, and a human reading it should trust the `ls` over the tally.

Kukks:
Superseded by #970, which contains every commit here, plus #929 and the rest of the carrier work. The final review pass on this PR recorded no open findings; the human protocol review it asked for applies to #970 instead. The branch is left in place.

PR #927, opened 2026-09-15
fix(sdk): stop sending Arkade headers to the indexer origin on Expo

### LINKED ISSUE #889: getExpoFetch's fallback sends Ark-server headers to the indexer origin

IgnacioPorte:
## The problem

`getExpoFetch` adds `X-Build-Version` and `X-SDK-VERSION` to every fetch it returns, and both Expo providers use it — but they reach different origins. arkd wants those headers; an indexer on its own origin rejects them in the CORS preflight, so the subscription request never gets sent.

The web providers already get this right: `RestArkProvider` uses `fetch`, `RestIndexerProvider` uses `baseFetch`. Since `ExpoIndexerProvider` extends `RestIndexerProvider` and only overrides `getSubscription`, it was sending the headers on its subscription but not on its other requests.

## Why it's bigger than the one-liner in #889

The issue asks to change the fallback `return fetch` to `return baseFetch`. But `getExpoFetch` has two branches and both added the headers — and the *other* one, the `expo/fetch` branch, is what actually runs in an Expo app. Fixing only the fallback leaves the common path broken.

## The fix

Headers belong to the request, not the transport, so `getExpoFetch` no longer adds any. `sseStreamIterator` already takes a per-call `headers` argument that the Ark call sites were passing `{}` to, so `ExpoArkProvider` passes `ARK_VERSION_HEADERS` there instead.

`expoIndexer.ts` is unchanged — sending no headers is now just the default, which is also the safer one: a future Expo caller has to opt in to reach a non-arkd origin wrongly.

## Severity

Latent. Default setups point both providers at the same `serverUrl`, so today the headers land on arkd anyway. It only bites someone running the indexer on a separate origin.

## Tests

`test/expoStreamHeaders.test.ts` checks what each provider actually puts on the wire, running the same cases against both branches. Two fail on master — same case, once per branch:

```
× Expo SSE streams over expo/fetch         > … > sends no Arkade headers on the subscription
× Expo SSE streams over baseFetch fallback > … > sends no Arkade headers on the subscription
```

The rest cover what must not break: arkd still gets its headers, the subscription keeps `Accept`/`Content-Type`, and the stream keeps its own abort signal so there's no `READ_TIMEOUT_MS` truncation.

Lint, typecheck, unit tests and build all pass.

Closes #889

## Summary by CodeRabbit

* **Bug Fixes**
  * Expo-based Arkade streaming requests now include SDK and build version headers.
  * Improved compatibility for Expo streaming through more consistent fetch transport and fallback behavior.
  * Streaming requests preserve abort-signal support for reliable cancellation.

* **Tests**
  * Added coverage for Expo streaming headers, fallback transports, abort signals, required subscription headers, and unavailable fetch implementations.


packages/ts-sdk/src/utils/fetch.ts
```diff
@@ -7,6 +7,17 @@ export const buildVersion = "0.9.9";
  */
 export const sdkVersion = `ts-sdk/${version}`;
 
+/**
+ * The version headers arkd's compatibility guard reads, for callers that need
+ * them as data rather than as a wrapper — {@link fetch} sets the same pair, so
+ * a header added to one belongs in the other. They go to the Ark server and
+ * nowhere else: another origin rejects them in the CORS preflight.
+ */
+export const ARK_VERSION_HEADERS: Readonly<Record<string, string>> = {
```
tiero: Please favor ARKADE instead of ARK

bordalix (approved):

PR #929, opened 2026-09-16
feat(swap): preserve assets when claiming a cross-rail receive

Kukks:
Task 6's asset half. Stacked on `feat/taxi-intents` (#925) — review that first; this diff is only the last commit.

## The gap

A cross-rail asset receive locks the asset behind the same preimage as the sats, and the claim dropped it. The path was sats-only at every layer:

| layer | before |
|---|---|
| `LockupVtxo` (`refund.ts:133`) | `{txid, vout, value, recoverable}` — the claim could not *see* an asset |
| `pushClaim` (`claim.ts`) | one aggregate sats output, no packet — `grep -c asset` → **0** |

**The failure mode is the bad one.** The claim leaf inspects no output, so nothing on-chain forces the declaration. arkd refuses the undeclared asset with `ASSET_NOT_FOUND` — and it does so *after* the preimage is public. The trader has revealed `P` and still holds nothing.

## What changed

`LockupVtxo` carries `assets`, `findLockupVtxos` maps them off the indexer, and `pushClaim` declares one group per asset: every input vin feeding the single payment vout.

A **total** rather than a routing decision, deliberately — the claim leaf inspects no output, so splitting would invent a policy the covenant does not have.

## Tests

Three cases: every asset declared and summed onto the payment output; each asset conserved (inputs sum equals outputs sum); and a lockup with no assets still emits a bare payment with no packet.

Confirmed by mutation rather than asserted — making `claimOutputs` never declare fails the declaration and conservation cases.

## What I have not verified

`ArkTxInput` has no `assets` field, so the **input** side of the ark transaction declares none. The packet references vins by index, which is what arkd reads, so this should be sufficient — but **a live asset receive against a real stack has not been run**, and that is the check I would want before trusting this. The unit tests prove the transaction shape, not that arkd accepts it.

Gate: ts-sdk 3075 passed +1 skipped, boltz-swap 617, swap **930** (was 927).

## Summary by CodeRabbit

- **New Features**
  - Swap claims now preserve and declare supported assets carried by locked funds, including aggregated quantities for each asset.
  - Asset information is retained when locating funds eligible for refunds.

- **Bug Fixes**
  - Claims without assets continue to produce the same payment output as before.

- **Tests**
  - Expanded end-to-end and asset-claim coverage helps catch type-checking and settlement issues earlier.


Kukks:
@coderabbitai review

Draft, stacked on #925. The diff is ready for review — please look at the last commits rather than the base branch's.

Kukks:
Superseded by #970, which contains every commit here plus the rest of the carrier work.

@arkana-ai-bot's finding on this PR carries over and is being fixed on #970: `pushRefundWithoutReceiver` builds no asset packet, so an asset-bearing lockup that times out unclaimed cannot be refunded. The branch is left in place.

PR #31, opened 2025-03-24
Add `examples/spillman.js` and `examples/vhtlc.js`

louisinger:
* `vhtlc.js` shows how to use the exported `VHTLC.Script` class to create and spend a Virtual Hashed Timelock contract.
* `spillman.js` is using the base class `VtxoScript` to create a Spillman channel and shows how to use it.

it closes #23 

@tiero please review

PR #931, opened 2026-09-17
fix(swap): consume the solver's carrier_sats, and stop blocking exact-out

Kukks:
Two client-side assumptions on the `arkade↔arkade` RFQ path that were correct when written and are not any more. Counterpart to **arkade-os/intent-solver#149**, which changes both facts on the solver side.

## 1. The carrier is now the Service's, not a constant

`requestArkadeSwap` returns `ASSET_CARRIER_SATS` — the SDK's compile-time `330` — as the dust a caller must attach to an asset deposit:

```ts
carrierSats: params.offerAsset !== undefined ? ASSET_CARRIER_SATS : 0n,
```

The reference solver now prices against the **Arkade Service's own dust**, read from `getInfo()`, and publishes it on the quote as `carrier_sats`. Those agree only while a deployment's dust happens to be 330. Anywhere else the caller funds a different carrier than the solver priced — on the very leg whose quote already nets that carrier into its amounts.

Now read off the quote when present:

```ts
const published = quote.carrier_sats === undefined ? 0n : BigInt(quote.carrier_sats);
return published > 0n ? published : ASSET_CARRIER_SATS;
```

- **Absent** → the constant still applies, so a solver that doesn't publish the field is unaffected. The existing asset-to-asset test pins this: its quote carries no `carrier_sats` and still expects `330n`.
- **Non-positive** → also falls back. A published zero would otherwise fund an asset deposit with no sats to ride on.
- **BTC deposits are still `0n`.** The solver publishes `carrier_sats` on *both* legs, but on `BTC→asset` it is a **charge already netted out of `to_amount`** — the client sends nothing extra. Consuming the field unconditionally would make every BTC-deposit caller overfund by one dust.

## 2. Exact-out was blocked client-side only

`arkadeSwapRequest` already accepted `amountSide` and then threw:

```ts
if (input.amountSide !== undefined && input.amountSide !== "from") {
    throw new Error('arkade↔arkade quotes are exact-in only: …');
}
…
amount_side: "from",
```

and `requestArkadeSwap` always asserted the response's `from_amount` against the requested figure. Both were right while the solver answered `exact_out_unsupported`. **It serves both sides now**, so this guard is the only thing still refusing exact-out — a caller cannot ask for a specific payout at all ("give me exactly 500 USDX"). The named side is passed through, and the assertion follows it.

The test that pinned the refusal is **inverted, not deleted** — it now asserts the request carries the side the caller asked for, in both directions.

## Scope

**Untouched:** the onchain and Lightning request builders. They have their own `amountSide` handling because an invoice fixes the payout there; this is the asset builder only.

**No breaking change.** `carrier_sats` is optional on `RfqQuote`, `amountSide` already existed and already defaulted to `"from"`, and every previously-valid call behaves identically.

## Test plan

- `rfq.test.ts` — a request names the leg the caller asked for, `"to"` and `"from"`.
- `register.test.ts` — `carrierSats` follows a published `1000`, falls back to `330` when the field is absent, and falls back again when a solver publishes `0`.
- The pre-existing asset-to-asset test is the backward-compatibility case and was not modified.

Gates: `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit` all clean. **912 tests in `packages/swap`** against **911** on master; `ts-sdk` (3081) and `boltz-swap` (617) unchanged.

Committed with `--no-verify` on the pre-commit hook's own instruction — in a worktree `pnpm --filter="...[HEAD]"` selects no projects, so its lint and test would run nothing. The gate it names was run by hand instead.

A sibling PR ports this onto #919's branch, so the v0.5 client layer gets it too.

## Summary by CodeRabbit

- **New Features**
  - Asset swaps support both exact-in and exact-out requests while preserving the selected amount side.
  - Optional maximum input and minimum output limits help protect against unfavorable quotes.
  - Asset deposits use the quoted carrier amount when available and positive, with a default fallback otherwise.
  - RFQ quotes can include carrier amount information.

- **Tests**
  - Added coverage for amount-side handling, quote limits, carrier amount selection, and fallback behavior.


Kukks:
Context update from the solver side — **no code change needed here**, but two things this PR's description now understates.

**1. The solver's carrier pricing became opt-in.** `arkade-os/intent-solver#149` now gates it behind `ASSET_CARRIER_PRICING`, **defaulting off**, so quoted prices don't move on upgrade. A solver therefore publishes no `carrier_sats` until an operator enables it.

That makes this PR's fallback the *common* path rather than the legacy one:

```ts
const published = quote.carrier_sats === undefined ? 0n : BigInt(quote.carrier_sats);
return published > 0n ? published : ASSET_CARRIER_SATS;
```

Absent → the constant, which is exactly right while pricing is off. Nothing to change; worth knowing the branch that runs by default is the `undefined` one.

**2. A `methods` field was proposed and dropped.** `arkade-os/solver-registry#31` briefly added one declaring RFQ vs offer per market. It's gone — a standing offer is a *settlement contract*, not a negotiation method, and `transports` already discriminates. So no discovery-shaped change lands here either.

What #31 does add is two **pricing inputs** for the path where a maker computes terms from the card instead of a quote: `fee_flat_base` and `charges_delivered_carrier`. Those are consumed inside `@arkade-os/solver-discovery`'s `planOffer`, which this package already depends on — so on `master` nothing here needs wiring, since the swap package takes an `OfferPlan` from its caller rather than computing one.

Kukks:
**Correcting my previous comment on this PR.** I said the offer path needed no wiring here because the swap package takes an `OfferPlan` from its caller. That was wrong, and I checked too shallowly before saying it.

`payment/crossAsset.ts` does hold the dust — it just fetches it one line too late to price with:

```ts
const plan = await deps.quote(market, found.give, asset.amount);
const [balance, dust] = await Promise.all([deps.btcBalance(), deps.dust()]);
return { plan, error: validatePlan(plan, balance, dust) };
```

`deps.quote(market, give, wantAmount)` has nowhere to accept a carrier, so a consumer cannot net it even deliberately; and `validatePlan` receives the plan, the market **and** the dust yet checks only balance, limits and the dust floor — never the price. A plan priced without the carrier therefore passes validation and is refused by the solver after the offer is funded.

Filed as **#935**, blocked on `arkade-os/solver-registry#31` releasing — the `carrierSats` option does not exist in the published `@arkade-os/solver-discovery` yet, so it cannot be written against it today.

**Nothing changes for this PR.** The RFQ path is unaffected: `carrier_sats` arrives on the quote and is consumed here. #935 is the client-side-pricing path only.

PR #932, opened 2026-09-17
fix(swap): consume the solver's carrier_sats, and stop blocking exact-out (on #919)

Kukks:
**Stacked on #919** — base is `cursor/asset-rfq-nostr-0036`, not `master`. Sibling of **#931**, which is the same change against `master`; two PRs because `master` ships `0.0.19` (`latest`, what `intent-solver` pins) and this branch is on the `0.1.0-rc` train. Counterpart to **arkade-os/intent-solver#149**.

## Why this one matters for #919 specifically

`client/quoteRfq.ts` already passes `amountSide: toRfqAmountSide(pinned.on)` — **the new client layer was built for both sides**. It just couldn't reach one, because `arkadeSwapRequest` threw on `"to"`:

```ts
if (input.amountSide !== undefined && input.amountSide !== "from") {
    throw new Error('arkade↔arkade quotes are exact-in only: …');
}
```

That guard was correct while the solver answered `exact_out_unsupported`. It serves both sides now, so the guard is the only thing left refusing exact-out. Removing it is what lets #919's layer do what it already asks for.

## The two changes

**Carrier.** `requestArkadeSwap` returned the SDK's compile-time `ASSET_CARRIER_SATS` (330) as the dust to attach to an asset deposit. The solver now prices against the Arkade Service's **own** dust from `getInfo()` and publishes it as `carrier_sats`. Those agree only while a deployment's dust is 330. Now read off the quote, falling back to the constant when the field is **absent** *or* **non-positive** — a published zero would fund an asset deposit with no sats to ride on.

BTC deposits stay `0n`: the solver publishes `carrier_sats` on both legs, but on `BTC→asset` it's a charge already netted into `to_amount`, so the client sends nothing extra.

**Exact-out.** The named side is passed through, and `requestArkadeSwap`'s assertion follows it instead of always reading `from_amount`. The test that pinned the refusal is inverted, not deleted.

## Porting note — the cherry-pick was not mechanical

This branch **drops the `arkServerUrl` argument**: `requestArkadeSwap` takes three arguments where master takes four. The cherry-picked test carried master's four-argument shape, so `"http://ark"` bound to `transport` and the transport to `params`, and it failed with `set at least one of wantAsset or offerAsset; BTC-to-BTC is not a swap`. Fixed to the three-argument form the neighbouring test here already uses. Flagging it because it's the kind of difference that would otherwise be "fixed" by loosening an assertion.

## One thing left, deliberately, for this branch to decide

`client/accept.ts` funds an asset deposit by **omitting the sats amount** and relying on `wallet.send`'s default carrier:

> `// An asset deposit rides the SDK's dust-sat carrier, so the sats amount is left to the default rather than set here.`

So it doesn't see `carrier_sats` either — same class of bug, different path. It doesn't go through `requestArkadeSwap`, and the client record never retains the figure, so wiring it means **adding a field to the record model in this PR's own design**. That's your call rather than mine, which is why I've flagged it instead of patching it.

## Test plan

- `rfq.test.ts` — a request names the leg the caller asked for, both ways.
- `register.test.ts` — `carrierSats` follows a published `1000`, falls back to `330` when absent, and falls back again on a published `0`.
- The pre-existing asset-to-asset test is the backward-compatibility case, unmodified.

Gates: `pnpm -r lint`, `pnpm -r build`, `pnpm -r test:unit` clean. **1518 tests in `packages/swap`** against a recorded **1517** at base `bb160483`.

Committed with `--no-verify` on the pre-commit hook's own instruction — in a worktree `pnpm --filter="...[HEAD]"` selects no projects, so its gate would run nothing. The gate it names was run by hand instead.


Kukks:
Context update from the solver side — **no code change needed here**, but two things this PR's description now understates.

**1. The solver's carrier pricing became opt-in.** `arkade-os/intent-solver#149` now gates it behind `ASSET_CARRIER_PRICING`, **defaulting off**, so quoted prices don't move on upgrade. A solver therefore publishes no `carrier_sats` until an operator enables it.

That makes this PR's fallback the *common* path rather than the legacy one:

```ts
const published = quote.carrier_sats === undefined ? 0n : BigInt(quote.carrier_sats);
return published > 0n ? published : ASSET_CARRIER_SATS;
```

Absent → the constant, which is exactly right while pricing is off. Nothing to change; worth knowing the branch that runs by default is the `undefined` one.

**2. A `methods` field was proposed and dropped.** `arkade-os/solver-registry#31` briefly added one declaring RFQ vs offer per market. It's gone — a standing offer is a *settlement contract*, not a negotiation method, and `transports` already discriminates. So no discovery-shaped change lands here either.

What #31 does add is two **pricing inputs** for the path where a maker computes terms from the card instead of a quote: `fee_flat_base` and `charges_delivered_carrier`. Those are consumed inside `@arkade-os/solver-discovery`'s `planOffer`, which this package already depends on — so on `master` nothing here needs wiring, since the swap package takes an `OfferPlan` from its caller rather than computing one.

Kukks:
One follow-up that belongs to **#919** rather than to this PR, flagging it here since this branch is stacked on it.

`client/quoteOffer.ts`'s `quoteFromFeed` prices through `@arkade-os/solver-discovery`'s `quoteOffer`. Once `arkade-os/solver-registry#31` lands, that planner takes a `carrierSats` input — the Arkade Service's dust — and applies it when the market declares `charges_delivered_carrier`.

Without it the client computes a want amount the solver refuses, **on an offer already funded on chain**, which is the same failure the carrier work exists to remove, arriving by the client-side-pricing path.

It needs the Service's dust plumbed into `quoteFromFeed`, which its current input shape doesn't carry — a design decision inside #919 rather than something to graft on from a stacked branch, so I've left it alone. The card deliberately publishes only *whether* the solver charges, not the amount: the amount is the Service's own dust, both parties to an Arkade-settled swap share that Service, and a live figure beats a cached card.

Kukks:
**Correcting my previous comment on this PR.** I said the offer path needed no wiring here because the swap package takes an `OfferPlan` from its caller. That was wrong, and I checked too shallowly before saying it.

`payment/crossAsset.ts` does hold the dust — it just fetches it one line too late to price with:

```ts
const plan = await deps.quote(market, found.give, asset.amount);
const [balance, dust] = await Promise.all([deps.btcBalance(), deps.dust()]);
return { plan, error: validatePlan(plan, balance, dust) };
```

`deps.quote(market, give, wantAmount)` has nowhere to accept a carrier, so a consumer cannot net it even deliberately; and `validatePlan` receives the plan, the market **and** the dust yet checks only balance, limits and the dust floor — never the price. A plan priced without the carrier therefore passes validation and is refused by the solver after the offer is funded.

Filed as **#935**, blocked on `arkade-os/solver-registry#31` releasing — the `carrierSats` option does not exist in the published `@arkade-os/solver-discovery` yet, so it cannot be written against it today.

**Nothing changes for this PR.** The RFQ path is unaffected: `carrier_sats` arrives on the quote and is consumed here. #935 is the client-side-pricing path only.

PR #936, opened 2026-09-17
test(sdk): wait on completeUnroll's own precondition in the unroll e2e

Kukks:
## Problem

`Common > Wallet > should complete unroll after unilateral exit delay` fails intermittently in `Integration (ts-sdk / ark-core)`:

```
Error: no available exit path found for vtxo b88ca51e...:0
 ❯ prepareUnrollTransaction src/wallet/unroll.ts:413:19
 ❯ Object.completeUnroll src/wallet/unroll.ts:336:26
 ❯ test/e2e/ark.test.ts:702:38
```

It is pre-existing and not branch-specific — the same signature appears on unrelated branches, e.g. run 35192456395 (`claude/determined-dirac-xkyy`) and run 35233018165 job 105243543566 (`fix/swap-carrier-and-exact-out-v05`, which passed on rerun). Both of those commits already contained the current guard, so the guard does not close it.

## Root cause

The test recomputed the CSV maturity condition by hand instead of using the one `completeUnroll` actually checks:

- the test read the timelock from `exits[0].params.timelock` (the tap tree);
- `prepareUnrollTransaction` resolves it through `resolveUnilateralPath`, which for a contract-backed VTXO takes the sequence from the **contract row**, not the tap tree.

Two properties turned any disagreement into this exact failure:

1. **The guard degraded silently.** The mining step was wrapped in `if (remainingBlocks > 0)`. If the recomputed requirement came out already-satisfied — or non-numeric — the test mined nothing, skipped the `waitFor`, and proceeded unmatured. `getTxStatus` returns `block_height`/`block_time` straight from the explorer payload with no validation (`src/providers/onchain.ts:318-322`) even though its return type promises `confirmed: true ⇒ blockHeight: number`, so `requiredHeight` becoming `NaN` makes `NaN > 0` false and takes the same silent path. The CI timing matches: the failing run spent 12503ms versus 12760ms for a passing one, i.e. no time was spent waiting.
2. **No headroom.** The guard mined to exactly `requiredHeight`, so a one-block divergence anywhere flipped the result. Confirmed at runtime on regtest: tip `163` vs required `163`, and tip `238` vs required `238` on two separate local runs.

The comment anchoring the duplication said "Keep this aligned with `availableExitPath()` selection logic" — `availableExitPath` no longer exists anywhere in `src/`, so the duplicate had already drifted from its referent.

## Change

Wait on `prepareUnrollTransaction` itself — the precondition `completeUnroll` checks — mining a block whenever it reports immaturity:

- it cannot drift from the code under test, because it *is* the code under test;
- it is side-effect free (it signs but does not broadcast; `completeUnroll` broadcasts separately at `unroll.ts:337`);
- any error other than `no available exit path found` is rethrown rather than mined through;
- `waitFor` throws on timeout, so the step can no longer silently no-op.

No test is skipped, filtered, retry-wrapped or excluded, and the `it` timeout is unchanged at 120000.

## Test plan

- Full `test/e2e/ark.test.ts` against a freshly reset regtest stack: **24/24 passed**.
- Gate: `pnpm -r lint` clean, `pnpm -r build` exit 0, `pnpm -r test:unit` exit 0 (3092 + 617 + 911 passed, 1 skipped).
- The target test goes from ~14.7s to ~20.5s locally, the cost of mining one block at a time rather than in a single batch.

## Not addressed

`should reject complete-unroll before unilateral exit delay matures` duplicates the same predicate at `ark.test.ts:559-575`, but it *asserts* the precondition rather than silently skipping, so it fails loudly instead of flaking. Left alone to keep this change scoped.

## Summary by CodeRabbit

* **Tests**
  * Updated unilateral-exit end-to-end test coverage to wait for a mature exit path by retrying transaction preparation.
  * Tests now advance the chain automatically when additional confirmation is required, with a 60-second timeout.


Kukks:
## CI evidence

Ran `Integration (ts-sdk / ark-core)` **20 times** on this branch (5 independent `workflow_dispatch` runs x 4 attempts each).

**`should complete unroll after unilateral exit delay`: 19/19 green.** The 20th run never reached the tests — its stack bring-up died on a Docker Hub pull reset (`docker compose up failed` fetching `btcpayserver/lnd`), so it is not a data point either way.

Four ark-core jobs went red across those 20. None were this test:

| cause | count |
|---|---|
| `Delegate Lifecycle > should track and spend VTXOs across delegate add/remove` | 2 |
| `Common > Wallet > should settle 2 clients in the same batch` | 1 |
| stack bring-up (Docker Hub pull reset, no tests ran) | 1 |

Both of those tests are separate pre-existing flakes in the same file — they are what runs 35168821579 and 35222685997 actually failed on, not the unroll test. Worth separate issues; out of scope here.

One detail that supports the diagnosis: the new wait is not a no-op. Across the 20 attempts this test ran anywhere from 12.7s to 24.5s, and the longer runs are the loop genuinely mining extra blocks past the point the old guard would have stopped at. The old code mined to exactly `requiredHeight` and never further.

## On the review points

Thanks — both bots' reads match mine.

1. **Module-internal import** — agreed, and I'd rather have the coupling to `src/wallet/unroll.ts` than a second hand-rolled copy of the predicate, which is the bug being fixed.
2. **`interval: 0` can over-mine** — correct, and observed (the 20.7s/24.5s runs). Harmless on regtest and bounded by the 60s budget, so leaving it: adding an interval would trade a real guarantee for cosmetics.
3. **`getTxStatus` not validating `block_height`** (`src/providers/onchain.ts:318-322`) — agreed this is a real latent issue and the reason `NaN` could propagate silently. Deliberately not fixed here to keep this test-only; it deserves its own PR since it changes SDK behaviour.
4. **The sibling `reject` test still reads `exits[0]`** and still carries the stale `availableExitPath()` comment — it asserts rather than conditions, so it fails loudly instead of flaking. Left alone to keep the scope tight.

PR #937, opened 2026-09-17
test(sdk): wait on send's own selection view between the two createCash calls

Kukks:
## The flake

`Integration (ts-sdk / arkade-assets)` → `test/e2e/arkadeCash.test.ts` → `ArkadeCash > should claim each arkadeCash independently`.

Intermittent, not branch-specific — the same test failed the same way on `fix/unroll-exit-delay-flake` (run 35249556069) and on `feat/asset-receive-claim` (run 35108398942). Neither branch touches arkadeCash or the wallet, and the guard it trips over predates both (`ae505277`).

## Root cause

The test sends twice from Alice and guarded the gap with:

```ts
await waitFor(async () => (await alice.wallet.getVtxos()).length > 0);
```

That predicate is already true when it is reached — `fundedWallet` waits on exactly it — and `mergeByKey` in `InMemoryWalletRepository.saveVtxos` never removes an outpoint, so the count cannot fall back to zero. The wait returned on its first poll having witnessed nothing, and the second `createCash` ran against whatever view happened to be current.

The precondition it should have established is that **the first send's tx has been indexed**. Until then the wallet's view of that tx is one-sided, and either half can break the next send:

| what is stale | what `send` does | server says |
|---|---|---|
| input still reported unspent (every read syncs a 24h window and upserts, overwriting the local `isSpent` from `updateDbAfterOffchainTx`) | re-selects the already-spent funding coin | `VTXO_ALREADY_SPENT (6)` |
| change output known only locally | selects a coin the server has never seen | `VTXO_NOT_FOUND (30)`, `got_vtxos: []` |

Both were observed: the first is the original CI flake; the second showed up on this branch while I was testing a narrower fix that only closed the input half.

## The change

`waitForCashFunded` already witnesses the thing that matters — it polls the indexer for an output of that same ark tx — so it now runs for `cash1` *before* `cash2` is created rather than after. One line moved; the vacuous guard is dropped.

Nothing was skipped, retry-wrapped, or given a longer timeout.

## Test plan

- `pnpm -r lint && pnpm -r build && pnpm -r test:unit` — green, unchanged from master (ts-sdk 3092 passed / 1 skipped, boltz-swap 617, swap 911).
- `tsc --noEmit` over the edited e2e file against the package tsconfig — clean. (`test/e2e/**` is excluded from both `tsconfig.typecheck.json` and the vitest `typecheck` globs, so no gate compiles it; checked separately.)
- Repeated `Integration (ts-sdk / arkade-assets)` runs on this branch — tally in a comment below.

## Follow-up, not fixed here

A lagging indexer un-spending a locally recorded spend is an SDK-level property, not a test artifact: two back-to-back `send` calls can race the same way for a real caller. `_pendingSpendOutpoints` only covers the in-flight window and is cleared in `_submitOffchainSpend`'s `finally`. Making the local spend survive a sync touches every repository backend and could strand coins when a submit did not actually land, so it is out of scope for a flake fix — raising it here rather than folding it in silently.

## Summary by CodeRabbit

* **Tests**
  * Updated the independent cash claim test to wait until the funding transaction is indexed before creating the next cash item.
  * Simplified spendable-outpoint polling in the test setup.


Kukks:
Correcting the record on this branch: the first approach here was wrong, and CI caught it.

`975414d1` / `a223fefc` waited for the funding VTXO to leave `getSpendableVtxos()`. That closes only the input half of the race — the set stops offering that coin as soon as the local `isSpent` write survives a sync, which can happen before the indexer has ingested the ark tx at all. Run [35268886741](https://github.com/arkade-os/ts-sdk/actions/runs/35268886741) then failed the same test with `VTXO_NOT_FOUND (30)` on `143df5c4…5b9129c0:1` — vout 1, the **change** output — with `got_vtxos: []`. Tally for that approach: 9 green, 1 real failure, 1 unrelated infra failure (Esplora never came up) across 11 runs.

`b159e917` replaces it. Both failure modes are one missing precondition — the second `createCash` must not start until the first send's tx is indexed — and `waitForCashFunded` already witnesses that, so it simply moves ahead of the second `createCash`. The diff is now one moved line instead of a bespoke helper.

@coderabbitai review

Kukks:
@arkana-ai-bot thanks — but that review covers `a223fefc`, and CI has since disproved one of its load-bearing claims. Flagging it rather than letting the LGTM stand unqualified.

The review concluded:

> The "momentarily empty `offered`" scenario is not a problem: `updateDbAfterOffchainTx` writes the change VTXO locally before the wallet returns from `createCash`, so the selectable set after the first send contains the change coin. The wait exits only when the funding outpoint is gone from that set, at which point the change coin is already there for the second `createCash` to consume.

The first sentence is right and the conclusion does not follow from it. "The change coin is there" was only ever true of the *local* repository. `updateDbAfterOffchainTx` writing it locally says nothing about whether the server knows it, and `send` does not just select the coin — it submits a tx spending it. Run [35268886741](https://github.com/arkade-os/ts-sdk/actions/runs/35268886741) failed on exactly that: `VTXO_NOT_FOUND (30)` for `143df5c4…5b9129c0:1`, vout 1 — the change output — with `got_vtxos: []` from `/ark.v1.ArkService/SubmitTx`.

The deeper miss is that the predicate was only ever a proxy. "Funding outpoint gone from `getSpendableVtxos`" is satisfied as soon as the local `isSpent` write survives a sync, which can happen while the indexer has not ingested the ark tx at all — so it could clear in precisely the state where neither half of the tx was visible to the server. Semantic alignment with `send`'s filter (point 1) was real but bought less than it appeared to: it aligned *which coins are offered*, not *whether the server agrees they exist*.

`b159e917` drops that approach for the underlying precondition — the second `createCash` must not start until the first send's tx is indexed — which closes `VTXO_ALREADY_SPENT` and `VTXO_NOT_FOUND` together. Re-review of the current head would be welcome.

Kukks:
## CI tally — `b159e917`

13 independent runs of `Integration (ts-sdk / arkade-assets)`, the job that owns `arkadeCash.test.ts`:

**13 / 13 green.**

<details>
<summary>run ids</summary>

`35275862531` `35275866708` `35275868006` `35275873543` `35275879301` `35275885484` `35275891083` `35275897645` `35275904389` `35275910717` `35275916970` `35275922873` `35275929462`

</details>

For contrast, the superseded approach (`a223fefc`) scored 9 green / 1 `VTXO_NOT_FOUND` / 1 infra failure across 11 runs.

Five of the 13 runs are red overall, all in buckets that do not run this file — `exit-providers-rotation` ×3, `settlement-delegation`, `ark-core`. Per the matrix in `.github/workflows/ci.yml`, `test/e2e/arkadeCash.test.ts` appears only in `arkade-assets`, so this change cannot reach them; they are separate flakes being handled elsewhere.

### Not verified here

The regtest stack is a shared singleton on fixed host ports and other work was using it, so nothing in this PR was run locally — the diagnosis is from source plus CI logs, and the evidence above is entirely from CI.

The one claim I would most expect to be wrong: that an output of the ark tx appearing in the *indexer* query implies the ark service's own vtxo store will accept a sibling output of that same tx in `SubmitTx`. They are the same origin (`localhost:7070`) and the same ingestion, and 13/13 is consistent with it, but I have not read arkd to confirm the two share a store.

Kukks:
@coderabbitai review

PR #939, opened 2026-09-17
test(sdk): stop the settlement e2e asserting on a re-read of getVtxos()

Kukks:
## The flake

`Integration (ts-sdk / settlement-delegation)` intermittently fails `Settlement - Auto-settle boarding UTXOs > should auto-settle new boarding UTXOs into Ark` with `AssertionError: expected 0 to be greater than 0`, always at the same place — **`settlement.test.ts:53`**. Confirmed on four unrelated branches, so it is not something any one of them introduced:

| run | branch | test 1 duration | outcome |
|---|---|---|---|
| [34438044203](https://github.com/arkade-os/ts-sdk/actions/runs/34438044203) | `fix/max-sendable-asset-carrier` | 13333ms | fail, alone |
| [34984476999](https://github.com/arkade-os/ts-sdk/actions/runs/34984476999) | `feat/taxi-intents` | 7451ms | fail |
| [35255893291](https://github.com/arkade-os/ts-sdk/actions/runs/35255893291) (attempt 1) | `fix/swap-carrier-and-exact-out` | 13387ms | fail, + 2 more |
| [35258254138](https://github.com/arkade-os/ts-sdk/actions/runs/35258254138) | `improve-e2e-tests` | 13562ms | fail, + 2 more |
| [35232236446](https://github.com/arkade-os/ts-sdk/actions/runs/35232236446) | `fix/swap-carrier-and-exact-out` | 19197ms | **pass** |

Note the pass is the *slowest* run of the five.

## Root cause

The test waited for a VTXO to appear and then **re-read** `getVtxos()` and asserted on that second read:

```ts
await waitFor(async () => (await wallet.getVtxos()).length > 0, { timeout: 60000, interval: 2000 });

const vtxos = await wallet.getVtxos();   // <- a separate read
expect(vtxos.length).toBeGreaterThan(0); // <- line 53, where it fails
```

The first read returning non-empty does not make the second one non-empty. This wallet enables a `VtxoManager` (`pollIntervalMs: 5000`), and `getVtxos()` deliberately hides the inputs of an intent while that intent is in flight: `filterSnapshotVtxos` drops every outpoint in `_pendingSpendOutpoints` (`src/wallet/wallet.ts`). That contract is already pinned by `test/wallet.test.ts` → *"getVtxos excludes outpoints present in `_pendingSpendOutpoints"*, and by *"the in-flight set is in-memory only"*, which asserts `getVtxos()` returns **length 0** for a wallet that does hold the VTXO.

So `getVtxos()` is non-monotonic by design for a wallet running background settlement, and asserting on a re-read is unsound. The failing runs being consistently *faster* than the passing one fits: they land the re-read inside a background settle window rather than after it.

## Why three tests fail together, not one

Every `dispose()` in the file is the last statement of its `it()` body, so it is unreachable once an assertion throws. When test 1 fails it therefore leaks its wallet — with a 5-second settle poll loop and an SSE subscription — which keeps working the single shared regtest `arkd` for the rest of the file. The two tests that follow are the two most sensitive to that: one waits on the esplora chain tip, the other sits in `beforeEachFaucet`'s `ark redeem-notes`, which needs a batch round of its own.

To be precise about what this is and isn't: the leak is confirmed from the source (that `dispose()` is unreachable on throw), but it is an *amplifier*, not a proven cause of the other two failures — in run 34438044203 test 1 failed and leaked in exactly the same way while both neighbours still passed. Closing it means one flaky assertion can no longer be the difference between one red test and three.

## The fix

1. Fold the assertion into the wait and assert on the snapshot that satisfied it. The wait condition is now a **settled** VTXO rather than any VTXO — strictly stronger than before, not more lenient.
2. Dispose in a `finally`, matching the idiom already used in `operatorOffline.test.ts`, so one failure can no longer cascade into its neighbours.

Nothing is skipped, filtered, retry-wrapped, or given a longer timeout.

## Not the cause

The `ProviderUnavailableError` / `bad port` / `POST http://127.0.0.1:9/v1/indexer/script/subscribe` lines in the same logs are unrelated noise. They come from `operatorOffline.test.ts`, which points a wallet at `http://127.0.0.1:9` **on purpose** — port 9 is on the Fetch bad-ports blocklist, which is how that test simulates an unreachable operator (`operatorOffline.test.ts:61-67`). They appear roughly four minutes after these failures, and that test passes.

## Test plan

- `pnpm -r lint && pnpm -r build && pnpm -r test:unit` — exit 0. ts-sdk 3092 passed / 1 skipped, boltz-swap 617 passed, swap 911 passed. Identical to the pre-change baseline captured on the same commit.
- 5 independent `Integration (ts-sdk / settlement-delegation)` runs on this branch: **5/5 green** (runs 35268949361, 35268966269, 35268975036, 35268981400, 35268988148). Further runs were cancelled to give the saturated runner pool back.

Five green runs is corroboration, not proof — the base rate of this flake is low enough that a short green streak is weak evidence on its own. The load-bearing argument is structural: the failure point is pinned to `settlement.test.ts:53`, the re-read line, across all four observed occurrences, and that line's premise is unsound against a contract the SDK's own unit tests assert (`getVtxos()` returns length 0 for a wallet that holds the VTXO while an intent is in flight). Removing the re-read removes the failure mode rather than making it rarer.

### Unrelated red check

`Integration (ts-sdk / exit-providers-rotation)` fails on this PR. It is not caused by this change and cannot be: that job's own log prints its `TEST_FILES`, and `test/e2e/settlement.test.ts` — the only file this PR touches — is not in it, so the job never loads the change. The failure is `test/e2e/unilateralExitDeepChain.test.ts > unilateral exit — deep chain ordering > prepare() emits a deep offchain send chain in physical-input topological order`, and the same job already fails intermittently on `master` (runs 35011213035, 34516376220). Being tracked separately.


Kukks:
@coderabbitai review

(The earlier CodeRabbit check reported `pass` but its comment was "Review limit reached" — no review actually ran.)

PR #940, opened 2026-09-17
test(sdk): wait for the indexer to commit the spend in the cross-contract e2e

Kukks:
## The flake

`Cross-contract spending > should spend VTXOs from both default and delegate contracts in a single send`, failing at `expect(spentDelegateVtxos.length).toBeGreaterThan(0)` with `AssertionError: expected 0 to be greater than 0`.

Intermittent, on branches that cannot have caused it:

| run | branch | test | line |
| --- | --- | --- | --- |
| [35256831870](https://github.com/arkade-os/ts-sdk/actions/runs/35256831870) | `fix/swap-carrier-and-exact-out-v05` (touches `packages/swap` only) | `...in a single send` | `ark.test.ts:1476` |
| [35215743624](https://github.com/arkade-os/ts-sdk/actions/runs/35215743624) | `cursor/program-artifact-contracts-9678` | `...funded after delegation` | `ark.test.ts:1592` |

Both green on re-run. Both tests in the describe flake, so both are fixed here.

## Root cause

`send()` resolves when `FinalizeTx` returns, and records the spent inputs locally (`isSpent: true`, `wallet.ts:6115`, `:6196`). arkd commits the spend to its **indexer** after that RPC returns. The arkd log from the failing run shows the ordering directly:

```
method=/ark.v1.ArkService/SubmitTx duration=7ms
method=/ark.v1.ArkService/FinalizeTx duration=3ms
method=/ark.v1.IndexerService/GetVtxos duration=0ms   <- the test's read
spent 2 vtxos                                        <- indexer commits, after
added 2 vtxos at depth 5
```

The test then called `manager.getContractsWithVtxos({ type: ["delegate"] })`, which runs an opportunistic sync on every call (`vtxoSyncMaxAgeMs` defaults to `0`, so `syncedWithin` returns `false` — `contractManager.ts:1710`, `:2373`), fetches from the indexer, and persists the result (`contractManager.ts:2508`). `mergeByKey` is last-write-wins by outpoint (`inMemory/walletRepository.ts:123`), so an indexer row that predates the spend **overwrites** the `isSpent: true` that `send` had just written. The before/after diff is then empty and the assertion fails.

The read is documented as eventually consistent (`contractManager.ts:701-708`), so this is the test asserting a guarantee the SDK does not offer — not an SDK defect. Ruled out: the retryable-sync-failure branch (`contractManager.ts:1719-1722`) serves *repository* state, which at that point still holds the local `isSpent: true`, so it would make the test pass, not fail.

## The fix

Same shape as #936 — drive the wait from the value being asserted so the two cannot drift, and make sure the wait cannot silently no-op.

1. The consumed set is computed **inside** a `waitFor` and asserted from that same variable, so the predicate and the assertion are one expression.
2. Step 4's wait had the matching latent defect: `vtxos.some((v) => !v.isSpent)` is already true of the pre-send set (the unspent inputs satisfy it), so it returned immediately without waiting for the change output at all. It now matches the change outpoint by the txid `send` returned (`wallet.ts:6147` writes the change as `txid: arkTxid`).

No timeouts raised (`waitFor`'s 25s default is unchanged), nothing skipped or retry-wrapped, no assertion weakened.

## Test plan

- `pnpm -r lint && pnpm -r build && pnpm -r test:unit` — exit 0. ts-sdk 3092 passed / 1 skipped, swap 911 passed, boltz-swap 617 passed; identical to the pre-change baseline on `9172dbc`.
- Repeated runs of `Integration (ts-sdk / ark-core)` on this PR; tally in a comment below.

- `Delegate Lifecycle > should track and spend VTXOs across delegate add/remove` fails the same way at `ark.test.ts:1314` (`spentDelegateOutpoints`) — run [35168821579](https://github.com/arkade-os/ts-sdk/actions/runs/35168821579). Same root cause, left alone to avoid colliding with concurrent flake work.
- `test/e2e/**` is excluded from every typecheck path (`tsconfig.json` excludes `**/*.test.ts`; `vitest.config.ts` `typecheck.exclude`). Compiling `ark.test.ts` surfaces 3 pre-existing type errors on `master` (lines 571, 574, 1185), untouched here.


Kukks:
@coderabbitai review

Kukks:
## CI tally

13 independent executions of `Integration (ts-sdk / ark-core)` on this branch.

| run | attempts | ark-core job |
| --- | --- | --- |
| [35269051499](https://github.com/arkade-os/ts-sdk/actions/runs/35269051499) (pull_request) | 1–4 | 4× success |
| [35269102566](https://github.com/arkade-os/ts-sdk/actions/runs/35269102566) (dispatch) | 1–4 | 4× success |
| [35269105730](https://github.com/arkade-os/ts-sdk/actions/runs/35269105730) (dispatch) | 1–4 | 3× success, 1 failure (see below) |
| [35274160356](https://github.com/arkade-os/ts-sdk/actions/runs/35274160356) (dispatch) | 1 | 1× success |

**Target tests: 26/26 green** (2 tests × 13 runs). Both appear as `✓` in the logs, so they ran rather than being skipped.

**Job-level: 12 success, 1 failure.** The one failure (`35269105730` attempt 3) is a different test and a different cause:

```
FAIL test/e2e/ark.test.ts > Delegate Lifecycle > should track and spend VTXOs across delegate add/remove
Serialized Error: { status: 1, ..., output: [ null, 'error: not enough funds to cover amount 10000\r\n', '' ] }
```

That is regtest faucet exhaustion, not the indexer race this PR addresses — and in that same attempt both `Cross-contract spending` tests passed (`✓`). Unrelated to this change, and distinct from the `:1314` follow-up noted in the PR body (which fails on the indexer race, not on funding).

## Review response

Thanks @arkana-ai-bot — both notes read correctly to me.

- **The trailing `expect` is a tautology.** Agreed, `waitFor` only returns on a `true` predicate. Keeping it deliberately: it names what the block is proving at the point a reader looks for the assertion, and `waitFor` throwing on timeout is the actual gate. Happy to drop it if a human reviewer prefers.
- **The follow-up fetches after each `waitFor` are separate round-trips.** Agreed and benign: a committed spend is monotonic, so a later fetch cannot regress to the pre-send view. The window only existed because the *first* read could precede the commit.

Kukks:
@coderabbitai review

PR #941, opened 2026-09-17
fix(sdk): keep a recorded spend when a lagging indexer sync writes it back

Kukks:
## This is a wallet defect, not a test fix

A VTXO the wallet has just spent can be restored to *unspent* in the repository by the next indexer sync, re-offered to coin selection, and then refused by the operator. **Any app making two `wallet.send` calls inside the indexer's read-your-own-write window can hit this** — it is not specific to any test.

Observed symptom (CI run [35215743624](https://github.com/arkade-os/ts-sdk/actions/runs/35215743624), `Integration (swap-rfq / all)`), on a plain send to an Arkade address with no swap behind it:

```
VTXO_NOT_FOUND: VTXO_NOT_FOUND (30): some vtxos not found
 ❯ maybeArkError            src/providers/errors.ts:180:24
 ❯ handleError              src/providers/ark.ts:1346:22
 ❯ RestArkProvider.submitTx src/providers/ark.ts:609:13
 ❯ submitOffchainTx         src/utils/arkTransaction.ts:701:22
 ❯ _Wallet._submitOffchainSpend src/wallet/wallet.ts:5758:54
```

## Root cause

1. `wallet.ts:6115` — a send records its inputs `isSpent: true`, with `spentBy` / `arkTxId`, via `saveVtxosForContract`.
2. `wallet.ts:5829` — the `_pendingSpendOutpoints` hold that hides those inputs from `getSpendableVtxos` is released in a `finally` the moment the send returns. After that only the repository row protects them.
3. `contractManager.ts:1710` + `:2373` — every `getSpendableVtxos` re-syncs against the indexer: `vtxoSyncMaxAgeMs` defaults to `0`, and `syncedWithin` returns `false` for `maxAge <= 0`.
4. The sync writes the indexer's rows into the repository, and the write is **wholesale**: `mergeByKey` does `incoming.forEach(item => next.set(key, item))`, replacing the row by outpoint.

So while the indexer has not yet caught up with an ark tx the operator has *already accepted*, step 4 discards the `isSpent: true` written in step 1. Selection then picks an already-spent coin.

`computeSyncWindow` (`src/utils/syncCursors.ts:120`) deliberately applies no upper bound, so this is genuine lag rather than a window that systematically excludes fresh rows.

## The fix

`saveVtxosForContract` pins a spend that carries local provenance (`spentBy` / `arkTxId`) against an incoming row that contradicts it. A spend is monotonic and is only recorded once the operator accepted the ark tx, so a lagging read is not evidence it was undone.

A recorded spend is one carrying `spentBy`, `arkTxId` **or** `settledBy`. All three mean "we have a local record of this coin being consumed": `updateDbAfterSettle` records a settle-spent input with `settledBy` alone (no `spentBy`, and no fresh `arkTxId`), so a guard on the first two would have missed the settle path entirely. `settledBy` is required on the indexer wire type (`providers/indexer.ts:175`) and already part of `hasTerminalSpend`'s `isSpent || spentBy || settledBy` (`wallet/vtxo.ts:346`). It is preserved onto the pinned row as well — setting `isSpent` while dropping the field that proves it would leave a row that fails the guard on the next sync, making the pin hold exactly once.

That helper is the one choke point every stale writer shares — the contract-manager sync, the Expo poll processor, and the service-worker handler — so the guarantee holds across repository backends instead of only the one a given suite happens to use. A row whose spend has no local provenance stays correctable by a later sync. The lookup runs on every save: an earlier revision skipped it for all-spent batches, but a sync for a fully-spent contract produces exactly that shape and could strip the provenance the pin is keyed on (caught in review, with a committed regression case). Provenance is merged rather than replaced — incoming wins where present, the recorded row fills only what is missing — so a fresher spend is never overwritten by an older record. The cost is one script-scoped read per save.

## Evidence

The new case is the primary evidence, and it **fails on `master` today** across all three backends:

```
× spent rows survive 'InMemoryWalletRepository'  > keeps the spend when a stale sync re-reports the row unspent
× spent rows survive 'IndexedDBWalletRepository' > keeps the spend when a stale sync re-reports the row unspent
× spent rows survive 'SQLiteWalletRepository'    > keeps the spend when a stale sync re-reports the row unspent
Tests  3 failed | 6 skipped (9)
```

The settle-spent case was written the same way — it failed on all three backends against the narrower `spentBy || arkTxId` guard before it was widened. With all three in place: `Tests 15 passed (15)`. That it fails on IndexedDB and SQLite too shows the defect is not an artifact of the in-memory repository the e2e uses.

Gate (`pnpm -r lint && pnpm -r build && pnpm -r test:unit`), baseline → after:

| package | baseline | after |
| --- | --- | --- |
| ts-sdk | 188 files, 3092 passed \| 1 skipped | 188 files, **3101** passed \| 1 skipped |
| swap | 40 files, 911 passed | 40 files, 911 passed |
| boltz-swap | 23 files, 617 passed | 23 files, 617 passed |

Exit 0 both times; 0 failing before, 0 failing after. The +9 are the three new cases across three backends.

Two test doubles (`test/wallet.test.ts`, `test/worker/expo/processors/contractPollProcessor.test.ts`) stubbed only the write half of `WalletRepository`; they now provide the `getVtxos` the interface requires at `repositories/walletRepository.ts:44` (non-optional, unlike `getVtxosForScript?`). Their assertions are unchanged — an empty prior history pins nothing.

## What I could not confirm

- **Which VTXO arkd actually rejected.** Two mechanisms fit the trace: a resurrected already-spent input (what this PR fixes), or a fresh change output arkd had not yet made spendable. The runtime signal favours the former — during the failing test `getSpendableVtxos` logged *no* `vhtlc-v2` exclusion, while both later tests in the same file logged the lockup VTXO created by the earlier funding, so the snapshot that selection used predated that transaction entirely. But arkd's own logs would settle it and the CI step captures only `docker logs --tail 200`, which in that run begins ~5s *after* the failure. I have not proven it.
- **CI green here does not prove the race is fixed.** The e2e that exposed it (`packages/swap/test/e2e/rfqVerbs.test.ts`) is not on `master` — it lives on the `feat/v0.5` line — so nothing in this PR's CI exercises the original path. The unit test is the evidence; that line inherits the fix when it merges.
- The original failure was rare: 1 occurrence across the 57 `swap-rfq` job runs whose commit actually contained that e2e file.

## Also unblocks a red master-branch test

`test/e2e/unilateralExitDeepChain.test.ts > unilateral exit — deep chain ordering` (`VTXO_ALREADY_SPENT`, currently red on #938 and #939) is the same bug. That was established at runtime, not assumed: a probe driving the real `saveVtxosForContract` — save unspent, save spent, save stale-unspent — fails on `master` (expected 0 spendable, got 1) and passes with this patch applied. The resurrected coin there traces to a batch tree root.

## Related

#938 hits the same root class from a different layer — two back-to-back `ark send` CLI calls each re-reading the spendable set from the indexer, where the just-spent vtxo was still a candidate before the `Spent` flag flipped (`VTXO_ALREADY_SPENT`). Same read-your-own-write lag on arkd's indexer; that one is a fresh process reading the indexer directly, this one is the SDK's repository being overwritten by the stale snapshot. This PR does not fix that case.

The back-to-back `ark send` failure in `ark.test.ts` is **not** this bug: it lives in the Go CLI (`pkg/client-wallet/wallet.go`), a separate process with no local spend record to discard, so no TS-side change can reach it. Same upstream property, different defect.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved recorded spend status when stale synchronization data incorrectly reports previously spent coins as unspent.
  - Retained associated transaction details, preventing spent coins from being mistakenly restored to an unspent state.

- **Tests**
  - Added coverage for spend-state preservation across supported storage backends and synchronization scenarios.
  - Updated wallet and contract polling test coverage for repository data retrieval.


Kukks:
@coderabbitai review

Kukks:
@coderabbitai review

Kukks:
@coderabbitai review

Kukks:
Both reviews landed and they contradict each other on the same line, so I tested it rather than picking a side.

## @coderabbitai — accepted, it reproduces

The `vtxos.every(isSpent)` early return really could strip the provenance the pin is keyed on. Probed directly against `saveVtxosForContract` on `InMemoryWalletRepository`:

1. save spent-with-provenance
2. save all-spent-**without** provenance (hits the early return, written through untouched)
3. save a stale `isSpent: false` row

→ `expected false to be true` — the coin comes back spendable. The shortcut disarmed the pin it was meant to make cheap.

Fixed in `dbb9e564`: the lookup always runs, and provenance is merged rather than replaced — incoming wins where present, the recorded row only fills what is missing, so a fresher spend is never overwritten by an older record. Added as a committed case (`keeps the spend when an all-spent sync strips its provenance first`), which fails without the change on all three backends.

## @arkana-ai-bot — one correction

Thanks for the cross-repo check and the `WalletRepository` confirmation. But this call was wrong:

> The `vtxos.every(v => v.isSpent === true)` early return (line ~113) is the right optimisation for the `updateDbAfterOffchainTx` and `updateDbAfterSettle` hot paths, which pass all-spent batches. ✓

Those are not the only callers that pass an all-spent batch. `saveVtxosForContract` is also the **sync's** write path (`contractManager.ts:2469/2508`), and a contract whose VTXOs are all spent yields an all-spent batch from the indexer — with no obligation to carry `spentBy`/`arkTxId`/`settledBy`. That is the sequence above. The reasoning held only if the three local writers were the sole producers of that shape.

Your two other findings I agree with and have not "fixed":

- **Persistent extra read** — real, and now slightly worse, since the send and settle writes no longer skip it. I took the read over the correctness hole. The `recorded.size === 0` short-circuit still covers fresh wallets. Your suggestion that indexer-confirmed spends could shed their local markers to collapse the map back to zero is a good follow-up, but it changes when provenance is dropped, which is the thing this PR exists to protect — not something to land in the same change.
- **Read-then-write is not atomic** — agreed, and agreed it pre-exists this PR; `saveVtxos` already replaced rows wholesale. Not made worse here.

## On the monotonicity assumption

You are right that this is the load-bearing claim and deserves human review. To be explicit about its limit: the pin is keyed on *local* provenance, so it only asserts monotonicity for a spend **this wallet recorded after the operator accepted it** (`updateDbAfterOffchainTx` / `updateDbAfterSettle`). A row that merely arrived `isSpent: true` from the indexer with no provenance is left correctable. So the assumption is not "arkd never un-spends anything", it is "a spend this wallet itself performed and persisted is not undone by a later read". If that narrower claim is wrong at the protocol level, this is the wrong fix.

Kukks:
@coderabbitai review

Kukks:
@arkana-ai-bot thanks for the retraction — noted, and the three-step sequence is committed as a regression case.

On your last informational gap (mixed batch, indexer-confirmed spend with no provenance alongside a locally-recorded one): you reasoned it was handled but untested. It is now tested rather than inferred — `fills missing provenance when the indexer confirms the spend` (`4b29c786`), asserting the recorded `spentBy`/`arkTxId` survive an indexer row that confirms the spend without carrying them, while an unspent sibling at the same contract is untouched. It passes on all three backends.

The two open findings stand as you describe them and I am not addressing either here: the persistent extra read is an accepted trade-off for the correctness hole, and the non-atomic read-then-write pre-dates this PR.

Note for the human reviewer @Kukks asked for: **CodeRabbit has not reviewed the final state.** Its last review was on `ab1b1d13` with the inline finding on `3d973865`; every re-trigger since has come back "Review rate limited", and its green check reflects that, not a clean review. The commits it has not seen are `dbb9e564` (its own finding applied) and `4b29c786` (this test).

Kukks:
Closing in favour of #943, which carries this diagnosis forward with a bounded fix.

The root-cause analysis here holds and is reproduced in the new PR: arkd commits a spend to its indexer after `FinalizeTx` returns, `getSpendableVtxos` re-syncs unconditionally, and `mergeByKey` replaces the row wholesale — so a just-spent input comes back unspent and is re-offered to selection.

What did not hold was the shape of the guard. Reading the pin back from storage made it permanent:

- nothing in the SDK clears `isSpent`, and `hasTerminalSpend` (`vtxo.ts:344`) hides a coin on leftover `spentBy` or `settledBy` alone, so a spend recorded in error hid the coin from selection and balance for good;
- `_runRestore` ends in `refreshVtxos({ includeInactive: true, after: 0 })`, which writes back through the same guarded path — so `restore()`, the operation a user would run to repair bad local state, preserved the bad state instead;
- the only escapes were a whole-bucket `deleteVtxos` / `deleteVtxosForScript` or a full wallet wipe.

That trades a frequent, loud, recoverable failure for a rare, silent, unrecoverable one — the wrong direction for a wallet.

Two further consequences were specific to reading from storage: it added a `getVtxos` call inside every write (a hard dependency for repositories that only implemented the write half, and a new per-save cost on the hot sync path), and it made the guard depend on whether a given backend returns spent rows — `IndexedDBWalletRepository.getVtxos` catches and returns `[]`, which would have disabled it silently there. `ContractManager.refreshOutpoints` also bypassed the guard entirely on the leg that cannot name a contract.

#943 keeps the record in memory under a 60s TTL and keys it by repository, so the guard covers exactly the race it was built for — a write this process made, read back before the indexer caught up — and expires on its own otherwise. It is opened as a fresh branch off master so this rejected design is not carried in the history.

PR #945, opened 2026-09-18
test(sdk): name both faucet recipients in the delegate lifecycle phase 3 wait

Kukks:
## Problem

Phase 3 of `should track and spend VTXOs across delegate add/remove` (`packages/ts-sdk/test/e2e/ark.test.ts`) fired two back-to-back `faucetOffchain` calls and then waited **once**, on `wallet3.getVtxos().length >= 2`, for both of them.

Two reviewers arrived at this from opposite directions: `arkana-ai-bot` flagged the single-wait-for-two-faucets shape on #938 as "the same class of race", and #942 hit the consequence from the other side.

**The wait cannot see either faucet.** Phase 2 sends `11_000` against two `10_000` vtxos, and `send` builds its change output from the wallet's *active* receive tapscript (`wallet.ts:5525-5529`, `5729-5735`) — which for `wallet2` is the delegate address. So `wallet3` enters phase 3 already holding one unspent delegate vtxo, and `>= 2` is satisfied by that change plus whichever faucet lands first, while the second is still in flight.

**And the assertion underneath was vacuous anyway.** `contracts3Before[0].vtxos` is the delegate contract's *full* list, which still contains phase 2's spent input. That entry is absent from the `!isSpent` set on the very first read, so the diff was non-empty before phase 3 sent anything — `expect(spentDelegate3.length).toBeGreaterThan(0)` could not fail.

#942 tried to fix the vacuity alone, by restricting the baseline to unspent vtxos. CI timed out at `ark.test.ts:1350` (job `105443629411`): with the baseline captured as just `[change]`, selection saw three candidates and took the two `10_000` vtxos, leaving the baselined one untouched. It was reverted rather than stacked on, with a standing note that the faucet wait has to name both recipients first. **Both halves have to move together**, which is what this does.

## Change

- Each faucet waits on **its own contract's** unspent count — the shape #938 used for the two-client settle test. A small `unspentOf(type)` helper reads per-type so the `default` and `delegate` faucets are named separately, and the baselines are captured before the faucets rather than assumed.
- The baseline is now the **unspent** delegate set, which makes the assertion load-bearing.
- The after-read is wrapped in `waitFor`, because arkd commits a spend to its indexer *after* `FinalizeTx` returns — the same wrap phase 2 already carries from #942.

With both faucets indexed, the delegate pool cannot be avoided: the default contract holds exactly one unspent `10_000` vtxo against an `11_000` send, so any selection must reach into the delegate pool whichever ordering `selectVirtualCoins` produces. That is what makes the fixed assertion robust where #942's attempt was not.

## Test plan

Local gate on this branch, all exit 0:

- `pnpm -C packages/ts-sdk run typecheck` → 0
- `pnpm -r lint` → 0
- `pnpm -r build` → 0
- `pnpm -r test:unit` → 0 — ts-sdk **3092 passed +1 skipped**, boltz-swap **617**, swap **911**

Identical to the counts recorded in #938/#939, so zero unit delta. e2e is not part of that gate, so `Integration (ts-sdk / ark-core)` is what proves this. The clean fix went **fully green** on the first run.

### Mutation check

The previous attempt at this assertion shipped green *while being unable to fail*, so a green run is not acceptable evidence here. A temporary probe commit (`4840eff7`, since reset away) removed the phase 3 `send` so that nothing could consume a delegate VTXO, and logged what the pre-fix formula would have produced against that same state. [Run 35329194107](https://github.com/arkade-os/ts-sdk/actions/runs/35329194107):

```
Test Files  1 failed (1)
     Tests  1 failed | 23 passed (24)

 × Delegate Lifecycle > should track and spend VTXOs across delegate add/remove  31382ms
   → timeout in waitFor
     waitFor  test/e2e/utils.ts:323:11
              test/e2e/ark.test.ts:1366:9

MUTATION PROBE vacuousBaseline=3 vacuousDiff=1 strictBaseline=2 allVtxos3=3
```

- **The new assertion is load-bearing**: remove the send and it fails. Exactly one test failed — the mutated one.
- **The old assertion could not fail**: `vacuousDiff=1` with *nothing spent*. The pre-fix diff was non-empty before phase 3 sent anything, measured rather than argued.
- **The per-recipient wait does its job**: `strictBaseline=2` — both phase 2's change and phase 3's delegate faucet are present. Under `getVtxos().length >= 2` that baseline could be `1`, which is exactly why #942's attempt timed out.
- `allVtxos3=3` matches the model above: one default VTXO against an `11_000` send, so the delegate pool cannot be skipped.

## Summary by CodeRabbit

* **Tests**
  * Improved delegate lifecycle test reliability by tracking unspent VTXOs separately for each faucet contract.
  * Added more precise waiting and verification for faucet funding and delegate spending steps.


Kukks:
@coderabbitai review

Kukks:
Heads up for anyone reading mid-flight: `4840eff7` is a **deliberate mutation probe, not a proposed change**, and will be reset away as soon as CI reports.

The previous attempt at this assertion (#942) shipped green while being unable to fail, so "CI is green" is not acceptable evidence here. The probe removes the phase 3 `send` so that nothing can consume a delegate VTXO, and logs what the pre-fix spent-inclusive baseline would have produced against the same state.

Expected result, both halves in one run:

- `Integration (ts-sdk / ark-core)` **fails** with `timeout in waitFor` at the phase 3 assertion → the strengthened assertion is load-bearing.
- The probe logs `vacuousDiff >= 1` → the old baseline produced a non-empty diff with no send at all, i.e. the assertion on master could not fail.

The clean fix (`8341a8e9`) already went fully green on this branch in the first run, including `ark-core`.

Kukks:
Mutation check done; probe commit reset away, the branch is back to the clean fix `8341a8e9`.

**Run:** [35329194107](https://github.com/arkade-os/ts-sdk/actions/runs/35329194107) — `Integration (ts-sdk / ark-core)`, probe commit `4840eff7` (phase 3 `send` removed, so nothing can consume a delegate VTXO).

### Result — the strengthened assertion is load-bearing

```
Test Files  1 failed (1)
     Tests  1 failed | 23 passed (24)

 × test/e2e/ark.test.ts > Delegate Lifecycle > should track and spend VTXOs across delegate add/remove  31382ms
   → timeout in waitFor
     waitFor  test/e2e/utils.ts:323:11
              test/e2e/ark.test.ts:1366:9
```

Exactly one test failed — the mutated one — so nothing else in `ark.test.ts` depends on this change.

### Result — the old assertion provably could not fail

The probe computed the **pre-fix** formula against that same zero-spend state:

```
MUTATION PROBE vacuousBaseline=3 vacuousDiff=1 strictBaseline=2 allVtxos3=3
```

- `vacuousBaseline=3` — the old baseline `contracts3Before[0].vtxos` (unfiltered) holds phase 2's **spent** input, phase 2's change, and phase 3's delegate faucet.
- `vacuousDiff=1` — with **no send at all**, the old diff is still non-empty. `expect(spentDelegate3.length).toBeGreaterThan(0)` would have passed having spent nothing. That is the vacuity, measured rather than argued.
- `strictBaseline=2` — the new unspent baseline contains *both* phase 2's change and phase 3's delegate faucet. This is the half #942 could not get: under `getVtxos().length >= 2` the baseline could be captured as just `[change]` (1 entry) while the second faucet was still landing, which is why its attempt timed out. Naming each recipient is what makes it 2.
- `allVtxos3=3` — matches the model in the PR body exactly: one default VTXO (10_000) against an 11_000 send, so selection cannot avoid the delegate pool.

The clean fix at `8341a8e9` went fully green in the first run, including `ark-core`.

Kukks:
@coderabbitai review

Kukks:
Thanks @arkana-ai-bot — checked all three observations against HEAD (`8341a8e9`). Two hold, one is stale, and the follow-up is narrower than it reads.

**1. `expect(contracts).toHaveLength(1)` inside `unspentOf`** — agreed, and that is deliberate. A throw propagates out of `waitFor` (`utils.ts:320` does not catch), so a missing contract fails fast with a real message instead of burning the 25s timeout and reporting `timeout in waitFor`, which would be a much worse diagnostic.

**2. Phase 2 has no `waitFor` around its post-spend read** — this one is **not accurate against the current tree**. #942 added exactly that wrap; on HEAD it is `ark.test.ts:1291-1303`:

```ts
let spentDelegateOutpoints: typeof delegateVtxosBefore = [];
await waitFor(async () => {
    const contractsAfter = await manager2.getContractsWithVtxos({ type: ["delegate"] });
    const delegateVtxosAfter = contractsAfter[0].vtxos.filter((v) => !v.isSpent);
    ...
});
```

That is the half of #942 that survived its revert — only the phase 3 half was rolled back. So phase 2 already has the treatment this PR gives phase 3.

**3. Phase 2 captures an unfiltered `.vtxos` baseline** (`ark.test.ts:1273`) — accurate, but worth being precise about why it is not the same defect. Phase 3 was vacuous because *phase 2 had already put a spent VTXO on the delegate contract*, so the diff was non-empty for free. At phase 2's own baseline the delegate contract holds exactly one entry — the faucet output — and it is unspent, so that assertion is load-bearing today. It is fragile rather than broken: it would silently go vacuous the moment anything spends a delegate VTXO earlier in the test.

I would rather not fold that into this PR. Tightening it is a one-line change but it is the *kind* of one-line change that took #942 a CI failure and a revert to get right, and it deserves its own mutation check rather than riding along on this one. Happy to open it as a follow-up.

Kukks:
@coderabbitai review

PR #946, opened 2026-09-18
test(sdk): wait on the first cash's own funding before the second createCash

Kukks:
Cherry-pick of `465fb524` (#937) onto the v0.5 line. One test file, +2/−2.

### Why

`Integration (ts-sdk / arkade-assets)` is red on #919 with `VTXO_ALREADY_SPENT`:

```
× arkadeCash.test.ts > ArkadeCash > should claim each arkadeCash independently
  → Failed to create ArkadeCash: send failed after the note may have been submitted.
  { code: 6, metadata: { vtxo_outpoint: '1e386048...:0' } }
```

Not caused by anything in #919 or #932 — those touch `packages/swap`, this is `packages/ts-sdk`. The wait between the two `createCash` calls is a no-op on this branch:

```ts
const cash1 = await alice.wallet.createCash(5000);
await waitFor(async () => (await alice.wallet.getVtxos()).length > 0);  // already true
const cash2 = await alice.wallet.createCash(3000);
```

`getVtxos().length > 0` is **already true** of Alice's pre-send state, so it returns immediately and waits for nothing. The second `createCash` then re-picks the first send's input before arkd has committed the spend to its indexer — arkd indexes the spend *after* `FinalizeTx` returns.

Waiting on the first cash's own funding is the real precondition.

### Scope

Deliberately narrow. `feat/v0.5` is 193 commits diverged from master and missing all six of master's flake fixes (#936, #937, #938, #939, #940, #942) — this brings only the one that turns #919's job green.

A full `master` → `feat/v0.5` merge is the better long-term fix but is **not** mechanical, so it should be its own coordinated piece of work:

- `packages/swap/src/rfq.ts` conflicts on a real API divergence — v0.5 has `createOffer(wallet, {…})`, master has `createOffer(wallet, arkServerUrl, {…})`.
- master's side of that hunk also carries #931's `maxFromAmount` / `minToAmount` gates, which would strand #932 (the same change targeted at this line) as redundant-but-conflicting.
- `settlement.test.ts` conflicts between master's #939 structure and v0.5's canonical-facts assertions (`isPreconfirmed` / `isSwept` vs `virtualStatus.state`).

### Verification

The cherry-pick applied cleanly and the resulting diff is identical to `465fb524`, which passed full CI on master.

**I did not run a local gate** — this is a fresh worktree with no `node_modules`, and installing the toolchain for a two-line test change wasn't worth it. Relying on this PR's CI, in particular `Integration (ts-sdk / arkade-assets)`.


Kukks:
Superseded by #947, which merges `master` into `feat/v0.5` in full rather than cherry-picking this one fix. #947 contains this exact change plus the other five indexer-lag fixes v0.5 was missing, so this PR is redundant.

PR #955, opened 2026-09-19
feat(sdk): let watchScript and unwatchScript take a set

Kukks:
Both `watchScript` and `unwatchScript` took one script at a time, and each call rebuilt the subscription — and on the add path, read the indexer too. A service that watches an address per payment pays that N times on every restart, even though the machinery underneath is already batch-shaped: `pollWatchedScripts` takes an array and `getAllNormalizedVtxos` chunks it at 32, and `withCoalescedSubscription` exists to fold many changes into a single update.

Both now accept `string | string[]`. A set is registered or dropped in one subscription update, with one batched indexer read on the add path. The single-script call is unchanged in behaviour and in signature, so no existing caller moves.

One part is worth a closer look. `withCoalescedSubscription` defers the subscription update to its flush, and its docblock asks a new caller to bring its own catch-up, since `scanContracts` relies on `Wallet.restore` doing a bulk `refreshVtxos` afterwards. For watched scripts the batched `pollWatchedScripts` is that catch-up: watch-only polling reads the indexer for those scripts directly instead of diffing a repository, so an output already sitting at a newly watched script is reported rather than lost to the deferred window. The update is also requested *inside* the scope rather than after it, because the scope only flushes an update something asked for and polling asks for none — I got that wrong first and the existing "does not prune wallet state on a watch-only spend" test caught it.

`unwatchScript` rebuilds only when something was actually removed, so unwatching a set of scripts none of which were watched stays a no-op.

The service-worker bus widens the same two payloads rather than gaining new message types.

Five tests: a set registers in one subscription update with no per-script read; a VTXO already present at a newly watched script is still reported; re-adding a set only updates labels; a set unwatches in one update leaving the rest subscribed; and unwatching nothing rebuilds nothing. The first asserts that every indexer read carries the whole batch rather than pinning a call count — registration polls and the flush re-syncs, and both are batched, so a count would assert incidental ordering.

`pnpm -r lint`, `pnpm -r build` and `pnpm -r test:unit` all pass (3111 + 619 + 913). Committed with `--no-verify` because the pre-commit hook's `...[HEAD]` filter selects no projects from a git worktree, and the hook itself prescribes running the gate manually in that case.

Context for why I wanted it: in `lnurl-server` we watch one Arkade address per user with an open payment. Today that is our own polling loop doing one indexer round trip per destination per tick, which saturates a 15s interval at roughly 128 users. Moving to `watchScript` replaces it with the stream plus your failsafe poll, and this is the piece that keeps startup from being N round trips.

*Note: this PR previously added a separate `watchScripts` method. Reworked to widen the existing ones instead, which is the smaller surface and the maintainer's preference. The earlier review comments about that design are stale.*


Kukks:
Worth pre-empting the obvious review question, since my first justification for this was weak.

Why a new method rather than widening the existing one to `watchScript(script: string | string[])`? I originally reached for the `createContract`/`createContracts` precedent, but that doesn't actually apply — those differ in return type (`Contract` vs `Contract[]`), which forces a split. `watchScript` returns `Promise<void>` either way, so there's no type-level reason there.

The real reason is that method parameters are bivariant. An embedder's existing `watchScript(script: string)` keeps satisfying a widened `watchScript?(script: string | string[])` signature — it compiles clean under `--strict`, with no error on either side — and then receives an array at runtime and treats it as a script. I checked that with `tsc` rather than assuming it.

That failure is reachable precisely because `watchScript` is optional, which its own docblock explains is so an embedder can bring its own `IContractManager`. An optional method is a capability a caller can test for (`if (!manager.watchScripts)`), the way `unsupportedByManager` already does elsewhere on this interface; a widened parameter is one it cannot test for at all.

Pushed that rationale into the docblock in 1a3af29 so it isn't only buried in PR discussion. Happy to go the other way if you'd rather have the smaller surface and are comfortable with that tradeoff for embedders — it's your interface, and the change is small either way.

Kukks:
Confirmed, and thanks — this is real. I instrumented both call sites rather than reason about it, and on a cold start the order is:

```
MARK connect-poll skipUpdate=false   <- startWatching
MARK registration-poll               <- pollWatchedScripts(added)
MARK connect-poll skipUpdate=true    <- the flush's connect(true) -> pollAllContracts()
```

So the newly registered scripts are read twice. Both reads carry the whole batch, so it is not per-script and does not affect the scaling claim, but it is redundant.

I have not fixed it, and I would rather say why than guess in someone else's state machine. The pre-poll is only redundant when the flush actually connects, which `tryUpdateSubscription` gates on `justGotSubscription && listenerParked`. That is decided *after* `updateSubscription()` runs, so it is not knowable before entering the coalesced scope. In the steady-state case — a subscription already exists, so no connect happens — the pre-poll is the only thing covering the new scripts until the next failsafe tick, and dropping it would trade one redundant read for a missed catch-up.

The obvious conditional (`if (this.subscriptionId !== undefined) await this.pollWatchedScripts(added)`) is the version I am unsure of: `subscriptionId === undefined` does not imply the listener is parked, and if both conditions are not met the connect never fires and nothing polls. You own that state machine and will know immediately whether that is reachable; I do not, and the failure mode is a silently late payment rather than a loud error.

Happy to take either direction:
- if `!hadSubscription` reliably implies the flush will connect-and-poll, I will add the guard and tighten the test back to asserting a single `getVtxos` call for the first registration;
- or if you would rather `connect(true)` not re-poll scripts a caller just polled, that is a change inside the watcher I would rather you steer.

Left as-is for now so the PR keeps one behaviour change rather than two.

Kukks:
@coderabbitai review

The incremental review after 1a3af29 came back rate limited, and that commit changed the docblock rationale for why this is a separate method rather than a widened `string | string[]` parameter — worth a look since it is the main design question on this PR.

Kukks:
Reworked to widen `watchScript` and `unwatchScript` to `string | string[]` rather than adding a separate `watchScripts` method — smaller surface, and it covers unwatch too, which the split version had left singular.

Force-pushed as a single commit, so earlier review comments on the `watchScripts` design no longer apply. The cold-start double-poll you flagged is unchanged and still open: on a cold start the flush's `connect(true)` re-reads the scripts the registration just polled. Both reads carry the whole batch so it is not per-script, and I have left it rather than guess at the guard, for the reason in the earlier comment.

@coderabbitai review

Kukks:
Thanks — took the order-dependence nit in 465db8d. The assertion now checks membership plus length rather than equality, since the claim under test is that no read is per-script, and that survives the batch being sorted or de-duplicated later.

Left the `--no-verify` note as context only: CI is the gating authority, and it runs the same lint/build/test on every push here. The hook is skipped solely because pnpm's `...[HEAD]` filter selects no projects from a git worktree, which is what the hook's own message says to do.

PR #957, opened 2026-09-21
fix(sdk): route HD unilateral exit sweeps through the descriptor-aware signer router

Kukks:
## Summary
- Route unilateral exit sweeps through the wallet's descriptor-aware signer router instead of `wallet.identity`
- Expose a public `Wallet.signInputsByWitnessScript` helper and keep `signOnchainBoardingTx` as an internal alias
- Add an opt-in "unknown script = mine" fallback so static wallets do not silently end up with an empty witness
- Add regression coverage for the rotated descriptor key path and the static fallback

## Why
Previously `UnilateralExit.prepare` signed HD exit sweeps through `wallet.identity`, which only holds the index-0 key. VTXOs on rotated receive descriptors therefore finalized with an empty witness, marking those VTXOs as skipped and shrinking the recoverable subset of the exit package. This mirrors the boarding sign path and keeps static wallets on the identity sign route.

## Testing
- `pnpm typecheck`
- `pnpm typecheck:tests`
- `pnpm vitest run test/exit-sweep.test.ts test/hdWalletCapable.test.ts test/wallet/inputSignerRouter.test.ts test/exit-prepare.test.ts` (56 tests pass)
- `pnpm build`


Kukks:
Replaced by #959 targeting master with only the exit signing changes.

PR #961, opened 2026-09-21
Fix/reduce indexer requests

bordalix:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved refund and claim handling by using the contract manager’s current lockup state.
  - Excluded terminally spent outputs from refund and recovery processing.
  - Prevented receive claims from running when a swap’s outcome is unknown.
  - Improved handling of swept and exited lockups during recovery and claims.

- **Tests**
  - Expanded coverage for swept, spent, exited, missing, and contract-managed lockup scenarios.
  - Added coverage for contract-managed funding and receive-claim outcomes.


packages/swap/src/refund.ts
```diff
@@ -284,34 +284,20 @@ export async function findLockupVtxos(
...
+    for (const vtxo of all ?? []) {
+        if (vtxo.isUnrolled) continue;
+        const key = `${vtxo.txid}:${vtxo.vout}`;
+        if (seen.has(key)) continue;
+        seen.add(key);
+        out.push({
+            txid: vtxo.txid,
+            vout: vtxo.vout,
+            value: Number(vtxo.value),
+            recoverable: !!vtxo.isSwept, // swept → recoverable, live leaf → spendable
```
gringokiwi: Any reason to use `!!`? Are there any cases where `isSwept` might be returned as the string `"false"` and thus show up as `recoverable: true`?

bordalix: `isSwept` is boolean or undefined.
`recoverable` is expecting a boolean.
by using the double bang we make `undefined` the boolean value `false`

tiero:
@arkana-ai-bot boltz-swap is going to be deprecated soon, so ignore

@bordalix just improve the comment and LGTM to merge

tiero:
@Kukks isn't this already been fixed by a PR of yours?

bordalix:
@arkana-ai-bot full review

bordalix:
@tiero, @Kukks PR only made the indexer query go through the contact manager, it didn't stop it.

On every runPass() (every 5 seconds) the SDK was querying the indexer twice:
- on readLockupFate()
- on driveReceiveClaim()

My latest commits have the objective of sparing the 2nd query, by levering the information we got from readLockupFate() and the swap refund lock time. We now only call the 2nd time when the fate is "open" or the swap passed it's refund lock time.

tiero:
Sounds good! Let's merge so we improve it progressively

PR #971, opened 2026-09-23
perf(sdk): batch boarding UTXO writes and bulk-insert SQLite wallet rows

Kukks:
Follow-up to #964. `getBoardingUtxosForSigners` saved each boarding address's UTXOs with its own `saveUtxos` call. It now makes one call with a map of address → UTXOs, and every `WalletRepository` backend writes that map in one go. On SQLite the same goes for `saveVtxos` and `saveTransactions`, which still inserted one row per statement.

## Why it matters

The VtxoManager's boarding poll runs `getBoardingUtxos()` every minute by default (`pollIntervalMs: 60_000`). Each cycle called `saveUtxos` once per boarding address, which on IndexedDB and Realm is one write transaction per address per minute, almost always with nothing in it. Now it is one call per cycle, and IndexedDB and Realm skip the transaction entirely when there is nothing to save.

## What changed

- **`WalletRepository.saveUtxos(utxosByAddress: ReadonlyMap<string, ExtendedCoin[]>)`** — a new overload beside `saveUtxos(address, utxos)`, with the same merge-by-outpoint semantics.
- **Backends** — IndexedDB: one readwrite transaction for the whole map. Realm: one `realm.write`. SQLite: multi-row `INSERT OR REPLACE` through one `insertOrReplace` helper that packs as many rows into each statement as 999 bound parameters allow (the limit of SQLite before 3.32 — the executor is the consumer's, so its build may be that old). In-memory: one pass. Legacy V1 adapter: key by key as before, since its `StorageAdapter` has no multi-set.
- **SQLite `saveVtxos` / `saveTransactions`** — through the same helper: 49 VTXOs or 111 transactions per statement instead of one. `saveVtxos` is the hot path; every ContractManager sync writes through `saveVtxosForScript`. A key repeated within one call still resolves to its last occurrence, since `INSERT OR REPLACE` applies row by row within a statement too.
- **`getBoardingUtxosForSigners`** — one `saveUtxos(map)` once every fetch has resolved, so a failed fetch still writes nothing (#964).

## Batched lookup on the onchain provider: not possible

Checked against each backend's source:

- **mempool's TypeScript backend** serves only `GET /address/:address/utxo` (`backend/src/api/bitcoin/bitcoin.routes.ts`).
- **mempool/electrs** has multi-address routes for history only — `POST /addresses/txs` and `/addresses/txs/summary`, capped at 300 — while `utxo` is per address or scripthash (`src/rest.rs`).
- **Blockstream electrs** has no multi-address routes.
- **Electrum**: `ws-electrumx-client` sends one frame per request, even from `batchRequest`, which is what #964's concurrent `getCoins` calls already put on the wire.

Deriving UTXOs from `POST /addresses/txs` would swap the authoritative `/utxo` endpoint for a paginated history computation that only some deployments serve (not mempool-backend-only ones such as the default mutinynet explorer). So `OnchainProvider` is unchanged. NArk's `IBitcoinBlockchain.GetUtxosAsync` is single-address as well.

## Test

- Contract suite (InMemory, IndexedDB), SQLite, Realm and the legacy V1 adapter: a map spanning several addresses — including an update to an existing outpoint and an address with no UTXOs — lands per address.
- SQLite: one call spanning more than one insert chunk (80 rows) keeps every row. The suite's in-memory SQL emulator learned multi-row `VALUES`, but it ignores the `VALUES` text, so a new block also runs the batch against real `better-sqlite3`. Breaking the tuple separator fails only that test (checked).
- SQLite bulk writes, on real `better-sqlite3`: statement counts for UTXOs (80 rows → 1), VTXOs (60 → 2) and transactions (120 → 2) — each failed on the per-row code (2, 60, 120) — plus a repeated key keeping its last row.
- Wallet: every boarding write that carries the test's second address is a single `Map` covering both addresses, and the coin lands in the repository.

## Verification

- Lint clean (ts-sdk incl. its vtxo/provider guards, boltz-swap, swap); `tsc --noEmit` clean.
- Unit, on the branch rebased onto master after #964 (`c17a7aa1`): ts-sdk **192 files, 3211 passed, 1 skipped** (11 of them new here); boltz-swap 619 passed; swap 913 passed.

## Compatibility

The overload is a compile-time break for custom `WalletRepository` implementations: they must accept the map form, and a backend that doesn't would receive the `Map` as its `address`. It is kept deliberately (maintainer's call) over an optional sibling method in the `saveVtxosForScript?` style: GitHub search found no implementations outside this repo, and the `arkade-os/wallet` test mock stays compatible. The release notes should call it out — for this 0.x package, that means a minor bump.

SQLite chunks are deliberately not wrapped in one transaction: each statement is atomic, and the executor is the consumer's connection shared across repositories, so a `BEGIN IMMEDIATE` on the `saveVtxos` path would take in other repositories' concurrent writes and roll them back with ours on failure.

## Summary by CodeRabbit

* **New Features**
  * Wallet UTXO saving now supports a single address or multiple addresses in one call across storage options.
  * Existing UTXOs are updated when matching entries are saved; new entries are added.
* **Performance**
  * SQLite batches wallet data writes, reducing the number of database operations.


Kukks:
@arkana-ai-bot — note the head has moved to 86ef11ac: SQLite `saveVtxos` and `saveTransactions` now bulk-insert too, through one `insertOrReplace` helper that sizes each statement to the 999-parameter budget (so `UTXO_ROWS_PER_INSERT` is gone).

**1. Overload vs optional method** — deliberate, and not mine to settle: the maintainer asked for `saveUtxos` itself to take the map. The trade-off and the `saveVtxosForScript?`-style alternative are in the description, and I've put your recommendation to them. (The package is 0.x, so the breaking signal would be the minor bump.)

**2. Chunks outside a transaction** — intentional. Each statement is atomic, and across chunks this is strictly more atomic than the per-row code it replaces; the rows are idempotent and re-synced every poll. Wrapping in `runInTransaction` isn't free here: the executor is the consumer's connection, shared across repositories (`sqlite/transaction.ts`), so a `BEGIN IMMEDIATE` window on the hottest write path (`saveVtxos`) would take in any plain write another repository issues meanwhile — and roll it back with ours on failure. I'll add a one-line comment saying so.

**3. Mock `rowCount` floor** — keeping it. `Math.max(1, …)` preserves the mock's previous behaviour exactly (it always inserted one row per `INSERT OR REPLACE`, whatever the params), and no caller issues an insert without parameters.

Kukks:
Decision on the overload: kept (maintainer's call), recorded under **Compatibility** in the description, with the release notes to flag it for custom `WalletRepository` implementations (a minor bump on 0.x). The transaction rationale is now a comment on `insertOrReplace` in 30d5ae6d.

Kukks:
@coderabbitai review

PR #973, opened 2026-09-23
test(swap): parse the root barrel once in the ./advanced superset check

Kukks:
## Summary

`keeps ./advanced a superset of the root's client vocabulary` called `moduleOf(ROOT_ENTRY)` inside its filter, re-parsing the root barrel (and its `export *` graph) with the TypeScript compiler once per root name missing from `./advanced`. The result never changes, so it's now computed once — the same pattern the file already uses at the `@deprecated` pointers block.

Same assertion; the test goes from ~5.0–6.2 s to ~110 ms locally. At 5 s+ it exceeded vitest's default timeout on a loaded machine, which failed `pnpm run test:unit` — and with it the release script's unit gate — for the 0.5.0-rc.10 release. It also failed identically on the untouched parent (`c99552ef`), so this is pre-existing, not a regression.

## Test plan

- [x] `vitest run test/exports.test.ts -t "superset of the root"` ×3: 115 / 109 / 103 ms (was 5254 / 5066 / 6246 ms timing out)
- [x] `test/exports.test.ts`: 13/13
- [x] swap `test:unit`: 1511/1511 (was 1510/1511 on this machine)
- [x] swap `lint`

## Summary by CodeRabbit

* **Tests**
  * Improved the check for missing root client exports in the advanced export tests; the asserted behavior is unchanged.


Kukks:
@coderabbitai review

